PR #2, opened 2015-10-12
Fix backward-compatibility for config versions

milliams:
Due to an mistake on my part when converting the version strings for Git (trying to use 'x.y.z' consistently everywhere rather than 'Ganga-x-y-z') it now breaks to run 6.1.11 and then go back and run 6.0.44 (for example). This is because the new_user_wizard runs and regenerates the config and places:

    # Ganga configuration file ($Name: 6.1.11 $). DO NOT remove this line.

on the first line which, when read by 6.0.44, fails to do a `version.split('-')[1]` on it and crashes.

Since we can't go back and change 6.0.44 etc. we need to make sure the config file from 6.1.12 can be read by earlier versions.

@rob-c did warn me about changing this version code but I thought I'd caught all the issues. However, this one slipped through.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -300,7 +300,7 @@ under certain conditions; type license() for details.
         with open(os.path.join(os.path.dirname(Ganga.Runtime.__file__), 'HEAD_CONFIG.INI'), 'r') as config_head_file:
             new_config += config_head_file.read()
         new_config += config_file_as_text()
-        new_config = new_config.replace('Ganga-SVN', _gangaVersion)
+        new_config = new_config.replace('Ganga-SVN', 'Ganga-'+_gangaVersion.replace('.', '-'))  #Add in Ganga-x-y-z format so that it is backward compatible.
         with open(config_file, 'w') as new_config_file:
             new_config_file.write(new_config)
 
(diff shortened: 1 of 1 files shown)


python/Ganga/Runtime/bootstrap.py
```diff
@@ -37,6 +37,17 @@
...
+
+def new_version_format_to_old(version):
+    """
+    Convert from 'x.y.z'-style format to 'Ganga-x-y-z'
+
+    Example:
+        >>> new_version_format_to_old('Ganga-6-1-11')
+        '6.1.11'
+    """
+    return 'Ganga-'+version.replace('.', '')
```
alexanderrichards: I think that there might be a couple of problem in this function as is:
* The doctest seems to be converting old to new which is the opposite of the function doc and name. In addition, the output expected seems to be missing the Ganga- prefix which the return line suggests should be present.
* I think you meant to have the replace char as a -, i.e.
```python
version.replace('.', '-')
```
* As you don't strip Ganga- from the `version` argument, the return would in the case of your doctest example be `Ganga-Ganga-6-1-11`


I think it's worth noting that this function is a general one that will guarantee that you have the old style as long as you strip the `Ganga-` before the replace, such that if you pass it the new `.` style you will get the same as it you pass it the `-` style, namely the old style so it should be safe to use in each case.

alexanderrichards:
The patch looks good to me, my in-line comment is now expandable as the diff has changed. Did you see the suggestion I made re a generalised func?

milliams:
Thanks @alexanderrichards I've fixed the errors with the code. I like the idea with the `standardise_version` function but I would probably only use it in the places where I've currently got `new_version_format_to_old`. In most places I'd prefer to be able to just use `_gangaVersion` and otherwise explicitly track those places where we're having to keep compatibility.

alexanderrichards:
@milliams fair enough. Patch looks good to me then.

PR #5, opened 2015-10-12
Remove the old JobRepository and JobRepositoryXML directories

milliams:
As far as I can tell, these are completely superseded by `python/Ganga/Core/GangaRepository` and are not referenced anywhere in the code any more. As discussed in the last meeting, this code was confusing us when trying to debug repository problems so it should be removed.


alexanderrichards:
Seems reasonable, If they are not referenced anywhere in the code then they are clearly candidates for dumping. Also I should think with something this Core related we will soon find out if it's broken :wink:

PR #7, opened 2015-10-13
Cleanup unused files

milliams:
There are a whole lot more unused files which I have come across. Most of these are old monitoring-related things but there is also some repository migration code there too.

I have tested submitting jobs and getting output and nothing seems to be broken.

This also reduces the size of python/Ganga by about 15% (nearly 10,000 lines).


alexanderrichards:
> This also reduces the size of python/Ganga by about 15% (nearly 10,000 lines).

Nice! I don't plan to look through all 10K lines deleted but I think that assuming that these lines are not referenced anywhere then we are probably safe.

rob-c:
I'm in favour of this, most of the stuff on the file list I've had my eye on either deleting or checking if I can even get the code to load, if we're aiming to drop it then the more the better at this stage imo

PR #8, opened 2015-10-14
Remove unreachable code

milliams:
There's three places where we currently have unreachable code. The first is in `should_wait_batch_cb` where in commit 9b78ad8 the function was made to always return `True`, regardless of anything. Perhaps this was just for testing, I'm not sure. @rob-c, do you remember?

The second is in `BoxRegsitrySlice` where the behaviour of `__getitem__` is such that it will always return the first match, regardless of how many there might have been. `RegistrySlice` however, will raise an warning if there is more than one match. This behaviour has been like this for many years now so I assume it's ok and so I removed the useless error checking.

The third case is trivial and is probably just vestigial.

### DIFF
--- a/python/Ganga/Core/__init__.py
+++ b/python/Ganga/Core/__init__.py
@@ -124,27 +124,9 @@ Do you want to force the exit (y/[n])? """ % (t_total, len(critical_thread_ids),
 
 
 def should_wait_batch_cb(t_total, critical_thread_ids, non_critical_thread_ids):
-    from Ganga.Core.MonitoringComponent.Local_GangaMC_Service import config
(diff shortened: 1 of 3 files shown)


rob-c:
I think this was added to prevent the an effectively infinite loop at some point, horrible hack but better than Ganga freezing on shutdown.
The bad thing is I didn't add a fixme comment.

I think you can simply remove the return True as this seems to safely shut down everything correctly now, I've tested and this simply waits for tasks to complete without asking for user permission to force the exit which is what is expected and I think the infinite loop problem has been fixed at some stage due to the shutdown now being done in a much more sane way.

milliams:
Thanks. I'll change that commit and remake the request.

PR #15, opened 2015-10-16
Remove unreachable code

milliams:
This is a retry of PR #8 with the change that @rob-c suggested.


alexanderrichards:
@milliams FYI instead of a new pull request you could have made a fixup commit and rebased it. This will change the history which is bad for a main develop/master branch but for a feature branch where it's unlikely that others will be simultaneously editing it, it can be cleaner than a separate pull.

alexanderrichards:
@milliams can this be merged now?

milliams:
@rob-c as discussed in the meeting, we'll stick with removing the code as it will keep it working the way it has been.

PR #28, opened 2015-10-30
Bugfix/working ganga list ganga object

rob-c:
Hi all,

I'm just making a pull request for this to get some feedback before I try and merge this into develop as it has the potential to shake up a few things. This does change the Schema type of the internal object _list within GangaList.
If this causes problems I will work out a fix for this but this change is something I'd strongly like to keep having made the change.

This branch addresses the problem that (Ganga)lists of GangaObjects can't be written to disk. It does however require that you make use of ComponentItem schema attributes for the list object you wish to write to XML.

The reason for this and the change of the schema for _list in GangaList is.
The SimpleItem attempts to wrap lists (and other objects) within <value> </value> tags which means that a change within the VStreamer to accommodate both simple values as well as more complex objects such as a class object becomes much more complex and could potentially break a lot of things.

As a bonus I've also cleaned up the XML which is produced by this branch so the XML formatting should be improved (I think) once this is merged.

Thanks,

Rob


markwslater:
@rob-c I may be misunderstanding, but can't we already persist Lists of Ganga Objects? I know that in Ganga/GPIDev/Lib/Tasks/ITask.py we store transforms as a list of Ganga Objects and these are persisted fine. They have to be labelled as ComponentItems but that's OK. I thought input/outputdata was the same as well. Am I missing something?

drmarkwslater:
[reposting as the correct account!]

@rob-c I may be misunderstanding, but can't we already persist Lists of Ganga Objects? I know that in Ganga/GPIDev/Lib/Tasks/ITask.py we store transforms as a list of Ganga Objects and these are persisted fine. They have to be labelled as ComponentItems but that's OK. I thought input/outputdata was the same as well. Am I missing something?

milliams:
I thought the same as Mark. I came across the fact a while back that anything that's a `GangaObject` has to be specified as a `ComponentItem`, otherwise it does not save any of the attributes. Having a `GangaList` of simple Python objects might work as a `SimpleItem` by chance but it's not by design.

alexanderrichards:
As I recall, lists of GangaObjects were fine but the problem was nested GangaLists.

rob-c:
If you have a GenericSplitter object which is splitting GangaObjects
(I know this isn't realistic as the GangaObject can't be passed to the object in a sane way without the backend knowing to expect this).

In this case in master the list of arguments passed to the GangaObject is the __repr__ equivalent of the GangaList objects as the "GangaList._list" where the list is stored is defined as a SimpleItem.
Using this branch the correct GangaObject is fully written to disk and read in again. Although this does mean that the Schema may have changed compared to what is already on disk.

@drmarkwslater 
In Ganga/GPIDev/Lib/Tasks/ITask.py you are explicitly storing the list of (possibly GangaObjects) as a ComponentItem which is defaulted to be a list "[]" rather than a GangaList, this will trigger a very different behaviour to storing the object as a SimpleItem.
I have no idea if this ever gets transformed back into a GangaList when data is read from disk I'll have a look in a bit.
The file Ganga/Core/GangaRepository/VStreamer.py uses the to/from XML functions and that explicitly handles SimpleItem and ComponentItem schema objects differently.

@alexanderrichards 
We may still have the problem where we want to write nested GangaLists, but now the GangaList can write a list of GangaObjects to disk and the GangaList itself inherits from a GangaObject so in principle I don't see a problem with using this in that way.
(Is suspect the only problem may be that the XML Parser may not fully descend into lists within lists but I think this is a problem localised to the parser)

alexanderrichards:
> (Is suspect the only problem may be that the XML Parser may not fully descend into lists within lists but I think this is a problem localised to the parser)

Yes as I recall that was exactly the problem.

rob-c:
@alexanderrichards In that case I have made a slight modification here to allow for some behaviour like this as it was required for the GangaList to be fully evaluated. I'll do some more proper testing at some point today as now I finally understand what's going on it hopefully should be a quick fix.

rob-c:
I just added the ability for the VStreamer to descend into lists of lists within the XML. This works for both my tests with [] and GangaList objects. This still does not work with SimpleValue schema attributes but as long as people are aware they should be using ComponentItem objects for potentially complex objects being written to disk I don't think this is too much of an issue.

I'll eventually run some tests of writing simple values to disk with an older branch and then loading them with this branch to verify that everything is working as expected will full backwards compatibility and if it does then I plan to merge this branch as long as there are no objections?

<aside>
This will be a backwards incompatible change so if anyone moves to use lists of lists in the XML beware that at best older Ganga versions will mark it as incomplete and some versions may crash on loading. But I think we can live with this.
</aside>

alexanderrichards:
> if it does then I plan to merge this branch as long as there are no objections?

If it's working and your happy with it then just go ahead and merge whenever your ready

rob-c:
I've had a look and still need to remove the warnings from GangaList and elsewhere which stop nested lists from being created. Once 6.1.13 is out today I should be able to finish this.

PR #31, opened 2015-11-04
Feature/IPython3

rob-c:
Hi all,

This is the branch I've been working on to enable IPython 3 support and drop support for the older 0.6.13 version.

This branch includes a few additional patches for one or 2 functions which behave differently under the change to IPython.

In addition to this I've done some cleaning up of code in the Runtime/bootstrap to identify static or nearly static methods. This gives me a clearer idea of how the bootstrap should be run from an integration test point of view and should make the code a little easier to understand I hope.

On top of this work I've also done some cleaning up of how the shutdown is handled and I think I've resolved the inability to access a global object during the final stage of shutdown.

Rob

### DIFF
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -439,13 +439,30 @@ def GPIProxyClassFactory(name, pluginclass):
         if cycle:
             p.text('proxy object...')
             return
-        p.text(self._display(True))
+
+        global proxyRef
+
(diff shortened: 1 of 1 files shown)


rob-c:
As mentioned in the meeting typing a multi-line statement causes the logger buffer to flush between IPython prompt lines under this patch. It would be nice to address this.

rob-c:
If there are no objections I plan to merge this branch this-afternoon after resolving some conflicts locally

PR #38, opened 2015-11-11
Use Python 2.7

milliams:
The main change in the pull request is a Bash script which tries to find Python 2.7 on the system and use it if possible. If Python 2.7 is not found, it raises a warning.

The logic for finding Python 2.7 is in `bin/find-python2.7` where it cycles though `python2.7`, `python` and `scl`.

`bin/ganga` is the shell script which uses this to start the (now renamed) `bin/ganga.py` which is unchanged from before.

I've taken (and edited a little) `bin/ganga-external-env.py` from Mark's `feature/python-importing-new` branch which does the work of finding where the externals are located and adds them to the `PYTHON_PATH`.

Finally, in `bootstrap.py` I've added a warning message if the Python version is less than 2.7 to point the user towards [the wiki](/ganga-devs/ganga/wiki/Python-2.7).

### DIFF
--- a/bin/find-python2.7
+++ b/bin/find-python2.7
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+#Find a valid version of Python and run with it
+#First try the binary named 'python2.7'
+#Then check if 'python' is version 2.7
+#Finally try 'python27' from Software Collections
+if command -v python2.7 >/dev/null 2>&1; then
+    PYTHON="python2.7"
+elif python -c "import sys; exit(0) if sys.version_info[0:2]==(2,7) else exit(1)"; then
+    PYTHON="python"
+elif scl --list | grep -q python27; then
+    PYTHON="scl enable python27 -- python"
+else
+    exit 1
+fi
+
+echo $PYTHON
--- a/bin/ganga
+++ b/bin/ganga
@@ -8,11 +8,10 @@ while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symli
     [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-
 BASEDIR=$( dirname "${DIR}" )
 
 #Add the path to the installed externals
-NEW_PYTHONPATH=`PYTHONPATH=$BASEDIR/python $DIR/ganga-external-env.py`
+NEW_PYTHONPATH=`$DIR/ganga-external-env.py $BASEDIR`
 export PYTHONPATH=$NEW_PYTHONPATH${PYTHONPATH:+:}${PYTHONPATH}
 
 if PYTHON=`$DIR/find-python2.7`; then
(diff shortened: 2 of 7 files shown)


JacekHoleczek:
Could you, please, implement a flag / option / environment variable which will "enforce" python's version?
So that I could prevent "bin/find-python2.7" from searching for python2.7 and use the system default version (which will be python 2.6 in the case of my remote cluster).

milliams:
@JacekHoleczek What is the problem with the Ganga UI being run with `python2.7` if it is available? This should not affect any remote scripts or job processes.

JacekHoleczek:
I think, if python 2.6 is the default system version then I can expect that it works, including a decent set of modules which I expect to be installed (although no "pip", of course).
That's a different story for any additional / alternative python2.7 which may come from an "unknown source" (e.g. compiled by somebody without doing any comprehensive tests, be this way or another broken, missing modules, ...).
So, as long as I am not forced to use python2.7 I would prefer to stay with the "default" one.
If I am forced to use python2.7, I might end in a situation that I am forced to compile and install my private version from scratch myself.

egede:
@JacekHoleczek   What I can maybe suggest is that we in the coming releases execute a deprecation warning when running with Python 2.6. We can still start whatever python is in your PATH though.

However, within not too long, Python 2.7 will become a requirement for us so you will at that point have to move forward (or become stuck with old Ganga versions).

@milliams So I suggest that you leave the "find Python 2.7" purely in a detection mode at the moment and then when we are ready for the Python 2.7 requirement we bring it into force. At that point starting with a lower version will make Ganga abort though.

JacekHoleczek:
I tried the newest Ganga 6.1.20 on a remote "user interface" machine which runs "SL 6".
The default "python", "python2" and "python2.6" (2.6.6) in /usr/bin".
There is also "python2.7" in "/usr/local/bin (2.7.6rc1).
Ganga, however, does NOT use this "python2.7".
When I start "ganga", I get TWO times the "WARNING  Ganga will soon be depending on Python 2.7.".
I tried to explicitly execute "python2.7 /full/path/to/ganga" and in this case I get this warning displayed ONE time, so even though I explicitly ask for "python2.7", ganga internally still runs something with the system default "python".
Note: I cannot create a "system wide" symbolic links from "/usr/local/bin/python2.7" to "python" and "python2" (I do not manage this "user interface" machine).
What's more important, I do not think I should make my private links in "${HOME}/bin" either. The "${HOME}" filesystem is shared between different machines, including PBS batch nodes. There is no guarantee that all these machines have "/usr/local/bin/python2.7" installed (as the "user interface" machine has).
So, could you, please, modify relevant ganga "controlling" scripts which are run on the "user interface" machine so that they "inherit" python version from the main "ganga" script (i.e. if I run "python2.7 /full/path/to/ganga", then they all should use "python2.7", too).

KonstantinSchubert:
I'll add my two cent here.

I don't think it is good behavior for a python package to search for an automatically use a python version that is not the default system version. That is the kind of "magic" which causes sleepless nights when debugging.

Also, relying on heuristics and default install locations to check which versions are available is prone to break, given how many flavors of linux exits.

One should simply use `sys.version_info` to check for the required python version and issue an informative error message if the version is not appropriate.
The error message can then instruct the user how he can configure ganga to use a non-default python version.

http://stackoverflow.com/a/446136/1375015

egede:
We require python 2.7 now, so this discussion is past the point.

PR #61, opened 2015-12-09
Make sure lock is released in monitoring

milliams:
When trying to get the new test system running in non-interactive mode (i.e. no active monitoring loop) I came across a race condition where a thread acquires a lock but never releases it. I wanted to check that this was a bug and not a feature that's malfunctioning elsewhere.

The lock is acquired on line 297 and is released on line 305 in the `if` block. However, in the `else` block it is never released and the function then returns, holding the lock forever.

I'd rather have changed this to a context manager but it's not possible to acquire a non-blocking `RLock` using a context manager.

@alexanderrichards, @rob-c Since you've been looking at the monitoring code before, perhaps you have a better idea?

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -2,6 +2,7 @@ import Queue
 import threading
 import time
 import copy
+from contextlib import contextmanager
 
 from Ganga.Core.GangaThread import GangaThread
 from Ganga.Core.GangaRepository import RegistryKeyError, RegistryLockError
@@ -220,6 +221,29 @@ class _DictEntry(object):
         return self.backendObj, self.jobSet, self.entryLock
 
 
+@contextmanager
+def release_when_done(rlock):
+    '''
+    A ``threading.Lock`` (or ``RLock`` etc.) object cannot be used in a context
+    manager if the lock acquisition is non-blocking because the acquisition can
+    fail and so the contents of the ``with`` block should not be run.
+
+    To allow sensible ``release()`` behaviour in these cases, acquire the lock
+    as usual with ``lock.acquire(False)`` and then wrap the required code with
+    this context manager::
+
+        if lock.acquire(blocking=False):
+            with release_when_done(lock):
+                #some code
+                pass
+
+    Attributes:
+        rlock: A lock object which can have ``release()`` called on it,
+    '''
+    yield rlock
+    rlock.release()
+
+
 class UpdateDict(object):
 
     """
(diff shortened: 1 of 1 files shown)


rob-c:
After reading this it looks like this is the lock which has been causing me problems that I couldn't find.
I have definitely observed a bug that after running for a while the monitoring seems to lock up. It goes through the motions of 'checking' jobs but doesn't actually do anything. I'm thinking it's possibly because one of the internal monitoring threads is locked with this.
(This could also potentially explain the stall I see from a non-active monitoring loop on shutdown which I still see from time to time)

I'm still not familiar with how the UpdateDict and timeoutcheck work. All I know is that they are pieces of code I try to avoid touching.
But, doesn't the releaseLocks method fix this at all? (I could be way off the mark in thinking this)

milliams:
You're right that `releaseLocks` should release this but the problem is that it is never called since I believe that the thread that would call it hangs at `_checkBackend` on line 918, waiting for the other thread to release the lock. The thread that holds the lock then moves on and ends up in a situation where it's waiting for the monitoring thread to be released so that it can continue and so both threads are waiting for each other. Classic deadlock.

That's my understanding anyway, I'd rather that this code all dies and was reborn in a less over-engineered way :)

rob-c:
@milliams I agree to fixing this properly with a cleaner simpler implementation asap!

alexanderrichards:
> I'd rather have changed this to a context manager but it's not possible to acquire a non-blocking RLock using a context manager.

We can easily make one:

```python
from contextlib import contextmanager

@contextmanager
def nonblocking_acquire(rlock):
    if rlock.acquire(blocking=False):
        yield rlock
        rlock.release()
```

alexanderrichards:
@milliams I also like contexts as I think this is what they were conceived for. I would be in favour of implementing something like the above and then using this where you mention. I'm sorry it's been a while since I actually looked at the monitoring loop code so I'd have to swat up a bit.

rob-c:
Would it be worth having a branch for a proposed new monitoring loop?

Adding some machinery to decide which monitoring loop to launch at start would be useful and would help in breaking out the monitoring loop code from elsewhere in the codebase.
Ultimately this would allow us to remove the old monitoring code once the new one is working in a relatively easy manner.
I think this can easily be made into a standalone class which has one method to startup and which shuts down in a much cleaner manner.

alexanderrichards:
just realised my example was too simplistic and wont work as it doesn't yield in the else case. Ah well I'm sure we can do something like this though

milliams:
@alexanderrichards The problem here seems to be that if the lock cannot be acquired then the code within the `with` block should not be executed. There is no way to do this with a context manager, see http://stackoverflow.com/a/12594323/96704.

@rob-c At the very least this will have to wait until after the credentials branch is merged and Jenkins is fully running. I think we need an in-person technical meeting to go through exactly what is running at any one time and what's in charge of what. However, I definitely agree with the idea of an explicit start and stop entry points for all these processes (as I think we've both found while working on `GangaUnitTest`!).

rob-c:
Yes I agree that this is a long term project but hopefully once the credentials branch is in we can do something more sensible than having the monitoring loop managing afs tokens and credentials as well as updating job statuses all through the same queue system.

Starting the monitoring loop at the moment shouldn't be too bad imo, the main difficulty has been in shutting everything down in a sane way.

OK, discussion aside I'm happy to merge this into develop and for it to be pulled into 6.1.14 for testing.

alexanderrichards:
> @alexanderrichards The problem here seems to be that if the lock cannot be acquired then the code within the with block should not be executed. There is no way to do this with a context manager, see http://stackoverflow.com/a/12594323/96704.

yup your quite right, momentary lapse on my part. As I recall it was stated in the refused PEP on this "The with shouldn't impersonate an if statement". However I see no reason why we cant try the lock first and then use a context to remember to release it.
```python
from contextlib import contextmanager

def release(rlock):
    yield rlock
    rlock.release()

l = RLock()
if l.acquire(blocking=False):
    with release(l):
        <blah>
```

note that a new context is needed as `RLock` defines `__enter__` already which will try to acquire the lock and at this point we only care about releasing it. Either this way or create a custom `RLock` with a nullop `__enter__` but I think this is clearer.

milliams:
@alexanderrichards I've updated it based on your suggestion. I renamed it something less generic and fleshed it out with some comments.

alexanderrichards:
@milliams looks good, much more readable IMO. Just one thing that I think you have overlooked which to be fair I didn't include in my original example. If the yielded block raises an exception then you need to also release the lock so to be even more specific, I think that you need to modify it slightly to reflect the following.

```python
from contextlib import contextmanager

@contextmanager
def release(rlock):
    try:
        yield rlock
    finally:
        rlock.release()

l = RLock()
if l.acquire(blocking=False):
    with release(l):
        <blah>
```

milliams:
@alexanderrichards Ahh, you're right. I'd seen that in the documentation but had forgotten about it. Funnily enough, without the try/finally, we might as well have not used a context manager at all :)

alexanderrichards:
> Funnily enough, without the try/finally, we might as well have not used a context manager at all :)

Indeed :smile:

alexanderrichards:
This looks good to me now, merge it when your happy.

PR #63, opened 2015-12-10
Don't bother checking for attributes we know exist

milliams:
As discussed in the meeting, I'm not sure we need to check for these things. It's probably best to wait until after 6.1.14 for this just to be safe though.

All `Job`s have a `backend`, all `IBackend`s have a `setup()` and `remove()` method and all `IApplication`s have a `transistion_update()` method.

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -961,25 +961,7 @@ class JobRegistry_Monitor(GangaThread):
             try:
                 log.debug("[Update Thread %s] Updating %s with %s." % (currentThread, getName(backendObj), [x.id for x in jobList_fromset]))
                 for j in jobList_fromset:
-
-            	    if hasattr(stripProxy(j), 'getNodeIndexCache') and\
-                        stripProxy(j).getNodeIndexCache() is not None and\
-                        'display:backend' in stripProxy(j).getNodeIndexCache().keys():
-
-                        name = stripProxy(j).getNodeIndexCache()['display:backend']
-                        if name is not None:
-                            import Ganga.GPI
-                            new_backend = eval(str(name)+'()', Ganga.GPI.__dict__)
-                            if hasattr(new_backend, 'setup'):
-                                j.backend.setup()
-                        else:
-                            if hasattr(j, 'backend'):
-                                if hasattr(j.backend, 'setup'):
-                                    j.backend.setup()
-                    else:
-                        if hasattr(j, 'backend'):
-                            if hasattr(j.backend, 'setup'):
-                                j.backend.setup()
+                    j.backend.setup()
 
                 if self.enabled is False and self.alive is False:
                     log.debug("NOT enabled, leaving")
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -1772,50 +1772,11 @@ class Job(GangaObject):
             # this is used by Remote backend to remove the jobs remotely
             # bug #44256: Job in state "incomplete" is impossible to remove
 
-            if stripProxy(self).getNodeIndexCache() is not None and 'display:backend' in stripProxy(self).getNodeIndexCache().keys():
-                name = stripProxy(self).getNodeIndexCache()['display:backend']
-                if name is not None:
-                    import Ganga.GPI
-                    new_backend = eval(str(name)+'()', Ganga.GPI.__dict__)
-                    if hasattr(new_backend, 'remove'):
-                        self.backend.remove()
-                    del new_backend
-                else:
-                    if hasattr(stripProxy(self.backend), 'remove'):
-                        stripProxy(self.backend.remove())
-            else:
-                if hasattr(stripProxy(self.backend), 'remove'):
-                    stripProxy(self.backend).remove()
+            stripProxy(self.backend).remove()
 
-            if stripProxy(self).getNodeIndexCache() is not None and 'display:application' in stripProxy(self).getNodeIndexCache().keys():
-                name = stripProxy(self).getNodeIndexCache()['display:application']
-                if name is not None:
-                    import Ganga.GPI
-                    new_app = eval(str(name)+'()', Ganga.GPI.__dict__)
-                    if hasattr(new_app, 'transition_update'):
-                        self.application.transition_update("removed")
-                        for sj in self.subjobs:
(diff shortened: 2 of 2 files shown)


rob-c:
Hi Matt,

The logic behind the change in Local_GangaMC_Service.py is to avoid loading every single job just to check the backend object associated with it.
(I should probably add a test to see if a backend has already been 'setup' or not. This is supposed to be run once per backend (or at least that's my understanding).)

The logic for this is for large repos (many hundreds of master jobs) running through the loading of each of these for the backend object is quite an expensive operation even when the subjobs aren't loaded compared to the fact that the smaller index files for the jobs have already been loaded at this point.
I would also like to avoid loading the jobs due to any hidden overheads that there may be such as memory footprint.

I know it's not ideal but I'm thinking of expanding the cache data per job in the future so that using this doesn't require something like an eval in order to be used.

I'm for the changes in Job if they don't cause a slowdown but again I don't really want to trigger the loading of Jobs from disk just to remove them. The backend may not expose the remove function through the GPI as it's just the same as the dummy remove function in the IBackend. If this hasn't been exposed I think loading the job just to call the dummy function is potentially bad.
Same goes for the Application and 'transition_update'.

Rob

milliams:
# backend.setup

There is, as far as I can tell, only one `IBackend` subclass which overrides the `setup` method and that is `Remote`. Furthermore, while it overrides it, it does not export it to the GPI. However, that is irrelevant since in the `_checkBackend`, it is operating on raw `Job` objects, not on their proxy. This means that code like:
```python
if hasattr(j, 'backend'):
  if hasattr(j.backend, 'setup'):
    j.backend.setup()
```
will *always* run the `setup()` method. In summary, *all* raw backends have a `setup` method and *no* proxied backends have a `setup` method.

The same facts and logic applies to `IBackend.remove()` too. `IApplication.transition_update()` is only ever overridden inside `MockExeApplication` which is only part of the test framework.

# Walkthrough

In `Local_GangaMC_Service.py` the code is currently getting all jobs which are `['submitting', 'submitted', 'running']`, then checks each of them for the following:

1. Does its proxy-stripped version have an attribute `getNodeIndexCache` *and* does `getNodeIndexCache()` not return `None` *and* have `'display:backend'` in the keys:
  1. If so, is `stripProxy(j).getNodeIndexCache()['display:backend']` not None
    1. If so, if we make an instance of that class (via an `eval`), does it have a `setup` attribute
      1. If so, run `j.backend.setup()`
      2. If not, do nothing
    2. If not (i.e. `'display:backend' is None`), does the `Job` have a backend attribute
      1. If so, does the backend have a `setup` attribute
        1. If so, run `j.backend.setup()`
  2. If not, does the `Job` have a backend attribute
    1. If so, does the backend have a `setup` attribute
      1. If so, run `j.backend.setup()`

So, at 1.i.a.a, 1.i.b.a.a, 1.ii.a.a the `j.backend.setup()` command will be run. It's my belief that all `Job`s have a `backend` attribute and all `IBackend`s have a `setup` attribute (as shown above) and so it can, at least, be simplified down to:

1. Does its proxy-stripped version have an attribute `getNodeIndexCache` *and* does `getNodeIndexCache()` not return `None` *and* have `'display:backend'` in the keys:
  1. If so, is `stripProxy(j).getNodeIndexCache()['display:backend']` not None
    1. If so, if we make an instance of that class (via an `eval`), does it have a `setup` attribute
      1. If so, run `j.backend.setup()`
      2. If not, do nothing
    2. If not (i.e. `'display:backend' is None`), run `j.backend.setup()`
  2. If not, run `j.backend.setup()`

and so the only branch in which `j.backend.setup()` is not run is 1.i.a.b which only happens if `stripProxy(j).getNodeIndexCache()['display:backend']` is the name of a class whose proxy does not have a `setup` attribute. At present this is *all* backends which makes this check quite redundant since if effectively becomes:

1. Does its proxy-stripped version have an attribute `getNodeIndexCache` *and* does `getNodeIndexCache()` not return `None` *and* have `'display:backend'` in the keys:
  1. If so, do nothing
  2. If not, run `j.backend.setup()`

and, realistically all the way to:

1. Does its proxy-stripped version `getNodeIndexCache()` not return `None`:
  1. If so, do nothing
  2. If not, run `j.backend.setup()`

which means:

1. Is the job lazy-loaded:
  1. If so, do nothing
  2. If not, run `j.backend.setup()`

So, at present the code will run `setup()` on all fully-loaded jobs (regardless of whether `setup` is exported) and will never run `setup()` on any job which is not yet loaded (i.e. being lazy-loaded).

This is all a lot of work for attempting to run `setup` for the `Remote` backend but if we need to do that then I think something like the following:

```python
if stripProxy(j).getNodeIndexCache() is not None \
        and 'display:backend' in stripProxy(j).getNodeIndexCache().keys():
    # The job is not fully-loaded so let's be more careful so as to not load if it not necessary
    name = stripProxy(j).getNodeIndexCache()['display:backend']
    if name is not None:
        import Ganga.GPI
        # Check the raw class of the backend to see if it overrides the setup method
        if hasattr(Ganga.GPI.__dict__[name]._impl.__dict__, 'setup'):
            j.backend.setup()
        else:
            # The backend does not override the setup() method so
            # let's not bother with running the base-class version
            pass
    else:
        # The job is lazy-loaded but does't give us the required information so let's setup anyway
        j.backend.setup()
else:
    # The job is fully-loaded so it's safe to just try running this function
    j.backend.setup()
```
Note, no `eval` is used, although it's still quite horrible.

rob-c:
@milliams I've made some changes in the release/6.1.14 branch to remove the eval functions in favour of looking up the backend object in the GPI using a helper function in the Proxy class (arguably not the best place to put this but as it returns a proxied object it feels correct).
Thanks for looking at this

milliams:
Having this in a helper function is definitely better but the problem is still that `getRuntimeGPIObject` returns the proxy-wrapped Remote backend and it does not export the `setup` method (nor do *any* of the other backends). At the very least we want to strip the proxy before checking for these attributes.

This same issue is still present in Job.py as well.

rob-c:
I would strongly prefer for the applications to expose but hide these methods through the proxy which would be more appropriate (I'll make the change in the next 6.1.14 pre which is needed) and avoids needlessly loading large jobs which is a constant performance problem, this code was moved here from the bootstrap as it was causing a lot of large expensive jobs to be loaded for no reason.

Can we discuss this at the meeting on Thursday in more detail?

milliams:
I'm abandoning this PR since the code it was affecting was changed underneath it and no longer applies to develop. I'll have to come back to the point of this PR at another point in the future.

PR #62, opened 2015-12-09
Fix validate method of MassStorageFile to allow resubmission of subjobs

chrisburr:
This issue has been reported on JIRA ([GANGA-2002](https://its.cern.ch/jira/browse/GANGA-2002)).

When calling `resubmit` on a subjob ganga throws the following error if the job has a MassStorageFile associated with it:
```
Ganga.Runtime.bootstrap            : ERROR    JobError: Error in MassStorageFile.outputfilenameformat field :  job is non-split, but {'sjid'} keyword found
```

I've only tested this fix with the `6.1.13` tag as I was unable to start ganga under the `develop` branch on lxplus. Is there any up to date documentation for how to do this?

PR #67, opened 2015-12-12
Fix various undefined names

chrisburr:
While searching for the undefined instances of `stripProxy` for #62 I found some other instances of undefined names.

I've tried to minimise the number of modified lines and avoid changing anything related to the interface with the exception being the movement of [`SplitterError` to `Ganga.Core.exceptions`](https://github.com/ganga-devs/ganga/commit/5fd7508da74b409d3963c97ce0edb20aa43140da).

I've opened this PR against `release/6.1.14` to keep the changelog simple, is develop a more suitable final destination for these changes?

A few cases remain where I'm unsure of a clear solution:
- [ ] [`config` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Core/__init__.py#L193) ([solution?](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Core/__init__.py#L155))
- [ ] [`err` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Core/exceptions.py#L109) ([solution](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Core/exceptions.py#L87)?)
- [ ] [`getDataFile` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/GPIDev/Lib/Dataset/GangaDataset.py#L77)
- [ ] [`lookup_input` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py#L273-L276)
- [ ] [`lookup_input` is always an empty list](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py#L343)
- [ ] [`config` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Lib/LCG/GridSimulator/simulator.py#L23)
- [ ] [`gangadir` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/bootstrap.py#L435-L439)
- [ ] [`override_credits` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/bootstrap.py#L1420)
- [ ] [`_display` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/bootstrap.py#L1422)
- [ ] [`tempFilePath` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/http_server.py#L468-L481)
- [ ] [`gaudi_env` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/GangaGaudi/Lib/Applications/GaudiBase.py#L127) (`gaudi_env = exec(in_file.read())`?)
- [ ] [`construct_merge_script_cmake` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/GangaLHCb/Lib/Applications/EnvironFunctions.py#L39)
- [ ] [`cmdbase` is undefined](https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/GangaLHCb/Lib/Applications/PythonOptionsParser.py#L62)

@rob-c Does [this change](https://github.com/chrisburr/ganga/commit/d08a46856936a01780ded112edb53b1fe987ab55#diff-f01461c81aa68e48a1ee33d1c865166aL286) fix the issue you mentioned or should I rebase the comment back into this branch?


rob-c:
@chrisburr https://github.com/chrisburr/ganga/commit/d08a46856936a01780ded112edb53b1fe987ab55#diff-f01461c81aa68e48a1ee33d1c865166aL286 this failed due to incorrect resolution of the method as it was looking for a method which had been name mangled to belong to the BKQuery class despite calling a static function within the baseclass... This unfortunately doesn't give the performance improvement it was designed to in some testing but I think this is due to a python limitation. The code would be nice to use but isn't required for the release and may 'never' work if the bug is due to some other Core changes.
I'm a bit hesitant to move the definition of SplitterError for 6.1.14 at this point but for 6.1.15 this looks sensible imo :)

Most of these changes look OK, but one or 2 involve imports which have historically caused problems with circular dependencies etc on startup so I'll check these commits before merging.
I'm curious how you've spotted these bugs, was it due to some form of static analysis or did you manually parse the code?

A few quick comments. I'll try and put some fixes in somewhere on Monday.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Core/__init__.py#L193 This requires the use of getConfig in order to get the current runtime config, I'll put together a fix
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Core/exceptions.py#L109 I don't think this is used atm (I could be wrong, but it should be defined as the exception above I would have thought
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/GPIDev/Lib/Dataset/GangaDataset.py#L77 appears to be a copy/paste error from LHCbDataset
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py#L273-L276 and https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py#L343 are likely due to missing a self decleration. However, we don't rebuild this table at shutdown as it's potentially an extremely expensive operation. After a ganga devs meeting we decided to disable this as it shouldn't cause any ill effects.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Lib/LCG/GridSimulator/simulator.py#L23 I have no knowledge of if this is used but would imagine again this should be using the global getConfig if it's sensible.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/bootstrap.py#L435-L439 I think this should be default_gangadir rather than gangadir looking at the code and the exception if this fails
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/bootstrap.py#L1420 hmm, pass this may have been lost in a new modernisation branch. This requires some detective work
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/bootstrap.py#L1422 we no longer rely on modifying the display hook I thought. If this default option is required we could fix this or simply add a message explaining the requested way to launch ganga is through a script or using ipython.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/Ganga/Runtime/http_server.py#L468-L481 Completely pass on this one as I'm not even sure if this code works or is used.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/GangaGaudi/Lib/Applications/GaudiBase.py#L127 I think this should be reading the env from the file and placing it in self.env so I think this needs a minor bugfix.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/GangaLHCb/Lib/Applications/EnvironFunctions.py#L39 This code is being abandonned in favour of a new LHCb GaudiRun app so this will never be called, even though the function is missing.
https://github.com/chrisburr/ganga/blob/bugfix/undefined-names-rebased/python/GangaLHCb/Lib/Applications/PythonOptionsParser.py#L62 I think this should be defined as 'gaudirun.py' as per the command it's analysing the code from

chrisburr:
Apologies for the circular import, it is now fixed by [`c4d53bd`](https://github.com/ganga-devs/ganga/commit/c4d53bdbbd49dc915b4ec07284bca18fbce65f53).

I wish I was able to manually parse this much code :laughing: I found the errors with [flake8](https://flake8.readthedocs.org/en/latest/) by running `flake8 --select=F821 . > errors.txt`. The default output is a little overwhelming on established projects but it might be worthwhile looking at some of its other error codes. Most text editors have have plugins available that can be used to show errors and warnings inline and I've previously used it in combination with [Travis CI](https://travis-ci.org/) to automatically scan when commits are pushed.

rob-c:
Ah, nice to know flake8 runs in a sensible way on this code now :) I had some problems running it a while ago but I forget the exact nature of the error.
I think I'll investigate using this myself over the holiday.

After looking at the SplitterError I'm concerned that this is defined in Ganga/GPIDev/Adapters/ISplitter.py I would strongly prefer not to move this into the Ganga.Core.exceptions or to remove it from ISplitter as well.
This "shouldn't" cause too many problems but from experience changing things like this generates more work and I'm keen to try and get 6.1.14 finished asap and before Friday to get it out ahead of the Christmas break.
The SplitterError should be compatible with the Adapters so if you can remove the definition from ISplitter and verify that nothing breaks I'd be happy to merge the branch for testing

chrisburr:
As this branch stands the error in `ISplitter` and `exceptions` shouldn't conflict as they're named `SplittingError` and `SplitterError` though it is a little messy.

Is [this](https://github.com/chrisburr/ganga/commit/dfaa83446e2886c0d9ce474e72dc4add18d73c15) the change you propose? If not I can remove the commit that moves `SplitterError` to `Ganga.Core`.

egede:
While I have not followed the details of these proposed changes, we should 
not create names that are only subtly different. We should also avoid name 
changes unless the old names are somewhat misleading.

Ulrik

rob-c:
@chrisburr Ah, my mistake for not looking too closely at this over the weekend, I thought this was a redefinition.
@egede Good point, although if all of the exceptions were better defined in one place keeping track of the names already used would probably be helpful. I'll have a look but I'm not convinced we gain too much by having both 'SplitterError' and 'SplittingError' defined, but as it's already used in some areas I'm happy to keep it.

OK, I think merging this branch in it's current state ahead of the testing for 6.1.14 is sensible, I'm doing some more debugging today and hope to tag another pre this-afternoon
Thanks

egede:
@rob-c Did not realise that we already had both exceptions! I think to put the different exceptions into a common place is a good idea. Could be a well defined place for each top level package (Ganga, GanLHCb etc.).

rob-c:
@egede We already have a lot of exceptions defined in Ganga.Core.exceptions so I think simply moving the definitions of other exceptions here would be beneficial. I think this work can be performed in tandem with moving a few interfaces of Ganga.GPIDev.Adapters probably in 6.1.15

alexanderrichards:
> We already have a lot of exceptions defined in Ganga.Core.exceptions so I think simply moving the definitions of other exceptions here would be beneficial.

While SplittingError is generic enough that this is fine. I would not recommend moving _all_ exceptions to Ganga.Core.exceptions as there may very well be package specific ones. Therefore we should probably as @egede says have a well defined place for these under each package.

rob-c:
@chrisburr Can you remove https://github.com/chrisburr/ganga/commit/9e6f76d8694d326ea6a2f2e51741fee984cd84fd so that I can merge this branch? Thanks :)

chrisburr:
@rob-c Done

alexanderrichards:
Hi all,
One last comment on this. I think that this merge (or at least part of it) should have targeted 6.1.15 rather than 6.1.14 which was in pre-release status. As a general rule we only want to merge bugfixes into a pre-release. With the caveat that I'm only skim reading the commit messages and quickly the diffs, commit 5fd7508da74b409d3963c97ce0edb20aa43140da for example doesn't look to be addressing a bug, merely code organisation. I would argue that without extensive testing moving code around during a pre-release is a bad idea. Certainly I don't think that this particular commit addressed a bug (although I'd be happy to be proved wrong :smile:)

rob-c:
@alexanderrichards I agree entirely, however as this is very unlikely to cause problems and more likely fixes some. (This exception should have never been defined within GangaLHCb and GangaDirac for instance but appears to be being constructed there)
Most of these fixes are very useful and good which is why I decided to forgo the problem of moving the exception now rather than waiting for 6.1.15

python/GangaLHCb/Lib/RTHandlers/XMLWorkerScript.py
```diff
@@ -1,5 +1,8 @@
 # XMLSummary parsing
 import os, sys
+
+from GangaLHCb.Lib.Applications.AppsBaseUtils import activeSummaryItems
```
rob-c: Rats, I should have spotted this sooner. This is not correct, we do NOT ship ganga to the Worker Node.

PR #90, opened 2016-01-12
Feature/move config startup

### LINKED ISSUE #87: Move config setup to __init__.py of Core package
In order to get the little ganga import working, the config system has to be completely set up just by doing:

```
import Ganga
```

Consequently, all the config options should be moved to __init__.py

drmarkwslater:
Fixes #87. Basically moves all the config options to __init__.py of the package so on import, all the config options are setup properly. Note that this has necessitated the moving around of a couple of other things, notably from Tasks and bootstrap.py. 

MOST IMPORTANTLY - this changes the location of the version string! The release scripts would need to be updated!


python/Ganga/Runtime/bootstrap.py
```diff
@@ -583,80 +574,23 @@ def set_cmdline_config_options(sects=None):
...
-        # so you can refer to them in the config file
-        # additionally they will be visible in the (write protected) [System]
-        # config module
-        syscfg = Ganga.Utility.Config.makeConfig('System', "parameters of this ganga session (read-only)", cfile=False)
-        syscfg.addOption('GANGA_VERSION', _gangaVersion, '')
-        syscfg.addOption('GANGA_PYTHONPATH', _gangaPythonPath, 'location of the ganga core packages')
-        syscfg.addOption('GANGA_CONFIG_PATH', self.options.config_path,'site/group specific configuration files as specified by --config-path or GANGA_CONFIG_PATH variable')
-        syscfg.addOption('GANGA_CONFIG_FILE', self.options.config_file, 'current user config file used')
-        syscfg.addOption('GANGA_HOSTNAME', hostname, 'local hostname where ganga is running')
+        from Ganga.Utility.Config import getConfig
```
milliams: Is this not already imported at the top of the file? One of them is `from Ganga.Utility.Config.Config` and the other is `from Ganga.Utility.Config` so perhaps that is why it wasn't noticed.

milliams:
I'm looking through this now. On the whole it looks fine (except the changes to Job.py which Mark says were made by mistake). I will run some local tests but I can't see a big reason not to merge.

rob-c:
Erm, I'm slightly concerned that this has rendered GangaLHCb unloadable, probably some minor fixes needed

drmarkwslater:
It also broke Atlas as well - fixed now :)

Issue #86, opened 2016-01-11
pycrypto needs to be re-built with gcc49


rob-c:
Hi all,

In releasing Ganga 6.1.14 for LHCb gcc49 I spotted that the external pycrypto is currently still built for gcc48.

I don't know how complex building this external is but I'd rather leave building of the externals to the experienced.

After a quick look I don't think this is particularly used in the Core so if it's easier to move this external to module rather than re-building it for gcc49 I propose this as a solution to the problem.

Thanks,

Rob

egede:
@rob-c  pycrypto is **only** used for the remote backend. 

@drmarkwslater Did you say that the remote backend was broken anyway?

drmarkwslater:
Basically yes - best thing to do at present is bypass it in the quickest way possible and I'll revisit the Remote backend at some point in the future

rob-c:
@drmarkwslater Shall I just remove this from the list of Core externals for 6.1.15 in this case?

drmarkwslater:
I think so - as I say, it's currently broken anyway and needs a rewrite!

milliams:
I would suggest removing it from the list of externals. The `import paramiko` is inside a function call anyway so it shouldn't cause any import errors for users.

rob-c:
@milliams I'm assuming you mean that we can drop both paramiko and pycrypto from the Core externals in this case?

milliams:
@rob-c Yes, I think we can. The only place they are used (pycrypto is only there as an indirect dependency of paramiko) is inside the `opentransport()` method of `Remote`.

milliams:
This was fixed by #94

PR #94, opened 2016-01-12
Removing pycryo and paramiko from Core externals for the time being

rob-c:
This is to try and address #86


milliams:
Looks good to me. This will only make things easier for the move to setup.py as well.

Closes #86

milliams:
Looks good to me. This will only make things easier for the move to setup.py as well.

Closes #86

PR #100, opened 2016-01-13
Bugfix/push session lock to back of thread shutdown

rob-c:
This makes sure that the SessionLocks are kept right up until the end of the thread shutdowns.

Maybe this shouldn't be being monitored by the GangaThreads since the Repos (could in principle) rely on this for shutdown.

At least now this is stopped after the Monitoring loop has finished it's work.

### DIFF
--- a/python/Ganga/Core/GangaThread/GangaThreadPool.py
+++ b/python/Ganga/Core/GangaThread/GangaThreadPool.py
@@ -169,9 +169,9 @@ class GangaThreadPool(object):
 
         logger.debug("ExternalTasks still running: %s" % queues.threadStatus())
 
-        logger.debug('Service threads to shutdown: %s' % list(_all_threads))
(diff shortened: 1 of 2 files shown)


python/Ganga/Core/GangaThread/GangaThreadPool.py
```diff
@@ -169,9 +169,9 @@ def __do_shutdown__(_all_threads):
 
         logger.debug("ExternalTasks still running: %s" % queues.threadStatus())
 
-        logger.debug('Service threads to shutdown: %s' % list(_all_threads))
+        logger.debug('Service threads to shutdown: %s' % ([i for i in reversed(list(_all_threads))]))
```
milliams: Just doing
```python
list(reversed(_all_threads))
```
would perhaps have been simpler.

PR #101, opened 2016-01-13
Bugfix/close stomp threads early

rob-c:
I was observing that the STOMP related communication threads are used once and then left on the GangaThread monitoring list.
This means that in some of my tests I have >100 threads being killed by the shutdown manager and whilst it can cope correctly with shutting down these non-critical threads before the critical ones it creates a lot of noise I'd rather not have hanging around until shutdown.
This kills, shuts-down and deletes the threads once they've done their jobs of sending single messages to the stomp server.

### DIFF
--- a/python/Ganga/Utility/stomputil/publisher.py
+++ b/python/Ganga/Utility/stomputil/publisher.py
@@ -131,6 +131,7 @@ def createPublisher(T, server, port, user='', password='', logger=None,
             self.backoff_max = 60
             # queue to hold (message, headers, keyword_headers) tuples
             self._message_queue = Queue()
+            self.__finalized = False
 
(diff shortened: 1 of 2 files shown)


python/Ganga/Utility/stomputil/publisher.py
```diff
@@ -301,6 +302,8 @@ def _finalize(self, timeout):
             @param timeout: Maximum seconds to clear message queue on exit.
                     Negative value indicates clear queue without timeout.
             """
+            if self.__finalized is True:
```
milliams: When dealing with booleans it's much nicer to just do:
```python
if self.__finalized:
    ...
```

Issue #96, opened 2016-01-12
Objects being wiped as soon as they're written


milliams:
I've hit upon a strange bug in the new integration tests. The situation is that I want to add a job to a `JobTree` and then check to see if was added.

In the integration test however, the code seems to be adding the job to the tree and then making it as dirty which prompts a flush to disk. In the process of flushing to disk (using `VStreamer._raw_to_file`) it calls `GangaObject.accept` which requests read access (via `GangaObject._getReadAccess`) which calls `Registry.__safe_read_access` which evaluates `if this_id not in self._loaded_ids` to `True` and so calls `Registry.load` which loads stale information from disk, overwiting the fact that a job was just added.

The registry is question is the one found via `jobtree._impl._getRegistry()` which is the same `getRegistry('jobs').metadata`.

Now, this is only seen in the integration test and so there might just be a slight misconfiguration there but it is 100% reproducible. Have we seen this sort of behaviour before or does anyone (@rob-c) have an idea of what might be causing it?

rob-c:
@milliams As per my email at the start of the year the ```_getReadAccess``` should be transparent and shouldn't perform a registry action if one is currently in progress. (It sounds like a flush to disk is in progress but half way through the object is loaded from disk again)

If the object is being flushed to disk then the Registry shouldn't be evaluating the methods to gain read access as it's a reasonable assumption that you've access if your doing something like writing the object to disk.

```_setDirty``` (or whatever the method in GangaObject is called) doesn't cause an immediate flush to disk due to the dirty count. Is the object correctly added to the registry in the same way that a Job object is added to the standard repo through the ```_auto__init__``` method?

I don't know how the handling of the different repos for one object fully works so I suspect that it's possible some error could be introduced through the Job being in both the registry and not the metadata.

I've not looked at the jobtree since I spend time looking at the registry so I don't want to say too much about how this class works.

**Comment by milliams** [comment]
It turns out that I can indeed reproduce this at the Ganga prompt. The difference is that if you have an existing ~/gangadir then doing:
```python
j = Job()
jobtree.add(j)
jobtree.listjobs()
```
will work correctly but if you completely delete ~/gangadir before running Ganga then the `add()` fails due to the bug I described above. I will look into this a little more.

The JobTree itself is only doing very little, it's perhaps how it interacts with `getRegistry('jobs').metadata` that is causing problems.

rob-c:
[...]

PR #102, opened 2016-01-13
Bugfix/minor xml init changes

### LINKED ISSUE #84: Remove additional load after flushing an object to disk
When an object is flushed to disk the object is then immediately loaded back into memory and overloads itself.

This is an expensive operation and I'm not seeing a very good motivation for this to being performed on every single flush so I'm opting to have this removed.

rob-c:
This reduces the amount of object loading that happens which is not required.

The workflow of flushing an object then reloading it from disk doesn't offer any real benefit over flushing the object to disk and then continuing with other work and in some scenarios causes some real slow down to occur.
Fixes #84

This also adds an object to the list of 'already loaded' objects in the Registry which should prevent the first call to getReadAccess or getWriteAccess from possibly losing some data if an object has incorrectly NOT been labelled as _dirty.
(When in doubt an object should rely on the methods in Core for deciding if it's become dirty or not and shouldn't be subverting this)
Possibly effects #96

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepository.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepository.py
@@ -193,13 +193,14 @@ class GangaRepository(object):
         """Internal helper: adds an empty GangaObject of the given class to the repository.
         Raise RepositoryError
         Raise PluginManagerError if the class name is not found"""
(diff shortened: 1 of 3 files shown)


rob-c:
May have to revert this due to a bug in object parents not being what I'm expecting. Investigating now

PR #105, opened 2016-01-13
Simplify logging in file utilities

milliams:
logging.DEBUG does not exist in our logging module and so this was causing
an exception. It wasn't really needed anyway so it has been removed from
here.

Without this, doing
```python
File('x')
```
would crash on latest `develop`.

### DIFF
--- a/python/Ganga/Utility/files.py
+++ b/python/Ganga/Utility/files.py
@@ -8,42 +8,28 @@
 Helper functions for operations on files.
 """
 
-## BE AWARE THIS IS LOADING Ganga.Utility.logging NOT python logging!! - rcurrie
-import logging
 import os.path
-import stat
 import Ganga
+from Ganga.Utility.logging import getLogger
 
 def expandfilename(filename, force=False):
-    "expand a path or filename in a standard way so that it may contain ~ and ${VAR} strings"
+    """expand a path or filename in a standard way so that it may contain ~ and ${VAR} strings"""
     expanded_path = os.path.expandvars(os.path.expanduser(filename))
     if os.path.exists(expanded_path) or force:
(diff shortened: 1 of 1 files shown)


drmarkwslater:
Looks good to me :)

rob-c:
I added this for keeping track of files being created which do not exist correctly.
I'm strongly in favour of keeping this as it helped me track down stupid things like files being pointed to that could never have existed

The DEBUG not existing is due to the wrong logging module being imported. I had fixed this in the past but it looks like some conflict with merging things back into develop caused this to be lost.

rob-c:
I was in favour of keeping this as a warning so that it's obvious to developers but on the debug due to not wanting to spam users with this.
If the warning is being dropped to a debug I'm happy to live with this I suppose, no point reverting the merge.

milliams:
@rob-c Since checking for `logger.isEnabledFor(logging.DEBUG)` doesn't and didn't work, do you have a suggestion for an appropriate way to check this? Perhaps this is another compelling case for `from __future__ import absolute_import`?

rob-c:
I'm happy to leave the merge in trunk as it is but I did resolve this by making use of absolute_import or an explicit path to the logging class I forget which exactly.
As it stands since this has gone in I'm happy just to filter the logging for these debug messages in future

milliams:
Ok, I'll leave it for now as it is. Also, sorry for not giving you more time to respond to the initial PR. I guessed this might have been related to code you'd worked on recently and so should have waited for your :+1:

rob-c:
It was just that I was half way through typing something as this got merged in front of me so the frustration was mainly that I'd just missed the window rather than towards the changes.

PR #107, opened 2016-01-13
Fix VPrinter's quoting so that it handles multiline strings in lists

milliams:
While `quoteValue()` handles multiline strings if they are by themselves, it fails if the string is inside a `list` or `GangaList`. This simplifies the logic of `quoteValue` by using `repr` for the string quoting behaviour and handing iterables manually.

This fixes the failure in Savannah13979.

### DIFF
--- a/python/Ganga/GPIDev/Base/VPrinter.py
+++ b/python/Ganga/GPIDev/Base/VPrinter.py
@@ -9,29 +9,20 @@ from Ganga.GPIDev.Base.Proxy import isProxy, isType, runProxyMethod, stripProxy
 
 from inspect import isclass
 
-def quoteValue(value, selection, indent=''):
+
+def quoteValue(value):
     """A quoting function. Used to get consistent formatting"""
-    # print "Quoting",repr(value),selection
-    if isType(value, type('')):
-        if selection == 'copyable':
-
-            value = value.replace('"', R'\"')
-            value = value.replace("'", R"\'")
-
-# DISABLED
-##             valueList = list( value )
-# for i in range( len( value ) ):
-##                 c = value[ i ]
-# if c in [ "'", '"' ]:
-##                     valueList[ i ] = "\\" + c
-##                     value = "".join( valueList )
-            if 1 + value.find("\n"):
-                # print 'Quote result',"'''" + value + "'''"
-                return "'''" + value + "'''"
-        # print 'Quote result',"'"+value+"'"
-        return indent + "'" + value + "'"
-    # print 'Quote result',value
-    return value
+    if isType(value, str):
+        # If it's a string then use `repr` for the quoting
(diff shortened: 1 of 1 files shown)


milliams:
Just to note that this is the last bug holding up the merging of the latest iteration of the new test system.

rob-c:
Looks good to me :)
I'm assuming that if a user had created an XML/export with the literal carriage return there is no way to recover this without manually editing their file?
(I'm not proposing that we do too much to auto-fix this problem from the past I'm just curious if I've understood it correctly)

milliams:
I don't believe that this affects the XML reading/writing as the change is only in `VPrinter` and not `VStreamer`. VPrinter is usually only used for printing information (it's more-or-less what's used when just typing `j` at the prompt). What `export` creates is a file like:

```python
#Ganga# File created by Ganga - Wed Jan 13 15:36:57 2016
#Ganga#
#Ganga# Object properties may be freely edited before reloading into Ganga
#Ganga#
#Ganga# Lines beginning #Ganga# are used to divide object definitions,
#Ganga# and must not be deleted

#Ganga# Job object (category: jobs)
 Job (
   comment = '',
   parallel_submit = False,
   do_auto_resubmit = False,
   name = '',
   inputfiles =[],
   backend =   Local (
     nice = 0
   ),
   application =   Executable (
     exe = 'echo',
     hash = None,
     env = [],
     args = ['a\n     b'],
     is_prepared = None
   ),
   inputdata =None,
   outputfiles =[],
   postprocessors =   MultiPostProcessor (
   ),
   info =   JobInfo (
     monitor =None
   ),
   splitter =None,
   inputsandbox =[]
 )
```
which, when `load`ing is just `eval()`ed.

If they had exported a job in the past few months with a literal carriage return in it then it will fail to load (this is true both before and after this PR). All we are changing here is the code for writing the file so that it is correctly `eval`able.

In an ideal world, the above printout (except the comments) is what should be printed by `repr(j)` as the rule for `repr` is that "this function makes an attempt to return a string that would yield an object with the same value when passed to eval()" [[1](https://docs.python.org/3/library/functions.html?highlight=repr#repr)]

rob-c:
Ah OK, I wasn't sure if writing a multi-line string into the XML can cause an XML corruption also or whether this has been investigated?

milliams:
It looks like the XML writer correctly writes the xml for multiline strings as:
```xml
<attribute name="args">
  <sequence>
    <value>'a\n     b'</value>
  </sequence>
</attribute>
```
so I think we're ok here. I checked this on both `develop` and this feature branch.

rob-c:
excellent :D I was just concerned that the user might have been able to destroy a job by modifying it's name or comments but that looks unlikely. OK, I'm happy

PR #108, opened 2016-01-13
Bugfix/fixing fqid bug/Subjobs not having Parent knowledge bug

rob-c:
Hi all,

After removing the extra load statements from the Registry commands I discovered an odd bug in the Jobs class which took a short while to solve.

When a Job containing subjobs was read from disk all of the information and parent knowledge was setup correctly for the objects created based upon the XML.

However, I discovered that once the subjobs are allocated within the Job class they have no immediate knowledge of their parent until the job is re-loaded from disk.

I've made a generic fix now such that when a subjob is added to job.subjobs the knowledge of the parent job is preserved. Hopefully this should prevent the strange oddities I was seeing from creeping up again.

This problem has only come to light now as the flushing is correctly delayed until the job is marked 'dirty enough'.

This primarily required some fixes to the Job and GangaList classes to maintain the parent knowledge.

Thanks,

Rob

### DIFF
--- a/python/Ganga/GPIDev/Lib/GangaList/GangaList.py
+++ b/python/Ganga/GPIDev/Lib/GangaList/GangaList.py
@@ -134,7 +134,11 @@ class GangaList(GangaObject):
         if name == "_list":
             if self.is_list(value):
                 if self.has_proxy_element(value):
-                    return [stripProxy(element) for element in value]
+                    returnable_list = [stripProxy(element) for element in value]
+                    for elem in returnable_list:
+                        if isType(elem, GangaObject):
+                            elem._setParent(self._getParent())
+                    return returnable_list
                 else:
                     return value
             elif self._list is None:
@@ -155,6 +159,15 @@ class GangaList(GangaObject):
             self._list = new_list
             self._is_a_ref = False
         return self
+
+    def _getParent(self):
+        return super(GangaList, self)._getParent()
+
+    def _setParent(self, parent):
+        super(GangaList, self)._setParent(parent)
+        for element in self._list:
+            if isType(element, GangaObject):
+                stripProxy(element)._setParent(parent)
 
     def get(self, to_match):
         def matching_filter(item):
@@ -421,7 +434,10 @@ class GangaList(GangaObject):
         return self.toString()
 
     def append(self, obj, my_filter=True):
-        self._list.append(self.strip_proxy(obj, my_filter))
+        elem = self.strip_proxy(obj, my_filter)
+        if isType(elem, GangaObject):
+            stripProxy(elem)._setParent(stripProxy(self)._getParent())
+        self._list.append(elem)
 
     def _export_append(self, obj):
         self.checkReadOnly()
@@ -442,6 +458,8 @@ class GangaList(GangaObject):
         return self._list.index(self.strip_proxy(obj))
 
     def insert(self, index, obj):
+        if isType(obj, GangaObject):
+            stripProxy(obj)._setParent(stripProxy(self)._getParent())
         self._list.insert(index, self.strip_proxy(obj, True))
 
     def _export_insert(self, index, obj):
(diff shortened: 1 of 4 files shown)


python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -156,6 +160,15 @@ def _on_attribute__set__(self, obj_type, attrib_name):
             self._is_a_ref = False
         return self
 
+    def _getParent(self):
```
milliams: Do we need this overload if it's only calling the base class version?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -993,10 +1001,24 @@ def getFQID(self, sep=None):
...
             return None
-        cur = stripProxy(self)._getParent()  # FIXME: or use master attribute?
-        while cur:
+
+        try:
+            if stripProxy(self).master is not None:
+                cur = stripProxy(self).master  # FIXME: or use master attribute?
+            else:
+                cur = None
+        except AssertionError:
```
milliams: Is this really raising an `AssertionError`? Do you know the origin of the assertion?

rob-c: Ah, this was from one iteration of the code which called getJobObject which will throw an assertionError when there is no job associated with it. I'll fix it now

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: I'm planning to investigate in future if I can let subjobs have a GangaList as a parent which would allow for something like GangaLists of Jobs to be stored in Box repo (currently not possible).

In this case I will likely be wanting to tie into the getParent/setParent functions at least during development. It's not required to keep it here, but I'm playing around with this interface and it saves me the time of adding it back in, in the future just to test that it's working with a debug statement.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -421,7 +434,10 @@ def __str__(self):
         return self.toString()
 
     def append(self, obj, my_filter=True):
-        self._list.append(self.strip_proxy(obj, my_filter))
+        elem = self.strip_proxy(obj, my_filter)
+        if isType(elem, GangaObject):
+            stripProxy(elem)._setParent(stripProxy(self)._getParent())
```
milliams: It should never be required to write `stripProxy(self)` as we are inside a class which is non a proxy class. If you find a situation where `self` is a proxy object then that is a major thing that needs fixing.

milliams: Also, just so I understand what's happening here; when an item is added to a `GangaList` its parent is set to be same as the parent of the `GangaList`? So in the case of a `Job` with subjobs, any subjobs have their parent set to be the job and not the `GangaList` itself? I think that makes sense, I just wanted to check.

rob-c: Yes, you've understood correctly.

I'm not a fan of this but this is the 'old style' behaviour and a lot of code is written to just naively expect that the ```_getParent``` method from a subjob will return the parent job.
I've modified the master attribute in the Job class to be more intelligent in future and I'm checking for any outstanding ```_getParent``` calls which I can replace with either ```job.master``` or ```job.getJobObject``` which would allow me to properly define the parent of elements in a GangaList as the GangaList itself.

I would rather do this work on a separate pull request however as I see it quickly growing as a task and at the moment I'm seeing problems with submitting simple jobs which is all I want to address here.

milliams: That makes sense. Thanks for the clarification.

milliams:
:+1: (that means I'm happy for you to merge :smile:)

PR #112, opened 2016-01-15
Bugfix/lazy loading improvements

rob-c:
This branch includes some helper functions which have been implemented in Job.py
These are called:

```
lazyLoadingJobApplication
lazyLoadingJobBackend
lazyLoadingJobStatus
```

These functions will return an object which is of the same type as that within the job which is being inspected.
This object is not guaranteed to have the same attributes (where applicable) as the Job object of interest. In order to get access to this info calling ```job.application``` or ```job.backend``` explicitly is required.

These functions greatly simplify the task of requesting an object without triggering the full loading of a job because the information of interest is in the node ```_index_cache```.

I've also put in a lot of modifications into the SubJobXMLList class and RegistrySlice such that calling:
```
jobs(x).subjobs
```
Doesn't result in loading a job from disk as the information is available in the job index.

Additionally I've added some minor changes to the code in the IBackend which delays the loading of a subjob until just before the job is requesting to be monitored which speeds up the monitoring loop as it avoids loading a large job with many subjobs before monitoring potentially a small number of the subjobs.

Thanks,

Rob

### DIFF
--- a/python/Ganga/Core/GangaRepository/SubJobXMLList.py
+++ b/python/Ganga/Core/GangaRepository/SubJobXMLList.py
@@ -116,7 +116,7 @@ class SubJobXMLList(GangaObject):
 
         raw_self = stripProxy(self)
 
-        import os.path
+        import os
         index_file = os.path.join(raw_self._jobDirectory, raw_self._subjob_master_index_name )
         if os.path.isfile( index_file ):
             index_file_obj = None
@@ -131,18 +131,23 @@ class SubJobXMLList(GangaObject):
                 if raw_self._subjobIndexData is None:
                     raw_self._subjobIndexData = {}
                 else:
-                    for subjob in raw_self._subjobIndexData.keys():
-                        index_data = raw_self._subjobIndexData.get(subjob)
+                    for subjob_id in raw_self._subjobIndexData.keys():
+                        index_data = raw_self._subjobIndexData.get(subjob_id)
                         if index_data is not None and 'modified' in index_data:
                             mod_time = index_data['modified']
-                            disk_location = raw_self.__get_dataFile(str(subjob))
-                            import os
+                            disk_location = raw_self.__get_dataFile(str(subjob_id))
                             disk_time = os.stat(disk_location).st_ctime
                             if mod_time != disk_time:
-                                raw_self._subjobIndexData = {}
+                                logger.warning("SubJob: %s has been modified, re-loading" % (subjob_id))
+                                new_data = raw_self._registry.getIndexCache( stripProxy(raw_self.__getitem__(i)) )
+                                raw_self._subjobIndexData[subjob_id] = new_data
                                 break
                         else:
-                            raw_self._subjobIndexData = {}
+                            logger.warning("Cannot find subjob index %s, rebuilding" % str(subjob_id))
+                            new_data = raw_self._registry.getIndexCache( stripProxy(raw_self.__getitem__(i)) )
+                            raw_self._subjobIndexData[subjob_id] = new_data
+                            continue
+                            #raw_self._subjobIndexData = {}
             except Exception, err:
                 logger.error( "Subjob Index file open, error: %s" % str(err) )
                 raw_self._subjobIndexData = {}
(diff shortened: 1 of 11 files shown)


python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -164,15 +169,24 @@ def __really_writeIndex(self):
...
+        import os
+
         all_caches = {}
-        for i in range(len(raw_self)):
-            this_cache = raw_self._registry.getIndexCache( stripProxy(raw_self.__getitem__(i)) )
-            all_caches[i] = this_cache
-            disk_location = raw_self.__get_dataFile(i)
-            import os
-            all_caches[i]['modified'] = os.stat(disk_location).st_ctime
+        for sj_id in range(len(raw_self)):
```
milliams: This should really be:
```python
for sj_id, sj in enumerate(raw_self):
    ...
```
and then any `raw_self.__getitem__(sj_id)` could become just `sj`?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -334,7 +350,16 @@ def _setParent(self, parentObj):
...
+        raw_self = stripProxy(self)
+        if index > len(raw_self) or index < 0:
+            return None
+
+        if index in self._subjobIndexData.keys():
+            return self._subjobIndexData[index]
+        else:
+            return raw_self._registry.getIndexCache( stripProxy(raw_self.__getitem__(i)) )
+
+        return None
```
milliams: This last bit should be:
```python
if index in self._subjobIndexData.keys():
    return self._subjobIndexData[index]

return raw_self._registry.getIndexCache( stripProxy(raw_self.__getitem__(i)) )
```
and I guess that `i` on that last line should be `index`?

Also, is there a reason for using `__getitem__` rather than `raw_self[index]`?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -116,7 +116,7 @@ def load_subJobIndex(self):
 
         raw_self = stripProxy(self)
```
milliams: As I said before, I don't think we should ever need to do `stripProxy(self)`.

milliams:
Overall I'm happy with this pull request, at least the purpose behind it. It makes sense to hide this complex stuff away behind some functions.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Unfortunately this is done in many, many places (for better or worse).

I'm hoping to address it by removing/changing this everywhere with another branch dedicated to changing this as when I replaced ```stripProxy(self)``` with ```self``` with a macro I saw Ganga have a heart attack due to a few different things getting broken.

milliams: Makes sense. This wasn't introduced in this PR so perhaps shouldn't be fixed in it.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: the use of __getitem__ was only because the class implements it and I was wanting to be explicit.

Yes however, that should be ```index``` rather than ```i``` on 360

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: If you're sure that the enumerate function only calls ```__len__``` and only explicitly calls ```__getitem__``` when required then I'm fine with that but I'd need to check as the ```__getitem__`` loads the subjob and hence if it's called rather than ```__len__``` it will break what the code is trying to do.

milliams: No, I imagine that it does indeed call `__getitem__`. I would suggest at least a comment on this bit of code to explain the unusual idiom.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
milliams: I would reserve directly calling dunder methods like `__getitem__` for those times where it really matters (i.e. when implementing them to avoid recursion etc.). Other than that, doing `self[index]` is much clearer.

PR #114, opened 2016-01-15
Add a setup.py

milliams:
This allows building of packages, uploading to PyPI and installing from there too. You can see an example PyPI at https://testpypi.python.org/pypi/ganga (ignore the version number). To test installing from this on lxplus for example you can do:

```bash
scl enable python27 -- bash
```
to start a bash session with Python 2.7 available and then run:
```bash
virtualenv ganga_venv
. ganga_venv/bin/activate
pip install ipython httplib2 python-gflags google-api-python-client stomp.py
pip install -i https://testpypi.python.org/pypi ganga
```
to install the package and:
```bash
ganga
```
to run it.

Once this is finalised, it will be possible to install with just:
```bash
virtualenv ganga_venv
. ganga_venv/bin/activate
pip install ganga
```
This also adds a README to the repo which is rendered correctly by GitHub.


drbenmorgan:
Can confirm this also works on Mac using Homebrew and the brewed python install (without virtualenv):

```
brew install python
pip install ipython httplib2 python-gflags google-api-python-client stomp.py
pip install -i https://testpypi.python.org/pypi ganga
ganga
```

Only thing I see when running `ganga` is a warning:

```
...
Ganga.Lib.LCG                      : WARNING  LCG-GLITE UI has not been configured. The plugin has been disabled.
<brewroot>/lib/python2.7/site-packages/IPython/config.py:13: ShimWarning: The `IPython.config` package has been deprecated. You should import from traitlets.config instead.
  "You should import from traitlets.config instead.", ShimWarning)
```

Not sure how significant that is, as I certainly get the prompt up and can submit a default job.

milliams:
@drbenmorgan Great. The "LCG-GLITE UI" warning is because it is looking for the LCG-GLITE UI by default (I assume you see this warning anyway). The second warning about IPython is because the codebase is currently written against IPython 1.2.1 but pip by default will download the latest which is IPython 4 which has deprecated a few functions.

Either I could require that pip only ever installs IPython 1.2.1 or we could see about not using `IPython.config` in Ganga somehow. As soon as we hard-require Python 2.7 then we can also require IPython 4 at which time this will be a solved problem.

drbenmorgan:
@milliams yep, the LCG-GLITE warnings also occur for non-pip Ganga, so not an issue.

If the `IPython.config` warning does not imply problems will occur elsewhere, it's not a major issue for me, especially if it'll go away in the not too distant future with the Python 2.7 minimum requirement.

Oh, I did also try

```
ipython 
...
In [0]: import Ganga
```
 and that works too, though I din't go any further than that.

milliams:
Unless anyone has any complaints, I'd like to merge this today. It doesn't touch any code so there's very little risk of it breaking anything.

alexanderrichards:
Sounds good to me

rob-c:
gets my vote :)

drmarkwslater:
@drbenmorgan Just to say, though `import Ganga` will work, it will do very little :) have a look at this branch:

https://github.com/ganga-devs/ganga/tree/little-ganga-import 

to see how far I've got :)

PR #117, opened 2016-01-17
Bugfixes/minor core bugfixes

rob-c:
The changes in this branch can be summed up into a few different cleanup tasks:

1. Remove excess code which uses ```_setParent``` in too many places.
2. Add ```stripProxy``` in places where proxied objects are passed but the proxy isn't stripped where needed.
3. Remove calls to ```__new__```.
 This unfortunately has exposed a LOT of problems. However this function has been (mostly) mis-used from what I can tell so now that I'm constructing jobs using the standard call to the class object i.e. ```j= Job()``` vs ```j=JobClass.__new__(); j.__init__()```. After doing this I see some logging errors go away and some other minor errors become exposed. The errors exposed so far have all been real errors, such as calling functions with wrong numbers of parameters. This may cause people to have a bumpy experience once this PR gets merged (hence why I'm waiting until Monday) But I'm keen to merge this ASAP!
4. Minor fixes causing stack traces to be thrown in normal use from basic testing.

I'm keen to get this in ASAP and have already merged this into my personal develop. I'm happy to fixup any problems in this PR but as for any problems that this exposes elsewhere I'm keen to stress that it's likely they should be fixed elsewhere.

Thanks,

Rob


rob-c:
I've also changed some logging startup code slightly to not throw errors for LHCb users on startup.

In addition I've managed to fix a repository sync bug which caused 'completed' jobs to not be correctly marked as 'completing still'

This branch also preserves the logging right until the end of the shutdown which is expected by design but has been highly temperamental until now.

This can cause proxied objects to be passed where they were previously not and can throw AttributeErrors but this can be fixed with appropriate stripProxy calls in the correct places.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -29,6 +29,18 @@
...
+do_not_copy = ['_proxyObject', '_proxyClass', '_data', '_index_cache', '_parent', '_registry', '_impl']
+
+def _getGangaList():
+    global _imported_GangaList
+    if _imported_GangaList is None:
+        from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
+        _imported_GangaList = GangaList
+    return _imported_GangaList
+
+
```
milliams: I really, really don't like that we're referring explicitly to `GangaList` (a subclass of `GangaObject`) inside the code for `GangaObject`. I see that this code above is there to avoid a circular import but this should really be fixed properly.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -928,43 +934,59 @@ def __setstate__(self, this_dict):
...
-                        from Ganga.Core.GangaRepository import getRegistry
-                        from Ganga.GPIDev.Base.Proxy import GPIProxyObjectFactory
-                        shareref = GPIProxyObjectFactory(getRegistry("prep").getShareRef())
+        global do_not_copy
+        for k, v in raw_self.__dict__.iteritems():
+            if k not in do_not_copy:
+                try:
+                    setattr(self_copy, k, deepcopy(v))
+                except Exception:
+                    setatttr(self_copy, k, v)
```
milliams: This should be catching the correct exception, not the general one. If this is there to catch problems in `deepcopy` then something like:
```python
try:
  v = deepcopy(v)
except Exception:  # deepcopy can raise almost anything so catch them all.
  pass
setatttr(self_copy, k, v)
```

However, catching `Exception` is really dangerous as it also catches `NameError` etc. which just hide bugs in our code.

python/Ganga/GPIDev/Base/Objects.py
rob-c: Yes this is what's going on.
If not this way can you suggest a better alternative to avoid the circular importing?
I'm all for a better solution as I'm constantly running into this in Python
I have tried placing the equivalent code in places where GangaList is called but end up with some of the code being called in the case of constructing GangaList so ended up duplicating this several times (hence the helper function)

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -924,6 +936,15 @@ def _getattribute(self, name):
 
     return type(name, (GPIProxyObject,), d)
 
+_metaClassHolder = None
+def _getMetaClass():
+    global _metaClassHolder
+    if _metaClassHolder is None:
+        from Ganga.GPIDev.Base.Objects import ObjectMetaclass
+        _metaClassHolder = ObjectMetaclass
+    return _metaClassHolder
```
milliams: What is the purpose of this function?

python/Ganga/GPIDev/Base/Objects.py
rob-c: Tbh I'm happy to remove the try/except here and have other classes deal with exceptions as this shouldn't be failing here (and often doesn't atm)

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Returns the ObjectMetaClass for a comparison (I don't know why the comparison is made I just do what the code used to do)
I had a problem just directly importing ObjectMetaClass in the past, plus wanted to avoid excessive calls to import unless this is safe to do?

python/Ganga/Utility/util.py
```diff
@@ -148,7 +148,10 @@ def execute_once():
 
 def hostname():
     """ Try to get the hostname in the most possible reliable way as described in the Python LibRef."""
-    import socket
+    try:
+        import socket
+    except ImportError:
+        return 'localhost'
```
milliams: Does this really ever fail? It's a built-in core Python module.

rob-c:
Rats, haven't quite fixed the repo syncing problem. Thought I had but not quite yet. Looks like the objects don't sync to disk after completing. I'll look into marking them dirty after my teaching today (after 12am or so)

python/Ganga/Utility/util.py
rob-c: oops, this was due to another bug i introduced, this can be removed

python/Ganga/GPIDev/Base/Objects.py
milliams: Really, all `GangaList` should be used for is to provide a proxy wrapper for a standard Python list. It's there so that we can store a list of raw objects behind the scenes and then when an object is requested, it will be proxy-wrapped automatically. Likewise, adding a proxied object should strip the proxy on its way in so the raw object is stored. This is needed because otherwise there's no way to provide a proxy interface which takes or returns a list. In principle `GangaObjects` should be storing internal things as python lists and not `GangaList`s.

Now, over the years this has been blurred and `GangaList` has been added all over the code where it is not needed and I'm trying to avoid making this worse.

python/Ganga/GPIDev/Base/Proxy.py
milliams: The problem here is perhaps coming from the import hierarchy of Proxy.py and Objects.py being backwards. Really, Objects.py should have no knowledge of proxies or anything to do with it, at all! `GangaObject`s should be able to operate in complete isolation of the proxy system. Proxy.py, however should absolutely know about `GangaObject`s since that is what they are wrapping and so Proxy.py should import Objects and not vice versa.

python/Ganga/GPIDev/Base/Objects.py
rob-c: given the class was here well before this commit I'm in favour of keeping it for now

rob-c:
@milliams Are the integration tests running as expected? I only ask as I don't know if the firewall is effecting the warning I'm seeing here?
(I've not looked through a socks proxy yet I'm doing a few other things)

I'm still a few commits behind develop and need to do some work/testing on this branch but I'm just curious what the status of the automated testing is?

python/Ganga/GPIDev/Base/Objects.py
rob-c: (aside)
As well as this, the ```GangaList``` does modify a certain amount of ```Node``` and ```GangaObject``` functions which it would be ideal to have merged into the Objects class to some extent (although writing code to deeply inspect lists can be a pain)

milliams:
@rob-c The automated testing is indeed working correctly as far as I can tell. Jenkins is still behind the CERN firewall so to open the link above (https://ganga-ci.cern.ch/jenkins/job/PR%20builder/18/) you'll need to proxy via lxplus or similar.

To save you the trouble, all the above tests failed because of:
```
 File "/usr/lib64/python2.6/unittest.py", line 269, in run
    self.setUp()
  File "/var/lib/jenkins/workspace/PR builder/python/Ganga/new_tests/GPI/GangaUnitTest.py", line 163, in setUp
    start_ganga()
  File "/var/lib/jenkins/workspace/PR builder/python/Ganga/new_tests/GPI/GangaUnitTest.py", line 72, in start_ganga
    Ganga.Runtime._prog.configure()
  File "/var/lib/jenkins/workspace/PR builder/python/Ganga/Runtime/bootstrap.py", line 574, in configure
    logger.debug("Config File Exception: %s" % str(x))
local variable 'logger' referenced before assignment
```
so it's a simple variable renaming bug.

Also, to run the tests now, you can do, from the root `ganga` directory:
```bash
python setup.py tests --type=all
```
where `--type` can be `unit`, `integration` or `all` and default to `unit`

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Tbh I'm all for moving the Proxy Construction entirely away from the Object class. This is a relatively minor task (albeit a very delicate one) which I think can be accomplished in a separate PR.

I think having a pure GangaObject without knowledge of ```_proxyClass```, ```_proxyObject``` or ```_impl``` references would be nice. I don't think this should be too difficult to remove them.
Then this would require the ```addProxy```( ```GPIProxyObjectFactory``` ) to actually construct new Proxy objects on the fly (not a task too difficult).

The problem only then becomes calling functions such as ```stripProxy```, ```getName``` and ```isType``` which should be minimised within the Proxy class anyway.

rob-c:
Alas CERN (or rather the LHCb python install strikes again):
```
  File "setup.py", line 1, in <module>
    from setuptools import setup
ImportError: No module named setuptools
```

OK, I'll fix the bug, strange how it doesn't crop up in the testing I've run with the old integration suite. Always good to have more tests :)

python/Ganga/GPIDev/Base/Proxy.py
milliams: That sounds like a good idea. If the proxy layer is doing its job properly (which we both know it doesn't always) then we should be able to remove all usage of `stripProxy` outside of Proxy.py (and maybe one or two other proxy-layer classes). Simplifying this logic inside `GangaObject` will only help us in the long-run as we look towards replacing it entirely.

milliams:
A `pip install setuptools --user` should do the trick. Otherwise, I've been doing everything inside a virtualenv which contains the extra packages to one place.

rob-c:
Given pip and python point to different places when trying to run in the LHCb env I'll try and do a manual install rather than use the automatic tools, but thanks for the suggestion ;)

rob-c:
@milliams All I get is:
```
9rcurrie@lxplus0100:->python setup.py tests --type=all
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: invalid command 'tests'
```

milliams:
@rob-c Have you rebased against the latest `develop`? I've added some things to setup.py in the last day to provide this. The changes only affected the tests and setup.py so the merge should be absolutely clean.

rob-c:
I haven't done so yet as this has several painful known conflicts I was hoping to avoid resolving without some tests to know if I've trashed part of the Core.
I'll merge the code by hand in this case and spend the time checking on the differences.

rob-c:
@milliams What's the status of the jenkins machine getting out from behind the firewall, I'm having issues accessing the test reports from the server.

milliams:
@rob-c The request has been put in and it's just waiting on CERN IT.

rob-c:
great, good news! I'm re-running the tests locally after torching my broken site-packages and starting afresh, I think my results are agreeing with the test server so I'll just go about working out what needs fixing.

rob-c:
Whilst I'm doing some cleanup and fixing minor bugs which I've introduced in this branch.

I've been able to remove most (all I think) references to stripProxy in this branch from Objects.py
I've put some fixes into the Proxy.py and it should now be doing a better job of stripping objects before they're passed into any GangaObject method and should be adding a proxy to all classes which derive from GangaObject when a result is returned from a getter function.

i.e.

```j._impl.application``` is a proper application whilst ```j.application``` will return a proxied application object.

This cleans up the problems of stripProxy having to be called in numerous places.
We can clean up the overuse of stripProxy in a future PR.
This should also mean that isinstance is quite a bit safer to use in a lot of places as it's rarely used on a proxy object at runtime.

I have also added a helper function such that:
```j._impl._getRegistry().has_loaded(j._impl)```
will return a boolean of whether the master job has indeed been loaded from disk:

i.e.
from a clean session with no monitoring:
```
j=jobs(13)
j._impl._getRegistry().has_loaded(j._impl)
```
will return False if job#13 hasn't previously been loaded (i.e. for inspection of the member variables)
and will return True once the job has been loaded for some reason.

This work was started as a combination of a minor code cleanup and light optimisation exercise but this exposed a few small oddities in the codebase which I've had a stab at fixing.

**Comment by rob-c** [comment]
This branch also appears to solve the repo syncing problems that people were earlier reporting :)

**Comment by rob-c** [comment]
If I'm correct this branch should now be passing all of the sensible unittests and should fix some outstanding issues which we don't currently have tests for.
I'm planning to merge this once the integration tests confirm that there are no problems unless there are any objections?

**Comment by milliams** [comment]
@rob-c I'd like to take one last look through the changes before it's merged. I'll do so first thing tomorrow morning and then we should be able to merge it.

File: python/Ganga/Core/GangaRepository/Registry.py [hunk 5]
```diff
@@ -1166,3 +1172,15 @@ def print_other_sessions(self):
         if len(other_sessions) > 0:
             logger.warning("%i other concurrent sessions:\n * %s" % (len(other_sessions), "\n * ".join(other_sessions)))
 
+    def has_loaded(self, obj):
+
+        try:
+            index = self.find(obj)
+        except:
+            return False
```
milliams: This exception handler should be more precise. I'm guessing it should be `ObjectNotInRegistryError`. Also, the function could do with a doc string.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -57,6 +57,8 @@ def __init__(self, jobDirectory='', registry=None, dataFileName='data', load_bac
             return
 
         self._subjobIndexData = {}
+        if parent is not None:
+            self._setParent(parent)
```
milliams: Just to be sure, this is only changing the parent if it already has one and is leaving it as `None` if it was before?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -695,7 +709,7 @@ def export(method):
     return method
 
 
-class ObjectMetaclass(type):
+class ObjectMetaclass(type):#, object):
```
milliams: I'm guessing this is just a mistake that it was left in?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -783,7 +797,7 @@ def __init__(cls, name, bases, this_dict):
         this_schema._pluginclass = cls
 
         # store generated proxy class
-        cls._proxyClass = proxyClass
+        setattr(cls, '_proxyClass', proxyClass)
```
milliams: What is this change fixing? The commit message was not clear.

python/Ganga/GPIDev/Base/Objects.py
milliams: This still needs fixing. In fact, I know the exception is never being raised since `setatttr` was misspelled :smile:

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -494,11 +503,13 @@ def __set__(self, obj, val):
...
         # apply attribute filter to component items
-        if item.isA(Schema.ComponentItem):
+        if item.isA(ComponentItem):
             val = self._stripAttribute(obj, val, getName(self))
 
         self._check_type(obj, val)
-        setattr(stripProxy(obj), getName(self), val)
+        
+        GangaObject = _getGangaObject()
+        GangaObject.__setattr__(stripProxy(obj), getName(self), val)
```
milliams: Why is the normal method lookup being subverted here? I don't think any `GangaObject` subclasses or instances should be overriding `__setattr__`.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -665,11 +647,14 @@ def updateStatus(self, newstatus, transition_update=True, update_master=True, ig
...
             self.status = newstatus
             self._commit()
+            logger.debug("Status changed from '%s' to '%s'" % (saved_status, self.status))
 
         except Exception as x:
             self.status = saved_status
             log_user_exception()
             raise JobStatusError(x)
+        finally:
+            pass
```
milliams: This `finally: pass` can probably be removed.

python/Ganga/GPIDev/Lib/Job/JobTime.py
```diff
@@ -66,6 +66,7 @@ class JobTime(GangaObject):
     _category = 'jobtime'
     _name = 'JobTime'
     _exportmethods = ['display',
+                      '_display',
```
milliams: Does `_display` need to be exported? I don't think we want end-users calling the function.

python/Ganga/Utility/logging/__init__.py
```diff
@@ -137,36 +176,67 @@ def _make_file_handler(logfile, logfile_size):
...
+        logger.error('Attribute Error: %s', str(err))
+        logger.warning('possible configuration error: invalid level value (%s), using default level', value)
+        return None
+
 
 # reflect all user changes immediately
 def post_config_handler(opt, value):
-
-    if '_customFormat' in config and config['_customFormat'] != "":
+    global config, _formats
```
milliams: Do these need to `global`. If we're not rebinding them then the name lookup should find the correct objects anyway?

python/Ganga/Utility/util.py
milliams: This still needs removing.

python/Ganga/new_tests/GPI/Bugs/Savannah10064.py
```diff
@@ -5,7 +5,7 @@
 
 class Savannah10064(GangaUnitTest):
 
-    def Savannah10064(self):
+    def test_Savannah10064(self):
```
milliams: We don't need to rename all these functions. The test runner is configured to find tests beginning with `Savannah` and `JIRA` so I'd rather keep these as they are.

python/Ganga/new_tests/GPI/Bugs/Savannah18729.py
```diff
@@ -13,7 +13,9 @@ def Savannah18729(self):
 
         tmpdir = tempfile.mktemp()
         os.mkdir(tmpdir)
-        self.fname = os.path.join(tmpdir, ' test.C')
+        ## Is this a test of files with a leading ' '  in the name? - rcurrie
+        #self.fname = os.path.join(tmpdir, ' test.C')
+        self.fname = os.path.join(tmpdir, 'test.C')
```
milliams: You're right; this was just a typo on my part. For the record, this is JIRA bug GANGA-343

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -125,9 +125,15 @@ def stop_ganga():
...
+            try:
+                j.remove()
+            except:
+                pass
         for t in templates:
-            t.remove()
+            try:
+                t.remove()
+            except:
+                pass
```
milliams: Are these exception handlers needed? I haven't seen problems myself. Otherwise they should use a more specific handler.

setup.py
```diff
@@ -6,7 +6,13 @@
...
 
 def readme():
-    with open('README.rst') as f:
+    import os.path
+    filename = 'README.rst'
+    if not os.path.exists(filename):
+        import inspect
+        script_path = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+        filename = os.path.join(script_path, filename)
+    with open(filename) as f:
```
milliams: By convention `python setup.py` is _always_ run from the directory it is in and so this is not needed. Many other things will break if this is not the case anyway.

milliams:
Apart from my comments above (and the one failing test which I will check tomorrow) I'm overall happy with the changes here. I'm very glad to see the large reduction in `stripProxy` calls and to see you've got the proxy layering under more control :smile: Great work.

However, alongside the intended changes here there's a lot of changes of `_name`→`getName` and various logging things. It made it a little hard to see the purpose of the PR at times and probably made the diff twice as long as it needed to be and so in future it would be good to separate those out into a separate PR. I understand that will slow things down but it will increase the flow of these things by keeping them as streamlined as possible.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -120,7 +124,10 @@ def simpleAttribute(self, node, name, value, sequence):
             # else:
             #    print 'no transformation'
             if isclass(value):
-                print(self.indent(), name, '=', repr(value), end = '', file=self.out)
+                if self._interactive is True:
+                    print(self.indent(), name, '=', str(value), end = '', file=self.out)
+                else:
+                    print(self.indent(), name, '=', repr(value), end = '', file=self.out)
```
milliams: As far as I can tell, this is reason for all those `interactive` flags being passed around that you've added. What is the reason for the difference and what does 'interactive' mean in this context. This would at least need a docstring explaining it but this change overall should probably have been in a separate PR so we could discuss it separately.

rob-c: Changing the behaviour of the jobtree etc to use a repr instead of a str changes the display behaviour of the job.
Before the ```repr``` function was used jobtree and subjobs would display correctly in the summary of a job. Now they display a really long line with formatting and ```\n``` characters and such. Using a string method to parse this gets back to the original intended behaviour.
Interactive matches the interactive mode of being on the IPython interactive prompt. We expect subtly different behaviour of some functions when we're using this for displaying the output. (Sure we could do some introspection and change the behaviour dynamically right at the repr/string level but this becomes overkill imo)
The default behaviour hasn't changed unless the VPrinter is called with pretty_repr for storing files to disk, in which case it's a simple matter to expand the API there to accept 1 additional boolean argument

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: In principle yes, however,
Do we care if the pull-down method in the test is failing as it shouldn't cause the test to fail should it?
Besides all we're doing is effectively an ```rm -fr ~/gangadir_testing``` so I don't think that it overly matters if the teardown fails due to something going wrong here. We could work out what the specific error type should be and wrap job.remove but I would say this only becomes important when testing the job.remove and we don't want to confuse test errors due to unrelated code failing imo.

python/Ganga/new_tests/GPI/Bugs/Savannah10064.py
rob-c: I wanted to rename those functions to be able to keep doing:
```nosetest Ganga/.../.../myTestScript.py```

If we want to drop this behaviour then I'm happy to revert this but it made life easier by being able to re-run a lightweight test without having to perform a complex setup or make a call to nosetests with too many arguments. I'm actually quite keen to fight to keep this style of workflow.

python/Ganga/GPIDev/Lib/Job/JobTime.py
rob-c: This can probably be reverted but was made to cope with a problem in the proxy

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: This was added in debugging and makes life easier when I need to comment out the Exception handling to get a stack trace.
I'm happy to revert but this is effectively used as a debug handle so I'll just comment it out

python/Ganga/GPIDev/Base/Proxy.py
rob-c: In principle we're not overloading it atm but I don't think we lose anything by using the methods provided by ```GangaObject``` (which are the same as ```object``` for now at least). The redirection is minor but it could allow for us to use this for debugging or other uses in the future imo.

python/Ganga/GPIDev/Base/Objects.py
rob-c: I was toying with something at the time and this can be removed to avoid un-needed conversations about how this doesn't work.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: The ```SubJobXMLList``` class doesn't always have it's parent set before it's fully constructed and used.
This class was written under the 'old-style' way of working so I'm planning to try and fix it up now that Core handles the ```_parent``` object and other functions in a much nicer/consistent way.

rob-c: Edit: I think this check is no longer needed tbh, but I could be wrong. Either way I plan to look into this class in the future for debugging whether subjobs are fully loaded or not.

python/Ganga/GPIDev/Base/Objects.py
rob-c: This was just a minor change which I made to highlight that this probably shouldn't be in Objects.py in the long-run.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: Lots of functions can use doc strings, I'll get around to this as a general point in the near future.
The error here needs to be better defined in the ```find``` method I suspect. I'll check there and update this.

rob-c:
This PR ballooned in order to NOT cause other tests to fail by the inclusion of this PR. There were things broken in Core that this was intending to address and this exposed other stuff.

Originally this was a PR for just removing excess use of ```_name```, ```__name__``` and adding relevant ```stripProxy``` and removing ```__new__```. The former change to use ```getName``` should and was mostly house-keeping but removing ```__new__``` in a few places exposed some really, really old horrible bugs which had very wide reaching repercussion.
The changes to the logging were made because I simply wasn't seeing logging messages when I knew they should be being displayed. I needed to fix this in order to debug a few problems. It would have been nice to do in a separate PR and commit this half way through the development but it was put in this initially as I didn't realise fully the scope of what fixing a minor bug in the Proxy would cause.

Removing the ```__new__``` calls in a lot of places caused full classes to be initialized where previously they had not severely changing the code paths which are called.
Unfortunately adding more and more ```stripProxy``` to cope with the fact that the proxy layer had some horrible bugs was like trying to resurrect the titanic. This introduced yet more and more bugs which all needed to be addressed before this PR was in a working state.
Eventually I ended up having to fix the problems in the Proxy itself which raised a few subtly different bugs but made life easier.

Yes the work on the ```interactive``` display strings could have gone in a separate PR but frankly we broke expected behaviour so I'm not too worried about the fact that this will likely _need_ to be reverted.

rob-c:
Making some small changes now and after the meeting and plan to merge once a few of the one-liners are in.

milliams:
Also, regarding the one failing test: I'm pretty sure it's the test that is wrong. It was a mistake caused by me copying the test incorrectly when porting them. `Savannah87262` should indeed test for `self.assertEqual(s[e.is_prepared.name], 1)` so feel free to make that change in this branch.

python/Ganga/Utility/logging/__init__.py
rob-c: You're probably right here so I'll remove the global

rob-c:
Strange how I can't make it fail for me however in private testing, I'll do some digging see if I can work out why it's not failing in my tests

python/Ganga/new_tests/GPI/Bugs/Savannah10064.py
milliams: Upon more though I'm happy to keep it how you've changed it. You're right that it's nice to be able to call `nosetests` simply without any complicated `--testmatch` setup.

rob-c:
I'm currently debugging a small problem I've seen highlighted from my test jobs which I submit for LHCb,
Once I've fixed these small bugs I'll commit the changes which include the fixes mentioned above.

rob-c:
OK. It turns out I was doing something stupid in the new offline Dirac splitter which led to problems when submitting 'full' LHCb jobs in my private testing.
Eventually I'll get around to integrating some of my private testing scripts into the new framework now it's up and running :)

At this point it's possible that there are some areas of bad code in Atlas/LHCb/other-plugins which are bypassing the proxy when setting attributes or constructing proxy objects 'behind the scenes' and allocating them internally to somewhere where there should be no proxy.
These areas of bad code should themselves be fixed as the Core now treats proxied objects in a consistent way.

I've not removed any ```stripProxy``` checks anywhere other than Objects.py for now but as long as we're careful we can start to remove the over-reliance of this function to fix problems with incorrect proxy objects.
The functions which can lead to problems are: ```addProxy```, ```GPIProxyClassFactory``` and ```GPIProxyClassFactory``` or anything obtained from within ```Ganga.GPI```. If these classes are passing an unwanted proxy somewhere it can be stripped by using ```stripProxy```.

I've fixed all of the integration test problems (and the new test system feels appreciably must faster than using the heavy-handed ```ganga --test``` framework. @milliams Great stuff!

rob-c:
Unless there are any more comments I plan to merge this branch today and then work through my outstanding pull requests and do some other minor house-keeping

setup.py
milliams: I still think this is not needed but it could be much simplified using `os.path.realpath(__file__)`.

milliams:
:+1:

Overall I'm really happy with these changes. I know it's been a lot of work but we need to make sure that we're fixing things in the correct way. I've got a few code simplifications in Objects.py queued up which should also provide some more performance.

I'm happy to merge so we can do some more thorough testing and maybe think about a release.

rob-c:
The main problem I think is that I started fixing things in the proxy and then ended up going the whole hog.
I'm wanting to do some more profiling and testing for LHCb jobs but I don't want to change anything too large before the next release to try and speed up the release cycles, that being said it would be nice to get some of the outstanding smaller PR in.

PR #119, opened 2016-01-17
Feature/xml read changes and fixes

rob-c:
I've tried to keep this branch un-polluted with respect to #117  but there is the possibility that some code has leaked in from this branch to here.

This PR has changes which load an XML object 'directly' into memory and avoids creating a dummy Job only to read in the contents of the XML file.
This is supposed to have some performance benefit, but this is not clear if it actually does in practice. It has the advantage of the scope of the object being much clearer as all that happens is:

a. Object is constructed in GangaRepositoryXML. Object has it's index_cache populated.
b. Object is only loaded when required.
c. Loading of object just calls setNodeAttribute of the Job object created in GangaRepositoryXML to change it's attributes.
d. Job object is now compared to the index_cache and the index_cache is removed.

This PR does result in more readable code in the GangaRepositoryXML by resolving the object from XML at the start of the load function rather than later on and then doing things like checking the index cache afterwards.

It would be nicer to have a system where we load an object if a boolean like 'isLoaded' is False and to generate the index_cache always on the fly from the object and just store the index from disk in a sensible cache location.

I think this is worth pulling quite quickly but I'd like to do this right after #117 just to check for any possible conflicts.


python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -76,17 +76,21 @@ def to_file(j, fobj=None, ignore_subs=''):
...
+def _raw_from_file(f, obj):
     # logger.debug('----------------------------')
     ###logger.debug('Parsing file: %s',f.name)
     xml_content = f.read()
-    obj, errors = Loader().parse(xml_content)
+    xml_lines = xml_content.splitlines()
+    for line in xml_lines:
+        line = line.lstrip()
+    xml_content = '\n'.join(xml_lines)
+    obj, errors = Loader(obj).parse(xml_content)
```
milliams: What's the purpose of this change?

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -299,28 +308,40 @@ def start_element(name, attrs):
...
+                        if not cls._schema.version.isCompatible(version):
+                            attrs['currversion'] = '%s.%s' % (cls._schema.version.major, cls._schema.version.minor)
+                            self.errors.append(SchemaVersionError('Incompatible schema of %(name)s, repository is %(version)s currently in use is %(currversion)s' % attrs))
+                            obj = EmptyGangaObject()
+                            # ignore all elemenents until the corresponding ending
+                            # element (</class>) is reached
+                            self.ignore_count = 1
+                        else:
+                            # make a new ganga object
+                            obj = cls()
```
milliams: For the record: this change here (or rather not calling `__init__` until now) was a the source of a slot of the problems with objects not having a `_parent` attribute. I fixed it in the same way in my local branch.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -728,93 +729,67 @@ def _actually_load_xml(self, fobj, fn, this_id, load_backup):
...
-                            # process modifying the repo
-                            #obj.setNodeIndexCache(None)
-                #obj.setNodeIndexCache(None)
+            if has_children:
+                obj.setNodeAttribute(self.sub_split, SubJobXMLList.SubJobXMLList(os.path.dirname(fn), self.registry, self.dataFileName, load_backup))
+            else:
+                obj.setNodeAttribute(self.sub_split, None)
+
+            # Check if index cache; if loaded; was valid:
+            if obj.getNodeIndexCache() not in [None, {}]:
```
milliams: This function should never return `None` as in `Node`'s `__init__` there is as `self._index_cache = {}`.

rob-c: It used to be incorrectly set to none, i don't know if this has been fixed yet. i'll fix this in a future pr once i've done some work on the index cache

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -329,26 +326,17 @@ def __deepcopy__(self, memo=None):
         # Due to problems on Hammercloud due to uncopyable object lets
         # explicitly stop these objects going anywhere near the __deepcopy__
 
-        cls = type(self)
-        c = Job.__new__(cls)
+        c = Job.__class__()
```
milliams: This will now call `__init__` twice I believe.

rob-c: i'll check but isn't this the same as just doing Job() which this should probably be. i was just trying to explicitly avoid the proxy

milliams: It's not clear from the excerpt above but the code now has:
```python
c = Job.__class__()
c.__init__()
```
and the explicit call to `__init__` is not needed.

Regardless, This should not involve the proxy at all since `Job` in the code above means `class Job(GangaObject)` unless you;ve loaded anything from the GPI explicitly so just `c = Job()` should be fine.

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: Yes I spotted that as well :) not to mention a bunch of other strange problems as it was called when you get an EmptyGangaObject (ironically enough)

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: Passing the object allows for the XML to be read from disk directly into the Job object in question, (or other GangaObject)
The string parsing was to remove whitespace to see how it effected file-size and loading speed.
Answer is not very much unfortunately, but I suppose every little helps.

milliams: I feel that this change causes a lot more complexity for very little gain. Maintainability always wins out over minor speed increases (and in fact most other things) in the long term. If we're loading a file with 1000 `GangaObject`s in it, this perhaps slightly speeds up the loading of just one of them. It's probably also cleaner to know we're always creating a `Job` object (or whatever) from scratch to avoid loitering cached data.

rob-c: I'm looking at the case of files with ~ 10,000 GangaObjects in them so I'm moving to avoid creating too many new objects as and when they're required in favor of doing things in place

milliams: The thing is we can only ever pre-create the root object as we don't know what nested objects there are in the tree. There is almost always going to be lots of nested objects but there will only ever be one root object in a file.

rob-c:
I think I'm going to delete this branch and start again.

As with a few things like this I learned a few things were different than I thought and I've got some good ideas how to make improvements.

I'll remove the PR and branch once I work out how to

PR #121, opened 2016-01-18
Speed up XML loading

milliams:
I spent some time getting the XML loading working faster. On my benchmark of:
```python
import cProfile
from Ganga.Core.GangaRepository.VStreamer import from_file, to_file

j = Job()
j.inputfiles = [LocalFile(compressed=True)]*1000
j.outputfiles = [LCGSEFile()]*1000

cProfile.run('to_file(j, open("test.xml", "w"))', 'write_profile')
cProfile.run('j = from_file(open("test.xml"))', 'read_profile')

assert j[0].inputfiles[100].compressed == True

import pstats
p = pstats.Stats('read_profile')
p.strip_dirs().sort_stats('time').print_stats(20)
```
I was finding load times of 7 seconds on latest develop. After these changes I'm now seeing it between 4 and 4.5 seconds.

The bulk of the changes in here are removing redundant checks in Objects.py for attributes which always exist (or at least do after making the objects be created correctly). It also changes some `isType` for `isinstance` (that alone gives almost a second speed-up).

Most importantly is that his PR adds 52 lines of code but removes 154.

I've run my local tests over this and have seen no regressions but I'd welcome a second and third pair of eyes.

It's possible that this clashes with #117 and #119 but since this patch is much smaller it shouldn't be a problem.


rob-c:
I think given that this is a smaller patch I'm happy to merge this ahead of #117 and look at the fallout there.

I am slightly concerned about the stripProxy possibly having a performance problem as I have highlighted.
We also had older code used to set _index_cache and _data to None (or even delete them) at times but if this code has gone I see no reason not to remove the checks around these objects.

rob-c:
Just out of curiosity @milliams have you run any integration tests on this branch?

milliams:
If we merge #122 then I will merge those changes into this branch and see how they do. I haven't given it a completely thorough test yet, no. I will certainly not be merging until the tests have completed.

milliams:
All the currently converted integration tests work correctly as well as all the unit tests.

rob-c:
I'm happy to merge this in that case :)

milliams:
Ok, great. I will go ahead and merge shortly. I suggest merging develop into your branches and pushing the changes to GitHub so that the integration tests can run.

rob-c:
I'm playing with that now, there's at least one thing has cropped up so I will likely need some time to get #117 into a sensible state

PR #124, opened 2016-01-20
Fix to expend environment variables in RUNTIME_PATH.

mjmottram:
Previously environment variables were not expanded, resulting in the GangaRootPath being prepended to paths beginning with an environment variable.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -766,10 +766,9 @@ under certain conditions; type license() for details.
 
             GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
             def transform(x):
-                return os.path.normpath(Ganga.Utility.files.expandfilename(
-                        os.path.join(GangaRootPath, expandvars(x))))
+                return os.path.normpath(Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath,x)))
 
-            paths = map(transform, filter(lambda x: expandvars(expanduser(x)), config['RUNTIME_PATH'].split(':')))
+            paths = map(transform, filter(None, map(lambda x: expandvars(expanduser(x)), config['RUNTIME_PATH'].split(':'))))
 
             for path in paths:
                 r = RuntimePackage(path)
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -762,14 +762,14 @@ under certain conditions; type license() for details.
 
             
             import inspect
-            GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
-            def transform(x):
-                return os.path.normpath(Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath,x)))
-
             from os.path import expandvars, expanduser
 
-            paths = map(transform, filter(lambda x: expandvars(expanduser(x)), config['RUNTIME_PATH'].split(':')))
+            GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
+            def transform(x):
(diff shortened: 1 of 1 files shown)


alexanderrichards:
ok to test

python/Ganga/Runtime/bootstrap.py
```diff
@@ -762,15 +762,15 @@ def initEnvironment(opt_rexec):
...
+
             GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
             def transform(x):
-                return os.path.normpath(Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath,x)))
-
-            from os.path import expandvars, expanduser
+                return os.path.normpath(Ganga.Utility.files.expandfilename(
+                        os.path.join(GangaRootPath, expandvars(x))))
 
             paths = map(transform, filter(lambda x: expandvars(expanduser(x)), config['RUNTIME_PATH'].split(':')))
```
alexanderrichards: @mjmottram This line looks like it's trying to do an expandvars/expanduser does this map not perform as expected?

mjmottram:
I guess it was using filter where map was more appropriate, the transform function itself was fine.

alexanderrichards:
Thanks Matt, we will discuss at the meeting tomorrow when the next release should be and I would anticipate that this fix would be included.

alexanderrichards:
I must confess to being a little confused by the code that was there already as looking at [Ganga.Utility.files.expandfilename](https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/Utility/files.py#L15) it seems that we are already doing an expand there as well. given that transform would only not be called before your patch when the path is filtered out I cant see how this was affecting you if your RUNTIME_PATH was not blank.

Having said that, clearly this patch fixes things for you so maybe I missed something.

python/Ganga/Runtime/bootstrap.py
```diff
@@ -762,14 +762,13 @@ def initEnvironment(opt_rexec):
...
+
             GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
             def transform(x):
                 return os.path.normpath(Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath,x)))
 
-            from os.path import expandvars, expanduser
-
-            paths = map(transform, filter(lambda x: expandvars(expanduser(x)), config['RUNTIME_PATH'].split(':')))
-
+            paths = map(transform, filter(None, map(lambda x: expandvars(expanduser(x)), config['RUNTIME_PATH'].split(':'))))
```
alexanderrichards: also (I know you didn't start it.. It may well have even been me ;-)) I think that all this use for map/filter/lambda is quite messy.

could you check if something like:
```python
paths = [os.normpath(os.normcase(os.expandvars(os.expanduser(p.strip())))) 
             for p in config['RUNTIME_PATH'].split(':') if p.strip()]
```

works in place here?

If so don't change your pull request and we can discuss in the meeting (since this must be a common thing to need to do) if it would probably be an idea to update [Ganga.Utility.files.expandfilename](https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/Utility/files.py#L15)

to something like:

```python
from os.path import abspath, expandvars, expanduser, normcase
def iexpandfilenames(filenames, force=False, ignore_empty=True):
    if not isinstance(filenames, Iterable):
        filenames = [filenames]

    for f in filenames:
        sanitised_path = abspath(normcase(expandvars(expanduser(f.strip()))))
        if ignore_empty:
            continue
        if os.path.exists(sanitised_path) or force:
            yield sanitised_path
        else:
            getLogger().debug("Filename: %s doesn't exist using it anyway" % f.strip())
            yield f.strip()

def expandfilenames(filenames, force=False):
    return [f for f in iexpandfilenames(filenames, force)]
```

and then this line would become simply:

```python
paths = [os.path.join(GangaRootPath,p) for p in 
              Ganga.Utility.files.iexpandfilenames(config['RUNTIME_PATH'].split(':'))]
```

alexanderrichards:
This passes the tests so I'll go ahead and merge it. 
Thanks again Matt

PR #126, opened 2016-01-20
hacks to get outputfiles and sandboxes working

### LINKED ISSUE #125: No Ganga/__init__.py shipped with job
Running the latest develop version, ```Ganga/Utilities/__init__.py``` and ```files.py``` are both shipped with the job, but no ```Ganga/__init__.py```, meaning that the job fails on the backend whenever attempting to load files relative to the _python/Ganga base directory (e.g. when adding an output sandbox).

mjmottram:
This fixes #125 , but it's not the most elegant.  Maybe it's worth breaking any ganga code shipped to a backend in the _python directory into standalone modules with no imports lurking within?

### DIFF
--- a/python/Ganga/Core/Sandbox/Sandbox.py
+++ b/python/Ganga/Core/Sandbox/Sandbox.py
@@ -33,9 +33,11 @@ def getGangaModulesAsSandboxFiles(modules):
     import inspect
     import sys
     from Ganga.Utility.files import remove_prefix
-    from Ganga.GPIDev.Lib.File import File
+    from Ganga.GPIDev.Lib.File import File, FileBuffer
 
     files = []
+    # HACK: append an empty __init__.py file for the Ganga base directory
+    files.append(FileBuffer("__init__.py", "", subdir=os.path.join(PYTHON_DIR, 'Ganga')))
     for m in modules:
         fullpath = os.path.realpath(inspect.getsourcefile(m))
         dir, fn = os.path.split(remove_prefix(fullpath, sys.path))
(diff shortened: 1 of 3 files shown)


milliams:
@ganga-ci add to whitelist

drmarkwslater:
I'm afraid this bug was introduced with #87 due to shifting all the config in to the `__init__.py`. I had assumed that just not including the Ganga module would be OK but obviously that's not the case. Hopefully tomorrow I'll try to do a more complete fix that basically catches this and manually adds a blank `__init__.py` to the sandbox. Apologies!

drmarkwslater:
I should have said - this will just affect the HACK line that automatically adds the `__init__.py`!

mjmottram:
Something different is definitely needed - for some reason the inclusion of a the empty FileBuffer __init__.py causes executables to fail upon submission (at least to LCG backends).  No idea why.

rob-c:
@mjmottram Could you try with a File object rather than a FileBuffer? It could be that the FileBuffer isn't being written to disk in the correct way (or to the correct place)

mjmottram:
Yep, that works.  Thanks.

mjmottram:
Oh dear, now I get <LogRecord: Ganga.Lib.LCG, 30, /users/mottram/processing/ganga/install/ganga_git/python/Ganga/Lib/LCG/LCG.py, 1653, "The job %d has reached unexpected the Cleared state and Ganga cannot retrieve the output."> as an error (either with the develop branch or with a version of this that adds `__init__.py` to the sandbox).

drmarkwslater:
#168 Should have a more complete fix in. I've also transferred the additional import error handling as well so as soon as that's in, we can close this PR I think.

drmarkwslater:
As #168 is now merged, I'll close this PR. @mjmottram do shout if this doesn't fix the problem!

alexanderrichards:
@drmarkwslater since this PR is closed without merging the `feature/backendImport` branch, is it safe to now delete that branch or is it being kept around for another PR?

drmarkwslater:
Yes, I think it can looking at the code

PR #127, opened 2016-01-20
Don't overwrite the existing documentation for a proxied GangaObject

milliams:
Currently, when a `GangaObject` is proxy-wrapped, its `__doc__` attribute is reset to a generic string pointing the reader towards the proxy object. In the process of auto-generating API and GPI docs, this is getting in the way.

Does anyone have a objection to removing this? I don't see that it adds any value.

### DIFF
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -690,8 +690,6 @@ def GPIProxyClassFactory(name, pluginclass):
 
     publicdoc = pluginclass.__doc__ + itbuf.getString()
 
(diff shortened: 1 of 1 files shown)


rob-c:
I'm unsure does this get exposed at the IPython prompt? if not then I'm happy to see it go it doesn't add much if it's getting in the way.

milliams:
The only way you'll see this is if you do something like
```python
help(Job._impl)
```
Since at the prompt you should never see any non-proxied objects it should not affect it at all. Proxied objects still have their enhanced doc strings.

rob-c:
ok, in that case I'm happy to get rid of this and continue. I don't see any real reason why someone would be looking here for documentation.

PR #128, opened 2016-01-20
Add initial version of Sphinx documentation

milliams:
I've been working the last few days on getting the new documentation suite up and running. Before you read too far, you can see the current state of it at [ganga.readthedocs.org](http://ganga.readthedocs.org/en/feature-sphinx/). That documentation is automatically regenerated from the branch and can be used to host all the docs in future.

The API docs are generated using the standard [`sphinx-apidoc`](http://www.sphinx-doc.org/en/stable/man/sphinx-apidoc.html) which creates a set of ReST files based on the source code.

The GPI docs are generated using a custom script which does a dummy bootstrap of Ganga and inspects the `Ganga.GPI` module. In the future this should be adapted to get the information from little ganga. The structure of these pages will likely evolve with time but I'm open to ideas on the best layout.

I'm planning on filling the user guide based on @drmarkwslater's tutorial on the wiki (or I can let him do it).

To generate the docs for yourself you can run `make html` from the `doc` directory. You will need to `pip install sphinx` but that should be all. It will take around a minute to generate the docs (can be sped up with `make html -j4`) mostly due to the API docs being so large.

This PR only touches the `doc` directory and a small change to _setup.py_ so there should be no clashes or introduced bugs.


alexanderrichards:
looks good Matt. You included the napoleon extension so we can do google style docs right? Also good job on the SVG logo. I think it's good to have it in the repo.

milliams:
Yep, `napoleon` is included and it should work. You can see an example of a function using the style at http://ganga.readthedocs.org/en/feature-sphinx/API/Ganga.Core.MonitoringComponent.Local_GangaMC_Service.html#Ganga.Core.MonitoringComponent.Local_GangaMC_Service.release_when_done

.gitignore
```diff
@@ -1 +1,7 @@
 *.pyc
+doc/_build
+.idea
```
alexanderrichards: Out of curiosity what creates a .idea file? :smile:

alexanderrichards:
> Yep, napoleon is included and it should work. You can see an example of a function using the style at

Excellent, The google style is much nicer.

I see no reason not to merge this as it doesn't touch and code

.gitignore
milliams: It's where PyCharm stores its configuration.

PR #130, opened 2016-01-21
always create a versions file if one does not exist

mjmottram:
I use multiple .gangarc (so that I can specify different ganga directories to separate workflows).  Moving to the latest develop version of ganga I cannot start ganga as no versions file exists.  Maybe there's a reason for not creating a new versions file if a custom config file is specified, in which case a different fix is needed.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -237,12 +237,11 @@ under certain conditions; type license() for details.
 #           file_opens(self.args[0],'reading script')
 
     @staticmethod
-    def new_version(update=True):
+    def new_version():
         versions_filename = os.path.join(getConfig('Configuration')['gangadir'], '.used_versions')
         if not os.path.exists(versions_filename):
-            if update:
-                with open(versions_filename, 'w') as versions_file:
-                    versions_file.write(_gangaVersion + '\n')
+            with open(versions_filename, 'w') as versions_file:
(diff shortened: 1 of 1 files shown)


alexanderrichards:
hmmm.. I recall that we definitely didn't want to update the config file if it is set explicitly as doing so would change the file that the user presumably picked precisely because of it's content (rather than using the default). I wonder if the same logic has been applied to the used_versions file erroneously.

@ganga-devs/core Anyone else have an opinion/thoughts on this? Did we intend to not update the used_versions if the user explicitly specifies a config file? Seems to me this may just have been copied across from the config updating algorithm and it's probably not the intended behaviour

milliams:
I had a look at this and there's two different things in the code. Creating a new `.used_versions` files is currently only done if a custom config file was *not* set explicitly. If there was already a file then the new version string was *always* added anyway. Then, if we're running a previously unseen version of Ganga (based on `.used_versions`), the config file is potentially updated with the new version number (again, only if the config file was *not* set explicitly).

@mjmottram's change here only affects the first of these two so explicitly setting the config file when starting Ganga will now always create a missing `.used_versions` file. I think this makes sense as `update` was only affecting the creation of a `.used_versions`, not its updating.

Personally, I'm happy with this change but I'll give someone else a chance to comment.

alexanderrichards:
It was my feeling too that this change is desired.

I'm happy to merge this.

milliams:
Thanks for the fix @mjmottram

PR #132, opened 2016-01-22
These changes will allow the unittests to be run from any location

rob-c:
This commit takes into account some slight differences between what is in setup.py and what is used within the LHCb SetupProject definition of GANGASYSROOT.
As LHCb is moving away from SetupProject and this will go away (I think it already can) I'm deciding that the 'correct' approach is what we use in the untittests.

I've refactored some of the code to allow for:
```
python $RELPATH/setup.py tests
```
to work as it examines the env to find Ganga and then passes this to the subprocess so that the convention of always cd-ing into the absolute path is no longer required.
(I want to run the tests often during development and I find it annoying to have to keep track of what folder I'm in instead of just executing a quick command to run all of the tests)


milliams:
Is the problem here that you want to be able to do:
```bash
SetupProject Ganga
python $GANGASYSROOT/../install/ganga/setup.py tests
```
If so is it not simpler to just have:
```python
# At the top of the file so readme() can use it
this_dir = os.path.dirname(os.path.realpath(__file__))

# At the end of RunTestsCommand.run
subprocess.check_call(' '.join(cmd), cwd=os.path.join(this_dir, 'python'), shell=True)
```
There's no need to set `PYTHONPATH` since the working directory is set.

Secondly I don't want to have LHCb-specific things in setup.py like `GANGASYSROOT`.

rob-c:
It's probably easier to extract the path once from ```__file__``` but I think this can be done once at the top of the file either way so I'll do that.

I found that if I wanted to run the literal command in the PR that the PYTHONPATH was NOT set in such a way that the Ganga module was importable for the unittests.
(integration tests perform a lookup in GangaUnitTest.py setup which I want to keep for running ```nosetests blah/GPI/Bugs/script.py```)

Finally:
I'll happily remove ```GANGASYSROOT``` in favour of inspecting the file or using ```__file__```. I'll do it as well from GangaUnitTest.py for the sake of consistency.

rob-c:
OK, unfortunately this is now pulling in changes from #150 so I would suggest holding off on reviewing this until that is in which depends on #138.

rob-c:
@milliams Are there any serious objections to this branch being merged other than the description doesn't strictly match the contents?

This branch contains some optimisation patches I intended to put into #150 (I accidentally messed up my develop branch earlier today)

The changes to the test system now make use of ```__file``` to determine what paths should be added to ```sys.path```. Are there any objections to this?

milliams:
I'd much rather have this as a PR with just the discussed changes in it. It should be easy enough to start a new branch and cherry-pick into the new branch commits 9736fda5ca7 and a6f20287ab65d with:
```bash
git checkout develop
git flow feature start runTestsFromAnywhere2
git cherry-pick 9736fda
git cherry-pick a6f2028
```
Also, there still some mention of `GANGASYSROOT` in `_getTestEnv` which probably needs removing.

The rest of the commits don't look like just optimisations, some of them are doing extra checking, refactoring etc.

rob-c:
Last time I touched cherry-pick I lost my whole repo due to corruption.
I'll look into this later today in that case

rob-c:
closing for #156

PR #133, opened 2016-01-22
Simplify descriptor get

milliams:
This is a cleanup exercise in `Descriptor.__get__`. The code before was overly nested as well as calling lots of functions it did not need to. The logic in these functions should be clean and easy to follow and hopefully this is getting there now.

I've noticed some speed-ups in XML loading with this as well.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -382,67 +382,44 @@ class Descriptor(object):
             raise AttributeError('cannot modify or delete "%s" property (declared as "getter")' % getName(self))
 
     def __get__(self, obj, cls):
+        name = getName(self)
+
+        # If obj is None then the getter was called on the class so return the Item
         if obj is None:
-            return cls._schema[getName(self)]
-        else:
-            result = None
-            getter = self._bind_method(obj, self._getter_name)
-            if getter:
-                result = getter()
-            else:
+            return cls._schema[name]
 
-                lookup_exception = None
+        if self._getter_name:
+            return self._bind_method(obj, self._getter_name)()
 
-                #logger.info("Looking for: %s in Cache" % getName(self))
-                _obj = obj
+        # ._data takes priority ALWAYS over ._index_cache
+        # This access should not cause the object to be loaded
+        obj_data = obj.getNodeData()
+        if obj_data is not None:
+            if name in obj_data:
+                return obj_data[name]
 
-                try:
-                    obj_index = _obj.getNodeIndexCache()
-                    if obj_index is not None:
-                        if getName(self) in obj_index.keys():
-                            return obj_index[getName(self)]
-                except Exception as err:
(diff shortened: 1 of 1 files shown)


rob-c:
A few questions in this case.

1. Is this getter used in the case of a purely transient parameter only stored in the ```__dict__``` and not in ```_data```?
2. Is returning the default value from the schema (potentially before the schema is fully populated) the safest thing we should be doing?

milliams:
1. This getter is part of a descriptor (in fact our type `Descriptor`) and so is only called when doing `obj.name` which will call `type(obj).name.__get__(obj, type(obj))` (or equivalent). The only time a `Descriptor` is created is in `GangaMetaclass.__init__` and it only creates them for schema items.
If there is a thing in the `obj.__dict__` called `'foo'` then doing `obj.foo` will not call the descriptor since `type(obj).foo` is not a descriptor (and in fact probably does not exist).

2. The code is already returning the default value if it wasn't found elsewhere. The alternative is that `j = Job(); j.name` would not work (since there would, at that point, be no `'name'` entry in `_data`). The problem here come from the fact that getting the default value is a very heavy operation as it involves querying the config system etc. (in case the defaults have been overridden) but I'm not sure of the best alternative.

rob-c:
1. OK, I think I understand
2. I'm not sure I'm keen on this behaviour.
Surely the following pseudo-example would fail in this case?
```
j=Job()
functionToChangeDefaultApplication()
assert j.application == Executable()
```
If the object is assigned on construction you know that the Job has been constructed according to the config recipie when it was constructed.
when itself was constructed which is potentially tricky.
I know this isn't the fault of the getter but I don't think that the expected behaviour should rely on the config unless the descriptor captures the config on initialization.

milliams:
Ok, so the only way to fix the problem you describe of creating an object, changing the defaults and then wanting the old defaults when asking the object for its values (which does make some sense) is to explicitly fill `_data` on GangaObject construction by calling `getDefaultValue`.

I've been wanting to avoid this second option as it makes creating lots of objects slow (like when loading a large XML file) whereas the current behaviour defers that until a property is actually requested (or at least it should).

rob-c:
OK, well I'm happy to merge asap :)

milliams:
Great, I'll do so now. Your comments do make sense though, I hadn't considered people changing the defaults for schemas on-the-fly like that. `getDefaultValue` (well, `_getDefaultValueInternal`) is probably my next target for cleanup since it's such a hog and is a bit of a mess

rob-c:
I've already added a large amount of caching in the getDefaultValueInternal function (which explains some of the mess)
The function was a real resource hog last I looked but I gave up as it's a difficult one to fix without changing some stuff quite low down.

milliams:
Fair enough, perhaps it's not going to get much faster then.

rob-c:
There's probably room for improvement I encourage someone else to take a look and see if we can change how it works from the ground up.

PR #134, opened 2016-01-22
Remove deprecated funtions etc.

milliams:
This is hopefully a completely safe PR as it only removes things that have been deprecated or not working for many years (since at least 2010). They have all had deprecation warnings printed and many aren't even user-reachable.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -231,7 +231,7 @@ class Job(GangaObject):
     _category = 'jobs'
     _name = 'Job'
     _exportmethods = ['prepare', 'unprepare', 'submit', 'remove', 'kill',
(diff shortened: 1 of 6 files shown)


alexanderrichards:
Looks safe to me

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -1911,10 +1911,6 @@ def doit(f):
             logger.debug("Remove Err: %s" % str(err))
             pass
 
-    def fail(self, force=False):
```
rob-c: This needs to be removed from _export_methods too for at least the Job class

milliams: Thanks @rob-c I'd missed this. I guess this is another reason to put together a patch to move to using the `@export` decorator I developed.

milliams:
If there are no further comment I'd like to merge this.

alexanderrichards:
none from me

PR #135, opened 2016-01-24
Cleanup ObjectMetaclass

milliams:
Mostly this is just moving some code from `ObjectMetaclass.__init__` into `GPIProxyClassFactory` since it is only relevant to proxies.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -686,15 +686,8 @@ class ObjectMetaclass(type):
 
         super(ObjectMetaclass, cls).__init__(name, bases, this_dict)
 
-        # ignore the 'abstract' base class
-        # FIXME: this mechanism should be based on explicit getName(cls) or alike
-        #if name == 'GangaObject':
-        #    return
-
-        #logger.debug("Metaclass.__init__: class %s name %s bases %s", cls, name, bases)
-
         # all Ganga classes must have (even empty) schema
-        if not hasattr(cls, '_schema') or cls._schema is None:
+        if cls._schema is None:
             cls._schema = Schema.Schema(None, None)
 
         this_schema = cls._schema
@@ -732,9 +725,6 @@ class ObjectMetaclass(type):
         # create reference in schema to the pluginclass
         this_schema._pluginclass = cls
 
-        # store generated proxy class
-        setattr(cls, '_proxyClass', proxyClass)
-
         # register plugin class
         if hasattr(cls, '_declared_property'):
             # if we've not even declared this we don't want to use it!
@@ -744,6 +734,9 @@ class ObjectMetaclass(type):
             # create a configuration unit for default values of object properties
             if not cls._declared_property('hidden') or cls._declared_property('enable_config'):
                 this_schema.createDefaultConfig()
+
+        # store generated proxy class
+        setattr(cls, '_proxyClass', GPIProxyClassFactory(name, cls))
 
 
 
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -694,6 +694,7 @@ class ObjectMetaclass(type):
 
         # Add all class members of type `Schema.Item` to the _schema object
         # TODO: We _could_ add base class's Items here by going through `bases` as well.
+        # We can't just yet because at this point the base class' Item has been overwritten with a Descriptor
         for member_name, member in this_dict.items():
             if isinstance(member, Schema.Item):
                 this_schema.datadict[member_name] = member
@@ -725,19 +726,16 @@ class ObjectMetaclass(type):
         # create reference in schema to the pluginclass
         this_schema._pluginclass = cls
 
-        # register plugin class
-        if hasattr(cls, '_declared_property'):
-            # if we've not even declared this we don't want to use it!
-            if not cls._declared_property('hidden') or cls._declared_property('enable_plugin'):
-                allPlugins.add(cls, cls._category, getName(cls))
+        # if we've not even declared this we don't want to use it!
+        if not cls._declared_property('hidden') or cls._declared_property('enable_plugin'):
+            allPlugins.add(cls, cls._category, getName(cls))
 
-            # create a configuration unit for default values of object properties
-            if not cls._declared_property('hidden') or cls._declared_property('enable_config'):
-                this_schema.createDefaultConfig()
+        # create a configuration unit for default values of object properties
+        if not cls._declared_property('hidden') or cls._declared_property('enable_config'):
+            this_schema.createDefaultConfig()
 
         # store generated proxy class
         setattr(cls, '_proxyClass', GPIProxyClassFactory(name, cls))
-
 
 
 class GangaObject(Node):
@@ -987,10 +985,9 @@ class GangaObject(Node):
     # this means that implicit (inherited) _name attribute has no effect in the derived class
     # example: cls._declared_property('hidden') => True if there is class
     # attribute _hidden explicitly declared
-    def _declared_property(self, name):
-        return '_' + name in self.__dict__
-
-    _declared_property = classmethod(_declared_property)
+    @classmethod
+    def _declared_property(cls, name):
+        return '_' + name in cls.__dict__
 
     # get the job object associated with self or raise an assertion error
     # the FIRST PARENT Job is returned...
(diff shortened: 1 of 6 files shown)


python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -676,72 +665,29 @@ def addToList(_input_elements, _final_list, action=None):
...
-        # FIXME: this mechanism should be based on explicit getName(cls) or alike
-        #if name == 'GangaObject':
-        #    return
-
-        #logger.debug("Metaclass.__init__: class %s name %s bases %s", cls, name, bases)
-
         # all Ganga classes must have (even empty) schema
-        if not hasattr(cls, '_schema') or cls._schema is None:
+        if cls._schema is None:
             cls._schema = Schema.Schema(None, None)
```
rob-c: Can we add a warning here? I saw at least one problem a while ago with the empty schema being used in the xml.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -771,19 +715,16 @@ def __init__(cls, name, bases, this_dict):
...
 
-            # create a configuration unit for default values of object properties
-            if not cls._declared_property('hidden') or cls._declared_property('enable_config'):
-                this_schema.createDefaultConfig()
+        # create a configuration unit for default values of object properties
+        if not cls._declared_property('hidden') or cls._declared_property('enable_config'):
+            this_schema.createDefaultConfig()
 
+        # store generated proxy class
+        setattr(cls, '_proxyClass', GPIProxyClassFactory(name, cls))
```
rob-c: Can this be moved to within the proxy so we're not examining objects for this there?
this is only needed in proxy construction i think

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -91,16 +91,6 @@ def __str__(self):
         return "JobError: %s" % str(self.what)
 
 
-class PreparedStateError(GangaException):
```
rob-c: Can we move this to somewhere closes to the Prep repository as it should be defined there imo or in Core.exceptions

python/Ganga/GPIDev/Base/Objects.py
milliams: It can't be moved just yet as there's currently no other way to call `GPIProxyClassFactory()`. What I'm planning on adding (and have started work on) is a decorator for `GangaObject` subclasses which adds this and registers them in the GPI (similar to how Alex's system will work). For now I just wanted to combine all this to a single line.

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: `PreparedStateError` is not currently used anywhere in the whole code base so I figured it was safe to remove it entirely. If you think we should keep it then I agree, it should be there instead.

python/Ganga/GPIDev/Base/Objects.py
milliams: `GangaObject` has the class attribute `_schema` set in the class definition so there's no way that the attribute will be missing unless this is not passed a `GangaObject`.

milliams:
retest this please

python/Ganga/GPIDev/Base/Objects.py
rob-c: I was thinking it would be nice to have a ```addProxyFactory``` function or similar which takes any of the knowledge of what it's called or what's in it out of the Objects.py then we can worry about placing this elsewhere is needed

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Apologies I hadn't had a strong enough coffee. I thought this was being added, please delete and ignore my comment :)

python/Ganga/GPIDev/Base/Objects.py
rob-c: In that case is this ever not passed to the ```GangaObject```?

python/Ganga/GPIDev/Base/Objects.py
milliams: My plan was to have a class decorator called `add_proxy_and_export_to_gpi` (except obviously a better name) which would be used like:
``` python
@add_proxy_and_export_to_gpi
class MyGangaObject(GangaObject):
    _schema = Schema(Version(1, 0))
    _category = 'MyGangaObjects'
    _name = 'MyGangaObject'
```
so we would add that decorator to any class which we want to be proxyable. It would also do the `exportToGPI()` bit as well.

This way, with just that we could do
```python
MyGangaObject()
```
in the GPI and it would work.

All of the proxy stuff would then be removed from Objects.py as it would be imposed on it externally instead.

rob-c:
@milliams  Just a thought, (probably completely incompatible with what we're currently doing)

Where are GangaObjects exported to the GPI is it clear?
Wouldn't it make more sense to add the proxy entirely at the point we chose to expose the object to the GPI?
It should be possible to extract the information for the proxyClass and proxyObject data there I would have thought.

python/Ganga/GPIDev/Base/Objects.py
rob-c: I somehow wonder if this onus of having a proxy should be placed on an effective request should be made from the ```GangaObject```. I'm more in favour of some some well defined layer where we chose to do something similar to export all known GangaObjects to the GPI in one single step but that's just my opinion.

rob-c:
I just fount it at line 882 of bootstrap.py

Can't we just change this function to either be something else or build in some ability to the exportToGPI step to fully construct and add a Proxy to GangaObject type objects and all of the required framework as the class is exposed to the GPI.

I've nothing against decorators but I'm much more in favour of just removing __all__ knowledge of if the object even wants to be proxied or not (let alone what a proxy is or looks like) from Objects.py If we leave something in even like a decorator it just leaves room for things to creep back in, in the future imo.

python/Ganga/GPIDev/Base/Objects.py
milliams: Do you mean "passed a `GangaObject`"? If so, no. There are no cases (at present) where anything other than `GangaObject` has its metaclass set to `ObjectMetaclass`.

python/Ganga/GPIDev/Base/Objects.py
milliams: If we decide that we want _all_ `GangaObject`s to be exported (which we currently don't since we set `_hidden` on many) then I agree but I think it's good to be explicit with the GPI where possible. For example we don't want `GangaObject` itself exported or any of the adapters or one or two other examples.

The way it works at the moment is based on `_hidden` and adding it to `allPlugins` and then the bootstrap iterating that and adding each to the GPI.

milliams:
That could work. We can't possibly completely get away from a `GangaObject` subclass caring about proxies since, for example, it has to choose which methods to export. However, we shuold be able to completely remove all proxy stuff from Objects.py.

I don't see any reason why doing (bootstrap.py:878):
```python
for k in allPlugins.allCategories():
  for n in allPlugins.allClasses(k):
    cls = allPlugins.find(k, n)
      if not cls._declared_property('hidden'):
        cls._proxyClass = GPIProxyClassFactory(cls._name, cls)
        exportToGPI(n, cls._proxyClass, 'Classes')
```
and removing similar from Objects.py wouldn't work in principle. It's definitely worth thinking about.

I also have a local branch in place which cleans up some of the mess in the proxy system which I'll PR soon.

rob-c:
Yes I can agree that we have to care about things like exporting functions and stuff and I think that's a very good place for decorators and a good reason to have some default list/dictionary/container in the Objects.py.

(I'm assuming in the above example you'd rather use ```setattr(cls, '_proxyClass' ...``` to avoid having to store a None ```_proxyClass``` attribute in Objects.py still?)

I don't think this is required for this PR and I think we've got the usual sentiment of lets clean up the Core but I'm keen to remove all ```'_proxy*'``` objects from everywhere but Proxy.py as I think should be implementation details.

I'm happy to merge when you are.

python/Ganga/GPIDev/Base/Objects.py
alexanderrichards: > My plan was to have a class decorator called add_proxy_and_export_to_gpi (except obviously a better name) which would be used like:

This is in fact, as you say, the way that I have done it in the new code I've been playing with

python/Ganga/GPIDev/Base/Objects.py
rob-c: Not really important but it sounded like the check for _schema isn't needed any-more if objects aren't being constructed with ```__new__``` and I think I removed all references of that from the Core

PR #138, opened 2016-01-25
Cleanup proxy attributes

milliams:
This is another step in the cleanup of Proxy.py. The primary thing here is enforcing which attributes of proxy classes and `GangaObject`s should be present and what they should point to.

The best summary of the state this change makes is in the [added docs](https://github.com/ganga-devs/ganga/blob/feature/cleanup_proxy_dict/doc/dev/index.rst#attributes-on-proxy-objects) and the [new test](https://github.com/ganga-devs/ganga/blob/feature/cleanup_proxy_dict/python/Ganga/new_tests/TestProxy.py#L47).

The automated tests pass just fine and I found no errors in my local testing but a change like this certainly needs some more eyes on it.

The bulk of the interesting changes are in commit 4ca5bfcc06a.


milliams:
retest this please

rob-c:
Apologies I thought this branch had already been merged, I wasn't paying attention. Aside from moving to prefer stripProxy I have a gut feeling that I don't want to see the 'implRef' parameter anywhere outside of Proxy.py. I suppose having it in the other Proxy layers isn't the end of the world but I think it isn't the nicest solution imo.

I'm happy to merge modulo my stripProxy comments :)

milliams:
Cheers @rob-c. I'll make the changes you suggested and update the PR.

rob-c:
@milliams Have you put the changes in? If so I'm happy to merge so I can try and get in my other PR

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -41,13 +41,12 @@ def _getGangaList():
 
 
 class Node(object):
-    _ref_list = ['_parent', '_registry', '_index_cache', '_proxyObject']
+    _ref_list = ['_parent', '_registry', '_index_cache']
```
rob-c: This is used to prevent any copy operation involving ```__dict__``` from going in infinite loops. Is this safe?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -30,7 +30,7 @@
 
 _imported_GangaList = None
 
-do_not_copy = ['_proxyObject', '_proxyClass', '_data', '_index_cache', '_parent', '_registry', '_impl']
+do_not_copy = ['_proxyObject', '_data', '_index_cache', '_parent', '_registry']
```
rob-c: This is used to prevent any copy operation involving ```__dict__``` from going in infinite loops. Is this safe?

milliams: Yes, it's safe to copy the `_proxyClass` itself since a class copy (as opposed to an object copy) does not recurse any deeper. `_impl` was removed since it's never present in a `Node` or `GangaObject`.

python/Ganga/GPIDev/Base/Objects.py
milliams: `_proxyObject` is protected from being copied by being in `do_not_copy`. At least it seemed to be based on my tests.

milliams:
The changes are in now so there's much less use of `implRef` and most of it has been changed to `stripProxy()`.

rob-c:
OK, I'm happy to merge in this case :)

PR #141, opened 2016-01-26
Cleanup schema

milliams:
This is some simple cleanup of Schema.py along with some reorganisation of the tests. There should be no functional changes here.


rob-c:
Looks good, Can't see any reason to hold up merging this

PR #142, opened 2016-01-26
Job.py: Reorganise imports and some PEP8ing

milliams:
Some simple PEP8ing of Job.py to reduce the trivial static analysis warning this file produces. The imports were extracted to module scope manually and then sorted using [isort](https://pypi.python.org/pypi/isort) which does a good job of keeping things neat.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -1,61 +1,54 @@
-##########################################################################
-# Ganga Project. http://cern.ch/ganga
-#
-# $Id: Job.py,v 1.13 2009-07-14 12:43:41 moscicki Exp $
-##########################################################################
+from __future__ import absolute_import
 
-from Ganga.GPIDev.Base import GangaObject
-from Ganga.GPIDev.Schema import Version, Schema, FileItem, ComponentItem, SimpleItem, GangaFileItem
-from Ganga.GPIDev.Lib.Job.MetadataDict import MetadataDict
-
-from Ganga.Utility.logging import getLogger, log_user_exception
-from Ganga.GPIDev.Adapters.IPostProcessor import PostProcessException, MultiPostProcessor
-
-from Ganga.Utility.Config import getConfig, ConfigError
-
-from Ganga.Core import GangaException
-from Ganga.Core.GangaRepository import RegistryKeyError, getRegistry
-
-from Ganga.GPIDev.Adapters.IApplication import PostprocessStatusUpdate
-
-from Ganga.Core.GangaRepository.SubJobXMLList import SubJobXMLList
-
-from Ganga.GPIDev.Base.Proxy import isType, getName, GPIProxyObjectFactory, addProxy, stripProxy, runProxyMethod, runtimeEvalString, getRuntimeGPIObject
-from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList, makeGangaListByRef
-
-from Ganga.Lib.Splitters import DefaultSplitter
-
-import errno
 import copy
+import errno
 import glob
+import inspect
 import os
 import re
-import shutil
-import sys
 import time
-import errno
-
 import uuid
-from JobTime import JobTime
 
+import Ganga.Core.FileWorkspace
 import Ganga.GPIDev.Lib.File.FileUtils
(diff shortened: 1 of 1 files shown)


rob-c:
I'm weary about moving anything from Ganga.GPIDev.Lib.File to the top level imports (they should be there anyway).
I seem to remember this caused problems as it would pull in GoogleFile which pulled in Job at the top level. I think that was fixed there but I don't know if any other Files make use of importing Job directly or not. (I don't think they should at the top level at least)

rob-c:
If it works I'm happy to see this committed :) Just wanted to highlight problems I saw in the past.

alexanderrichards:
> GoogleFile which pulled in Job

I'm more interested why the GoogleFile needs Job. This sounds to me like it could be re-factored.

milliams:
@alexanderrichards it needs it at the moment in [`_on_attribute__set__`](https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/GPIDev/Lib/File/GoogleFile.py#L167). I'm sure this could be fixed properly though.

@rob-c Any conflict like that would exhibit immediately on module load and there's no problem that I can see. Logically, Job should be on a higher level than File objects since job depends on so many things. Having all these imports at the top also helps highlight exactly how big the over-dependency problem in Job.py is.

rob-c:
@milliams I agree Job.py touches a little too much, for instance it shouldn't even go near the Monitoring system imo and the Files stuff should all be through a common API rather than rely on single file types but this has been discussed before.

milliams:
The tests are passing so if people are happy, I'd like to merge.

Ignore the failing tests from Travis CI. That was me testing something which will need some more work. I'd like to use Travis as a secondary testing system for the unit tests since it uses an Ubuntu system so it might find some things not seen on SL.

alexanderrichards:
> I'd like to use Travis as a secondary testing system for the unit tests since it uses an Ubuntu system so it might find some things not seen on SL.

Sounds good. Nice to have a non-CERN based system as well. Would this be just for the unit tests? If not how would you run atlas and lhcb tests?

Looks good to merge to me

milliams:
@alexanderrichards I think it would just be for the unit tests and perhaps full tests that just use Local/Executable since we of course cannot get any grid proxies. In general perhaps it would be good to categorise the tests in some way so that we can be specific about this. Perhaps using the [attrib plugin](http://nose.readthedocs.org/en/latest/plugins/attrib.html).

alexanderrichards:
yes, using something prescriptive like the attrib plugin is the correct way to do this. In general this decorator is simple and of the form:
```python
class attrib(object):
    def __init__(self, *args, **kwargs):
        self._args, self._kwargs = args, kwargs
    def __call__(self, func):
        for a in self._args:
            setattr(func, a, True)
        for k, v in self._kwargs.iteritems():
            setattr(func, k, v)
        return func
```
It is not part of the standard unittesting framework however as it's a nose extension. Do you happen to know if there is a standard equivalent? Are we happy to start relying on nose features? It's an extra dependency albeit only on our testing system.

milliams:
Personally I'm happy to start relying on nose features since they're only required in test mode. I will add development dependencies like this to `requirements.txt` so that they are not required for all users. I'll try to add some information about how to install Ganga for developers and run tests etc. to the dev guide too.

alexanderrichards:
just one further comment. If we are likely to use the attrib plugin with a standard marker it might be repetitive and more typing to keep decorating as such:

```python

    @attrib(needs_proxy)
    def test_one(self):
        pass

    @attrib(needs_proxy)
    def test_two(self):
        pass
```

when we could consider having a simple:
```python
def needs_proxy(func):
    func.needs_proxy = True
    return func
```

making the test code:
```python
    @needs_proxy
    def test_one(self):
        pass

    @needs_proxy
    def test_two(self):
        pass
```

This could be arguably more explicit and nicer to look at and is certainly less code for the tester to write and wouldn't need an extra dependency on nose. Just a thought but is seems like there aren't too many drawbacks. 

Thoughts...?

milliams:
I agree this makes sense. I think putting things like this into a new test framework-type place in the code would be good.

I've started talking to Ulrik about getting a Grid Robot certificate so hopefully we'll be able to actually start testing this soonish.

alexanderrichards:
> I think putting things like this into a new test framework-type place in the code would be good.

agreed, a sort of tests/utils.py type place.

PR #143, opened 2016-01-26
Fix splitter

milliams:
Running a simple job like seen in TestArgSplitter.py was failing with errors. This fixes them.

@rob-c The only question I have is that `j.subjobs` is a `GangaList` by line 487; I forget if that is expected at this point?

### DIFF
--- a/python/Ganga/GPIDev/Adapters/IBackend.py
+++ b/python/Ganga/GPIDev/Adapters/IBackend.py
@@ -484,7 +484,7 @@ class IBackend(GangaObject):
                     try:
                         subjobs_to_monitor = []
                         for sj_id in this_block:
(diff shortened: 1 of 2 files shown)


rob-c:
Yes I'd expect this to be a GangaList at this stage

rob-c:
There is some funny behaviour as to whether you're addressing the RegistrySlice vs the true object. but I would strongly suspect it should behave in the same way as a GangaList regardless for all intent and purpose.
There is the potential for a stray proxy to enter in the latter vs the former so it may be worth being careful.

milliams:
Ok, so this PR is correct then. Am I ok to merge?

rob-c:
Yes please do :)

PR #147, opened 2016-01-26
Add support for Travis CI

milliams:
This adds support for running the unit tests on Travis CI which can be seen at https://travis-ci.org/ganga-devs/ganga. This gives us a secondary source of testing on a different system (Travis uses Ubuntu) which reduces the chance of faulty assumptions.

It will only run the unit tests (and maybe in future full GPI tests where limited to `Local`/`Executable`) but its better than nothing. There is a chance of using it in future to test running on OS X too.

I've also just added support for code coverage reporting using [Codecov](https://codecov.io). It's only based on the unit tests but it can supplement any other coverage reporting we have.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: python
+python:
+  - "2.6"
+  - "2.7"
+  - "3.4"
+  - "3.5"
+install: pip install -r requirements.txt
+script: python setup.py tests --type=unit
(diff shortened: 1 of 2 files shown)


milliams:
retest this please

milliams:
Does anyone have any opinions on me merging this? Having quick feedback on the tests from another system can only help us with avoiding introducing bugs.

alexanderrichards:
these changes seem innocuous so I see no reason not to merge

PR #148, opened 2016-01-27
Remove our textwrap copy. It's been in Python since 2.3.

milliams:


rob-c:
I'm in favour of merging this as I suspect that the python implementation is faster than what we have.

alexanderrichards:
ditto

milliams:
It almost certainly is. The copy in our codebase I think was copied from Python 2.3 and was therefore probably a licensing violation anyway (PSFL vs GPL).

rob-c:
ouch, erm, hadn't considered where the origins of that were. I should probably double check the OrderedDict in that case (Ganga/Utility/externals). I've referenced the source material and it's origin but I've not explicitly stated that the drop in code is under MIT vs our codebase being the GPL.

I think once we move to Python 2.7 we can drop this, but until then does anyone feel it's required to expressly state the MIT license here?
It's interesting that even though the source may be licensed this way it itself does not contain the MIT license as I would expect from reading the document so there may be some issues here.

milliams:
I would recommend adding the MIT license test to the top of the OrderedDict module just to be safe. I don't think there should be a problem then.

PR #149, opened 2016-01-27
Make `__construct__` consistent

milliams:
This neatens up the slightly inconsistent usage of `__construct__` across the code base by forcing everyone to always pass a list to it. The list can be empty of course.

Apart from that there were a bunch of overrides of it which were doing nothing so they have been removed for neatness reasons.

Finally, I moved the base class version of `__construct__` from `Node` into `GangaObject` since the code in `Node.__construct__` was not needed and really it's dealing with `GangaObject`-type things.


rob-c:
Thank goodness someone got around to making this consistent :)
I think the only place where the ```__construct__``` is now called is from ```def _init``` in Proxy.py. Do you know if I'm correct?
I'm happy to drop it from here as we fully construct objects there as required now and pass the correct arguments to the constructor.

alexanderrichards:
In the re-write I've been working on we lose `__construct__` anyway so I'm also happy to see it's influence diminished :wink:

rob-c:
I'm keen to see it removed as it appears to at best mimic an ```__init__``` but if it's at least consistent this is a first step :)

milliams:
`__construct__` is also called from IPostProcessor's `postprocessor_filter()` as well as from `GangaDataset`, `Bookkeeping` and `LHCbDataset`. The full list can be found with `git grep "__construct__(" | grep -v def | grep -v super`.

There's some use to `__construct__` so that you can do `LocalFile('~/test.txt')` without having to explicitly pass it as `LocalFile(namePattern='~/test.txt')` and it's used extensively in files and datasets (`git grep "def __construct__"`).

[on a related note I'd love to get rid of the Filter mechanism for constructing objects from strings but that will have to wait]

rob-c:
Shall we merge this for 6.1.15 and get it out of the way?

milliams:
Yes, I'll merge it now and then start the release if there's nothing else.

Issue #73, opened 2015-12-18
New LHCb Gaudi application type


egede:
A new LHCb application type should be created. It should work with the new structure for the private relase areas based on lb-dev and lb-run.

## Connection to old Gaudi types
There should be no connection to the old application types for Gaudi based applications.

## New GaudiRun application type
* The new application type should be called *GaudiRun* (as a way to distinguish it from the existing 
*GaudiPython* type). It will not need any specialisation for Brunel/DaVinci etc.
* The schema will have
  *  is_prepared = None ,
  * extraopts = None ,
  *  args = [-T]  ,
  *  platform = 'x86_64-slc6-gcc48-opt' ,
   * lb-runtOptions = '' ,
  *  location = '' ,
  *  optsfile = []  
 The *location* attribute is where the release is located. There default value should now be empty for this and the job submission should fail if empty.

## Implementation
 * Marco Clemencic will provide script that can encapsulate environment.
 * Marco Clemencic will provide build target which can pack up local shared libraries.
 * Caching of the environment for running should initially be avoided as there are some tricky issues
 * Implementation that works without local shared libraries should be possible initially without the
  build target above in place.

rob-c:
Hi Ulrik,

I'm thinking that this would be a nice required schema component if for no other reason than for helping users with book-keeping and knowing what job was what.

e.g:
```j.application = GaudiRun(app='DaVinci', location='my/DaVinci/path')```

I think if we ask for it we should avoid using it as much as possible in the App to keep the code generic but having this upon construction makes life a bit easier imo.

It would also allow us to use Bender with the new GaudiRun application as it requires different runtime scripts which we'd need to use only with this application.

This is looking like quite a light weight application to write if we are relying on the build targets from the cmake style apps. I'm going to try and duplicate the functionality that I see in the existing Applications and hopefully this shouldn't take too long.

egede:
> I'm thinking that this would be a nice required schema component if for no other reason than for 
> helping users with book-keeping and knowing what job was what.

@rob-c I do not think this works. It is possible with the `lb-dev` command to create release areas that are no longer associated with any of the *traditional* applications.

alexpearce:
[...]

PR #150, opened 2016-01-27
Bugfix/speed up jobs

rob-c:
After profiling the 'jobs' command I've fixed some bottlenecks in the code.

One of the annoying ones is that a lot of transient objects were being constructed when calling what should have been the lightweight 'addProxy' method.

This branch mainly addresses:

Stop creating Objects which will be removed/overwritten in the addProxy step. This effects the proxy ```def _init``` which has been modified.

Some code cleanup in the Registry. This doesn't represent anything more than intelligently not calling the same functions multiple times if not required.

Adding some stronger locking of repository transactions from within the registry. This cleans up the code and adds a second lock around disk transactions to try and mitigate the race-conditions some users see.

Simplifying getName.

Adding more caching to the Schema to reduce the time taken by getting the default Value of an object.

Removing calls to ```getLogger``` within some functions which are called many, many times. It adds a significant weight through frame inspection which can be avoided during running for 99% of users. I've left the code in commented out if anyone needs to re-enable it for debugging purposes but it was turning up in profiling quite a lot.

Redesign of TypeCheck to make it much faster and introduce some basic string lookup/eval caching.

Overall I see a large boost when running 'jobs' and many other commands at least interactively and I suspect when submitting large jobs now.

Rob


milliams:
It looks like your changes to Proxy.py here are similar to those I made in #138. I think it would be a good idea to get #138 reviewed and merged before making these (potentially conflicting) changes on top.

milliams:
The latest commits seem to overwrite many of my changes to Proxy.py and also seem to include your changes to `GANGASYSROOT` etc. I'm assuming this was not intentional?

rob-c:
You've jumped the gun to me explaining what is going on in the branch.
I'm concerned that the tests are failing in horrible ways atm which needs addressing.

I've not reverted all of your changes and I'm intending to be able to run addProxy on a vanilla ```GangaObject``` derivative to add the proxy from scratch if needed. I'm also trying to be explicit in making sure that any new proxy created in ```_impl``` is referenced by the object being returned from addProxy.
I'll update in a few minutes once I've worked out what broke all of the integration tests as this is confusing me somewhat.

As for pulling in #132 The changes are to mainly get rid of problems I'm having in actually running the tests. I'm happy to revert this after the fact but this got committed accidentally.

rob-c:
Also #132 no longer references GANGASYSROOT as per the commits merged here.

rob-c:
OK, all of the tests should be running now and I've fixed a few problems along the way. There are 2 tests failing from Jenkins, but I'm investigating this locally on my setup.

@milliams I've changed the ```GPIProxyObjectFactory``` method I mentioned earlier to add proxies to GangaObjects which all (in principle) can accept them being added.
I've changed the ```def _init``` method to try and make it clearer that there is nothing special about the object passed from ```GPIProxyObjectFactory``` vs the pluginclass() which has been constructed internally.

I am more inclined to drop #132 at this point as the main changes there with regard to the use of ```__file__``` and ```sys.path``` can be hashed out here as I don't see a reason not to keep the integration and unittests setup this way so that they can be run agnostically of where they're run from.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +624,55 @@ def helptext(f, s):
...
 
-        # The first thing we do is, if we were called with GPIProxyClassFactory(_existing=some_ganga_object)
-        # to simply create a link between them and us and nothing else.
-        if '_existing' in kwds:
-            # Link the two together in the proxy layer
-            setattr(self, implRef, kwds['_existing'])
-            setattr(kwds['_existing'], proxyObject, self)
-            return
+        ## Zero-th fully initialize self before moving on
+        super(type(self), self).__init__()
```
milliams: This is potentially dangerous as any subclasses of a proxy class would then enter an infinite loop. Admittedly we don't ever subclass a proxy class but in fact this would prevent it.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -561,16 +570,25 @@ def GPIProxyObjectFactory(_obj):
...
+        if isProxy(_obj):
+            return _obj
+        elif not isProxy(_obj):
+            pass
+    else:
+        pass
 
-    return proxy_class(_existing=_obj)
+    proxy_class = this_class(_proxy_impl_obj_to_wrap=_obj)
+    return proxy_class
```
milliams: I'm not sure what the changes to this function are adding? You seem to have reverted to an earlier version of your code sine these comments have reappeared.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +624,55 @@ def helptext(f, s):
...
-        ## SECOND WE NEED TO MAKE SURE THAT OBJECT ID IS CORRECT AND THIS DOES THINGS LIKE REGISTER A JOB WITH THE REPO
+        if proxy_obj_str in kwds.keys():
+            instance = kwds[proxy_obj_str]
+        else:
+            ## FIRST INITALIZE A RAW OBJECT INSTANCE CORRESPONDING TO 'pluginclass'
+            ## Object was not passed by construction so need to construct new object for internal use
+            instance = pluginclass()
 
         # at the object level _impl is a ganga plugin object
+        setattr(self, proxyObject, self)
```
milliams: We don't want proxy objects to have a `_proxyObject` attrbiute. What purpose could it solve? It would just point to itself.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +624,55 @@ def helptext(f, s):
...
+        instance.__dict__[proxyObject] = self
+        instance.__dict__[proxyClass] = type(name, (GPIProxyObject,), d)
+
+        if proxy_obj_str in kwds.keys():
+            return
+
+        ## SECOND WE NEED TO MAKE SURE THAT OBJECT ID IS CORRECT AND THIS DOES THINGS LIKE REGISTER A JOB WITH THE REPO
+
+        auto_init_str = '_auto__init__'
+        if hasattr(instance, auto_init_str):
```
milliams: *All* `GangObjects` have an `_auto__init__` so we don't need to check for it.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -730,10 +767,15 @@ def _init(self, *args, **kwds):
...
+        #import time
+        #b4=time.time()
         import cStringIO
         sio = cStringIO.StringIO()
         stripProxy(self).printSummaryTree(0, 0, '', out=sio, interactive=interactive)
-        return str(sio.getvalue()).rstrip()
+        returnable = str(sio.getvalue()).rstrip()
+        #a4=time.time()
+        #logger.info("Time Taken to generate printSummaryTree: %s" % str(a4-b4))
+        return returnable
```
milliams: I'd rather not add too much commented-out code if we can avoid it.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -267,8 +276,13 @@ def _getDefaultValueInternal(self, attr, val=None, check=False):
                 # if a defvalue of a component item is an object (not string) just process it as for SimpleItems (useful for FileItems)
                 # otherwise do a lookup via plugin registry
 
+                category = item['category']
+
+                global _found_category
```
milliams: We don't need the `global` here unless we're rebinding it the variable.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -243,14 +248,18 @@ def _getDefaultValueInternal(self, attr, val=None, check=False):
                 defvalue = _found_attrs[stored_attr_key]
             else:
                 defvalue = config[attr]
-                _found_attrs[stored_attr_key] = defvalue
+                from Ganga.GPIDev.Base.Proxy import stripProxy
+                ## Just in case a developer puts the proxied object into the default value!
+                _found_attrs[stored_attr_key] = stripProxy(defvalue)
```
milliams: Does this ever actually happen? It really shouldn't and if it does we should raise an exception and tell the developer off.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -450,7 +464,7 @@ def __check(isAllowedType, name, validTypes, input_val):
         if not isAllowedType:
             #import traceback
             #traceback.print_stack()
-            raise TypeMismatchError('Attribute "%s" expects a value of the following types: %s\nfound: "%s" of type: %s' % (name, validTypes, input_val, type(input_val)))
+            raise TypeMismatchError('Attribute "%s" expects a value of the following types: %s\nfound: "%s" of type: %s' % (name, validTypes, input_val, str(type(input_val))))
```
milliams: This `str()` in unnecessary since `%s` always performs the string conversion.

python/Ganga/GPIDev/TypeCheck.py
```diff
@@ -7,60 +7,83 @@
...
-        except ImportError:
-            isType = isinstance
-
         ## Return None when None
         if _t is None:
             if val is None:
                 return True
 
-        if not isType(_t, str) and isclass(_t):
+        if type(_t) is str:
```
milliams: We shouldn't use `type() is ..` but rather `isinstance`

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I've changed the function to always add a proxy to all GangaObject derivatives whilst returning Proxies which are already assigned to objects. As for where the proxyClass constructor is pulled from we're both using GangaObject ```__class__``` structure to get the corresponding proxy_class (this_class) which is what I'm hoping to avoid.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I was under the impression we don't do this by design unless I'm mistaken. I can change the super call if required as I can't see anything bad in calling it in principle.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: OK, happily removed

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Fine, happy to simply call it directly. Also can change the functions below to call it based upon type

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Fair enough, I've commented this back in, in a private commit anyway as I was seeing bottlenecks in the summary tree function.
Any feeling if this is wanted or not?

python/Ganga/GPIDev/Schema/Schema.py
rob-c: I'm happy to rather test for isProxy and raise an error but I was hoping to avoid this particular problem for now in favour of getting things working.
If we want to take a stance on it, it likely would have turned into more work I have to do on this branch.

python/Ganga/GPIDev/Schema/Schema.py
rob-c: happily removed in this case

python/Ganga/GPIDev/TypeCheck.py
rob-c: happily changed

python/Ganga/GPIDev/Base/Proxy.py
alexanderrichards: my feeling is probably not

python/Ganga/GPIDev/Base/Proxy.py
milliams: The code already does that with (line 564):
```python
if hasattr(_obj, proxyObject):
  return getattr(_obj, proxyObject)
```
as for `getattr(_obj.__class__, proxyClass)` it could just be replaced with `getattr(type(_obj), proxyClass)`.

Currently,:

1. if a `GangaObject` instance is passed then if it already has a proxy object assigned, it returns that.
2. Otherwise, if the object class does not define a `_proxyClass` then it raises an error. 
3. Finally it uses the `GangaObject` instance's type's `_proxyClass` to make a new proxy object which wraps the existing `GangaObject` instance.

Which of those are you changing?

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Fair enough, happily removed as the bottleneck was found and removed.

python/Ganga/GPIDev/Base/Proxy.py
milliams: I wonder if changing it to:
```python
GPIProxyObject.__init__(self)
```
would be more explicit and safer. Otherwise I agree that we never subclass these proxy classes so there's probably no danger.

python/Ganga/GPIDev/Schema/Schema.py
milliams: Is it actually a problem though? Do we have any where in the code, people setting default values to be proxy objects? I feel it would take quite some gymnastics to make happen.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I'm happy to revert 154 over the proposed code at 582, I wasn't sure what this was doing.

I'm still not keen on getting the proxyClass in this way from an object which in principle doesn't need to know about it so I feel my comments stand but I'm happy to get the proxyClass instance this way for now. (I'd rather construct it as required here if needed, or get from a list of known proxy-class types).

If the object is of GangaObject type (we know it is from 560) it currently always has the proxyClass so testing for it feels a bit excessive. I'm also of the opinion that this should be something that this function is concerned with so I'm not wanting to add a reliance on getting things from inside a GangaObject here.
I'd rather just get the proxyClass otherwise than worry about raising an error when it doesn't exist (yet?).

3 doesn't change it simply is a string change which I hold no strong ties to, as long as the string is unlikely to be used by a developer on construction of a raw proxied object from 'in front' of the proxy.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I'm happy to do that in this case :)

python/Ganga/GPIDev/Schema/Schema.py
rob-c: Happily removed

python/Ganga/GPIDev/Schema/Schema.py
rob-c: If a developer adds a possible type of ```Proxy.File``` class and then sets a value of ```File``` in some part of their code the TypeCheck will fail. (```File``` is not an instance of ```Proxy.File```)
(It used to simply ignore this and assume as you're setting an object which looks like it's a Ganga class it's probably OK. I'm happy to revert this functionality but this is adding inconsistent behaviour into TypeCheck imo. If we know it's a Ganga class we stop caring about checking the type?...
).
I've added a better checking mechanism into TypeCheck which checks if the Ganga object is an instance of the class but raised at least one issue so I was wanting to be explicit here.

rob-c: the developer should only be doing this if they're using Ganga.GPI in their code which I would strongly argue is something to be avoided

python/Ganga/GPIDev/Base/Proxy.py
milliams: You're right that at present `_proxyClass` is always present on a `GangaObject`. In future I was planning on not always making it so, such that we could have some `GangaObject`s as non-proxiable (such as `GangaObject` itself, the adapters etc.). In this world we would have to be explicit about which classes we export.

I would want to keep any changes to the proxy system in a separate RP so that we can discuss them in their own context and not mixed up with speeding up running `jobs`.

python/Ganga/GPIDev/Schema/Schema.py
milliams: Exactly, the way to fix this is to remove usage of `Ganga.GPI` in code and to remove any use of `Proxy.File` in type lists of schemas.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Yes I agree, I'm not talking about introducing this here. I'm simply explaining that is the intent behind the comments which I can happily remove as long as we're aware that this change is coming.

python/Ganga/GPIDev/Schema/Schema.py
rob-c: Yes I agree but I was hoping to do this in a future PR vs adding that into this. I can revert it but this may add problems which this is looking to avoid

rob-c:
I should have addressed most (or all) of the issues by now. I would like to see #156 go in first if possible which would allow me to rebase and remove the changes to the tests in this branch

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -13,6 +13,9 @@
 
 logger = getLogger()
 
+_reg_id_str = '_registry_id'
+_id_str = 'id'
+
```
milliams: I'm not a big fan of this technique of putting attribute names in string and using `getattr` everywhere. Unless we're planning on changing the attribute name it's solving a non-existent problem. It's a level of indirection which I'm not sure is worth it given the noise it adds to the code.

I'm saying that we should change the code back or anything I'm just thinking we want to use it with care.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -199,13 +204,32 @@ def __iter__(self):
...
     def __len__(self):
+
+        import os
+        this_time = os.stat(self._jobDirectory).st_ctime
+
+        if len(self._stored_len) == 2:
+            last_time = self._stored_len[0]
+            if this_time == last_time:
+                return self._stored_len[1]
+
```
milliams: What is `_stored_len` and what is it used for. I feel adding code like this should really be accompanied by some comments otehrwise in 5 years, no-one will have a clue (e.g. our conversation in the meeting about the prepared state).

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -564,13 +573,12 @@ def GPIProxyObjectFactory(_obj):
     if hasattr(_obj, proxyObject):
         return getattr(_obj, proxyObject)
 
-    if not hasattr(_obj.__class__, proxyClass):
-        from Ganga.Core.exceptions import GangaException
-        raise GangaException("%s is NOT a Proxyable object" % type(_obj))
+    ## This is defined within Objects.py, we could probably store this elsehere
+    ## (We probably do) but as this is guaranteed to be accessible for GangaObjects I will use this
```
milliams: *Very* minor point: PEP8 [suggests](https://www.python.org/dev/peps/pep-0008/#block-comments) using a single `#` and a space for block comments.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +614,51 @@ def helptext(f, s):
...
         setattr(self, implRef, instance)
-        setattr(instance, proxyObject, self)
 
+        ## Need to avoid any setter methods for GangaObjects
+        ## Would be very nice to remove this entirely as I'm not sure a GangaObject should worry about it's proxy (if any)
+        instance.__dict__[proxyObject] = self
+        instance.__dict__[proxyClass] = type(name, (GPIProxyObject,), d)
+
+        if proxy_obj_str in kwds.keys():
+            return
```
milliams: I feel a comment here to make it explicit that we don't run the rest of the `_init` in the case of proxying an existing object is needed.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: I'll add a doc line if it's not obvious

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +614,51 @@ def helptext(f, s):
...
 
-        # at the object level _impl is a ganga plugin object
+        ## Avoid intercepting any of the setter methos associated with the implRef as they could trigger loading from disk
         setattr(self, implRef, instance)
-        setattr(instance, proxyObject, self)
 
+        ## Need to avoid any setter methods for GangaObjects
+        ## Would be very nice to remove this entirely as I'm not sure a GangaObject should worry about it's proxy (if any)
+        instance.__dict__[proxyObject] = self
+        instance.__dict__[proxyClass] = type(name, (GPIProxyObject,), d)
```
milliams: `_proxyClass` is already set at `GangaObject` construction so I don't think we want to be overwriting it here. If we change how this works later we should make this change then and not now.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +614,51 @@ def helptext(f, s):
...
+
+        if proxy_obj_str in kwds.keys():
+            return
+
+        ## SECOND WE NEED TO MAKE SURE THAT OBJECT ID IS CORRECT AND THIS DOES THINGS LIKE REGISTER A JOB WITH THE REPO
+
+        auto_init_str = '_auto__init__'
```
milliams: You define this and then never use it.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -243,14 +248,22 @@ def _getDefaultValueInternal(self, attr, val=None, check=False):
...
             defvalue = item['defvalue']
+            from Ganga.GPIDev.Base.Proxy import isProxy
+            if isProxy(defvalue):
+                raise GangaException("(2)Proxy found where is shouldn't be in the Config" % stored_attr_key)
+            ## Just in case a developer puts the proxied object into the default value!
             _found_attrs[stored_attr_key] = defvalue
 
         # in the checking mode, use the provided value instead
-        if check:
+        if check is True:
```
milliams: `check` should only ever be a boolean so I don;t know if it's worth being explicit here.

python/Ganga/new_tests/TestProxy.py
```diff
@@ -69,9 +70,14 @@ def test_dict_attributes(self):
 
         # Proxy class instance
         self.assertFalse(hasattr(self.p, '_proxyClass'))
-        self.assertFalse(hasattr(self.p, '_proxyObject'))
+        self.assertTrue(hasattr(self.p, '_proxyObject'))
```
milliams: This probably still needs changing.

python/Ganga/new_tests/TestProxy.py
```diff
@@ -79,7 +85,7 @@ def test_dict_attributes(self):
         self.assertFalse(hasattr(proxied, '_impl'))
 
         # A proxy instance should only have _impl in its dict
-        self.assertEqual(self.p.__dict__.keys(), ['_impl'])
+        self.assertEqual(self.p.__dict__.keys(), ['_impl', '_proxyObject'])
```
milliams: This probably still needs changing.

python/Ganga/new_tests/TestProxy.py
```diff
@@ -69,9 +70,14 @@ def test_dict_attributes(self):
...
         self.assertFalse(hasattr(self.p, '_proxyClass'))
-        self.assertFalse(hasattr(self.p, '_proxyObject'))
+        self.assertTrue(hasattr(self.p, '_proxyObject'))
         self.assertTrue(hasattr(self.p, '_impl'))
 
+        # Proxied class
+        self.assertTrue(hasattr(self.p_class, '_proxyClass'))
+        self.assertFalse(hasattr(self.p_class, '_proxyObject'))
+        self.assertFalse(hasattr(self.p_class, '_impl'))
+
```
milliams: This is doing the same thing as `# Proxy class` a few lines above (and in fact should fail since it's looking for the wrong attributes).

rob-c: Above self.p is an initialized instance self.p_class is a raw class unless we expect for this to be evaluated into an instance?

python/Ganga/new_tests/TestProxy.py
rob-c: Yep, changed locally, will commit

python/Ganga/new_tests/TestProxy.py
rob-c: Wait, there is something concerning here.
If line #73 fails and is assertTrue it created a proxyObject attribute for the instance which effects this test.
(which when 73 was different I had to modify this)

I'm slightly bugged that this will pass/fail or change what is a pass or fail based on line #73. It should stand by itself as a pass/fail.

python/Ganga/new_tests/TestProxy.py
milliams: I mean:
```python
# Proxy class
proxy_class = TestGangaObject._proxyClass
self.assertFalse(hasattr(proxy_class, '_proxyClass'))
self.assertFalse(hasattr(proxy_class, '_proxyObject'))
self.assertTrue(hasattr(proxy_class, '_impl'))
```
Now, if you wanted to change that to be:
```python
# Proxy class
self.assertFalse(hasattr(self.p_class, '_proxyClass'))
self.assertFalse(hasattr(self.p_class, '_proxyObject'))
self.assertTrue(hasattr(self.p_class, '_impl'))
```
that'd be fine.

rob-c: These are 2 different things surely? One is the proxyClass of the class object the other is the class itself. (Granted if/when we get rid of the proxyclass this becomes mute as it just points to itself) I was wanting to be explicit.

python/Ganga/new_tests/TestProxy.py
milliams: No, a proxy object should *never* have a `_proxyObject` attribute. A proxy should just point to its implementation. Nothing should ever add one and if it does then it's wrong.

It used to be different (before my recent changes) but now a proxy object never point to itself (or anywhere) with `_proxyObject`. See http://ganga.readthedocs.org/en/latest/dev/index.html#attributes-on-proxy-objects

rob-c: I'm NOT talking about the Proxy implementation here.

I'm concerned that if line 73 reads assertTrue then I get I get a proxyObject attribute ADDED by the test. It doesn't matter that 73 then fails.
The failing test is somehow adding a ```_proxyObject``` attribute to the object being tested here at line 88.

If line 73 reads assertFalse then this test should NOT have a ```_proxyObject``` and the conversation is mute, I'm concerned about the test behaviour here, not ganga.

python/Ganga/new_tests/TestProxy.py
milliams: Sorry, I got a bit confused. I thought `self.p_class` was a proxy class but in fact it's a standard `GangaObject` since `addProxy` does nothing to class objects? I was confused by the odd name and the unneccesary use of `addProxy` in the `setUp()`.

In that case these new lines are just duplicating
```python
# Proxied GangaObject class
self.assertTrue(hasattr(TestGangaObject, '_proxyClass'))
self.assertFalse(hasattr(TestGangaObject, '_proxyObject'))
self.assertFalse(hasattr(TestGangaObject, '_impl'))
```

python/Ganga/new_tests/TestProxy.py
milliams: Before your changes, line 73 was:
```python
self.assertFalse(hasattr(self.p, '_proxyObject'))
```
because `self.p` is a proxy object wrapping a `TestGangaObject` instance. It should not have a `_proxyObject` since it is itself a proxy.

If you are seeing a proxy object with a `_proxyObject` attribute then the code needs to be fixed, not the test.

If
```python
self.assertFalse(hasattr(self.p, '_proxyObject'))
```
then (it doesn't have `_proxyObject`) and so
```python
self.assertEqual(self.p.__dict__.keys(), ['_impl'])
```
should be fine?

python/Ganga/new_tests/TestProxy.py
rob-c: I'm happy to remove this if the object is being parsed into an instance but I thought we were running these tests as unittests without the full machinery of Ganga which pulls in string parsing from the interpretor?
I'll remove the test save confusion here and write another test to see what's going on with objects when they're constructed.

python/Ganga/new_tests/TestProxy.py
rob-c: The object being tested DOES NOT have a _proxyObject hence the assetTrue test would fail. After the test failing when I test the same object I expect not to have a _proxyObject (still) the assetEqual fails because it now has one. I don't know where it gets it from but it does during the test.

At least when I run it. This is a concerning bug and I don't know what could lead to this, but frankly as I reverted the change I'm letting this one fall by the wayside unless anyone else wants to do any digging.

python/Ganga/GPIDev/Schema/Schema.py
rob-c: Yes but it's callable from external functions so I'd rather be implicit in-case we ever change our way of working with this class.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I'm wanting to remove registry_id and try replacing it with 'id' directly. I don't know how good/bad that would be so I'm wanting to play around with this parameter on a private branch.
Doing this as an exercise showed up at least one bug though so the result is getting committed even if it doesn't effect much

rob-c:
OK, there is currently a problem with the tests, once I've fixed it this should be ready to go

rob-c:
OK, I believe that should fix most of the problems. I'm running the tests locally now and I'll see if I agree with the Jenkins tests.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +614,53 @@ def helptext(f, s):
...
 
-        # at the object level _impl is a ganga plugin object
+        ## Avoid intercepting any of the setter method associated with the implRef as they could trigger loading from disk
         setattr(self, implRef, instance)
-        setattr(instance, proxyObject, self)
+
+        ## Need to avoid any setter methods for GangaObjects
+        ## Would be very nice to remove this entirely as I'm not sure a GangaObject should worry about it's proxy (if any)
+        instance.__dict__[proxyObject] = self
+        instance.__dict__[proxyClass] = type(name, (GPIProxyObject,), d)
```
milliams: Only thing left, do we want to be overwriting this? We already set it in the metaobject `__init__` so I'd rather not repeat that here. In future it might make sense to move it here but for now we should just do this in one place.

rob-c: Given that this should probably be done here I'll comment it out as currently done in GangaObject but I'm strongly opposed to it being done there at all but we'll leave this until another PR

milliams: I think commenting it out is a good idea for now but the question of where and to what we add proxies needs some careful thought.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -606,40 +614,54 @@ def helptext(f, s):
...
-        # at the object level _impl is a ganga plugin object
+        ## Avoid intercepting any of the setter method associated with the implRef as they could trigger loading from disk
         setattr(self, implRef, instance)
-        setattr(instance, proxyObject, self)
+
+        ## Need to avoid any setter methods for GangaObjects
+        ## Would be very nice to remove this entirely as I'm not sure a GangaObject should worry about it's proxy (if any)
+        instance.__dict__[proxyObject] = self
+        ## THIS SHOLD BE DONE HERE BUT IS DONE IN GANAGOBJECT, PLEASE ADDRESS
+        #instance.__dict__[proxyClass] = type(name, (GPIProxyObject,), d)
```
rob-c: One would imagine we're adding a Proxy at the level of ```addProxy``` but apparently pre-processing for this action goes on elsewhere.

rob-c:
OK, all of the tests are passing for me, I'm ready to go

rob-c:
@milliams Shall I merge?

milliams:
The tests pass so it gets my :+1:

PR #151, opened 2016-01-27
Bugfix/add MIT to Odict

rob-c:
Adding the MIT license to the ODict. This will be removed post 2.7 but I would like to be more explicit that this isn't necessarily GPL code. (Although I believe it's GPL compatible?)

Please wait to review this until #132 is in as my git history is slightly messed up and short of re-basing branches and moving individual commits between branches it's best to deal with multiple PR in a sane order.
(My apologies)


milliams:
Indeed it is compatible with the GPL but we currently say that "Ganga is GPL" but if it contains MIT code then that isn't quite true.

rob-c:
I'm left wondering what the exact license is, I don't know how heavily this borrows from Python 2.7 which isn't MIT so there is the possibility for some problems there.
The best solution is to move to Python 2.7 and then this discussion becomes mute, however at least for the time being I've removed our GPLv2 code from being directly interfaced with the MIT code and we simply pull in an use this object as is.
I think we still have freedom to modify/distribute under MIT and I think we're not claiming to have written this code or hold copyright on it. I don't know how the MIT/GPL/Python languages all play when you consider it's all processed at runtime. If it was pre-compiled and linked against I think we're fine so if the .pyc contains instructions from MIT files that we use then this seems like less of an issue, but if the .pyc from our GPL code contains some instructions from the MIT code explicitly then I think there are problems.

In short I don't think it's a major issue as the official Python docs point users to use this for their projects but I'm wanting to try and be as reasonably correct as possible.

egede:
We are moving in the correct direction with the license issue by going to Python 2.7. Can I please ask that we do not worry more about this transient issue for now (and spend time on our other problems).

rob-c:
any objections to merging this and closing this?

milliams:
I was waiting for all the unrelated commits which were mixed up in this PR to be cleared. As it stands, I'm happy.

:+1:

PR #156, opened 2016-01-28
Feature/run tests from anywhere2

rob-c:
This PR is taking relevant code from #132 which should allow for the unittests to be run from any location.

I'm keen to get this in before #150 unless there are any objections?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,16 +7,23 @@ import sys
 
 ganga_python_dir = os.path.abspath('python')
 sys.path.insert(0, ganga_python_dir)
-from Ganga import _gangaVersion
+if not os.path.isdir(ganga_python_dir):
+    ganga_sys_root = os.environ.get('GANGASYSROOT', None)
+    if ganga_sys_root is not None:
+        python_rel_path = 'python'
+        ganga_python_dir = os.path.abspath(os.path.join(ganga_sys_root, python_rel_path))
+        if not os.path.isdir(ganga_python_dir):
+            python_rel_path = '../install/ganga/python'
+            ganga_python_dir = os.path.abspath(os.path.join(ganga_sys_root, python_rel_path))
 
+sys.path.insert(0, ganga_python_dir)
+from Ganga import _gangaVersion
 
 def readme():
     import os.path
     filename = 'README.rst'
     if not os.path.exists(filename):
-        import inspect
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -5,18 +5,16 @@
...
 
-
 def readme():
     import os.path
     filename = 'README.rst'
     if not os.path.exists(filename):
-        import inspect
-        script_path = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
-        filename = os.path.join(script_path, filename)
+        filename = os.path.abspath(os.path.join(ganga_python_dir, '..', filename))
```
milliams: This could just be
```python
filename = os.path.abspath(os.path.join(file_path, filename))
```

setup.py
```diff
@@ -42,8 +40,19 @@ def finalize_options(self):
...
+
+        test_env = os.environ.copy()
+        ## Make sure that the PYTHONPATH is correct for the tests to pick up 'import Ganga'
+        test_sys_path = ''
+        for _dir in sys.path:
+            test_sys_path = test_sys_path + _dir + ':'
+        test_env['PYTHONPATH'] = test_sys_path
+
+        return test_env
+
```
milliams: Do we actually need to do this? If the `cwd` of the `subprocess.check_call` is set to the `python` directory then we don't need to set it in the Python path since Python always looks in the CWD.

rob-c: This is so that I can run ```python $SOMEPATH/setup.py tests``` without this the env is not setup correctly

milliams: Are you sure? If the working directory is set properly then the `import Ganga` stuff should be fine. What error do you see?

rob-c: Certain. In the LHCb environment without it I get "ImportError: No module named Ganga.GPIDev.Base.Objects" when I don't modify the PYTHONPATH here.

setup.py
rob-c: I'll make the change but it shouldn't effect much

rob-c:
Any reason not to commit this? I'm keen to get the PYTHONPATH change in so I can keep running the tests locally as I do development.

milliams:
:+1: Go for the merge but I'm very surprised that the CWD isn't affecting things as it should. I will need to look into this afterwards.

PR #159, opened 2016-01-28
Bugfixes/remove annoyance

### LINKED ISSUE #152: Selecting on status produces warning
When running jobs.select(status='blah') https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py#L137 produces an error (although the selection is still successful) that is confusing for users.

rob-c:
Fixes #152, based on #150 so wait until that's in and I'll rebase


milliams:
In future I'd *really* like to avoid these PRs with changes from all other branches mixed in. They should be reviewed independently and without forcing other PRs in first. It shouldn't be difficult to just take a branch off `develop` before starting each new PR?

rob-c:
@milliams I would have changed it if #150 was going to take more than ~5-10min to finish but I was working on my branch which didn't have massive unfixed bottlenecks which made working stressful.
As it stands this a 1 file 2 commit pull.

Any objections to the merge?

milliams:
Go for the merge.

PR #161, opened 2016-01-28
Feature/test for lazy loading

rob-c:
This is adding a new unittest for testing the lazy loading of jobs.

I've also put in a fix for a bug that sometimes occurs when looking at lazyily-loaded jobs.

Don't know what's required for adding this into the testing on Jenkins but it works for me on my system.

### DIFF
--- a/python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
+++ b/python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
@@ -1,0 +1,69 @@
+from __future__ import absolute_import
+
+from ..GangaUnitTest import GangaUnitTest
+
+class LazyLoadingTest(GangaUnitTest):
+
+    def setUp(self):
+        """Make sure that the Job object isn't destroyed between tests"""
+        super(LazyLoadingTest, self).setUp()
+        from Ganga.Utility.Config import setConfigOption
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
+
+    def test_a_JobConstruction(self):
+        """ First construct the Job object (singular)"""
+        from Ganga.Utility.Config import getConfig
+        self.assertFalse(getConfig('TestingFramework')['AutoCleanup'])
+
+        from Ganga.GPI import Job, jobs
+        j=Job()
+        self.assertTrue(len(jobs) == 1)
+
+    def test_b_JobNotLoaded(self):
+        """ Second get the job and check that getting it via jobs doesn't cause it to be loaded"""
+        from Ganga.GPI import jobs
+
+        self.assertTrue(len(jobs) == 1)
+
+        print("len: %s" % str(len(jobs)))
+
+        j = jobs(0)
+
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+        raw_j = stripProxy(j)
+
+        has_loaded_job = raw_j._getRegistry().has_loaded(raw_j)
+
+        self.assertTrue(has_loaded_job == False)
+
+    def test_c_JobLoaded(self):
+        """ Third do something to trigger a loading of a Job and then test if it's loaded"""
+        from Ganga.GPI import jobs
+
+        self.assertTrue(len(jobs) == 1)
+
+        j = jobs(0)
+
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+        raw_j = stripProxy(j)
+
+        ## ANY COMMAND TO LOAD A JOB CAN BE USED HERE
+        raw_j.printSummaryTree()
+
+        has_loaded_job = raw_j._getRegistry().has_loaded(raw_j)
+
+        self.assertTrue(has_loaded_job == True)
+
+    def test_d_JobRemoval(self):
+        """ Fourth make sure that we get rid of the jobs safely"""
+        from Ganga.GPI import jobs
+
+        self.assertTrue(len(jobs) == 1)
+
+        jobs(0).remove()
+
+        self.assertTrue(len(jobs) == 0)
+
+        from Ganga.Utility.Config import setConfigOption
+        setConfigOption('TestingFramework', 'AutoCleanup', 'True')
+
--- a/python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
+++ b/python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
@@ -17,13 +17,13 @@ class LazyLoadingTest(GangaUnitTest):
 
         from Ganga.GPI import Job, jobs
         j=Job()
-        self.assertTrue(len(jobs) == 1)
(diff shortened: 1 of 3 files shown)


python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
```diff
@@ -0,0 +1,69 @@
...
+        print("len: %s" % str(len(jobs)))
+
+        j = jobs(0)
+
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+        raw_j = stripProxy(j)
+
+        has_loaded_job = raw_j._getRegistry().has_loaded(raw_j)
+
+        self.assertTrue(has_loaded_job == False)
```
milliams: There's an `assertFalse` for these situations

python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
```diff
@@ -0,0 +1,69 @@
...
+
+        has_loaded_job = raw_j._getRegistry().has_loaded(raw_j)
+
+        self.assertTrue(has_loaded_job == True)
+
+    def test_d_JobRemoval(self):
+        """ Fourth make sure that we get rid of the jobs safely"""
+        from Ganga.GPI import jobs
+
+        self.assertTrue(len(jobs) == 1)
```
milliams: There's an `assertEqual` for these situations.

python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
rob-c: It's a first draft to see if it worked, I'll spruce it up tomorrow at some point

python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
```diff
@@ -0,0 +1,69 @@
...
+
+        from Ganga.GPI import Job, jobs
+        j=Job()
+        self.assertTrue(len(jobs) == 1)
+
+    def test_b_JobNotLoaded(self):
+        """ Second get the job and check that getting it via jobs doesn't cause it to be loaded"""
+        from Ganga.GPI import jobs
+
+        self.assertTrue(len(jobs) == 1)
```
milliams: I assume by this that the tear down (which wipes the repo etc.) is only run between runs of test classes and not between test functions? Otherwise it would have been cleared. I forget how these things work.

milliams:
Apart from the minor things of more precise assert types I'm happy for this to be merged. I guess we should wait until the release is finalised though.

The tests you added in this should have been run as part of the [test run](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/84/) on this PR automatically so all should be good.

python/Ganga/new_tests/GPI/Loading/LazyLoadingTest.py
alexanderrichards: Nope, `setUp` and `tearDown` run per test function. If you want you can give the class methods `setUpClass` and `tearDownClass` which work on a class basis.

milliams: Ahh, so it's because we have `setConfigOption('TestingFramework', 'AutoCleanup', 'False')` set I guess?

rob-c: I think it would be nice to add at least a tearDownClass function to obliterate the test repo once we're finished with it.
Are we definitely using only Python 2.7 on the Jenkins test box yet? If so I'll add a suitable step here to make sure that there are no possible conflicts between different repos.

@milliams Are some of the unittests running in parallel? If this is done are the tests all running on the same repo or separate repos?

milliams: We're still using 2.6 on Jenkins since for now we need to make sure that it works there. I am tempted however to make unittest2 a requirement for the testing so that we can use all the nice features in Python 2.7's `unittest` module

Currently all tests are running in series so they are all running on the same repo (`~/gangadir_testing/`). In the medium-term I'd like to investigate running every test on a separate repo (so `GPI/Loading/LazyLoadingTest.py` would run with `gangadir` as `~/gangadirs/GPI-Loading-LazyLoadingTest/` or something). This way it will be possible to run multiple tests in parallel and if one is a long running Grid job then it won't hold anything up.

rob-c: This could be accomplished with some well placed config commands in one of the setUp or setUpClass I think. I'll look into this.
I'd be tempted to just add a hard remove command into a tearDownClass which uses os level commands to remove the files from disk after Ganga has shutdown.

milliams: I agree on both points. Getting the fully-resolved name of `self` in `GangaUnitTest.setUp` (like in this [SO answer](http://stackoverflow.com/a/2020083/96704)) should allow us to specify a gangadir uniquely.

I think that a full remove of the gangadir after the test has fully finished is not a bad idea. For non-multipass tests it might actually be good to wipe it after each test function though. It needs a bit of thought to do this in a consistent and obvious way.

This is for a separate PR though.

rob-c:
OK, changes should be enough to get this merged

milliams:
I'm happy for this to be merged once this release if completed.

milliams:
I've just tested this locally and while the test passes, if I run the commands interactively:
```python
j = Job()
```
[exit and restart Ganga]
```python
j = jobs(0)
raw_j = stripProxy(j)
raw_j._getRegistry().has_loaded(raw_j)
```
it prints `True`. This is with a wiped gangadir and not running any other commands.

So there's something up with the testing differing from the runtime. Do you see this behaviour on your local checkout?

rob-c:
do you have the monitoring loop running?

rob-c:
fyi ```Localhost``` from the backend string can't be parsed into ```Local``` so Local type jobs have to be loaded so that the backend can be inspected within the monitoring system.
Not the biggest of bugs as the local backend is the machine you're on so the job should be relatively lightweight

milliams:
I expect that I do have the monitoring loop running as I have not disabled it. It is indeed a `Local` job so your assessment might well be correct (and indeed I see the correct behaviour for the `Interactive` backend).

I guess this is another reason to remove the `GangaObject._name != GangaObject.__name__` thing which seems to only cause problems for almost no gain.

rob-c:
It will require some minor education of the users but I think the gains for us outweight the inconvenience of doing a tab-complete.

We can always add a hook to add the class twice to the GPI for compatability if the ```_name``` and ```__name__``` are different?

rob-c:

milliams:
:+1:

PR #163, opened 2016-01-29
Cleanup/move proxy stuff into proxy

rob-c:
Hi all,

I'm expecting this to cause some debate as it's proxy related, but I'm very keen to get this PR in at the start of 6.1.16 as I think it helps smooth some things out.

This patch is intended to remove the standing proxy object from the Objects.py file and make the exporting of objects to the GPI more robust.

I've added ```getProxyClass``` to ```Proxy.py```.
This function will return a proxy object if the class has been expressly written with one. Aka the repositories.
If the class is a GangaObject subclass it will add a proxy as required constructing one on the fly.

If a ```GangaObject``` is exported to the ```GPI``` a proxy is automatically added.
If a class which is a subclass of ```GangaObject``` is exported to the ```GPI``` a proxy is automatically added to the class (if it's needed) through ```getProxyClass```.
If anything else is exported it is just exported to the ```GPI``` as normal.

I've moved the code which deals with constructing the proxy of object purely into Proxy.py as suggested above.

I have also removed the setting of ```_proxyObject``` for a ```GangaObject``` instance.
The only reason this was referred to in the Objects.py was to make sure that we avoid copying it and ending up in an inifinite loop.
I appear to be able to submit jobs and work with Ganga just fine without it so I'm strongly of the mind to promote dropping this entirely from within ```GangaObject```.

Unless someone knows of a good reason why the ```GangaObject``` should be able to access it's proxy?

In summary:

```GangaObject()._proxyObject``` I want to get rid of (I'm happy to keep it but I don't personally like it.
```GangaObject._proxyClass``` I want to keep as an optional property of the class. This allows for classes such as the Registries to overload the proxy with something more customised to their needs.
```exportToGPI``` will now automatically add proxies to objects and classes that need them when it's called.

I am now able to do a direct ```from Ganga.GPIDev.Base.Objects import GangaObject``` at the top of ```Proxy.py``` so I'm feeling a bit better about how these 2 systems are now separated,

I'm happy to have some conversation about keeping stubs for ```_proxyObject``` in the ```GangaObject``` and a few details but I'm keen to get most of this in as it gets rid of some horrible circular dependencies.

### DIFF
--- a/python/Ganga/Core/GangaRepository/SubJobXMLList.py
+++ b/python/Ganga/Core/GangaRepository/SubJobXMLList.py
@@ -232,7 +232,10 @@ class SubJobXMLList(GangaObject):
     def __len__(self):
         """ return length or lookup the last modified time compare against self._stored_len[0] and if nothings changed return self._stored_len[1]"""
         import os
-        this_time = os.stat(self._jobDirectory).st_ctime
+        try:
+            this_time = os.stat(self._jobDirectory).st_ctime
+        except OSError:
+            return 0
 
         if len(self._stored_len) == 2:
             last_time = self._stored_len[0]
--- a/doc/dev/index.rst
+++ b/doc/dev/index.rst
@@ -40,13 +40,11 @@ Attributes on proxy objects
 
 Proxy classes and the object that they are proxying have a set number of attributes which should be present.
 
-If a ``GangaObject`` class has been registered as being proxyable then it will have the property ``_proxyclass`` set on it which will point to the relevant :class:`~.GPIProxyObject` subclass.
+If an object inherits from ``GangaObject`` the class can have the property ``_proxyClass`` set which will point to the relevant :class:`~.GPIProxyObject` subclass. This is created on demand in the ``addProxy`` and ``GPIProxyObjectFactory`` methods.
 The proxy class (which is a subclass of ``GPIProxyObject`` and created using :func:`~.GPIProxyClassFactory`) will have the attribute `_impl` set to be the relevant ``GangaObject`` subclass.
 
 When an instance of a proxy class is created, the `_impl` attribute of the instance will point to the instance of the ``GangaObject`` that is being proxied.
-This ``GangaObject`` instance will have an attribute ``_proxyObject`` which points back to the proxy.
 
-Simply making an instance of a proxyable ``GangaObject`` class will not automatically create a proxy object though and so will not have ``_proxyObject`` present.
 
 Repository
 ----------
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -209,7 +209,11 @@ class Registry(object):
         while this_id in self._inprogressDict.keys():
             logger.debug("Getting item being operated on: %s" % this_id)
             logger.debug("Currently in state: %s" % self._inprogressDict[this_id])
-            time.sleep(0.05)
+            #import traceback
+            #traceback.print_stack()
+            #import sys
+            #sys.exit(-1)
+            #time.sleep(0.05)
         self._inprogressDict[this_id] = action
         if this_id not in self.hard_lock.keys():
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

File: python/Ganga/Core/GangaRepository/SubJobXMLList.py [hunk 1]
```diff
@@ -232,7 +232,10 @@ def __get_dataFile(self, index, force_backup=False):
     def __len__(self):
         """ return length or lookup the last modified time compare against self._stored_len[0] and if nothings changed return self._stored_len[1]"""
         import os
-        this_time = os.stat(self._jobDirectory).st_ctime
+        try:
+            this_time = os.stat(self._jobDirectory).st_ctime
+        except OSError:
+            return 0
```
milliams: This doesn't seem related to this change at all? I think this is from one of your other PRs? Could we keep the PR clean please?

milliams:
Overall this is a good idea but I do have some thoughts.

`_proxyObject` is currently used so that if you create a proxy object, strip the proxy and then add a proxy again it will just reuse the existing proxy object rather than creating a new one. I'm not sure how much this really save as in theory this should only happen when crossing the proxy barrier which is all in the user's control. I'm happy either way but I do like this caching which could be significant in loops etc.

I'm not sure we always want to be able to add a proxy to any `GangaObject` subclass. Some of them like `GangaObject` itself or the adapters should not have proxies added. Perhaps since this isn't differentiated in the current code should be a change for later?

rob-c:
The 'proxy' for the GangaObject subclass is the ```_proxyClass``` not anything new.

I've been careful with both ```_name``` and ```__name__``` conventions as per the final state of the PR

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Frankly this was making it impossible to test this PR so either I hold off on the PR or don't move forward. It's effectively a 1 line fix anyway.

rob-c:
My strong feeling toward the ```_proxyObject``` classes is that it should be so lightweight that we don't miss it not being passed around with the actual ```GangaObject```. We always have to make one when constructing an ```GangaObject``` from 'in front' of the Proxy layer anyway so this is no more of less expensive with this PR.
The only potential performance loss is due to possible multiple decelerations of ```_proxyClass``` objects.

We could be intelligent and store a dictionary of proxied classes (```_proxyClass```) and perform a dictionary lookup whenever we need one, but I suspect that there would be little performance loss anywhere as I don't know of anywhere in the code that these are constructed many, many times rather than in ```exportToGPI```.

milliams:
OK, in that case I'm largely convinced about the removal of `_proxyObject`. we should bear it in mind in the future as a potential optimisation if it is necessary.

One thing I would ask is that you update the dev guide at [doc/dev/index.rst](https://github.com/ganga-devs/ganga/blob/develop/doc/dev/index.rst) with this information so we make sure that things don't get out of sync and make it part of the PR?

rob-c:
This patch now does away with proxyObject entirely unless someone strongly pushes for it to be re-introduced.
I've not added a dictionary cache for ```_proxyClass``` classes as I'm not convinced this is worth the complexity of adding a cache for this.

I tried updating the documentation and then lost the file I'd spent 2hr editing so I'll get back to this in the near future

rob-c:
I'm keen to get this into 6.1.16 asap, are there any objections to merging this branch?

milliams:
I'd like to see the tests pass before you merge but apart from that it's a :+1:

rob-c:
A reasonable request :) I'll run them locally to check in that case too

rob-c:
OK, I think I'm seeing a problem here in copying attributes which should be proxied. I'll have a look into this before pushing to merge the branch

rob-c:
OK, I think I've fixed this branch, I'll wait for the automated tests to confirm then I'll move to merge either this-evening or tomorow

milliams:
Tests pass so I'm :+1:

PR #165, opened 2016-01-29
Bugfix/fix for glitches in repo

rob-c:
This PR fixes some bugs I've been seeing in private testing. These are all repo handling bugs including https://github.com/ganga-devs/ganga/commit/ddcd3cec49261069cdb426259131cdd0780f614c which was exposed in my private testing.

I would prefer to get these fixes into 6.1.15 if there is time to go to a second pre as (at least) one of these bugs will prevent me from releasing to LHCb.

Apologies that these weren't found sooner but it was only from interactive use that some of these problems were apparent.

### DIFF
--- a/python/Ganga/Core/GangaRepository/SubJobXMLList.py
+++ b/python/Ganga/Core/GangaRepository/SubJobXMLList.py
@@ -232,7 +232,10 @@ class SubJobXMLList(GangaObject):
     def __len__(self):
         """ return length or lookup the last modified time compare against self._stored_len[0] and if nothings changed return self._stored_len[1]"""
         import os
-        this_time = os.stat(self._jobDirectory).st_ctime
+        try:
+            this_time = os.stat(self._jobDirectory).st_ctime
+        except OSError:
+            return 0
 
         if len(self._stored_len) == 2:
             last_time = self._stored_len[0]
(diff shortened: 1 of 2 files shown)


python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -407,10 +413,13 @@ def getAllCachedData(self):
         #logger.debug( "Cache: %s" % str(self._subjobIndexData.keys()) )
         if len(self._subjobIndexData.keys()) == len(self):
             for i in range(len(self)):
```
milliams: This is what `enumerate()` is for.

milliams:
I'll need to check the core tests but at least the ATLAS test are going to need another run because of a bug in `bootstrap.py` so I'm happy to get this included in 6.1.15. It's possible to make PRs against other branches (i.e. directly into `release/6.1.15`) but it really doesn't matter in this case. I'll merge `develop` into the release branch once this is merged and run another set of tests on Monday.

I'm pretty much happy for you to merge when you're ready.

:+1:

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Probably. But I don't know 100% for certain that all that enumerate calls is len here as I don't want to iterate over the class at all in this step for a few minor reasons. I'll do some investigating in the future but as it stands this PR isn't responsible for this so I'm going to do this elsewhere.

rob-c:
I know it's technically possible to do a PR against any branch but my experience with this has been varied. Merging now.

PR #167, opened 2016-02-02
This reverts 320e247e1ad which was causing stale data to be displayed

milliams:
Commit [320e247e1ad](https://github.com/ganga-devs/ganga/commit/320e247e1add935c09dfe3dede1b87e305f4ce66#diff-7056ed3fcb714dad4d540fcd686efe61R403) changed the behaviour of `j.status` so that it was *always* preferring the index cache (if present) over the real data, even if there was real data present and so was presenting stale information. This is what's been causing a lot of the problems of jobs 'reverting' to `submitted` when they should be `running` or `completed` etc. (at least as far as my testing can see).

The check for the index cache is done in `Descriptor.__get__` anyway so this shouldn't cause full loading.

In discussion with Mark we agreed that there shouldn't be any cases where the index cache is populated on a fully-loaded job, especially one that's only just been created in memory but it *was* happening. That will have to be a fix for another time.

This PR is against the release branch so I'll see if that works correctly.

If people could review this one quickly so that it can get put into the release, that'd be great.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -401,8 +401,6 @@ class Job(GangaObject):
         # Attempt to spend too long loading un-needed objects into memory in
         # order to read job status
         if name == 'status':
-            if stripProxy(self).getNodeIndexCache():
-                return stripProxy(self).getNodeIndexCache()['display:status']
             return object.__getattribute__(self, 'status')
 
         # FIXME Add some method of checking what objects are known in advance of calling the __getattribute__ method


drmarkwslater:
Looks good for me!

rob-c:
As long as this doesn't break the lazy loading I'm happy

rob-c:
Can we just generate this indexcache on the fly in future. It will get rid of problems like this in a much better way by making sure that the indexcache is never out of date (which is a separate bug imo)

drmarkwslater:
[this may go beyond the scope of the PR btw!]

Personally, I think we should explicitly **NOT** have an index cache available if the job is fully loaded simply because then we've got explicit duplication of data and this could lead to similar problems of which is being read. I would say that either:

- We have just one _dict (or whatever) and a flag that says it's fully loaded or not. If you request data outside the lazy loaded data, it triggers a full load, changes the flag and overwrites the cache.
-  We have two dictionaries but we explicitly blank the cache if the Job is fully loaded.

We then restrict the lazy loading/caching to:

- Ganga start up: If the index cache is present on disk, load that. If not fully load the job
- When anything changes in the job, fully load it (or automatically fully load it from monitoring) and blank the cache (both in memory and on disk) to avoid confusion
- At Ganga shutdown, any fully loaded jobs get written out along with their index cache.

That's my 2 cents anyway :)

rob-c:
Yes Mark, but then we have to worry about setting and clearing the index cache correctly under different conditions which in my mind introduces more problems.

In the getIndexCache methods I suggest checking if the job is loaded or not in the registry (now possible).
If it is loaded, attempt to generate the cache from memory.
If it's not then first attempt to get the data required from the 'true' index cache which is stored elsewhere, and if that fails revert to loading the job and re-generating the index cache.

Relying on the index cache being empty or the job keeping track of itself leads to design headaches which are better solved by a direct question to the registry to determine if the job is loaded.
The GangaObject shouldn't really have to care about this if the Node is where the IndexCache is stored imo.

I'm extremely against trying to load a job from disk at the start of Ganga. This leads to some severe performance problems with large jobs that I've been trying to mitigate.
Ganga shutdown already does this.

drmarkwslater:
>Yes Mark, but then we have to worry about setting and clearing the index cache correctly under >different conditions which in my mind introduces more problems.

But isn't the the cache just created at Ganga startup (and no other time) and cleared when you fully load the job? I may be wrong but I can't think of any other circumstances...

>Relying on the index cache being empty or the job keeping track of itself leads to design headaches >which are better solved by a direct question to the registry to determine if the job is loaded.
>The GangaObject shouldn't really have to care about this if the Node is where the IndexCache is stored >imo.

But don't we have more headaches if we're relying on two data sources (both in memory and on disk) that are storing the same data always being in sync? This sounds like a worse problem to me :)

>I'm extremely against trying to load a job from disk at the start of Ganga. This leads to some severe >performance problems with large jobs that I've been trying to mitigate.
>Ganga shutdown already does this.

I was only suggesting fully loading jobs that had an active status though this would get done automatically when the monitoring loop got to it anyway though I guess.

I think discussing this in the meeting is the best option - I'm afraid I can't make this Thursday so maybe next week? Or organise a separate meeting?

rob-c:
The index is created as part of the flush process as well when writing the job.

I'm extremely happy to remove the cache once the job has loaded as I don't want to see 2 dictionaries with the same information either.
With that in mind I still think this should be taken care of by the getIndexCache method which returns the index cache of the job (created on the fly) when the job is loaded and never consult a secondary cache used for lazy loading.

But I don't think that there should be commads who call 'getIndexCache' that should ever care. They just want the index cache so we should generate and provide it rather than what we're currently giving them ```None or ```{}```.

Yes I'm happy to see the job loaded when in the monitoring loop or when we need information about the job we don't have when it's needed. I just don't think that loading a job if the index is missing/corrupt, 'just-in-case' is a good thing to be doing.
It helps lead to the issues I'm seeing now where semi-loaded jobs will overwrite the 'corrupt/bad' job on disk and I see data loss.

I'll come up with a PR for what I'm proposing at some point as I'm envisioning this taking all of about 20 lines of code so I think this would be better discussed there.

rob-c:
@milliams please feel free to commit from me :)

PR #168, opened 2016-02-02
Bugfix/ganga init transfer

### LINKED ISSUE #125: No Ganga/__init__.py shipped with job

drmarkwslater:
This fixes #125 by basically removing the whole mechanism and just sending the 'files.py' file and referencing it appropriately. This **may** bite us in the future as we can no longer 'easily' send Ganga modules with jobs. However, users can't do this anyway and we've gone this far without needing to send any more so I think it's worth the risk :)


milliams:
retest this please

milliams:
While I am in general in favour of merging this I think we should wait until we have sorted out the other issues we're working on at the moment. I would also want to run some manual LCG/Dirac tests to make sure it's still working there, Do people agree?

drmarkwslater:
That sounds like a good idea to me - let's make sure the new tests are working and private testing is going well, etc. first.

drmarkwslater:
retest this please

drmarkwslater:
Any problems with merging this?

milliams:
Since we don't yet have any automatic tests for remote jobs (LCG or Dirac) I'd like to make sure that this hasn't broken anything there. Apart from that, I'm happy.

drmarkwslater:
Just run a private test for both Condor and LCG and both worked without issue. OK to merge?

rob-c:
I don't think we make use of any of this in Dirac so go from me

alexanderrichards:
:+1:

milliams:
:+1:

PR #172, opened 2016-02-02
Bugfix/fixing bugs in 6.1.15 testing

rob-c:
I've fixed some bugs I came across in testing ahead of 6.1.15.

This includes some fixes into GangaList, Proxy and the Registry.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -365,22 +365,24 @@ class Descriptor(object):
             return cls._schema[name]
 
         if self._getter_name:
-            return self._bind_method(obj, self._getter_name)()
+            returnable = self._bind_method(obj, self._getter_name)()
+            if isinstance(returnable, GangaObject):
+                returnable._setParent(self)
+            return returnable
+
 
         # First we want to try to get the information without prompting a load from disk
 
         # ._data takes priority ALWAYS over ._index_cache
         # This access should not cause the object to be loaded
         obj_data = obj.getNodeData()
-        if obj_data is not None:
-            if name in obj_data:
-                return obj_data[name]
+        if name in obj_data:
+            return obj_data[name]
 
         # Then try to get it from the index cache
         obj_index = obj.getNodeIndexCache()
-        if obj_index is not None:
-            if name in obj_index:
-                return obj_index[name]
+        if name in obj_index:
+            return obj_index[name]
 
         # Since we couldn't find the information in the cache, we will need to fully load the object
 
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -962,8 +962,7 @@ class GangaRepositoryLocal(GangaRepository):
     def unlock(self, ids):
         released_ids = self.sessionlock.release_ids(ids)
         if len(released_ids) < len(ids):
-            logger.error(
-                "The write locks of some objects could not be released!")
+            logger.error("The write locks of some objects could not be released!")
 
     def get_lock_session(self, this_id):
         """get_lock_session(id)
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -990,7 +990,8 @@ class Registry(object):
                 oid = self.find(obj)
                 if getattr(obj, _reg_id_str) in self.dirty_objs.keys():
                     self.repository.flush([oid])
-                    del self.dirty_objs[getattr(obj, _reg_id_str)]
+                    if getattr(obj, _reg_id_str) in self.dirty_objs:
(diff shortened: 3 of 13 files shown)


milliams:
Cheers @rob-c I'll merge these in and run another batch of testing.

rob-c:
@milliams I've got a small addendum to make before the merge, spotted a few more things, I'll update this once they're in

milliams:
Ok, let me know when they're ready.

rob-c:
OK, that should be it. Hopefully this should push up the passing test count a little in the next test cycle :)
Feel free to merge when ready

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -365,22 +365,24 @@ def __get__(self, obj, cls):
             return cls._schema[name]
 
         if self._getter_name:
-            return self._bind_method(obj, self._getter_name)()
+            returnable = self._bind_method(obj, self._getter_name)()
+            if isinstance(returnable, GangaObject):
+                returnable._setParent(self)
+            return returnable
```
milliams: Do we really want to be setting the parent here? If there are cases where the parent of an object is not set then we should fix that at its source.

rob-c: I would have assumed that an object created in a getter method doesn't necessarily have to know about it's parent but I would have thought it nice to know the parent of such objects which aren't necessarily going through a setter method.
e.g. if I write a getter method to return a new Backend rather then always storing one (for whatever reason) I could imagine situations where you'd want to know which job object created this backend.
(The parent will never be set if the object is never assigned to the class within the getter)

I'm happy to revert if this causes issues but I was playing around with some classes and this cropped up as possible problem that I saw.

milliams: I feel it would be technically cleaner if we just made sure that the parent was being set correctly everywhere but I see your point that the getter can be a special case. There's no need to revert and I will merge this soon.

milliams: In fact this change has caused some big problems since it's setting the parent of returned object to be `self` which is a ` Descriptor object. I will have to revert or change this as it's breaking 6.1.15.

rob-c: In this case shouldn't it be ```_setParent(obj)``` rather than not setting the parent if the object is a GangaObject, otherwise it could (does) have the parent of ```None```?
I'm curious how you spotted this breaking anything?

milliams: It seems like sometimes `obj` is a proxy object and doing `stripProxy` didn't seem to work properly. It think the reason this got past the testing is that I didn't rerun the new tests on the release branch after merging this and somehow the legacy tests weren't hitting this (or they were and that's what was causing the LHCb problems).

rob-c: I think this is a specific bug that some piece of code is inserting a proxy object into Core where it shouldn't be for the obj to be proxied here.
If this is related to GangaLHCb then it's likely a LHCb bug which I'd say we fix in 6.1.16.

milliams: Given this bug is causing infinite loops and hanging when submitting even simple test jobs I'd like to push out 6.1.16 ASAP, maybe even today. It's causing python/Ganga/new_tests/GPI/TestArgSplitter.py to hang forever (reproducible on the interactive prompt).

rob-c: If this is bad enough to push out a second release so quickly can we at least add an issue to find out how/why this fairly innocuous line is causing infinite loops.
I'm happy to see a 6.1.16 without it but I think this warrants a closer look after this

rob-c: Is this happening in vanilla Ganga?

milliams: It's causing infinite loops because `Job._getMasterJob` is apparently:
```python
def _getMasterJob(self):
    parent = self._getParent()
    while parent is not None:
        if isType(parent, Job):
            break
        parent = self._getParent()
    return parent
```
which loops forever if it doesn't find a parent of the correct type which is a very silly way of doing it without proper checks in place.

rob-c: This leads me to suspect that a job is being created without knowledge of it's parent using a getter method then? This confuses me as to why this would be a good idea to have this happen.

milliams: The problem is with your code which sets the parent in `__get__`, it overwrites any existing parent which is there already. So objects which had correct parents are getting their parents changed to something wrong. At least if we remove this code then and existing parents will remain which up until now has worked fine.

The code you added to `__get__` here is only needed is the getter is creating a new object and returning it parentless which I guess does not happen very often.

rob-c: OK, if you want to revert this I'll make a more advanced check to only assign parents if they're missing.

PR #175, opened 2016-02-04
Fix objects' parents being set to Descriptors

milliams:
It was causing the parent of a returned object to be set to the `Descriptor` rather than the object itself. There's still some problems with proxies existing here so for now a revert is safest.


rob-c:
I'd somehow feel happier if we had an ```assert(returnable._parent is not None)```, although maybe for 6.1.16 as I would like to catch any objects being created out of the blue with no idea of what parent object they belong to.

rob-c:
Feel free to merge asap from me

PR #176, opened 2016-02-04
Typo fixes in Registry code, 6.1.X compatibility fixes to LHCbTasks code

olupton:
- Fix two typos in Registry.py
- Fix import and when a warning is printed in LHCbTransform code
- Adjust handling of mc_num_units so the parent transform's GaussSplitter can be used to change the first event number.


rob-c:
Thanks for the fixes :) I'll update in a bit my bugfix branch as I think we have a few fixes in core due to go in asap

PR #179, opened 2016-02-05
Bugfix/lhcbfixes

rob-c:
Hi all,

These are the fixes that I've been working on to debug (amongst other things) the LHCbTasks system.

Some of these changes will effect others but this is mostly bugfixes or minor changes to the code.

Everything seems to be working with this branch and it's able to submit some jobs

Thanks,

Rob


rob-c:
@milliams Are there any objections to merging this? The only testing problems I see are the tests timing out, when I run locally everything appeared OK

milliams:
I'm getting multiple failures locally and at least some of the CI failures are not timeouts such as [Savannah44116](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/130/testReport/junit/GPI.Bugs.Savannah44116/Savannah44116/Savannah44116/) and [Savannah4781](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/130/testReport/junit/GPI.Bugs.Savannah47814/Savannah47814/Savannah47814/)

rob-c:
I'll do some investigating and see if this is due to this PR for me then.

rob-c:
Are you certain these tests are failing they aren't for me but they seem to fail right after a wait until completed function.
(which is odd as both tests take <20sec when I run them on lxplus)
I'll do some digging but I don't think the PR is to blame for these failures. (I've been wrong before)

milliams:
These tests are definitely timing out for me when running the full suite as `python setup.py tests --type=all`. The should not be timing out as they should only take a fraction of a second to run so waiting 60+ seconds is obviously a sign something is broken. It could be broken in the monitoring or it could be caused by the `GangaUnitTest` setup/teardown but if it works fine in `develop` but not in this branch then the branch is causing some change.

Running the test individually as `nosetests python/Ganga/new_tests/GPI/Bugs/Savannah15630.py --testmatch="(?:\b|_)([Tt]est|Savannah|JIRA)"` works fine so it's probably something to do with the registry?

rob-c:
@milliams they work fine for me when doing ```nosetests Ganga/new_tests/...../SomeTest.py```
but fail if I execute a combined nosetests command across multiple files the tests timeout. (I'm suspecting a monitoring bug)

Surely each test should be run in an isolated way to protect against bugs spilling over if this is the case?

The only changes to do with the registry have been to cleanup the ```if self.hasStarted()``` tests which should not be effecting performance. (and a minor fix related to removing jobs from a repo which by design can't effect what is going on here).

milliams:
Indeed they *should* be isolated but apparently they are not. There is various code in `GangaUnitTest` which tries to reset the state of the registries/repositories/monitoring but given that those systems weren't written with the idea of resetting it's apparently very fragile. Regardless of the reason, we can't merge this branch until it is sorted out. I can try to take a look this afternoon if you can't work it out before.

rob-c:
This still happens when the tests are isolated enough that they run in a unique directory to one another.

The Monitoring service should not persist beyond a ```tearDown``` as I call the standard atexit functions which should kill all running threads and flush and close the registries.

With that in mind; there are problems with some Monitoring threads not exiting cleanly. (I suspect I understand why in another branch but have yet to backport the 'fix' to something closer to develop)
I'll have a stab at implementing this and see if it helps.

I've even got a tearDownClass method in my copy of GangaUnitTest which wipes the repo from disk every time Ganga exists which still doesn't solve this problem, so the problem doesn't appear to persist on disk.


I appreciate merging this branch is a pain if it causes issues from the unittests persective but if the unittests are broken by the _fixes_ in this branch then I'm beginning to think this should be a separate issue to address what is broken there. Otherwise I risk adding a lot of unittest changes to this code which weren't intended to be part of this branch.

milliams:
I agree. I suggest starting a new branch from `develop` which fixes the test suite. Once that's working we can merge it into `develop` and then also into this branch to get things working here.

I appreciate that it's annoying that test suite brokenness is slowing down the fixes in this branch but we can't let ourselves get into the situation where we have *any* failing tests in `develop` (even if it's the tests themselves that are broken).

rob-c:
tbh I'm tempted to just run the tests with something similar to:

```for i in ./Ganga/new_tests/GPI/Bugs/Savannah1*.py; do nosetests ${i}; done```

This works but I've no idea how it would(or not) integrate with Jenkins.

I've got some ideas on how to debug the sleep_until function as I've been exploring the monitoring loop recently but I've got a few other tasks to be getting on with so I'll have a look at this later today at some point

**Comment by rob-c** [comment]
Tbh, I'm planning to do the work on here, cherry-pick the fixes once I have them and create a separate PR and I'll try to avoid polluting this PR with those changes before that PR goes in.

**Comment by drmarkwslater** [comment]
Just a quick note: I don't think this will have a significant bearing on this PR but after discussions with Matt, this seems like a good way forward: #180. In other words, don't spend a lot of time trying to fix GangaTestCase as hopefully we'll remove it in the not too distant future!

**Comment by rob-c** [comment]
@drmarkwslater There is very little not understood in the GangaUnitTest it only protects against things like running the bootstrap twice.

There is a definite bug in the Monitoring however as the monitoring is displaying the jobs here as 'submitted' and is refusing to progress them to a further state despite being capable of doing so.

This doesn't appear to be related to anything in the GangaUnitTest unless I'm mistaken as this is a problem during a unittest with just 1 test inside it so should have the same environment as a 'normal' interactive ganga session.

This bug in the monitoring is very similar to other strange behaviour I've seen elsewhere with the system so I'm very keen to understand it.

**Comment by rob-c** [comment]
OK,
I seem to tracked this down and as a result I'm going to take a break for a few hours before I clean up my code and commit the changes.

Simply put I was seeing a re-entrant lock in the registry causing issues. This was causing single threads to lock up entirely even through they should have had full access to acquire a the lock and proceed normally.
(How much of this is due to our extensive use of the python locking system I don't know, this could very well be a bug of some sort external to us).

That being said I see no real reason for the re-enterant lock within the Registry any more.
I've since implemented a classic hard ```threading.Lock``` around anything which causes disk transactions due to people getting repository/xml corruptions and this only locks soft methods such as ```__len__``` and requesting file locks (which are transactions locked at the filesystem level anyway).

With this in mind I have removed the ```RLock``` from my Registry and don't appear to see any ill effects. In-fact I'm seeing the contrary, everything seems to be better behaved at the monitoring level.

I've a lot of code to clean up before I push this patch so I may need to do this tomorrow morning after a rest.

**Comment by milliams** [comment]
That's great that you've found the potential source for the bug. If you've got a possible fix then feel free to take your time committing it. As long as I know it will on the way in the next day or two, the release can wait.

**Comment by rob-c** [comment]
I've updated this branch to match trunk and I don't think there are any major issues here.
I'm keen to get this PR into 6.1.17 work as soon as 6.1.16 is released so the fixes here can be advertised to LHCb as a Dev release.

**Comment by rob-c** [comment]
@milliams This branch is failing with the codecov test but when I click the link I see a passing tick on the site. Is this a problem with the PR or the test?

**Comment by milliams** [comment]
The codecov failure is simply telling you that this PR has reduced the coverage overall by 0.01% and it's counting that as a 'failure'. This is completely ignorable and does not count against the PR in my opinion. The fact that the link takes you to a page with a passing tick is likely due to the fact that there's some small uncertainty in how it measures coverage and so sometimes it's coming up as -0.01% (fail) and sometimes as -0.00% (pass).

Secondly I should note that the coverage only counts the 'unit' tests' coverage so it doesn't account for the many integration tests we have.

**Comment by rob-c** [comment]
retest this please

**Comment by rob-c** [comment]
I'm looking to merge this branch again asap, I'm going to try and look through the code changes later today but all of the tests are passing and this is mainly fixes for LHCb so I don't expect too much to change.

File: python/Ganga/GPIDev/Adapters/IPostProcessor.py [hunk 1]
```diff
@@ -85,8 +85,8 @@ def __str__(self):
 
     def append(self, value):
         self.addProcess(value)
-        self.process_objects = sorted(
-            self.process_objects, key=lambda process: process.order)
+        if hasattr(self.process_objects, 'order'):
+            self.process_objects = sorted(self.process_objects, key=lambda process: process.order)
```
milliams: Out of interest, which postprocessors don't have an `order` parameter?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -877,7 +877,8 @@ def _auto__init__(self, registry=None, unprepare=None):
         self._setDirty()
 
         super(Job, self)._auto__init__()
-        stripProxy(self.info).uuid = str(uuid.uuid4())
+        if self.info is not None:
+            stripProxy(self.info).uuid = str(uuid.uuid4())
```
milliams: Would it make sense instead to ensure that a job always has a `JobInfo` class in `self.info`?

milliams:
I'm also curious as to what bug changing `defvalue=[]` to `defvalue=GangaList()` in every schema fixes?

rob-c:
This fixes storing of any proxied objects passed into the schema by accident.
Ideally the proxy would take care of stripping things but we have other problems with regard to proxies lists which I just want to avoid by moving to GangaLists for now.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Probably, but I don't want to go patching another subsystem of the Job object.

python/Ganga/GPIDev/Adapters/IPostProcessor.py
rob-c: I couldn't tell you but the issue of the post-processors has come up before now and I'd rather look at it in a PR dedicated to changes of that system.

rob-c:
Are there any objections to merge? There is some outstanding work to be done with regard to job attributes which I think should be handled in separate PRs but I'm keen to get this in as it's the main part of 6.1.17 from an LHCb perspective imo.

python/Ganga/GPIDev/Adapters/IPostProcessor.py
milliams: I just want to make sure we're not just hiding a real problem which will end up showing up in other, more subtle ways.

drmarkwslater: Actually, I can give you an example right now :smile: :+1: 
```
j = Job()
j.postprocessors.append( Notifier() )
```

Now this is obviously another bug but I agree with @milliams that if this code was added, we could get some odd behaviour...

milliams: So this looks like it's probably a proxy-layer violation? Somehow, proxied objects are being passed to this function.

rob-c:
OK, I've addressed some issues and the tests pass, any objections to getting this merged so I can push for wider testing?

drmarkwslater:
I personally can't see a problem with merging but as it's a large set of changes, do we want to leave off until tomorrow at the meeting to make sure everyone's OK'ed it? I think this is the major PR for 6.1.17 (excluding whatever I end up doing for fixing Tasks :smile: ) so it might be worth being a bit more cautious. Of course, if everyone chimes in here and now then I'd say go ahead!

rob-c:
OK, well I'll likely keep this updated with changes from develop so I may just move lhcb _HEAD_ to this branch ahead of the merge in that case, I'm keen to get feedback from users to help fix any more problems.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -184,7 +184,7 @@ class Job(GangaObject):
 
     _schema = Schema(Version(1, 6), {'inputsandbox': FileItem(defvalue=[], typelist=['str', 'Ganga.GPIDev.Lib.File.File.File'], sequence=1, doc="list of File objects shipped to the worker node "),
                                      'outputsandbox': SimpleItem(defvalue=[], typelist=['str'], sequence=1, copyable=_outputfieldCopyable(), doc="list of filenames or patterns shipped from the worker node"),
-                                     'info': ComponentItem('jobinfos', defvalue=None, doc='JobInfo '),
+                                     'info': ComponentItem('jobinfos', defvalue=JobInfo(), doc='JobInfo '),
```
milliams: I'm not sure that this should be needed. The default value of `defvalue` is `None` and that should trigger the logic in `Schema._getDefaultValueInternal` to load a default-constructed object of the default plugin for the category. If this isn't working then it suggests a deeper problem.

Alternatively setting it to `defvalue='JobInfo'` should work but it *shouldn't* be needed.

rob-c: And again I'm getting the strong opinion that this is a mess to get the default value for an object over just writing some python code.
I don't know where or how the category will decide that a JobInfo is the best thing to put here as there is simply no mention of a category or equivalent.

This is done elsewhere for the JobTime etc so I'm sticking with this if the object isn't being correctly assigned. Do we ever expect this not to be valid?

milliams: The category for the `ComponentItem` in this case is `'jobinfos'` and the default plugin for a category is the first one put into it. Now, since `JobInfo` is the only class in the plugin category `'jobinfos'` it will be the default and therefore the instance created. I'm not a big fan of the plugin system but that is my understanding of it. If this isn't working then there is some deeper problem.

rob-c:
OK, To kick off some discussion of this PR for 6.1.17 final:

Are there any sections of code or changes which people have objections to?
If so I'm happy to revert these particular changes (if they're small) and leave these bugs as unfixed but I'd like to get in as much as possible from this PR into 6.1.17.

Thanks,

Rob

drmarkwslater:
It would be nice to split the Task stuff into a separate PR as I think that would make things easier to review but from what I can tell it should be OK to merge. Anyone else any thoughts?

milliams:
I'm still concerned about the `JobInfo` problem as well as the underlying bug behind the reason for needing all the `GangaList` changes. I feel these are both just hiding underlying problems and need some investigation.

rob-c:
The GangaList changes happen largely behind the scenes anyway and a blanket solution to going through many methods and checking that they employ stripProxy correctly where appropriate.
This can be changed in the future but solves many headaches for people _now_.

As for the jobinfo yes, there may be a deeper problem but users ultimately want to submit jobs.
I'll try getting rid of this one and running the examples from @olupton to see if this has gone away as a result of other fixes in core since this started (now quite some time ago).

As for splitting the PR, can do if people feel really strongly about it. But It'll lose the history of the development as I'm not intending to track down each commit/fix I'll just break out the final set of changes which represent <400 lines of code changed.

egede:
So is this going in in full or not. Seems like it is time to decide.

drmarkwslater:
I would think so as I don't have strong objections. However, I'd have this be the last merge before release to make sure we have all tests running properly.

milliams:
Personally without knowing what it actually fixes it's difficult to say. There's a lot of code churn here and I'm still convinced that it's necessary. The `GangaList` changes still feel like they're just hiding a problem.

There's clearly some good stuff in here like the changes to `LHCbUnit.py` but there's so much other stuff mixed in it's hard to work out what's doing what. I'd suggest putting some of these changes in (being sure to include those from Olli) but we need to understand all the changes we are making.

rob-c:
The GangaList changes are hiding the fact that the interfaces in the Tasks system (Core and LHCb) are passing around proxied and un-proxied object inconsistently atm. This reduces the problem as the GangaList has it's own layers of stripProxy before storing elements.
Yes It would be ideal to track this down and clean every interface but frankly we should add a proxy decorator which classes can import and then it becomes a much quicker task to fix what's going on.

A good 75% of the code is needed to get the LHCb tasks working as expected but I was seeing deadlocks in testing I was unable to get past.
Hopefully this has improved based upon latest changed but I still need to get the XML tests fixed then I'll come back to this.

This branch also adds a minor change to the Schema to allow the use of GangaList in the Schema. The only real changes dropped in a minimal implementation are some switches to GangaList in the tasks section, the code 'cleanup' in the Repo and a few large sections of commented out code being removed.

My proposal is that I'd like to merge this if the deadlocks in #271 have gone away in recent develop I'd like to push for merge. If #271 is still a problem then this code (which largely deals with bugs only experienced by Tasks) is not worth putting in as users won't be able to use it.

eta about another hour or so before I can say if #271 is an issue

milliams:
I feel, as discussed in #184 that we should move away from the use of `GangaList` behind the proxy and instead move towards using it like it was intended - as a way of safely storing a collection of `GangaObject`s in GPI-land while maintaining the proxy-safety. Making changes to increase the usage of `GangaList` is just making the code more complicated when, if we had no `GangaList`s in Python-land many things would become better understood.

One of the first things I would fix in this PR is changing all the `from Ganga.GPI import Job; j = Job()` to importing the real `Job` object and just using that.

rob-c:
@milliams Frankly I have more issues with 'getting rid' of GangaList as we start having headaches such as object parent and other stuff getting to be an annoyance. I would like to fight for GangaList to be more widely used as well as having something like GangaDict in the future for similar reasons.
This is a conversation better left for a round table meeting in person imo.

I would really NOT like to start muddling about with asking for a Job behind the scenes unless you want the Tasks system to start caring about the correct use of ```_auto__init__``` and at that point we're starting to ask 'is the Tasks is handling jobs correctly?' which starts to become a complete overhaul.
(Not to mention tasks not by default having access to anything generated from within the proxy using filters)

If this is starting yet another conversation, we can drop this until 6.1.18 as this minor set of fixes in the Tasks will quickly start to escalate from getting it to work into a full overhaul.

python/Ganga/GPIDev/Adapters/IPostProcessor.py
```diff
@@ -83,10 +82,11 @@ def __str__(self):
         else:
             return str(GPIProxyObjectFactory(self.process_objects))
 
-    def append(self, value):
+    def append(self, _value):
+        from Ganga.GPIDev.Proxy import stripProxy
```
egede: @rob-c The import doesn't work. Should it not be 
```

```

rob-c: Indeed, thanks for testing, as it stands I see other problems with this PR which may be show stopper. We also need a proper PR looking into the use of post-processors as they're unfortunately very fragile and have been brought up at least 3 times during 6.1.17 development/bugfixing

egede:
I changed to 6.1.18. Let us get the other few things in and get the release out!

rob-c:
I would prefer to see this go in for 6.1.18. This branch gets the tasks into a 'working' state despite not fixing some underlying problems. I am more keen to get something in a working state out to users to get people to finally move off 6.0.44 and prior 6.1.xx versions unless we're confident we can focus the main (all) efforts for 6.1.19 to getting the tasks fully working.

rob-c:
This PR was left open until the LHCb tasks were working. At the moment they're throwing an error on launch from develop.
Are we fixing LHCbTask for 6.1.19? in which case I should probably extract out the LHCb fixes and fold them back into a new PR.

Either way the PR in this form isn't going to be merged but will probably lead to a newer slimmer PR to address some of the unique problems identified by LHCb Task users.

milliams:
There's definitely some things from this PR which we would want to use. Perhaps extracting the changes relevant to fixing LHCb tasks would be a good start. I think having them working for 6.1.19 would be a good idea if we can manage it.

rob-c:
So we are aiming to fix LHCbTasks for 6.1.19 then?

In this case I'll spend the time to break out the relevant changes into a smaller PR as a lot of this is either defunct or has been superseded by a lot of work which has come later than this.

This will take me at least a day if not until Monday as I was hoping to work on the GangaLSST branch I still have locally.

milliams:
@drmarkwslater might have an opinion but given that we're now getting core tasks working again, it makes sense to have LHCb tasks working too. We can defer it to 6.1.20 if we want but I think it's something that we *must* fix before 6.2.

rob-c:
Apologies, I'm not aware of 6.2 but Fixing these things as of 6.1.20 yes. I'm happy to delay until 6.1.20 for LHCbTasks I just want to prioritise other work if possible.

drmarkwslater:
With #377, LHCb Tasks should now be fixed as well (though I'm willing to be proved wrong!).

rob-c:
Superseded by other work, Closing

PR #182, opened 2016-02-09
Bugfix/fix repo lockup bug

rob-c:
This PR addresses the issue where ```_checkBackend``` method within the Monitoring Loop would randomly lock up.

It appears that a combination of this method on Monitoring thread, the JobsRepisotory and the Registry access on a ```MonitoringWorkerThread``` was causing the repository to enter a state where the re-entrant locking stopped behaving as expected.
It's unclear if this is actually a re-entrant lock being caught between threads requiring a task to be completed but it appears that the registry is performing a task on the worker thread and then gets stuck and unable to proceed.

I've traced the full monitoring loop through the tests with the current thread name on in the debug logging. (this was difficult to parse but hopefully worth the effort). This showed the worker threads to stall at a lock statement in the registry and never seem to progress any further even though it looks like all other threads (and the main thread) have moved onto other tasks.

As it stands I've since implemented hard non-rentrant locks around methods such as flush and load to avoid in-thread corruption issues so these older locks aren't needed here. Other than that these RLocks are mainly around _relatively_ safe commands such as ```__len__``` etc, which if we knew them to be particularly unsafe i.e. length of a repo when jobs are being added, we can in future add a 'better' locking mechanism at the registry level.

Given this was causing the backend monitoring JobAction in the monitoring loop to lock and only seems to effect this I'm hoping this _might_ have actually fixed some of the mystery behaviour in the Monitoring.

Edit:
In addition I've included some minor changes in the monitoring loop which were helpful as well as some minor changes to the GangaUnitTest to allow for unittests to run in their own directory(registry) to isolate them from problems or corruptions from other jobs.


python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -81,16 +86,16 @@ def _execUpdateAction(self):
             if self.should_stop():
                 break
 
-            tpFreeThreads -= 1
             #setattr(threading.currentThread(), 'action', action)
             log.debug("Qin's size is currently: %d" % Qin.qsize())
             log.debug("%s running..." % threading.currentThread())
-
+            self._alive = True
```
milliams: How does this differ from [`is_alive()`](https://docs.python.org/2.6/library/threading.html#threading.Thread.is_alive)? It seems like a potentially confusing naming.

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -82,8 +83,12 @@ def start_ganga():
 
     # [PollThread]autostart_monThreads=False has turned this off being done automatically.
     # The thread pool is emptied by _ganga_run_exitfuncs
-    from Ganga.Core.MonitoringComponent.Local_GangaMC_Service import _makeThreadPool
-    _makeThreadPool()
+    from Ganga.Core.MonitoringComponent.Local_GangaMC_Service import ThreadPool, _makeThreadPool
+    if len(ThreadPool) > 0:
+        del ThreadPool[:]
+        ThreadPool = []
```
milliams: I have a feeling that this line will perform a name rebinding, not reassign the module level variable to point to something else. It shouldn't be needed anyway if the previous line is doing its job.

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -169,3 +186,13 @@ def tearDown(self):
...
         sys.stdout.flush()
+
+    @classmethod
+    def tearDownClass(cls):
+        if cls.wipe_repo is True:
+            import shutil
+            try:
+                shutil.rmtree(self.gangadir)
+            except:
+                pass
```
milliams: This is better done with a `ignore_errors=True` for the `rmtree` call.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
rob-c: tbh, a global counter was even more confusing, especially when it wasn't ever reset to zero on ganga exit/cleanup. I can change the name but as you're talking about different classes and have an extra '_' I was thinking this was enough.

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -5,7 +5,7 @@
     import unittest
```
milliams: If we're using `tearDownClass` etc. then this will have to become:
```python
try:
    import unittest2 as unittest
except ImportError:
    import unittest
```
or similar so it can continue to run on Python 2.6.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: The _makeThreadPool was never intended to wipe/clear the ThreadPool only to append to it, which is what concerned me about it's behaviour.
This seemed to work, although I can just as easily change the method to reset the thread pool list rather than raise an exception which it now does.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
milliams: What I mean is: what does `self._alive` mean? Is it related to the standard Python `threading` concept of `is_alive()` or is it a different thing? If it's different then I think that calling it something which makes this clear is a better idea.

milliams:
Overall this PR seems like a good idea. I like the idea of removing as much locking as we can since locks (and [especially recursive ones](https://groups.google.com/forum/#!topic/comp.programming.threads/tcrTKnfP8HI[51-75]) are fraught with danger.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: I was under the impression this simply wasn't called if you were running python 2.6 (if it works with unittest2 then I'm happy to add this as it fixes some problems)

rob-c:
Basically as a result of this work I no longer trust RLocks, especially for anything critical.
@milliams I'll make the changes you suggest and update the PR.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: We're already doing this from what I can see aren't we?

milliams: Ah, yes. GitHub has truncated the file and so it looked like we weren't. All is good.

rob-c:
@milliams This branch should be working as expected and passing all of the continuous tests.

Any objections to merging this branch so we can begin doing further testing on it?

This commit also appears to be associated with a LARGE performance improvement.

I can submit and have 100 GangaLHCb Dirac executable test jobs finish in <30min which is faster than previous releases by a loooong way.

rob-c:
And by finish I mean marked as fully completed correctly within Ganga with a large bottleneck being that I only ran with 3 threads available on the queues system :)

milliams:
I'm concerned that the tests took so long to run. On Jenkins they took 10 minutes when they normally take around 2 or 3. I'll have to run them locally to check it wasn't just transient slowness.

rob-c:
tbh I'm seeing this branch significantly reducing overhead, unless the running of the monitoring thread and the tearDownClass add an overhead to the unittests.
Is this a reason to hold on merging this?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -169,3 +182,10 @@ def tearDown(self):
         stop_ganga()
         import sys
         sys.stdout.flush()
+
+    @classmethod
+    def tearDownClass(cls):
+        if cls.wipe_repo is True:
+            import shutil
+            shutil.rmtree(self.gangadir, ignore_errors=True)
```
milliams: This will probably not work since this is a `classmethod` and so does not have access to `self`.

rob-c: Fixed the typo. Interesting we have lost the logging by this point so debugging that may have been difficult

rob-c:
Fixing the typo in the tearDownClass took a few seconds off running a random test for me, does this help?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -169,3 +182,10 @@ def tearDown(self):
         stop_ganga()
         import sys
         sys.stdout.flush()
+
+    @classmethod
+    def tearDownClass(cls):
+        if cls.wipe_repo is True:
+            import shutil
+            shutil.rmtree(cls.gangadir, ignore_errors=True)
```
milliams: I'm not sure this will still work as in `setUp` you're setting an *object* attribute `gangadir` but here you are reading the *class* attribute `gangadir` which are different. Put a `print(cls.gangadir)` in `tearDown` and run `nosetests` with `--nocapture` to see that it's always set to `None`.

milliams:
Even after reverting all the changes to `GangaUnitTest` locally it is still taking 3 times longer than it used to so it's definitely some change either in the Registry or, more likely, in the monitoring.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: Well in that case I'm happy to drop the method in this case, the main change I wanted was in setUp anyway.
Unless you can devise a fix, I'm assuming ```self.__class__.gangadir = something``` should work well enough?

rob-c:
If this is slower due to the monitoring now working, I'm calling it an improvement and moving on

python/Ganga/new_tests/GPI/GangaUnitTest.py
alexanderrichards: hi all, came to convo late so may not have all the facts but why not set it in a setUpClass classmethod?

```python

@classmethod
def setUpClass(cls):
    cls.gangadir = <blah>
```

milliams: Since it's not trivial to get working I suggest we just remove this function for now. I feel the correct way to fix this would be a basedir which is set on the class with test gangadirs inside it so we can just wipe the basedir each time. Something to sort out later though.

rob-c:
@milliams OK the teardown is now working as expected, although some error is preventing the repo from being wiped from disk. I don't want to replace it with an ```rm -fr``` but I'm tempted

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: Frankly I'd rather keep it as it should work bar a file remove command not working.
This isn't failing to get the correct arguments to shutil.rmtree so I would like to keep this for the time being

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -101,12 +106,23 @@ def _execUpdateAction(self):
...
 def _makeThreadPool(threadPoolSize=THREAD_POOL_SIZE, daemonic=True):
+    global ThreadPool
+    if ThreadPool and len(ThreadPool) != 0:
+        #from Ganga.Core.exceptions import GangaException
+        #raise GangaException("Cannot doubbly init the ThreadPool! ThreadPool already populated with threads")
+        log.debug("Found a thread pool already in existance, wiping it and startig again!")
+        for i in ThreadPool:
+            del i
+        del ThreadPool[:]
+        ThreadPool = []
```
milliams: It is indeed these changes which is now causing the test suite to run a lot slower. If it's fixing things then I agree that it's not itself a problem as 'correct' beats 'fast'. Two things though. Firstly, I believe that:
```python
for i in ThreadPool:
    del i
del ThreadPool[:]
ThreadPool = []
```
could all just be replaced with
```python
del ThreadPool[:]
```

The second and most important thing to realise here though is that just because the threads have been 'deleted' (had this reference removed) that doesn't mean that they will stop running. If one if them was doing something which is hanging (maybe talking to DIRAC is taking many minutes) then running this code will not halt that process, it will continue to run in the background. This could cause situations where we have two DIRAC monitoring threads running simultaneously.

In Python there is essentially no way to kill a thread dead if it's busy. The only safe thing to do in reality is to wait for the threads to to finish.

rob-c: This could be replaced if it was actually working in the disable and stop methods but as this ```del ThreadPool[:]``` is not working there we're left with a TreadPool object when starting up again.
I suspect this could be a case of the object being marked for deletion but still existing as ThreadPool is not set to ```[]``` at the end of the shutdown.

milliams: My point still stands that we need to be careful to realise and make note of that fact that even though this clears the thread pool, all the background threads will run to completion, no matter what we do. The only *safe* course of action is to wait on the threads finishing.

rob-c: The threads should have been joined and exited and cleaned up within the Monitoring loop of the previous run. If they have not this is a definite bug in the shutdown methods.
The array is (or should be by design) empty but shouldn't be made larger by effectively restarting ganga. The array should contain 4 possible running threads if this is requested by the user not (4* number of times ganga started the monitoring).
Whether I delete it or simply not add to it should be immaterial but I'm wanting to add new threads as the threads from the prior run should have been terminated and destroyed not, simply restarted here.

milliams: Ok, that makes sense. I guess the only thing left is some debug statements notifying if `stop_and_free_thread_pool` did not manage to close all its threads before timing out but that's not something that's needed here.

rob-c:
OK, I'm going to add back in a new lock for the setter and flush methods to avoid job object corruption before this PR gets accepted. ETA a day or so maybe to test this.

rob-c:
OK, I've added a heartbeat to the running threads in the monitoring loop which should throw a warning if they stall for 180sec or more.
I've also added a basic lock to the setter and registry.flush methods.

This is safe enough to submit a simple job and see it complete but I've not tested it yet with anything complex like job splitting or assigning complex attributed to the job so this may fail in spectacular ways if these are used.
I'm trying to give some thought to the problem of locking a parent whilst we set and object within an object and from what I can think we surely shouldn't be doing this?
(Maybe we do by accident with GangaList)

Most of the time we do something like:
```
j=Job()
j.application=DaVinci()
j.application.opt=blah
```
or
```
j=Job()
dv=DaVinci()
dv.opt=blah
j.application=dv
```

Even for:
```
j=Job(application=DaVinci(opt=blah))
```

This is behind the scenes no different to the above.

I suspect we could be doing something bad with GangaList but even there we _should_ be creating a new object (without knowledge of the parent) and then only once setting the object to belong to the parent once it's been modified in some way.

rob-c:
I'll try and do some more testing this-evening/tomorrow and see if this is good enough to ask for merger.

rob-c:
Minor update this branch seems safe enough to split a large Bender job and submit it as well as having the flush/write-protection lock from the setter.

This appears also to have kept most of the performance improvements which is nice even though this is secondary to getting this working safely.

splitting a job of 8k LFN into ~80 subsets and submitting them in about 20min which is nice.

I'll check the jobs flush and do everything else they should before I call this ready for merging but the simplest case of a lock per GangaObject-per-Repo seems sensible :)

rob-c:
OK,

This branch creates a separate lock for the top most parent of a GangaObject instance (i.e. the ```Job``` in ```jobs``` or the ```Task``` in ```tasks```). This lock is then used to lock GangaObjects from changes whilst setting an attribute or flushing the object to disk.
This is also safe enough to provide a lock for GangaObjects which haven't yet been assigned to a registry. The Registry id or the ```id(myObj)``` is used to identify a unique lock.

There are also additional child locks created an used which make use of the parent to create a separate lock. This second lock is only used to prevent getting an object whilst it is being set.
This is due to the fact it is perfectly reasonable to want to get objects which are being flushed but not get objects which are being manipulated at this time.

The various locks are stored in dictionaries within GangaObject and are accessed using static functions ```aquireParentLock```, ```releaseParentLock```, ```acquireChildLock``` and ```releaseChildLock```, which _for now_ are also in GangaObject.
These static functions make use of inspecting the parent object and in the case of the child lock the name of the attribute or child being possibly locked.

These seem to be safe enough to use to perform full LHCb related job submission and things appear to work as expected. I can also verify that the acquire and release methods are being called quite regularly when running as adding in some debug statements gives a lot of feedback from my testing.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -50,26 +51,53 @@ def __init__(self, function, args=(), kwargs={},
...
+def checkHeartBeat():
+
+    latest_timeNow = time.time()
+
+    for this_thread in ThreadPool:
+        thread_name = this_thread._thread_name
+        if heartbeat_times[thread_name] > 0:
+            last_time = heartbeat_times[thread_name]
+        else:
+            last_time = global_start_time
```
milliams: A `defaultdict` might make this cleaner?

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -754,6 +726,10 @@ def _flush(self, _objs=None):
             objs = []
 
         obj_ids = []
+
+        for obj in objs:
+            obj._acquireParentLock()
+
```
milliams: I feel very uneasy doing this without a context manager. Would it be possible to acquire the lock per object, work with it and then release it? Otherwise, does the locking on the `GangaObject` not suffice? It should otherwise we have a problem.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -365,6 +368,19 @@ def _check_getter(self):
             raise AttributeError('cannot modify or delete "%s" property (declared as "getter")' % _getName(self))
 
     def __get__(self, obj, cls):
+
+        self_name = _getName(self)
+        GangaObject._acquireChildLock(self_name, obj)
```
milliams: Again with all these locks, they should be done with context managers.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
rob-c: I can't see the wood for the trees, how so?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -365,6 +368,19 @@ def _check_getter(self):
...
+
+        try:
+            returnable = self.__really_get(obj, cls)
+        finally:
+            GangaObject._releaseChildLock(self_name, obj)
+
+        return returnable
+
+    def __really_get(self, obj, cls):
```
milliams: I'd rather avoid all these levels of indiretion in the function calls. Either we should just lock around the contents of `__get__` with a context manager or put together a small decorator which locks for a function.

python/Ganga/GPIDev/Base/Objects.py
rob-c: Point me to a tutorial and I'll happily oblige

alexanderrichards:
```python
import threading
l = threading.RLock()
with l:
    <do stuff>
```
[locking contexts](https://docs.python.org/2/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement)

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -785,6 +810,88 @@ def __construct__(self, args):
...
+        parent_str = GangaObject._parentString(parent)
+
+        if parent is not None and parent_str in parentLockDict:
+
+            parentLockDict[parent_str].release()
+            return
+
+        raise GangaException("Cannot release an object if it has no parent associated with it: %s" % parent_str)
+
+
```
milliams: Instead of all these functions, could we not store the lock inside the root object for each tree rather than in a global dictionary? so
```python
@staticmethod
def _releaseParentLock(child):
    ## Top level parent
    child._getRoot().lock.release()
```
except with a little checking if there no parent?

rob-c: Not a bad suggestion, but do we really want to keep a lock for every GangaObject we create?
I'll make this change now

python/Ganga/GPIDev/Base/Objects.py
rob-c: I may clean this up into a decorator but for now I'm going to work with this until it's working

python/Ganga/GPIDev/Base/Objects.py
alexanderrichards: This seems to be converging on the solution that I have for the new GangaObject which is good I suppose :smile:

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
milliams: Instead of setting `heartbeat_times[thread_name] = -1` and then checking this and setting it to `global_start_time` if it's negative, make `heartbeat_times = defaultdict(lambda: global_start_time)` and just ask it for last seen times. If it's never been asked then it will return the `global_start_time`, otherwise it will return the true last seen time.

I'm not completely sure but I think it should work and would remove some lines of code.

alexanderrichards: defaultdict is essentially a way of hiding all the typing of setdefault :wink:

python/Ganga/GPIDev/Base/Objects.py
rob-c: Great this is now giving some nonsense errors about __exit__ from the threading.Lock() object

...time to debug

milliams: For reference, `__exit__` is the function that is called when the context manager closes or finishes. In this context it is the method which will call `release()`.

rob-c: I know and it should be an attribute which is assigned to a ```threading.Lock()``` but I'm still getting the error from passing a lock to a ```while myLockObject:``` line of code :/

rob-c: Hmmm, object is of None type... looks like another bug in the order in which things are constructed vs used

alexanderrichards: > while myLockObject: line of code :/

don't you mean with?

rob-c: yes

rob-c:
OK, I've moved the parent locks into a per-GangaObject thread Lock.

The child locks are also stored/requested from the top level parent from ```_getRoot()``` still making use of the parent inspection to lock the appropriate child when it is being modified.

I think I'm calling this finished.

rob-c:
(bar some stuff with defaultDicts which I'll do this-afternoon but that's code cleanup vs proof of concept)

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -788,7 +768,8 @@ def _flush(self, _objs=None):
             for obj_id in obj_ids:
                 self.unlock_transaction(obj_id)
 
-            self._lock.release()
+            for obj in objs:
+                obj._releaseParentLock()
```
milliams: I think this is no longer needed as it should be called by the `with` statement on line 730?

milliams:
That's looking much better. I wonder though if we need locks on both the root object an the object itself? It looks like if you're *get*ting a parameter then only the object itself is locked and if you're *set*ting one then both the root object and the object itself is locked? Perhaps that makes sense but we need to make sure we're documenting our intent here.

The child locks are not stored on the children themselves but rather on the root object inside `_child_locks`? What if you create an object and then change its parent or even just *give* it one? All those locks are now ignored?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -785,6 +813,46 @@ def __construct__(self, args):
...
+        child_str = "/".join(reversed(names))
+
+        return child_str
+
+    @staticmethod
+    def _getParentLock(child):
+        ## Top level parent
+        parent = child._getRoot()
+        parent_lock = parent._getLock()
+        return parent_lock
```
milliams: This is really `_getRootLock`, right? We should make sure we're naming things appropriately to avoid confusion.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: No, I want to prevent stupid oddities like the registry trying to load the object and flush it from 2 threads. This shouldn't happen but the other locks will only mean the resulting mess  is consistent at the atomic level of single get/set actions. the XML repo (specifically VStreamer functions) subvert some of the setter methods by design.

rob-c: This should never happen but I've seen enough cases, of 'should not happen' problems at the registry level that I'm wanting to keep the locks around repo access in the registry as strict as possible.

milliams: We just have to be very careful to not release a lock if it's already unlocked since that will raise a `ThreadError` or worse, release a lock that someone else is holding because they are using the object. We need to be sure that acquires and releases are always paired and always exception-safe.

rob-c: This is why they're all in final statements

milliams: Ok, but it will release it on line 772 and then again after line 731? Or am I missing something?

milliams: Ahh, I see. `_releaseParentLock` doesn't exist so this will just raise an exception instead of double-releasing?

rob-c: Apologies, was looking at the wrong source code, there is a bug here

milliams:
I had to kill the [test run](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/168/) as it was taking over 30 minutes and was showing errors anyway. I see the errors (heartbeat-related) locally too so they should be fixable.

rob-c:
I'm busy until about 3pm today then I'll do some more debugging. This works for basic interactive things and monitoring a simple job and submission but I've not done too much testing yet.

rob-c:
OK, I've fixed a handful of smaller bugs along the way in addition to putting in my latest implementation of the locking around ```__set__```, ```__get__``` and ```_flush```.

I've had troubles with context managed threading locks here as per the comments in the code so I'm sticking to this implementation for now.

rob-c:
OK, all of the tests should be passing now on the latest commit I'm keen to get any feedback people may have on implementing the locks this way?
(If anyone can suggest how this may work with context management I'm happy to test but ```with parent_lock:``` and ```with child_lock:``` statements causes problems from my testing as these locks are potentially used within a nested context across multiple threads.)

milliams:
(It looks like the tests are hanging still)

We need to make sure that we can have nested lock acquisition so that we can claim an explicit lock on an object and at the same time call a `__get__` function without that hanging. In my implementation (which I will push today hopefully) I use an `RLock` for this and it seems to be working well so far. It means I can do things like:
```python
for _ in range(100):
    with o.lock():
        num = rand.randint(0, 1000)
        o.a = num
        time.sleep(1)
        self.assertEqual(o.a, num)
```
in hundreds of parallel threads and have it work. Perhaps try a similar thing in yours?

rob-c:
Hi, the tests were hanging due to a bug I introduced into the safe save logic, just fixed it, it's because the test and the main code still behave differently.

rob-c:
@milliams  All passing now I think?

rob-c:
As Matt says there is the possibility of trailing locks if the parent or registry of an object changes. I'm giving some thought as to how this could possibly be addressed in this branch.
I think I should be able to put something together to take care of this but other than this I think everything in this example works as expected right now, not seeing any lock ups in my testing.

milliams:
I'll have a proper look through this tomorrow and will upload my version at the same time (confined to just thread-safe `GangaObject`).

I have a thought though, under what circumstances do we actually want to *change* the parent or registry of a `GangaObject`? Do we actually do this at the moment in the code, and if so, is it for a good reason? If we could confine the thread-safety requirements to just schema items, everything will be a lot easier.

rob-c:
```box.add(myJob)``` for instance should be allowed I would think for changing a repo.
(Changing the parent has to happen for: ```myApp = Application(some_opt="foo"); j.application=myApp``` to work)
I appreciate changing repo is unusual but any code relying on this never happening is going to end up exposing problems at some point imo.

We also want to keep the Registry ```_flush``` command from locking ```__set__```. However having a way of accessing a fixed lock on a per Root way would be a nice simple extension to the ```safe_save``` locking Mark introduced.

**Comment by milliams** [comment]
So does `box.add(myJob)` change the registry for `myJob` so that it's no-longer associated with `jobs`? I'd thought that in this case the object was still registered with `jobs`? Can it be in two registries at once? As for `myApp = Application(some_opt="foo"); j.application=myApp`, that's not *changing* the parent, only setting it. Do we ever take an object which has a parent and re-parent it?

With my branch for `_flush` it will be possible to do a manual lock like:
```python
def _flush(self, obj, ...):
    with obj.lock:
        do_things()
        save_to_disk(obj)
```
so that no other threads can interfere. This isn't dissimilar to what you are doing here.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -78,76 +78,104 @@ def get_backupFile(input_filename):
...
-                if not os.path.isfile(fn):
-                    new_file = open(fn, 'a')
-                    new_file.close()
+    was_locked = False
+    if not parent_lock.locked() and isinstance(obj, GangaObject):
+        parent_lock.acquire()
+        was_locked = True
+    elif not isinstance(obj, GangaObject):
+        parent_lock.acquire()
+        was_locked = True
```
milliams: Is this function ever called with non-`GangaObject`s? `to_file()` only works on `GangaObject`s anyway so I'd be surprised.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -101,12 +132,33 @@ def _execUpdateAction(self):
...
+        #from Ganga.Core.exceptions import GangaException
+        #raise GangaException("Cannot doubbly init the ThreadPool! ThreadPool already populated with threads")
+        log.error("Found a thread pool already in existance, wiping it and startig again!")
+        del ThreadPool[:]
+        ThreadPool = []
+
+    if heartbeat_times is not None:
+        heartbeat_times = None
+
+    heartbeat_times = defaultdict( lambda: global_start_time )
```
milliams: This line can just be moved to where `heartbeat_times` is defined at the top of the file.

alexanderrichards:
> If we could confine the thread-safety requirements to just schema items, everything will be a lot easier.

This is pretty much how my re-write works

milliams:
I'll start another issue about the parent and registry reassignments since I don't want to clog this one up but in short there's no cases (that I found while running the test suite) where a registry is set to anything except `None` if it wasn't already `None` and the only time it seems to be set to `None` is when the Job is being destroyed. There *are* some cases where parents are changed but they're always from an object of type `T` to another object of type `T` so I'm guessing that this is a temporary thing during a copy/clone operation. If we could rule that these variables are never changed then everything will be simpler.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
rob-c: Not really, I wanted to do it here to avoid disabling/re-enabling problems as the heartbeat will be stable if turned on for the first time after >3min

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: It is in testing

milliams: Is that in Mark's new test? I found that too and I think it's better fix one small test that change a core function like this to allow for it. I've got a PR incoming to change Mark's `TestSafeSave` test.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
milliams: Well `global_start_time` will be reset when this function is called but you're right that you want to reset this dict to blank when restarting the monitoring. I recant my question.

rob-c:
My logic for having multiple threads was to make use of a registry ID for each object where it exists as it makes debugging easier. Yes I could change to have each object look after it's own lock which simplifies things but does potentially make debugging more difficult.
If the object is in a registry then the registry decides the location of the thread-lock and if the the object ID is used it is a 'memory-like' address and so is clearly a GangaObject which hasn't been added to a registry.

I'll look into the failing test for now.

rob-c:
This isn't going to be merged as a result of a developer meeting today. Some improvements from this may make their way into future PR.

milliams:
Thanks a lot for the work that went into this though @rob-c. It really helped with seeing where the main pain points were.

rob-c:
@milliams can you kill the tests for this branch, they've got stuck and I'm unable to find anything useful in Jenkins to abort the test. Since we're not planning to use this branch we can safely drop them as we don't care.
Thanks :)

PR #183, opened 2016-02-09
Migrate most merger tests

milliams:
This is another in the line of test migrations to the new system. This isn't all the merger tests but it's most of them and the last few need some care.


rob-c:
If these are working I'm in favour of merging them asap

drmarkwslater:
:+1:

PR #186, opened 2016-02-10
Added a global lock for the safe_save function to make sure XML files…

### LINKED ISSUE #185: XML files disappearing
I think I've tracked this problem down. I think the only way that the data file can be blanked and yet the data~ file still be present is because in the `safe_save` method in `GangaRepositoryXML.py`, the backup file is created from a rename of the `data` file before the new `data` file is created from a rename of `data.new`. If this function has been called from multiple threads simultaneously, this can lead to a race condition where the `data.new` file is overwritten by the second thread before the `data` file is created from it. The original `data` file is removed to produce the backup file and then, because the `data.new` file has gone, it can't create the proper `data` file, fails silently, thus leaving just the backup file. 

This has two issues to address:

1. safe_save isn't thread safe. Though it never really could be, we could make it a lot better by shifting things around a bit and being a bit cleverer with how things are handled. I try to get this done tomorrow.

2. safe_save should **never** be called by multiple threads. This may be sorted by the locking changes that are being discussed for the repo but there should at least be some form of global lock on accessing a repo by more than one thread.

Note that this was found through Tasks because I suspect there are explicit flushes coming from multiple threads (still got to track these down) but I think this is the only way the similar reports of XML files disappearing from the job repo can be explained as well.

drmarkwslater:
… don't disappear.

Fixes #185.

### DIFF
--- a/python/Ganga/new_tests/GPI/Bugs/Issue185.py
+++ b/python/Ganga/new_tests/GPI/Bugs/Issue185.py
@@ -1,0 +1,31 @@
+from ..GangaUnitTest import GangaUnitTest
+
+class Issue185(GangaUnitTest):
+
+    def test_Issue185(self):
+        "Test that XML files don't disappear"
+        import threading
+        import os
+
+        from Ganga.Core.GangaRepository.GangaRepositoryXML import safe_save
+
+        def my_to_file(obj, fhandle, ignore_subs):
+            fhandle.write("!" * 1000)
+
+        # Create lots of threads that will keep hitting safe_save
+        testfn = '/tmp/xmltest.tmp'
+        ths = []
+
+        if os.path.isfile(testfn):
+            os.remove(testfn)
+
+        for i in range(0, 500):
+            ths.append( threading.Thread(target=safe_save, args=(testfn, [], my_to_file ) ) )
+
+        for th in ths:
+            th.start()
+
+        for th in ths:
+            th.join()
+
+        self.assertTrue(os.path.isfile(testfn))
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -10,6 +10,7 @@ import os.path
 import time
 import errno
 import copy
+import threading
 
 from Ganga.Core.GangaRepository.SessionLock import SessionLockManager
 
@@ -79,6 +80,10 @@ def get_backupFile(input_filename):
 
 def safe_save(fn, _obj, to_file, ignore_subs=''):
 
+    # Add a global lock to make absolutely sure we don't have multiple threads writing files
+    # See Issue 185
+    safe_save.lock.acquire()
+
     obj = stripProxy(_obj)
 
     check_app_hash(obj)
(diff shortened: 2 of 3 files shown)


rob-c:
Hi Mark, I'm keen to get this in asap :)

Do you know from reading this code do we ever have a point where the 'data' file doesn't exist temporarily? I just wonder if a user hitting Ctrl+C at the wrong time would be exposed to this again a bit more?
(I know we can only do so much to prevent this)

milliams:
Firstly the lock should be acquired and released using a context manager otherwise we will end up with dangling locks.

Secondly, do we really want a global lock or just one per object/file? If there any problem with writing job1 and job2 at the same time?

drmarkwslater:
Stupid - now changed to using context manager!

For the moment, I would prefer (and it looks like @rob-c agrees) to use a global lock so we can be absolutely sure this has gone away. This may hit performance a little but hopefully not a lot. After we've got more testing for these things, then I would suggest revisiting it. E.g. I suspect this lock won't b needed anyway given the locking that's being added for the monitoring but better to be safe than sorry at the moment :)

Btw, I've added a safe_save unit test that recreates the bug to this PR as well.

python/Ganga/new_tests/GPI/Bugs/Issue185.py
```diff
@@ -0,0 +1,31 @@
+from ..GangaUnitTest import GangaUnitTest
+
+class Issue185(GangaUnitTest):
```
milliams: For consistency with the others, let's call these `GitHub185`.

drmarkwslater:
Forgot to say: the code to move the `data` file to `data~` and moving the `data.new` file to `data` is straight after each other. If someone `Ctrl+C`'s out they would have to be **very** unlucky but could then also recover from `data.new` anyway.

python/Ganga/new_tests/GPI/Bugs/Issue185.py
milliams: Also, since this doesn't use any of the Ganga runtime machinery I'd rather this was done as a standard UnitTest, not in the `GPI` directory. In a while I'm going to have to rearrange these anyway but if this is already in the right pace then it will be easier.

In fact, I would just call this `test_safe_save` and keep it as a generic unit test for the function rather than tying it to a specific issue.

rob-c:
OK, Erm We should probably change the load mechanism to check for a ```data.new``` existing as I think at the moment if it fails to load ```data``` it then goes directly to ```data~``` which will miss out ```data.new```. Although this feels like splitting hairs as we'd be quite unlucky to get this.

python/Ganga/new_tests/GPI/Bugs/Issue185.py
drmarkwslater: Where is the place to put this to make sure it's run then? It certainly doesn't need the GPI but from what I could glean from Jenkins, we're only running tests in `new_tests/GPI`...

drmarkwslater: By tests I mean the PR tests!

drmarkwslater:
I agree - it's a minor point but should be made of a note of :)

python/Ganga/new_tests/GPI/Bugs/Issue185.py
milliams: You're right, it was just running over `new_tests/GPI` but I've just changed it so that it runs over all of `new_tests` so this will include all unit tests too. I'm not sure why I hadn't set it up like that before.

drmarkwslater:
I think that addresses everything - if so, let me know and I'll merge, or @milliams as release manager can do the honours :)

milliams:
There's a brief problem with the Jenkins tests which I'm working to resolve before we can test and merge this. I'll let you know.

milliams:
I'm happy to merge but we should bear in mind the future potential for a per-object lock (or perhaps it becomes unnecessary with the `GangaObject`-level locking?)

:+1:

rob-c:
@milliams There is a per-object lock in the Registry which only works per-object per registry and doesn't protect against the bug this is fixing.
The locks in the GangaObject should prevent objects being modified during flushing whilst allow for appropriate getter methods to work.

PR #188, opened 2016-02-10
Migrate registry and repository tests

milliams:
Given the work that we're doing on this at the moment, I thought it was worth converting over the Repository and Registry tests so that we can keep track of some of the basics.


rob-c:
Do these just work with the new test framework? I thought more work would have been required in porting these, looking good :)

milliams:
They do indeed. They were originally written in a medium-new unittest style so porting them wasn't too bad. They don't use the GPI so they're simple too. I made them a `GangaUnitTest` as I though it best if the background machinery is running for the test.

rob-c:
If these are working please feel free to merge them I'm keen to run tests

PR #189, opened 2016-02-11
Fix node tests

milliams:
These two changes fix bugs that I have come across in the course of working on the thread-safe `GangaObject`. They're both real bugs in the code which were just introduced by accident at some point.

The schema change is for those cases where you have a `GangaObject` which is hidden (and so has no `default_*` config) which has schema `ComponentItem`s items with default values. This fix allows these to work whereas before it would always try to look in the config system.

The second is a small bug which completely breaks `Node.__deepcopy__`. The reason this has not shown before is that no-one actually calls this function as `GangaObject` overrides it and doesn't call the base class. Really this whole method should be killed but while there's a test for it I thought it worth fixing. Longer-term I think we should just combine `GangaObject` and `Node` into one class but that's not for now.

These aren't directly stability fixes (as they would raise exceptions) but they are real bugs so I'd like to merge this as soon as possible as I don't want these fixes mixing with my up-coming PR for the thread-safe `GangaObject`.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -101,12 +101,10 @@ class Node(object):
         for elem in this_dict.keys():
             if elem not in do_not_copy:
                 this_dict[elem] = deepcopy(this_dict[elem], memo)  # FIXED
-            else:
-                this_dict[elem] = None
 
+        obj.__dict__ = this_dict
         if self._getParent() is not None:
             obj._setParent(self._getParent())
-        setattr(obj, '_index_cache', {})
         setattr(obj, '_registry', self._registry)
         return obj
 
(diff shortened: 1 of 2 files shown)


rob-c:
@milliams good stuff feel free to merge asap.

Thinking about mid to long term, do we want to keep Node around?

I know it takes care of parent and registry assignment in a neat place but I can't help but feel there may be a better way.

milliams:
I think long-term we want to get rid of `Node` as it's just confusing and there's such a mix of attributes between the two classes. I'm pretty sure that in Alex's rewrite it's gone anyway. The reason it first existed was just to do the visitor pattern `accept()` thing but it's been so abused over the years that it's silly having a distinction any more. Also, if we do away with the visitor pattern then it's not needed anyway and all the useful stuff can be moved into `GangaObject`.

alexanderrichards:
> I think long-term we want to get rid of Node

:+1:

rob-c:
good we're all in agreement here then :)

PR #190, opened 2016-02-12
Fix testsafesave

milliams:
This is few minor fixes to @drmarkwslater's `TestSafeSave` test to make sure that the things being passed around are valid for the `save_save` function. The main change is using a `LocalFile` instead of a `list`.


rob-c:
Could you add an equivalent to ```assert(not os.path.isfile(tmpfile+'.new')``` just for completeness at the end?

Other than that feel free to merge from me :)

milliams:
Good idea, I'll add that in.

PR #191, opened 2016-02-12
Make GangaObject thread-safe

milliams:
This adds thread-safe locks to all `GangaObject`s for all schema attributes as well as getting and setting the parent. In addition to this you can, at any time, claim a lock on an object with a `with obj.lock:` context manager and have only the current thread be able to access it.

As well as locking the object itself, it will also claim the locks on all its parents and grandparents (etc.) for safety. This PR makes no distinction between read locks and write locks but I'm not convinced there's a need to distinguish.

This doesn't make any changes to the registry etc. yet but it would be very easy to use the locks here to guarantee thread-safety there too but I haven't yet seen any issues.

This PR obviously competes with some of @rob-c's work in #182 but I feel that this approach is simpler and this PR is much more focused.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -147,10 +147,12 @@ class Node(object):
             p._lock.acquire()
             ancestors.append(p)
             p = p._parent
-        yield
-        for p in reversed(ancestors):
-            p._lock.release()
-        self._lock.release()
+        try:
+            yield
+        finally:
+            for p in reversed(ancestors):
+                p._lock.release()
+            self._lock.release()
 
     # get the root of the object tree
     # if parent does not exist then the root is the 'self' object
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -13,6 +13,10 @@
 #   and followed by
 #    obj._setDirty()
 
+import threading
+from contextlib import contextmanager
+import functools
+
 import Ganga.Utility.logging
 
 from copy import deepcopy, copy
@@ -37,7 +41,7 @@ logger = Ganga.Utility.logging.getLogger(modulename=1)
 
 _imported_GangaList = None
 
-do_not_copy = ['_index_cache', '_parent', '_registry', '_data']
+do_not_copy = ['_index_cache', '_parent', '_registry', '_data', '_lock']
 
 def _getGangaList():
     global _imported_GangaList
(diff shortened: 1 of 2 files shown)


milliams:
The tests were passing on my local machine so I will need to give this some more thought.

rob-c:
This potentially exposes problems with being able to flush and set an attribute at the same time no?

I think locking the set method with an explicit Lock rather than an RLock removes the possibility of bad things being done behind the scenes and somehow feels much safer than relying on the RLock which was causing problems in the Monitoring initially with lockups.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -339,6 +377,19 @@ def getNodeIndexCache(self):
 ##########################################################################
 
 
+def synchronised_descriptor(get_or_set):
+    """
+    This decorator should only be used on ``__get__`` or ``__set__`` methods of a ``Descriptor``.
+    """
+    @functools.wraps(get_or_set)
+    def decorated(self, obj, type_or_value):
+        if obj is None:
```
rob-c: Could this be not a GangaObject here?

milliams: This check is needed since if a descriptor is [called on an class](https://docs.python.org/2/howto/descriptor.html#invoking-descriptors) (rather than an object) e.g. `f = LocalFile.subfiles`, the first argument passed to `__get__` is `None` and the second argument is the type. If we're requesting a value from a class then we don't want to lock the object (since we have none to lock).

Unless I misunderstand the question?

rob-c: Nope, that's fine I was just curious how this could arise

milliams:
retest this please

milliams:
At the minimum it is possible to replace @drmarkwslater's new lock in `safe_save` with a `with _obj.lock:` and it should be sufficient (in fact I have this done locally). If we'd rather have this lock applied higher in the call chain, actually inside the `_flush` function itself then that's possible too.

If we could have a test which tries to perform a registry thread attack, that would be useful. Otherwise there's no way to know if we've solved it. Do you have anything which works for this?

Either way I feel it might be worth saving that for another PR so that we're tackling a single problem at once.

milliams:
@ganga-ci retest this please

rob-c:
@milliams I'd take a look at #192 for a heavy test involving lots of threads and repo activity in the same place

rob-c:
I'm very weary of locking he safe_save on a per-object level. It should be fine provided that the RLock is correctly acquired and released but then the Registry should also never have locked up if this is the case.
If it works in testing I'm happy to move in that direction but for 6.1.16 I think I'm keen to stick a global lock there and release as we know it's guaranteed safe. Then we can worry about improving it later.

This can cause issues if there is a background flush at the same time you try to do something else with the object from my testing.

milliams:
Ok, I will keep a global lock on `safe_save` for now and I will look into where I should put a per-object lock in `_flush`.

rob-c:
If this branch is independent of the changes which have gone into the repository shall we merge this for now?
Code looks good to me unless there are any outstanding changes :)

milliams:
This is independent of any changes need to the registry. Now, it doesn't fix the lockups in the registry (and in fact, potentially could cause some more until the registry is tweaked) but it does provide us with the tools to do so. I believe the design of the system is sane and once the registry tweaks are in we'll be pretty stable again.

I'm happy to merge now if there are no objections.

rob-c:
please feel free to merge

PR #192, opened 2016-02-12
Adding new Queued submit test

rob-c:
This is a new submit test which perfoms a lot of parallel submits from a jobs repo at the same time and expects the jobs to all come back as fully completed within a reasonable time scale.

This also includes some minor modifications to the GangaUnitTest which allows for the Config to be edited ahead of bootstrap which is required for things like running with different numbers of threads in the queues system.

This should always complete in <5min as it seems to take of the order of 200s on an lxplus node.

This should be relatively sane and good for the current thread-safety testing.


python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -93,8 +97,10 @@ def start_ganga():
     logger.info("Checking Credentials")
 
     if missing_cred:
-        raise Exception("Failed due to missing credentials %s" %
-                        str(missing_cred))
+        raise Exception("Failed due to missing credentials %s" % str(missing_cred))
+
+    from Ganga.GPI import queues
+    queues.unlock()
```
milliams: Is it possible to access the queues object directly and not through the GPI here?

rob-c: I want to make sure the queues system which was constructed in the GPI is unlocked.
This is the queues instance which is exposed to the user.

This should probably not be exported by ```exportToGPI(... ThreadPoolQueueMonitor()...)``` as in bootstrap. But this would require some work with singletons which is more hassle than it's worth imo to address this line.
(This would allow us to only start the queues up when they're needed would be nice :) ).

rob-c: (Yes I could subvert this and access some globals in ThreadPoolQueueMonitor.py but this makes for something even more nasty which would be more fragile)

milliams:
This test is not being run since the file is named incorrectly. This is also the case with `LazyLoadingTest.py` and `ParserTest.py` which you added in the past. None of these tests have been running since their names have to start with `Test` or `test`. Could you rename them so they run? The only test names that I have done differently are the Savannah and JIRA ones and those are accounted for individually (and even those it is probably worth renaming?).

python/Ganga/new_tests/GPI/GangaUnitTest.py
milliams: No, I see now how `queues` is constructed so this is fine for now. In future we should have a `start_queues()` and `stop_queues()` function which does something global.

rob-c:
hmm, nosetests picks up the file correctly when I just run over the file, is some different logic being used in Jenkins?

milliams:
It could be, I'll look into it.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: Yes I agree, I think this would be much nicer to have but I think this is work related to startup/shutdown re-working and breaking up of bootstrap that's going on :)

rob-c:
I'll rename the tests either way, it's a minor point to do that

milliams:
I think it's a combination of things but mostly being caused by different versions of nosetests. The two things that seem to need to be done for it to work on all versions I've come across are name the file so it starts with `[tT]est` and make sure that there is an `__init__.py` file in the directory.

rob-c:
OK, I've changed the test names and added a short test for the Dirac backend which checks the monitoring command as well as submitting the job.
For the jobs using Dirac I don't think we should need to wait for these jobs to complete so I've just added a step to be sure that every job has been run through the monitoring system at least once.
This is all across multiple threads as well so could potentially cause things to choke/break due to errors.

rob-c:
These tests run successfully on my branch for thread safety at least. I'd be keen to get these into trunk for testing purposes as well as the fact there are some bugfixes/improvements in the GangaUnitTest in this branch.
Assuming these don't highlight anything dramatic am I safe to merge?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -5,7 +5,7 @@
     import unittest
 
 
-def start_ganga():
+def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
```
milliams: Could we document these new parameters, either here or in `GangaUnitTest.setUp` below?

milliams:
Really, we want these tests to fail so that we know any changes we make with the multi-threading stuff is fixing the problem? Or are these tests not supposed to be demonstrating the repository contention bug? If they're just here to test behaviour that we expect should work already and we just want to keep it working then we should be able to merge them if they pass.

rob-c:
They will be failing with develop, I am testing them in my other branch

rob-c:
I know the test should probably be added with/after the fix so that no test is failing but given we're wanting to do a lot of testing on possibly different ways of solving this I'm very keen to get the test in first so we can prove to ourselves everything works in the PR.

milliams:
Ok, so if they are not passing on `develop` I don't think we should merge them but I will include them locally (and probably push it as well) to test my branch. I don't want anything in `develop` which is failing.

rob-c:
In that case I'll open an additional PR for the fixes in GangaUnitTest which will benefit develop

milliams:
That would be great.

rob-c:
OK, these tests might take >5min in their current format, but from testing on a branch where they work, they do complete in <10min

egede:
This pull request is labelled 6.1.16 but is not yet merged. Does that mean that there is still work to do for 6.1.16 or is it to be delayed?

rob-c:
I was strongly hoping to get this in to test the Dirac backend against lockups.
I'm working on it atm

rob-c:
OK, I think I'm happy to delay most of the content of this particular PR until 6.1.17 as the scope of the patch is starting to grow into fixing a lot of smaller bugs which are cropping up for me (mainly in testing).

I think using this branch however I have spotted a minor bug which may need to be fixed in 6.1.16 so I'm going to spend a short amount of time attempting to fix this.

rob-c:
@milliams I think I'm seeing tests executing in parallel for some reason. Are the tests definately not running in parallel on the same repo?

I see:
```
Fourth make sure that we get rid of the jobs safely ... ok
test_a_JobConstruction (GPI.Monitoring.TestMonitoring.TestMonitoring) ... FAIL
Fourth make sure that we get rid of the jobs safely ... ok
```

I expect to see:
```
Fourth make sure that we get rid of the jobs safely ... ok
test_a_JobConstruction (GPI.Monitoring.TestMonitoring.TestMonitoring) ... ok
```

milliams:
@rob-c The first few lines there are from `TestLazyLoading`:
```
First construct the Job object (singular) ... ok
Second get the job and check that getting it via jobs doesn't cause it to be loaded ... ok
Third do something to trigger a loading of a Job and then test if it's loaded ... ok
Fourth make sure that we get rid of the jobs safely ... ok
```
and the next few are from the new `TestMonitoring`:
```
test_a_JobConstruction (GPI.Monitoring.TestMonitoring.TestMonitoring) ... FAIL
Fourth make sure that we get rid of the jobs safely ... ok
test_c_disableMonitoring (GPI.Monitoring.TestMonitoring.TestMonitoring) ... ok
test_d_anotherNewJob (GPI.Monitoring.TestMonitoring.TestMonitoring) ...
```
and seem to be doing the correct thing.

rob-c:
Ah, apologies I forgot to edit the TestMonitoring. Ignore me :)

rob-c:
This is getting stupid.

I keep getting:
```
test_a_JobConstruction (GPI.Monitoring.TestMonitoring.TestMonitoring) ... FAIL
```

Which is the first test to fail that I'd like to debug and I'm getting no useful information about what failed. In comparison to this I'm not seeing failures on my machines that I'm testing on so is there any way of preserving the failed tests rather than aborting these 'builds'?

milliams:
If the test is hanging then it will never give the results of the failure. Running locally it's easier for you to check with `--nocapture` etc. but that's not so easy on ganga-ci. Are you running your tests locally in the same way as on Jenkins? i.e. inside a clean virtualenv running:
```bash
pip install -r requirements.txt
python setup.py tests --type=all
```

rob-c:
I've tried multiple machines with different configs but I'll give this a shot. As to why I can't reproduce this failure still escapes me.

rob-c:
OK, on my personal machine I get no errors and on lxplus it doesn't look like I get errors either... I'll keep searching for problems.

rob-c:
(Which is clearly wrong on my machine without Dirac but something is going bad there and I don't want to waste time on it)

rob-c:
OK,
I'm now seeing:
```
2016-02-24 15:23:36,777 GANGA_Update_Thread_MonitoringWorker_0_1456327416747 Ganga.Core.MonitoringComponent.Local_GangaMC_Service:1041 ERROR   : Monitoring Error: ObjectNotInRegistryError: Object 'Job' is a duplicated version of the one in this registry: jobs !
```
which is something slightly more tangible to go digging around

rob-c:
Once #445 is in I'll look into sprucing up the parallel Dirac submit tests from here as it would be nice to run that nightly.
(Most of this PR appears to be merged in one form or another so this will get much smaller after being rebased)

rob-c:
@milliams Is there a mocked DIRAC server coded up for the other tests? I'll adapt the DIRAC parallel tests to use this if we have one, if not I'm wondering if it's worth keeping around this PR for much longer.

rob-c:
I think I can abandon this PR now. This was the first implementation of a parallel submit and complete test using the queues system. It's changed drastically into the tests we're currently using for the parallel thread safety atm and I don't think we're closer to having a full mocked DIRAC API so spamming the DIRAC backend with hundreds of hello world jobs is probably a bit bad. With this in mind I'm closing the PR and cleaning up

PR #193, opened 2016-02-12
Moving tests and attempting to get them to run with Jenkins

rob-c:
This should be enough to get some outstanding unittests to work with Jenkins.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,7 +59,7 @@ class RunTestsCommand(Command):
         if self.type in ['unit', 'all']:
             cmd.append('Ganga/new_tests/*.py')
         if self.type in ['integration', 'all']:
-            cmd.append('Ganga/new_tests/GPI --testmatch="(?:\\b|_)([Tt]est|Savannah|JIRA)"')
+            cmd.append('Ganga/new_tests/GPI --testmatch="(?:\\b|_)(Test|Savannah|JIRA)"')
 
         if self.coverage:
             cmd.append('--with-coverage --cover-erase --cover-xml --cover-package=.')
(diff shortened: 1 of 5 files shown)


milliams:
I think we should keep the file names and class names as upper camel case for consistency.

rob-c:
Then can we remove the lowercase t from the nosetests search in this case if we're conforming to a standard

milliams:
We could, the reason I kept it is that the default nosetests match is `(?:\b|_)[Tt]est` so I kept that and just appended the `Savannah` and `JIRA` matches so it becomes `(?:\b|_)([Tt]est|Savannah|JIRA)`.

rob-c:
OK, I expect these to run correctly, any problems with me merging them if there are no problems highlighted?

setup.py
```diff
@@ -59,7 +59,7 @@ def run(self):
         if self.type in ['unit', 'all']:
             cmd.append('Ganga/new_tests/*.py')
         if self.type in ['integration', 'all']:
-            cmd.append('Ganga/new_tests/GPI --testmatch="(?:\\b|_)([Tt]est|Savannah|JIRA)"')
+            cmd.append('Ganga/new_tests/GPI --testmatch="(?:\\b|_)(Test|Savannah|JIRA)"')
```
milliams: I think in fact we need to keep this as lower case `t` as well so that it can match the function names correctly. Let's keep this line as it was.

rob-c: Wouldn't this present with no tests being run, yet the tests are running aren't they? (I assume they are simply due to the delay and everything reporting fine)

rob-c: Never mind I just checked they're missing... I'll revert

rob-c:
OK, waiting for the tests to re-run then I'm planning to merge

PR #194, opened 2016-02-12
Allowing tests to run in separate repos, allowing config to be editte…

rob-c:
…d before fully initializing Ganga and other minor changes

These are changes and bugfixes I've been using for other tests and I'm keen to get them into develop.

These should protect tests from bleeding into each other as well as some minor improvements to the GangaUnitTest itself to offer greater flexibility.

### DIFF
--- a/python/Ganga/new_tests/GPI/GangaUnitTest.py
+++ b/python/Ganga/new_tests/GPI/GangaUnitTest.py
@@ -5,7 +5,7 @@ except ImportError:
     import unittest
 
 
-def start_ganga():
+def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
 
     import sys
     import os.path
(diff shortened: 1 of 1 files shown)


rob-c:
Any objections to merging after the tests complete?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -14,10 +14,10 @@ def start_ganga():
     file_path = os.path.dirname(os.path.realpath(__file__))
     ganga_python_dir = os.path.join(file_path, '..', '..', '..')
     ganga_python_dir = os.path.realpath(ganga_python_dir)
-    sys.path.insert(0, ganga_python_dir)
+    if len(sys.path) >= 1 and ganga_python_dir != sys.path[0]:
+        sys.path.insert(0, ganga_python_dir)
```
milliams: For my own interest, why is `sys.path.insert(0, ganga_python_dir)` not sufficient? Is it just to avoid adding it multiple times? In that case maybe:
```python
if ganga_python_dir not in sys.path:
    sys.path.insert(0, ganga_python_dir)
```
or do we really need our path to be at the beginning of the Python part?

rob-c: This is to avoid adding it multiple times.
We don't 'need' it to be at the start, but by convention it has been for a long time (at least on LHCb) and changing it would be something I want to be careful with.

milliams: Ok, in that case I'm happy for this to be merged.

PR #196, opened 2016-02-15
Bugfix/cleanup queues launch/shutdown

rob-c:
This PR deals with improving the startup/shutdown of the queues system.

All I've done is take the code from Shutdown/bootstrap and move it into a startup/shutdown method within the Queues section of the code which is more inline with how it should be handled.

There is still a check for only 1 running queues system but this should be correctly destroyed on shutdown now.

This means I can remove some custom code from the Shutdown method and cleanup the startup of the queues in the bootstrap as this appears to make use of the existing functionality already available.

As this was primarily to make the shutdown more sane I've also removed some lines from GangaUnitTest which are no-longer needed due to improvements in the shutdown of Ganga.

### DIFF
--- a/python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
@@ -1,0 +1,32 @@
+
+_global_queues = None
+
+def startUpQueues():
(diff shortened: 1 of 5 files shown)


rob-c:
Any objections to me merging this?

PR #197, opened 2016-02-15
Cleanup test naming

rob-c:
This PR tries to bring all of the tests into a uniform testing name scheme. This is mainly a cleanup task but has some advantages for developer(s).

This basically ports the class and function names so that they match the default search patterns from nosetests.

I'd like to has as little as possible difference between what is run on the test system and what is run by:

```nosetests path/to/test.py```

i.e. running the test with no extra options passed to nosetets that aren't needed by the developer during testing.


rob-c:
I've just realised this branch got forked from my ```develop``` branch which accidentally had my queues changes merged in the branch.
Apologies for the noise this will cause, but I think github will track the changes through the different branches correctly as it has done in the past.

rob-c:
Strange, I'm not seeing all integration tests pass locally on my machine despite them working on Jenkins.
@milliams Do you know any reason why the following command may cause errors on some systems?

```python ./setup.py tests --type=integration```

rob-c:
My tests appear to be passing now as on Jenkins. I'm assuming this to be some probable race condition, I'm happy to say all tests are now passing :)

milliams:
If given a chance I would have said that this is not going to be needed since I'm going to be switching away from nosetests to pytest in the near future anyway. It doesn't hurt too much but I'd really like to reduce the amount of code going into develop if we can.

PR #198, opened 2016-02-15
Cleanup/remove state tracking

rob-c:
This PR deals with some code in the Core which has been bugging me for a while.

I've completely removed all referenced to ```__setstate__``` and ```__getstate__``` from within the Code.
These functions appear to have at one time been used to perform a 'safe' copy by keeping object references when an object is marked as being un-copyable.

This has led to problems and I'm sure is responsible for the parent being lost at some point in the copy process due to the setParent None statements which can creep in from this.

I've also improved the handling of the index caches.
This has come up before so I'll explain what goes on now.

 1. ```getIndexCache``` will always return an index cache if an object is in a repository. If it is not it will return ```{}```
 2. ```getIndexCache``` makes use of repository knowledge to know if an object is fully loaded from disk.
  1. If the object is fully loaded from disk the index cache is re-generated using the object in memory. (Only safe thing to do in-case it has changed).
  2.  If the object is _not_ fully loaded the stored index cache is used to prevent loading of the object from disk.
 3. ```getIndexCache``` should always return a dict style object with a default of being empty. All GangaObjects should have knowledge of this and this should not be set to None.
 4. ```setIndexCache``` has a silent debug message which warns if the IndexCache is being set on a live object, this is something we should avoid doing.
 5. ```getIndexCache``` makes use of an internal cache which will be set to ```{}``` once a job is loaded from disk to prevent people inspecting the object and finding inconsistent data.

This should hopefully prevent problems in the future of polling a job and finding it's state to be inconsistent.

Whether we want to include this in 6.1.16 or 6.1.17 is up for debate. If this causes absolutely no problems in testing I'll push to get this into 6.1.16 as it gets rid of a lot of very old code which was causing problems.


milliams:
This will have to wait until at least 6.1.17 as it's touching too much code to be known to be safe. 6.1.16 is a *bugfix-only* release to get things stabilised.

rob-c:
I know 6.1.16 is mainly bugfix but this is an attempt to address several bugs which keep coming up over and over again.
Personally I'm keen to get it in to avoid any possibility of getNodeIndexCache causing any future problems.

As I say if it causes all tests to fail in strange/wonderful ways then I'm happy to not push for inclusion but I expect this to do a lot more good than possible harm.

drmarkwslater:
I'm afraid I agree with @milliams. Only PRs that are required for fixing the known major issues should be going in as we cannot risk introducing more bugs than we're fixing and may not even be testing for. As soon as we're happy with 6.1.16 and, more importantly, so are the users, then we can put in all the additional PRs that are getting held up.

rob-c:
OK, I'm happy to hold off on this one. It's not essential to the stability (we think we've fixed the IndexCache bugs anyway) so I'm happy to wait until 6.1.17 so that this will (hopefully) make them go away for good.
As for the get/set state code I'm concerned that it is at some level in the copy mechanisms as it appears to be a copy in and of itself.

rob-c:
On a hopefully unrelated note does anyone know why the continuous integration stuff from travis is failing here?

milliams:
@rob-c If you click on the link it takes you to the [results](https://travis-ci.org/ganga-devs/ganga/jobs/109329175) where you can see the error.

rob-c:
Frankly this concerns me that this branch is revealing a bug which in principle has been there for a long time. The Node should _not_ have had knowledge of the ```_registry``` and the fact it's causing problems in the test only in this branch and now worries me.

This removes the ```__getstate__``` from the deepcopy which shouldn't have been there and replaces it with something sane.

Is there a very good reason we don't want to be merging this if it's fixing problems on develop?

milliams:
Yes, unless there's an observable bug in `develop` which this commit fixes, it's definitely not going in yet. Even then, we need to be careful about submitting anything.

I'm not sure there's a real bug there in that while `Node` should not know anything about `_registry`, infact all `Node`s are `GangaObject`s and vice verse so there's no real bug likely present.

rob-c:
I think I'm a bit concerned that we don't know entirely what the ```__getstate__``` was doing as it was overridden in effectively random places in the code and changing the behaviour of the Node copy commands which are now being called.
This is causing strange behaviour in at least the LogicalFile https://github.com/ganga-devs/ganga/pull/198/files#diff-7ecc138d063d806ff34d36e5506f0074L65 due to the fact that these functions have been copied without a clear correct behaviour from them.
(I know this _should_ be deprecated but this likely causes strange behaviour to LHCb users coming from 6.0.44)

The whole business with the ```getIndexCache``` is intended to fix entirely rather than, making it go away for now in 6.1.16 the 3rd stability issue of jobs/tasks are being represented in a bad state in the display summaries due to the object not displaying it's current information in favour of some cache which 'should be up to date'.
The fact we don't see it doesn't necessarily mean we have fixed that one as I think we were potentially seeing more than just the job.status being stale, even though it now appears fixed.

This should make develop more stable or at least more predictable I would strongly argue.
The changes are (aside from some code cleanup) minor in the sense that they're confined to the copy methods of the Node as well as the get/setIndexCache methods which without this are still very fragile.

rob-c:
Superseded by #225 and #226

PR #200, opened 2016-02-15
Feature/monitor heartbeat checking

rob-c:
This branch attempts to pull out the heartbeat code from #182 as it's likely that this branch could be superseded by other code due to work performed elsewhere.

This PR includes the code to inform the user if a worker thread is working and has been performing the same task for > 3min.
This 3min is currently not user configurable and is hard-coded as 180.

This PR also includes some very minor changes to improve the shutdown of the Monitoring loop at the end of it's life and some minor changes relating to the construction/teardown of the thread-pool which effects the GangaUnitTests derived tests as well as elsewhere.

This has already been discussed for inclusion in 6.1.16.

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -28,6 +28,7 @@ log = getLogger()
 
 config = getConfig("PollThread")
 THREAD_POOL_SIZE = config['update_thread_pool_size']
+heartbeat_stall_time = config['HeartBeatTimeOut']
 Qin = Queue.Queue()
 ThreadPool = []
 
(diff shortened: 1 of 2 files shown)


milliams:
In principle I'm happy for this to be merged. I'll need to review the code first but then we should be good.

rob-c:
There's nothing here not already in #182 which has been in for a while, the code has been through many iterations already.
I'm just breaking #182 into code I am certain to keep vs code I would like to keep.

egede:
Is it easy to make the 180 s configurable? Just in case we figure out that it is too short and that users get flooded with fake positives.

rob-c:
@egede I'll get on that now before I push to merge the PR, we may keep it at 5min or higher for 6.1.16 while it's in testing

milliams:
I guess a config item in `[PollThread]` would make sense. We can keep this low for testing but I agree with @rob-c for making it longer for the release,

rob-c:
OK, should be configurable now, and set to 300s by default.

rob-c:
I think this should be fixed for merging?

milliams:
:+1:

PR #202, opened 2016-02-17
Bugfix/cleanup proxy ganga fixes

rob-c:
This branch was designed to clean up some code handling the use of proxies and migrate everything to use ```addProxy``` and ```stripProxy``` so that the use of proxies in a lot of places becomes much more transparent.

This branch also includes a rudimentary implementation of the loading of the config system for 'little ganga' based upon the bootstrap code in 'big Ganga'.

This required some minor changes but reduces some of the differences which crept in between ganga and Ganga recently.

As a result of this 'ganga' now correctly loads my repository rather than making a new one in the wrong place on disk and I'm able to create and submit a Dirac job.

There are some issues to do with exiting after importing little ganga in this branch which I'm yet to look into but this work will probably be of interest to people in it's current form.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -603,12 +603,13 @@ under certain conditions; type license() for details.
 
         config = getConfig("Configuration")
 
+        ## This is now handled properly in the config system in __init__
         # detect default user (equal to unix user name)
-        import getpass
-        try:
-            config.options['user'].default_value = getpass.getuser()
-        except Exception as x:
-            raise Ganga.Utility.Config.ConfigError('Cannot get default user name' + str(x))
+        #import getpass
+        #try:
+        #    config.options['user'].default_value = getpass.getuser()
+        #except Exception as x:
+        #    raise Ganga.Utility.Config.ConfigError('Cannot get default user name' + str(x))
 
         # import configuration from spyware
         from Ganga.Runtime import spyware
@@ -640,10 +641,10 @@ under certain conditions; type license() for details.
 
             def _accept(fname, p=re.compile('.*\.ini$')):
                 return (os.path.isfile(fname) or os.path.islink(fname)) and p.match(fname)
+
             files = []
             if dir and os.path.exists(dir) and os.path.isdir(dir):
-                files = [os.path.join(dir, f) for f in os.listdir(dir) if
-                         _accept(os.path.join(dir, f))]
+                files = [os.path.join(dir, f) for f in os.listdir(dir) if _accept(os.path.join(dir, f))]
             return string.join(files, os.pathsep)
 
         def _versionsort(s, p=re.compile(r'^(\d+)-(\d+)-*(\d*)')):
@@ -669,6 +670,7 @@ under certain conditions; type license() for details.
                         select = os.path.join(this_dir, d)
                         config_files.append(_createpath(select))
                         break
+
         if os.path.exists(self.options.config_file):
             config_files.append(self.options.config_file)
         Ganga.Utility.Config.configure(config_files, system_vars)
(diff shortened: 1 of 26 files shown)


python/Ganga/Runtime/bootstrap.py
```diff
@@ -1069,12 +1070,12 @@ def convert_merger_to_postprocessor(j):
 
         # JobTree
         from Ganga.Core.GangaRepository import getRegistry
-        jobtree = GPIProxyObjectFactory(getRegistry("jobs").getJobTree())
+        jobtree = getRegistry("jobs").getJobTree()
```
drmarkwslater: Shouldn't this have an `addProxy` call on it as well?

python/Ganga/Runtime/bootstrap.py
```diff
@@ -1069,12 +1070,12 @@ def convert_merger_to_postprocessor(j):
...
         exportToGPI('jobtree', jobtree, 'Objects', 'Logical tree view of the jobs')
         exportToGPI('TreeError', TreeError, 'Exceptions')
 
         # ShareRef
-        shareref = GPIProxyObjectFactory(getRegistry("prep").getShareRef())
+        shareref = getRegistry("prep").getShareRef()
```
drmarkwslater: Again: Shouldn't this have an `addProxy` call?

python/ganga/__init__.py
```diff
@@ -16,28 +14,33 @@
...
+from Ganga.Utility.files import expandfilename
+from os.path import expandvars, expanduser, normpath, dirname, abspath, join, isfile, islink, exists, isdir
+from inspect import currentframe, getfile
+from os import environ, listdir, pathsep
+import re
 
-logger = getLogger(modulename=True)
+logger = getLogger('ganga')
 
+logger.info("Hello from ganga")
```
drmarkwslater: This should probably be removed :)

rob-c: ah yes, typo on my part, was there for testing as for some reason passing 'modulename=True' didn't display anything for me

python/Ganga/Runtime/bootstrap.py
rob-c: getJobTree is defined within the job registry is it not?
If you're referring to the 'exportToGPI' does all of the difficulties of wrapping modules/objects/classes/my-sanity in a proxy for you now so you don't have to add proxies to the object before passing it to here as this layer expects to always add a proxy when the object in question is being exposed to the user.

python/Ganga/Runtime/bootstrap.py
rob-c: same logic as above

python/Ganga/Runtime/bootstrap.py
drmarkwslater: Good point - hadn't noticed that so actually the original `GPIProxyObjectFactory` was unnecessary. Ignore me!

drmarkwslater:
Looks good - I suspect as time goes by we may refactor the config loading somewhere else but certainly as a stop gap to load up `.gangarc` properly, it looks fine to me!

rob-c:
Yes I was looking at this and decided I can probably refactor this to load the user and global config from the same function and move out a load of static code to a helper class but I think I'm more keen to worry about that when ```import ganga``` just works, and it does for LHCb atm.
I've an incoming fix for pprint and queues which I'll upload in a few min

python/Ganga/Runtime/bootstrap.py
rob-c: I'm wanting to move away from ```GPIProxyObjectFactory``` to just use ```addProxy``` since we're now allowing for users to add a proxy to a class. Hence the changes which started this.
I think the ```addProxy``` should eventually cope with adding proxies to objects and classes atm but it should make dealing with adding/removing proxies a lot more transparent (I hope)

rob-c:
This PR has been superseded with development elsewhere and I think some lessons were learned after it's initial implementation.

PR #206, opened 2016-02-18
Cleanup/removing  construct

rob-c:
After some playing around with cleaning up the way that the Proxy is used elsewhere in the code I decided to bite the bullet and remove the ```__construct__``` functions everywhere.

This method wasn't written for all classes and for the classes it was used for it appeared to be misused over using the class constructor.

I would like to see this go into 6.1.17 as these functions have at least caused confusion and have at worst caused strange conflicting results if the ```__construct__``` method re-initialized an object or did something not always in ```__init__```.

If a class wants to implement something in a sane way and execute 'special' code when it's 'in front' of the proxy it should do it in ```_auto__init__```.

A good example of this is the Job class.

```Job.__init__``` fully creates a new job instance
```Job._auto__init__``` is only called when the job is created from a Job proxy and adds the job to the repository in this step.

```Job.__construct__``` was designed to be a copy constructor, but not a copy constructor from inspecting the code and was at the least deadweight and at worst caused errors due to the copy methods being broken.

The Proxy layer now also makes use of a copyFrom method which only copies persisted copyable attributes from the schema between classes.
This also provides some level of safety in that if the schema don't match it should throw an appropriate error.

This gets rid of a lot of dead code and makes it much clearer how some classes are being constructed/initialized.

This method used to be abused elsewhere but after a lot of good changes in the way that we handle XML files and proxies these pieces of code have been re-written to be much more sensible and have dropped this badly used method. The result is that it was only ever called in the proxy constructor.

### CODE REVIEW

**Comment by milliams** [comment]
If I remember correctly, `__construct__` is used for file objects so that we can do `LocalFile('/my/file')` without having to specify the name of the schema item. I think that as much as I'd like to get rid of `__construct__` we need to keep this functionality.

**Comment by rob-c** [comment]
I'd really, __really__ like to see this method go away. Once it's gone it's so it's much clearer what's actually going on.
(modulo a minor change I need to make in the Proxy class for handling constructing an object.)

The construct method in the case you mention would only be operating on the un-named options passed in that case and the constructor should be handling this correctly.
Frankly I expect the following line to produce the same thing:

```
LocalFile(name="somerequiredName")
LocalFile("somerequiredName")
```

and whilst they do at the moment (or should), they're doing it in a very round about way that's difficult to decipher.

If the ```__init__``` method doesn't handle the flat args then that should be treated as a bug and resolved which I'll happily address that on this branch.
I want to do some testing before I push for merger but this is something I'm keen to see in 6.1.17 or 6.1.18.

egede:
As we agreed today, there will be no "structural" changes going into 6.1.17. We will use that release to add in some features for the experimental plugins and hopefully conclude that we have the stability in place to create a 6.2.0 release. We can then consider to start some of these structural changes again. No harm at all at having them sitting as pull requests though.

drmarkwslater:
I would agree with @rob-c I think - is there any reason that whatever `__construct__` does **can't** be handled by `__init__`? I would naively assume that's where it should be put. If there are cases where it the construction can't be handled then we maybe just put in specific exceptions for those, or at least name the function something a bit clearer! Just my 2 cents :)

rob-c:
I would be happy to delay this until 6.1.18 if there is a strong call for this, but this is code that leads to strange behaviour due to the ```__construct__``` being treated as effectively an ```__init__``` replacement by the proxy system (and in the parsing of XML files).
I'm extremely keen to remove this as most of this is dead-code or worse code with unknown/unclear effects as in LHCbDataset and Job both relatively high profile classes.

(This function probably explains the remaining oddities between constructing a class normally vs constructing it from the proxy layer)

There is a bug in the use of copyFrom as this PR doesn't handle string arguments correctly, I'll update this at some point soon and we can look at the result.

rob-c:
This PR will require some work to move the logic into the correct ```__init__``` methods as it's tripping a few of the normal tests which is good to see the normal tests are beginning to highlight the effect of changes.

rob-c:
I've dropped this from 6.1.20.

The main difficulty with dropping the ```__construct__``` functions is the need to migrate functionality to specific classes. Mainly this is the ability to construct many of the ```IGangaFile``` objects using no keywords. This would make the proxy init much cleaner imo but it's a nicety over something to get into 6.1.20

rob-c:
This has been superseded by better work elsewhere.

PR #207, opened 2016-02-19
Fixes for GangaUnitTest

milliams:
In the process of fixing the registry lock-ups I've come across some bugs in the test suite which prevents the tests from running correctly. This primarily affects the queues starting and shutting down correctly but also makes the per-test config system work properly. These are all needed in order to run the `TestQueuedSubmit` test which will be incoming.

This will need to go in before I can merge my registry fixes as it allows the test to run correctly.

### DIFF
--- a/python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
@@ -14,7 +14,7 @@ def startUpQueues():
         exportToGPI('queues', _global_queues, 'Objects')
 
         import atexit
-        atexit.register((-10., shutDownQueues), ())
+        atexit.register((-10., shutDownQueues))
 
     else:
         logger.error("Cannot Start queues if they've already started")
--- a/python/Ganga/new_tests/GPI/GangaUnitTest.py
+++ b/python/Ganga/new_tests/GPI/GangaUnitTest.py
@@ -36,11 +36,17 @@ def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
     import Ganga.Runtime
     this_argv = [
         'ganga',  # `argv[0]` is usually the name of the program so fake that here
-        '-o[Configuration]RUNTIME_PATH=GangaTest',
-        '-o[Configuration]user=testframework',
-        '-o[Configuration]gangadir='+str(gangadir_for_test),
-        '-o[Configuration]repositorytype=LocalXML',
-        '-o[TestingFramework]ReleaseTesting=True',
+    ]
+
+    # These are the default options for all test instances
+    # They can be overridden by extra_opts
+    default_opts = [
+        ('Configuration', 'RUNTIME_PATH', 'GangaTest'),
+        ('Configuration', 'gangadir', gangadir_for_test),
+        ('Configuration', 'user', 'testframework'),
+        ('Configuration', 'repositorytype', 'LocalXML'),
+        ('TestingFramework', 'ReleaseTesting', True),
+        ('Queues', 'NumWorkerThreads', 2),
     ]
 
     # FIXME Should we need to add the ability to load from a custom .ini file
@@ -58,13 +64,15 @@ def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
     except:
         do_config = True
 
+    # For all the default and extra options, we set the session value
+    from Ganga.Utility.Config import setConfigOption
+    for opt in default_opts + extra_opts:
+        setConfigOption(*opt)
+
     if do_config:
         # Perform the configuration and bootstrap steps in ganga
         logger.info("Parsing Configuration Options")
         Ganga.Runtime._prog.configure()
-        from Ganga.Utility.Config import setConfigOption
-        for extra_opt in extra_opts:
-            setConfigOption(*extra_opt)
         logger.info("Initializing")
         Ganga.Runtime._prog.initEnvironment(opt_rexec=False)
     else:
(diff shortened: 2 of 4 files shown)


milliams:
retest this please

milliams:
retest this please

python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
```diff
@@ -14,7 +14,7 @@ def startUpQueues():
         exportToGPI('queues', _global_queues, 'Objects')
 
         import atexit
-        atexit.register((-10., shutDownQueues), ())
+        atexit.register((-10., shutDownQueues))
```
rob-c: Thanks for putting this in, I had forgotten to launch a PR for this.

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -58,13 +64,15 @@ def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
     except:
         do_config = True
 
+    # For all the default and extra options, we set the session value
+    from Ganga.Utility.Config import setConfigOption
+    for opt in default_opts + extra_opts:
+        setConfigOption(*opt)
```
rob-c: Nice, never thought of using it this way :)

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -79,6 +87,10 @@ def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
         else:
             logger.info("InternalServices still running")
 
+        # The queues are shut down by the atexit handlers so we need to start them here
+        from Ganga.Core.GangaThread.WorkerThreads import startUpQueues
+        startUpQueues()
```
rob-c: Is this always needed? I forget where 'startUpQueues' is in the bootstrap code but I don't think the function prevents from the queues system being 'started' twice which may lead to threads being lost.
Is it worth adding a test within 'startUpQueues' for this?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -162,7 +179,7 @@ def setUp(self, gangadir=None, wipe_repo=None, extra_opts=[]):
         # This is called before each unittest
         if gangadir is None:
             import os
-            gangadir = os.path.join('$HOME/gangadir_testing', self.__class__.__name__)
+            gangadir = os.path.join('$HOME/gangadir_testing')  # TODO make separate dirs per test suite, `self.__class__.__name__`
```
rob-c: Is there a reason to avoid using this?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -175,6 +192,9 @@ def setUp(self, gangadir=None, wipe_repo=None, extra_opts=[]):
         self.__class__.wipe_repo = self.wipe_repo
         start_ganga(gangadir_for_test=gangadir, extra_opts=extra_opts)
 
+        from Ganga.GPI import config
+        print('gangadir', config['Configuration'].gangadir)
```
rob-c: Is this effectively a debug statement?

python/Ganga/new_tests/GPI/GangaUnitTest.py
milliams: `startUpQueues` is in `initEnvironment` in bootstrap.py so currently it is only run if `do_config` is `True`. If we could move it elsewhere then this would not be needed.

`startUpQueues` has a `if _global_queues is None:` before doing anything so it's safe to run if there are already queues running.

rob-c: Ah, so it does, in that case never mind my comment :)

Once 'ganga' has replicated 90% of the functionality of the bootstrap we can just use that rather than trying to fix too much in the bootstrap now :)

python/Ganga/new_tests/GPI/GangaUnitTest.py
milliams: At the moment the test harness thinks each test is being run with its own gangadir and so it's removing those directories as it goes. However, each test actually always runs in the gangadir of the first test that was run (at least for the registry) and so things are not being cleaned up correctly. This is because the registry location-setting start-up code is only run once and it is never updated. If we fix that in the future then we can go back to running with a gangadir per test.

python/Ganga/new_tests/GPI/GangaUnitTest.py
milliams: Yes, I will remove it.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: Is this the result of running multiple unittests from the one ```nosetests``` command?
I think in this case this means that the ```tearDownClass``` needs to be expanded to _de-configure_ Ganga once each test is over.
(I think this is a topic for a separate PR but we some tools here ready for when this is addressed :) )

Good spot with regards to that, I never considered the registry would be bleeding due to this :)

milliams: In the near future I'd like to switch to using pytest as our test runner which has [the option](https://pypi.python.org/pypi/pytest-xdist) of running each test in an isolated subprocess. I think this wil help fix a lot of these test problems we're seeing. I've got a branch working with a PR incoming.

python/Ganga/new_tests/GPI/GangaUnitTest.py
milliams: Indeed. I've got plans for a much more elegant fixture using pytest and little ganga.

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: Ah, that sounds quite nice :)
In that case it's not worth investing too much time in making ```tearDownClass``` too complex

rob-c:
OK, looks good from me :)

milliams:
retest this please

milliams:
retest this please

PR #208, opened 2016-02-19
Make updateDict_ts so that it's recreated each time

milliams:
When stopping and restarting the monitoring, some jobs were not being
monitored as the system had stale data claiming that they were already
covered and so no new monitoring should be performed.

By making updateDict_ts a member of the JobRegistry_Monitor class it will
be recreated every time the monitoring is started and so will recheck
everything.

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -17,7 +17,7 @@ from Ganga.GPIDev.Base.Proxy import isType, stripProxy, getName, getRuntimeGPIOb
 from Ganga.GPIDev.Lib.Job.Job import lazyLoadJobStatus, lazyLoadJobBackend
 
 # Setup logging ---------------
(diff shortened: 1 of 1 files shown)


rob-c:
Do we know if this is safely re-created for disable/enable Monitoring commands?

milliams:
I don't understand; locking *which* thread?

rob-c:
```SynchronisedObject``` has a thread lock

**Comment by milliams** [comment]
Ahh, I understand. No, at this point the `JobRegistry_Monitor` thread is the only thing that is accessing the object so it should be safe to remove the locking. I think I'll leave that for another time as I don't want to change any more right now than necessary. It would be a simple case of changing:
```python
self.updateDict_ts = SynchronisedObject(UpdateDict())
```
to
```python
self.updateDict_ts = UpdateDict()
```
but in the name of caution I will defer this.

rob-c:
I'm happy to defer this I'm just interested in getting opinions on how to deal with it.
I think we only ever have one JobRegistry_Monitor but I'd rather not change this now and get bitten later.
Looks good to merge from me.

PR #209, opened 2016-02-19
Fix registry deadlocks

milliams:
This branch attempts to fix the various deadlocks surrounding the registry. It makes quite a few changes to the registry so it might be easier for some of them to go through commit-by-commit since they are logically grouped there,

Mostly this is making sure that the registry locks are all exception-safe as well as some simplification surrounding the flushing to avoid common deadlocks there. Overall this branch will be flushing to disk less often than before but it should still always be flushing everything on shutdown. I have a WIP for an auto-flushing thread for each registry but that will have to wait for another PR.

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -244,44 +244,13 @@ class Registry(object):
             Raise RegistryKeyError"""
         logger.debug("__getitem__")
         try:
-            self.lock_transaction( this_id, "_getitem")
-
-            real_id = None
-            if type(this_id) is int:
-                if this_id >= 0:
-                    ## +ve integer, should be in dictionary
-                    real_id = this_id
-                    this_obj = self._objects[this_id]
-                else:
-                    ## -ve integer should be a relative object in dictionary
-                    real_id = self._objects.keys()[this_id]
-                    this_obj = self._objects[real_id]
-            else:
-                ## NOT an integer, maybe it's a slice or other?
-                this_obj = self._objects[this_id]
-
-            logger.debug("found_object")
-
-            found_id = None
-            if hasattr(this_obj, _id_str):
-                found_id = getattr(this_obj, _id_str)
-            if hasattr(this_obj, _reg_id_str):
-                found_id = getattr(this_obj, _reg_id_str)
-            if found_id is not None and real_id is not None:
-                assert( found_id == real_id )
-
-            logger.debug("Checked ID")
-
-            return this_obj
-
+            return self._objects[this_id]
         except KeyError as err:
-            logger.debug("Repo KeyError: %s" % str(err))
-            logger.debug("Keys: %s id: %s" % (str(self._objects.keys()), str(this_id)))
+            logger.debug("Repo KeyError: %s" % err)
+            logger.debug("Keys: %s id: %s" % (self._objects.keys(), this_id))
             if this_id in self._incomplete_objects:
                 return IncompleteObject(self, this_id)
             raise RegistryKeyError("Could not find object #%s" % this_id)
-        finally:
-            self.unlock_transaction(this_id)
 
     @synchronised
     def __len__(self):
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -626,40 +626,6 @@ class Registry(object):
         else:
             self._flush(objs)
 
-    def _dirty(self, _obj):
-        """ Mark an object as dirty.
-        Trigger automatic flush after specified number of dirty hits
-        Raise RepositoryError
-        Raise RegistryAccessError
-        Raise RegistryLockError"""
-        logger.debug("_dirty")
-        obj = stripProxy(_obj)
-        self._updateIndexCache(obj)
-
-        if self.find(obj) in self._inprogressDict.keys():
-            self.dirty_objs[getattr(obj, _reg_id_str)] = obj
-            self.dirty_hits += 1
-            return
-
-        self._write_access(obj)
-        self._lock.acquire()
-        try:
-            self.dirty_objs[getattr(obj, _reg_id_str)] = obj
-            self.dirty_hits += 1
-            if self.checkShouldFlush():
-                self._backgroundFlush([obj])
-            # HACK for GangaList: there _dirty is called _before_ the object is
-            # modified
-            for this_d in self.changed_ids.itervalues():
-                this_d.add(self.find(obj))
-        except (RepositoryError, RegistryAccessError, RegistryLockError, ObjectNotInRegistryError) as err:
-            raise err
-        except Exception as err:
-            logger.debug("Unknown Flush Exception: %s" % str(err))
-            raise err
-        finally:
-            self._lock.release()
-
     @synchronised
     def _flush(self, objs):
         """
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -612,20 +612,6 @@ class Registry(object):
 
             self.unlock_transaction(obj_id)
 
-    def _backgroundFlush(self, _objs=None):
-
-        if _objs is not None:
-            objs = [stripProxy(obj) for obj in _objs]
-        else:
-            objs = [obj for obj in self.dirty_objs.itervalues()]
-
-        if False:
-            thread = threading.Thread(target=self._flush, args=())
-            thread.daemon = True
-            thread.run()
-        else:
-            self._flush(objs)
-
     @synchronised
     def _flush(self, objs):
         """
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -296,8 +296,6 @@ class Registry(object):
 
     def turnOnAutoFlushing(self):
         self._autoFlush = True
-        if self.checkShouldFlush():
-            self._backgroundFlush()
 
     def isAutoFlushEnabled(self):
         return self._autoFlush
@@ -354,7 +352,6 @@ class Registry(object):
         returnable = self._objects
         return returnable
 
-    @synchronised
     def ids(self):
         """ Returns the list of ids of this registry """
         logger.debug("ids")
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -1,3 +1,4 @@
+import functools
 from Ganga.Utility.logging import getLogger
 
 from Ganga.Core import GangaException
@@ -154,6 +155,18 @@ class IncompleteObject(GangaObject):
         return "Incomplete object in '%s', ID %i. Try reload() or remove()." % (self.registry.name, self.id)
 
 
+def synchronised(f):
+    """
+    This decorator must be attached to a method on a ``Registry``
+    It uses the object's lock to make sure that the object is held for the duration of the decorated function
+    """
+    @functools.wraps(f)
+    def decorated(self, *args, **kwargs):
+        with self._lock:
+            return f(self, *args, **kwargs)
+    return decorated
+
+
 class Registry(object):
 
     """Ganga Registry
@@ -225,11 +238,11 @@ class Registry(object):
         self.hard_lock[this_id].release()
 
     # Methods intended to be called from ''outside code''
+    @synchronised
     def __getitem__(self, this_id):
         """ Returns the Ganga Object with the given id.
             Raise RegistryKeyError"""
         logger.debug("__getitem__")
-        #self._lock.acquire()
         try:
             self.lock_transaction( this_id, "_getitem")
 
@@ -269,37 +282,22 @@ class Registry(object):
             raise RegistryKeyError("Could not find object #%s" % this_id)
         finally:
             self.unlock_transaction(this_id)
-        #    self._lock.release()
 
+    @synchronised
     def __len__(self):
         """ Returns the current number of root objects """
         logger.debug("__len__")
-        #self._lock.acquire()
-        try:
-            return len(self._objects)
-        finally:
-            pass
-        #    self._lock.release()
+        return len(self._objects)
 
+    @synchronised
     def __contains__(self, this_id):
         """ Returns True if the given ID is in the registry """
         logger.debug("__contains__")
-        #self._lock.acquire()
-        try:
-            return this_id in self._objects
-        finally:
-            pass
-        #    self._lock.release()
+        return this_id in self._objects
 
     def updateLocksNow(self):
         loger.debug("updateLocksNow")
-        #self._lock.acquire()
-        try:
-            self.repository.updateLocksNow()
-            return
-        finally:
-            pass
-        #    self._lock.release()
+        self.repository.updateLocksNow()
 
     def trackandRelease(self):
 
@@ -353,9 +351,9 @@ class Registry(object):
         if self._flushLastTime is None:
             self._flushLastTime = timeNow
 
+    @synchronised
     def checkShouldFlush(self):
         logger.debug("checkShouldFlush")
-        self._lock.acquire()
 
         timeNow = time.time()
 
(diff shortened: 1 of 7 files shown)


python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -225,81 +236,35 @@ def unlock_transaction(self, this_id):
...
-
-            logger.debug("found_object")
-
-            found_id = None
-            if hasattr(this_obj, _id_str):
-                found_id = getattr(this_obj, _id_str)
-            if hasattr(this_obj, _reg_id_str):
-                found_id = getattr(this_obj, _reg_id_str)
-            if found_id is not None and real_id is not None:
-                assert( found_id == real_id )
```
rob-c: I'd really like to see these assert statements kept if possible. This reduced problems during testing/development in the past and highlighted some issues.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -323,17 +288,6 @@ def trackandRelease(self):
...
                  flush_thread = threading.Thread(target=self._flush, args=())
                  flush_thread.run()
-
-                 for obj in self.dirty_objs.itervalues():
-
-                    if self.shouldReleaseRun is False:
-                        break
-
-                    _args = ([obj])
-                    release_thread = threading.Thread(target=self._release_lock, args=_args)
```
rob-c: Removing this will cause issues when LHCb users try to submit with multiple ganga instances due to the Prep repository not releasing it's whole table lock until shutdown.

rob-c: I'm not strongly arguing to keep this, I'd just like to highlight that this causes other issues.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -393,68 +345,50 @@ def checkShouldFlush(self):
...
         return decision
 
     def _getObjects(self):
         logger.debug("_getObjects")
-        #self._lock.acquire()
         returnable = self._objects
-        #self._lock.release()
         return returnable
 
     def ids(self):
```
rob-c: Should this be protected or does this cause issues?

python/Ganga/Core/GangaRepository/Registry.py
milliams: Unfortunately, they were causing deadlocks as they were inspecting the object inside a registry lock which was conflicting with another thread locking the registry while inside an object lock. They shouldn't be required if we have a test covering the case.

python/Ganga/Core/GangaRepository/Registry.py
milliams: Ok, I'll consider leaving this in. How much do we support submitting from multiple Ganga instances running on one repository?

python/Ganga/Core/GangaRepository/Registry.py
milliams: It probably should be locked but since it was already unlocked I was trying to not add any more locks than necessary. It will need a thorough review of the flow of information before we can go ahead and lock everything correctly.

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: I am fairly sure that we don't allow submission from multiple Ganga sessions anyway (at least to the best of my knowledge). You can have multiple Ganga sessions but only one has write access. Isn't this why we have the session files created?

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: I agree with @milliams - this should be in a unit test rather than production code. `__getitem__` should be as simple as possible.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: OK, I guess we can wait for 6.1.17 for a review of what should or shouldn't be locked...

python/Ganga/Core/GangaRepository/Registry.py
rob-c: This shouldn't be happening by design and concerns me quite a bit as I never had problems here in testing.
Why should getting an attribute from within an object the registry is in control over cause a deadlock?

python/Ganga/Core/GangaRepository/Registry.py
rob-c: The session file (when working) allows for multiple ganga sessions to ask what is locked/unlocked and to operate on job objects that were previously locked.
I know this is a minor use case and it often leads to repository corruptions but these corruptions are more to do with bad use of the locks rather than the principle being broken.

I can add the equivalent line into the prep repo in a PR to allow for the repo to be released after each operation, as it's only the prep repo this performance hit is relatively minor.

I've had briefly 2 or more monitoring threads on different systems operating on multiple different job objects without too many problems. (Ganga shouts loudly when it can't get access to a job for monitoring but eventually moves onto the next job).

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I agree we should make this as simple as possible but we've already seen that unittests don't cover 100% of all possible problems and I am worried if a simple assert statement here can cause a deadlock.

drmarkwslater: That may have been the case previously (though I'm not certain) but now, if we have a lock per object then you are acquiring the lock by accessing the object within something that is already locked. This makes life very complicated at best and will cause problems at worse. To keep things as simple as possible and avoid any chance of deadlocks, I don't think the Registry should ever access the internals of an object except for checking if it's dirty and setting the id. That's my 2 cents anyway :)

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: Oh OK, that's either changed or I never really understood it properly :)

python/Ganga/Core/GangaRepository/Registry.py
rob-c: @drmarkwslater I don't see where the object is locked here. The Registry is locked by the decorator from what I'm reading and we have a RLock which is in the ```__get__``` method but these shouldn't be interfering as the Registry doesn't inherit from GangaObject.

I think all that is happening is that within a registry lock we're calling for an attribute of a GangaObject which should manage it's own (separate) (R)locks.

If the ganga object is taking control of locking/unlocking the repo then I think this should be addressed as I don't think it's the place of the GangaObject to lock/unlock the repo at all.

If I've misunderstood something I'm keen to be corrected as this is all a bit complex.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I think in the interest of stability I'll opt for keep this removed here and I'll add a quick 2 or 3 line PR for the prep repository which is less invasive than adding yet another thread of repo actions.

rob-c: OK, I think the line ```self._releaseWriteAccess()``` is present at the end of the Prep registry methods ```increase``` and ```decrease``` so I'm happy to keep this code removed :)

python/Ganga/Core/GangaRepository/Registry.py
milliams: What is happening is that two separate threads are attempting to claim two locks in the reverse order. Thread 1 claims lock A and then tries to get lock B, meanwhile thread 2 claims lock B and then tries to get lock A. They will both then wait forever. This is a deadlock.

In our case lock A is an object lock on a specific `GangaObject` and lock B is the lock on the registry.

drmarkwslater: So, a possible example would be:

* Thread 1 does something to an object that locks that object but later (before releasing the object lock) requires a lock on the registry
* Thread 2 then requests the same object from the registry before the other thread has a lock on the registry, thus locking it first.
* Thread 1 is then waiting to get the registry lock from Thread 2
* But, Thread 2 (because of the request for the registry ID from the object) is waiting on for Thread 1 to release it's lock on the object.
* and you're stuck...

Now, I'm aware I haven't listed specifics but seeing as this **can** happen (and Matt was seeing it), for the sake of some code that I would prefer to be in a test case anyway, it's better to simplify everything as much as possible.

drmarkwslater: So how @milliams put it was a lot better and more succinct :)

milliams:
In general I'm now starting to see problems with the session locks. The problem with them is that they are done using flock (or lockf) which lock files per process. This means that two threads in the same process are not protected by the session lock. I'm seeing situations where multiple threads are trying to interact with the session lock simultaneously which is causing some strange interference.

I guess the first way to fix this would be to put a thread lock on the `SessionLock` class which technically would be a solution. However, this is starting to get very complicated and it's not clear to me what the canonical source of information about an object is, what are read-only views and what are free, lockless objects. Currently a `GangaObject` can be in multiple processes, multiple registries, multiple threads and maybe none of the above.

drmarkwslater:
Are these problems with the Session Locks a show stopper? I'd prefer to put off changing **another** major subsystem of Ganga until we have the Registry/Threading under control :)

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I've some reservations about the fact I suspect this is glossing over a potential problem by reducing the amount of possible code that could conflict but I'm happy to see this go in if the unittests are passing for the parallel stuff at last.

rob-c:
The thread lock actions were designed to be safe between multiple different instances of ganga, not thread safe within themselves.
I would opt for a quick and dirty (and effective) global lock around all file transactions in this layer if it doesn't come with a large performance penalty.

milliams:
I'm hoping that I can work around them because I'd really like to not have to touch `SessionLock` at all. At the moment they're sometimes causing indefinite hangs on exit. I imagine it's due to some mismatched lock acquisition/release so hopefully should be resolvable. I think I just need to make sure that all the session lock access goes through the repository and synchronise those points.

rob-c:
This is beginning to sound like a slightly invasive change within the XML repository.

What are the errors you're seeing Matt, are they endemic of file locks being incorrectly handled or of multiple threads hammering them into an inconsistent state?

milliams:
The problem I'm seeing is:
```python
# ThreadID: 140315541059328
File: "/usr/lib64/python2.7/threading.py", line 783, in __bootstrap
  self.__bootstrap_inner()
File: "/usr/lib64/python2.7/threading.py", line 810, in __bootstrap_inner
  self.run()
File: "/home/matt/ganga/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py", line 543, in run
  self.__monStep()
File: "/home/matt/ganga/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py", line 606, in __monStep
  self.callbackHookDict[cbHookFunc][0](**cbHookEntry.argDict)
File: "/home/matt/ganga/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py", line 1079, in _checkActiveBackends
  activeBackends = activeBackendsFunc()
File: "/home/matt/ganga/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py", line 926, in __defaultActiveBackendsFunc
  fixed_ids = self.registry.ids()
File: "/home/matt/ganga/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py", line 79, in ids
  return [k for k in self.objects.keys() if minid <= k <= maxid]
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 446, in keys
  returnable = self.ids()
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 395, in ids
  changed_ids = self.repository.update_index()
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 582, in update_index
  self._write_master_cache(shutdown=isShutdown)
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 434, in _write_master_cache
  if len(self.lock(arr_k)) != 0:
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 968, in lock
  return self.sessionlock.lock_ids(ids)
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/SessionLock.py", line 823, in lock_ids
  self.session_write()
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/SessionLock.py", line 665, in session_write
  On computers with very high load and on network filesystems, try to avoid running concurrent ganga sessions for long.\n '%s' : %s" % (self.fn, x))
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepository.py", line 66, in __init__
  Repository_runtime.shutdown()
File: "/home/matt/ganga/python/Ganga/Runtime/Repository_runtime.py", line 169, in shutdown
  registry.shutdown()  # flush and release locks
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 166, in decorated
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 960, in shutdown
  self.repository.shutdown()
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 268, in shutdown
  self._write_master_cache(True)
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 434, in _write_master_cache
  if len(self.lock(arr_k)) != 0:
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 968, in lock
  return self.sessionlock.lock_ids(ids)
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/SessionLock.py", line 823, in lock_ids
  self.session_write()
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/SessionLock.py", line 649, in session_write
  fd = self.delayopen(self.fn)
File: "/home/matt/ganga/python/Ganga/Core/GangaRepository/SessionLock.py", line 611, in delayopen
  time.sleep(0.1)
```
so it's hanging in `delayopen` for a very long time. I see that earlier in the stacktrace it went through `SessionLock.py:665` which is the "Own session file not found" error. It's then doing a shutdown which is hanging. There's a chance I've caused this error with a mistake with getting read/write access so I need to check all that thoroughly.

rob-c:
The own session lock not found is far more endemic of multiple session lock management threads running by accident.
I think it's impossible for Ganga to remove it's own session file by accident.
Is there any possibility that this could be being run in the same place as another running test of equivalent or for the shutdown to not have completed correctly and bled a session lock managment thread from another session?

The delayopen was added due to what appeared to be an AFS issue but now I'm not so sure if you're asking if this layer is thread safe.

Can you try adding a global Lock object and acquire and release it in ```global_lock_acquire``` ```global_lock_release``` in Session lock? this would constitute a 4-5 line change in SessionLock but may give us some stability we need.

rob-c:
I've just tried playing with this change and it doesn't adversely effect performance on lxplus from what I can tell and gives a global hard lock around most of the SessionLock disk transactions which could interfere.

milliams:
@rob-c I've just done the same (well, with an `RLock`) and it seems to be helping. I'm still having to tweak exactly which Registry methods are synchronised (as some still cause deadlocks if not careful) but it's currently running without stalling. The ganga-ci machine is a bit slower so it's much more likely to show up there if anything is still present.

I am still seeing a very occasional `forbidden status transition of job 170 from "completed" to "running"` which I think is caused by the monitoring picking up a job either twice or too early. It's very infrequent but I haven't tracked it down yet (and will likely leave it for another PR anyway).

rob-c:
If we're still seeing the state changing in a strange way I'd be very keen to incorporate #198 which does away with possibility that we're returning a stale cache value for the job.

rob-c:
OK, if some form of additional lock is required for now in the SessionLock I'm not opposed to this :)

I think a proper review of all registry transactions is ultimately needed with some scope for global locks which stop multiple repositories hitting the same potentially unsafe piece of backend code but this has been agreed to be investigated for 6.1.17 or 6.1.18 to correctly address stability issues rather than tracking down the most common occurrences.

Once you've got these additional locks in this branch I'm keen to see it merged so I can start hammering it with some intensive LHCb tasks and see if anything crops up :)

drmarkwslater:
I would personally prefer to leave out the rather large change that #198 is doing for later but if it can directly be seen to improve the state change bugs that still seem to occasionally happen and not introduce any other bugs, then I guess it would be OK. I'll leave @milliams to decide as he can see if those changes stop the issues he's seeing or not. I'm aware though that we have already made a lot or progress as it is and so it may be worth just going with what we have and getting 6.1.16 out...

rob-c:
(Avoiding too much conversation related to another PR here) I'm happy to wait for #198 but I wrote it to address these issues but I'm happy to see it dropped back to 6.1.17 (although preferably not too much later)

milliams:
Ok, the tests seem to have passed without problems. I'm happy to wait for any more comments on this PR but if there aren't any then I'm happy to merge.

I think for now we hold off on #198 but it sounds like the next logical place to go if we're still seeing problems. Once this PR is merged we should all run some intensive local testing and if we're seeing any persistent and reproducible problems then we'll try to fix then quickly before release. Otherwise, if we're hitting the 99% good point then I think we go for release.

drmarkwslater:
I'm happy to merge :+1:

rob-c:
Feel free to merge from me :)

rob-c:
Are we ready to start testing and tag 6.1.16? I'll start running some intensive tests on develop this evening now. Thanks for the work Matt :)

milliams:
I think we're ready to start testing, yes. Whether we're ready to tag will depend on the results of the testing :smile: I think we give it go and we can make a decision on Wednesday if nothing has cropped up. If something serious comes up then we fix it and start again.

rob-c:
Sounds good :) Thanks for finding and fixing various bugs in the different parts of the repo!

PR #210, opened 2016-02-22
Reorganise tests

milliams:
This is the beginning of some clean-up work on the test infrastructure to make things work a little better in the future. The work started with the intention of allowing us to run [doctests](https://docs.python.org/2/library/doctest.html) on Ganga functions but that didn't work with the setup that we have in Ganga and nosetests. In the process I found out that nose is [now unmaintained](https://github.com/nose-devs/nose/issues/976) and so I started looking to an alternative.

[Pytest](http://pytest.org) provides two main things, a test runner (like nose does) and a test fixture system (similar to `unittest`). For now this PR only changes the runner but in future it might be good to use the fixture system too. In addition in the future pytest allows us to run tests in isolated subprocesses which will avoid the setup/teardown problems we've been having occasionally.

These changes are almost completely benign but nonetheless this will wait for at least 6.1.17.


alexanderrichards:
> The work started with the intention of allowing us to run doctests on Ganga functions but that didn't work with the setup that we have in Ganga and nosetests.

I have no strong feelings about staying with Nose vs Pytest but more for curiosity I would have thought that you can incorporate doctests into nose as you can do so even within the standard unittest framework using `doctest.DocTestSuite` and treating it like any other unittest suite.

milliams:
The problem that nose was having is that since doctest imports every module individually it was getting into very strange situations where, for example, there was a class in a file in a directory of the same name (e.g. `class MTRunner` in `.../MTRunner/MTRunner.py`) it would try to import `MTRunner` and then do somethin with it and it would say something like `AttributeError: module 'MTRunner' has no attribute 'start'`. This is a known bug with nosetests which is never going to be fixed.

The main reason for me avoiding `doctest.DocTestSuite` is that it's too manual. The point of using nose or pytest is that it can automatically find all the tests in our codebase and having to add boilerplate to run doctests (which are supposed to be almost boilerplate-free) seems a bit heavy. Given that pytest finds them without problems (and is maintained and will have bugs fixed) swung the favour their way.

I did look at [nose2](https://github.com/nose-devs/nose2) which would otherwise have been good but it was suffering the same import bug as nose.

alexanderrichards:
excellent, sounds sensible to me

alexanderrichards:
> The point of using nose or pytest is that it can automatically find all the tests in our codebase

I'm not trying to be pedantic, I'm just wanting to make sure that we are all on the same page, Is this really the main reason to use these frameworks? As the standard unittest framework also have test discovery.

https://docs.python.org/2/library/unittest.html#test-discovery

I was of the opinion that the main reason was the ease of integrating things like Coverage and creating JUint/XUnit test results for publishing via jenkins? Am I wrong?

milliams:
Ok, you're right that it's not perhaps the main reason. The discovery mechanism of nose and pytest is perhaps a little more advanced than unittest's built-in one and the interface is much nicer. The provision of JUnit/coverage is indeed a very important thing and indeed all the third-party runners provide such features.

My aim is to make writing and maintaining tests as simple and easy as possible but automating as much of the work as possible. It also means using the tools and frameworks that the rest of the Python world is using as it relieves some of the pressure on maintaining our own code (see `GangaTest`).

When our main requirements are: automatically find tests, output junit xml, output coverage xml and run doctests I found that pytest came out on top. It also allows us to run tests in isolation in the future.

alexanderrichards:
Thanks Matt, sorry for the noise. It's mainly for my own understanding really :smile: For all your above reasons it sounds like pytest is a winner and I certainly agree that the old GangaTest needed replacing

milliams:
In the name of slowly starting work on the next release I'd like to start with merging a few of the more harmless PRs, particularly those which primarily focus on increasing testing. Given that this PR does a little bit or reorganisation (moving non-GPI unit tests into their own directory) I think it makes sense to do this one first. After this I'd like to push for #183 as well.

We want to be careful about what changes we are making for 6.1.17 as we want to ensure that it is nothing but an improvement on 6.1.16 so I feel that a bugfixes-only rule (to first order) is still a good idea.

The tests for this PR are failing since this PR changes where the tests put the XML test summary so before merging I will tell Jenkins to look elsewhere and rerun the tests to make sure they're working fully.

alexanderrichards:
:+1:

milliams:
retest this please

python/Ganga/Core/GangaRepository/GangaRepositorySQLite.py
```diff
@@ -34,7 +34,7 @@ def startup(self):
             os.makedirs(self.root)
         except OSError as x:
             pass
-        self.con = sqlite.connect(os.path.join(self.root, "database.db"))
+        self.con = sqlite3.connect(os.path.join(self.root, "database.db"))
```
rob-c: Are we ever using this backend? Should we really be touching code we will likely drop if not?

python/Ganga/new_tests/Unit/TestGangaObject.py
```diff
@@ -296,9 +296,9 @@ def write_read(thread_number):
                     child_num = rand.randint(0, 1000)
                     o.b.a = child_num
                     time.sleep(rand.uniform(0, 1E-6))
-                    self.assertEqual(o.a, num)
```
rob-c: I thought we were moving to make use of the ```assertEqual``` functions? What are we choosing to standardize on?

python/Ganga/Core/GangaRepository/GangaRepositorySQLite.py
milliams: Well `sqlite` does not exist and so when pytest is searching the codebase to look for tests it tries to import this module and fails. This change was just to get it importable.

setup.py
```diff
@@ -6,15 +6,17 @@
 import sys
 
 file_path = os.path.dirname(os.path.realpath(__file__))
-ganga_python_dir = os.path.join(file_path, 'python')
```
rob-c: Given this is used in multiple places doesn't it make more sense to define it once?

python/Ganga/new_tests/Unit/TestGangaObject.py
milliams: pytest has a feature called [assertion introspection](https://pytest.org/latest/assert.html) whereby it can understand plain assert statements and provide very detailed feedback about the failure. Within `unittest.TestCase` subclasses we can continue to use `self.assertEqual()` but this allows us to write tests unconstrained by `unittest`.

setup.py
milliams: Perhaps, yes. I was partly trying to clear up the global namespace. Significantly though, the `_get_test_env` function will likely be able to be removed soon since pytest is a little cleverer about setting `PYTHONPATH` etc.

rob-c:
I'm happy to merge this today as it only really touches the unittests.

Would it make sense to have the line:
```
#!/bin/env python
```
at the top of ```setup.py``` and mark this file as executable?

I'm only asking as when I run tests locally I'm just trying to be as lazy as possible and I keep making this change by hand.

milliams:
The reason I have not included it is that it's the convention followed by everyone else. I had a look at a few well-known Python projects out there and few of them specified the shebang. I'm not against the change per-se but I have asked a [question on SO](http://stackoverflow.com/questions/35701131/why-does-setup-py-usually-not-have-a-shebang-line) to see if there's a good reason not to do it.

rob-c:
https://docs.python.org/2/distutils/setupscript.html Seems to have it in their example, and I agree with the comment it's harmless and is just to allow me to be lazy :) I'm happy to wait to see if there is a convention

python/Ganga/new_tests/Unit/TestGangaObject.py
rob-c: Just for completeness, which are we favoring in the future, ```assertEqual``` and equivalent of plain old ```==```?

Issue #213, opened 2016-02-23
Proxy round-tripping is broken


milliams:
In latest develop
```python
from Ganga.GPIDev.Base.Proxy import addProxy, stripProxy
l = LocalFile()
assert l is addProxy(stripProxy(l))
```
does not pass but it used to, even as recent as 6.1.14. This was broken I think in #163 (specifically 55d7098cdac079) where I pointed it out but didn't realise the extent of the change. It was presented then as a performance change but really we want to be careful about object identity breaking.

I think that for correctness and compatibility with previous version of Ganga I would like to change this behaviour back by attaching a `_proxyObject` attribute to a `GangaObject` from Proxy.py if it's been wrapped. Of course we want to avoid `GangaObject` having any knowledge of this and so I also propose changing the "do not copy" logic in Objects.py to an "only copy these" logic.

This is not important enough to warrant a change in 6.1.16 but would make sense soon after.

rob-c:
You're changing the proxy so the proxy is not identical. This was by design.

I'd worry if this was changing the object behind the scenes but the proxy if being created destroyed as per the earlier PR to cleanup the code.

Adding this back in again just to keep this assert causes us to add it as an exemption to the copy methods to avoid circular loops and other horrible things.

Is there a reason we ever test this or expect this assert to be true as I'm not?

egede:
@rob-c So what you say is that we need to make sure that

```
from Ganga.GPIDev.Base.Proxy import addProxy, stripProxy
l = stripProxy(LocalFile())
assert l is stripProxy(addProxy(l))
```
works?

rob-c:
My comment was that I didn't expect it to work. Where is this assert used?
The users shouldn't be managing our own proxies.


As for considering that objects returned from methods which generate objects and wrap them I'm seeing some advantage to this allowing for people to run ```assert a is b``` when they get an item aka:

```
class someClass:
...

  def __getitem__(self, num):
    self.objects = [listofunproxiedObjects]
    return addProxy(self.objects)
...

a=myClassInstance[5]
assert a is myClassInstace[5]

```

The above will at the moment fail unless we cache the proxy...


OK, I'm beginning to fall more in line restore this old behaviour rather than fix code to correctly store the proxied object in places as fixing that everywhere is more painful than handling it in the Objects.py class.

rob-c:
Fixed by #214

PR #214, opened 2016-02-23
Restoring the use of proxyObject, now much cleaner

rob-c:
This should address #213.

I would call this a bugfix based upon the example of the discussion I give in the issue above.

As for whether this is included in 6.1.16 of 6.1.17 I leave that for others to decide.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -41,7 +41,7 @@ logger = Ganga.Utility.logging.getLogger(modulename=1)
 
 _imported_GangaList = None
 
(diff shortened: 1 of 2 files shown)


rob-c:
Actually I'm tempted to bump this to 6.1.16 as this fixes:
```
j=jobs(0)
j2=jobs(0)
assert j is j2
```
in develop this fails but in this branch it works again

drmarkwslater:
Looks OK to merge to me :+1:

milliams:
I'm happy for this to be merged. @drmarkwslater has one or two PRs to fix show-stoppers incoming and once those (or at least one of them) are merged I think we're getting close to a release.

PR #215, opened 2016-02-24
Add some GangaList tests

milliams:
This is some more porting of the old test suite. This includes the basic `GangaList` tests but is not all of them. Importantly this excludes the nested list tests since those are now failing since we can now support nested `GangaList`s. I imagine that I will simple remove that test but otherwise I will need to edit them to check the nested lists fully.


rob-c:
@milliams I think as long as any test for the nested list inspects the object before and after it's flushed to disk between 2 tests this should be good enough to check the machinery is working.

PR #217, opened 2016-02-24
Make Job copy work correctly

### LINKED ISSUE #216: Job.copy doesn't respect inputdata
As mentioned in #212, inputdata isn't copied correctly when doing j.copy.

drmarkwslater:
Cleans up the `__copy__` methods and uses a `deepcopy` for j.inputdata. Also added a test for job assignment and copy that fails on develop and passes here. Fixes #216.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -209,9 +209,6 @@ class Node(object):
 
         visitor.nodeEnd(self)
 
-    def __copy__(self):
-        copied_obj = self.clone()
-
     # clone self and return a properly initialized object
     def clone(self):
         new_obj = deepcopy(self)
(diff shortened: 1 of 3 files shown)


python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -105,7 +105,8 @@ def __copy__(self, memo=None):
             else:
                 this_dict[elem] = None
         #obj.__setstate__(this_dict)
-        obj.getParent(self._getParent())
+        if self._getParent() is not None:
+            obj._setParent(self._getParent())
```
rob-c: Is this really needed?
If the object has no parent it's copy should still have no parent, or is the _parent correctly initialized to ```None```? (I suspect it might be from looking at the ```cls()```)

drmarkwslater: Fair point - I had basically blindly copied this from `__deepcopy__` but now I think about it, `obj` wouldn't have a parent set and so it should be `None` anyway. I'll remove the additional check!

rob-c: OK, cool, I was just wondering if this was needed due to some bug or not that I was unaware of :D

rob-c:
Looks good to me, I'm happy to merge :)

milliams:
:+1:

PR #219, opened 2016-02-24
Transform status not persisting

drmarkwslater:
Needed to change how GangaLists were being created in `makeGangaListByRef` as simple assignment for tasks.transforms was resulting in the status not being loaded correctly. Also added a Task test to check for general Task creation, start and persistency. Fixes #219


python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -52,8 +52,9 @@ def makeGangaListByRef(_list, preparable=False):
     result = GangaList()
     if len(_list) == 0:
         return result
-    temp_list = [stripProxy(element) for element in _list]
-    result._list = temp_list
+
+    result._list = []
```
milliams: `result._list` defaults to `[]` in the schema so this line is probably not needed.

rob-c:
Go for merge from me, looks like something surprisingly subtle, appologies I suspect this bug was likely my fault at some point in dealing with the atomicity problems in ```__set__```

drmarkwslater:
I may hold off at the moment as the tests seem to occasionally hang :disappointed: I'll try to have a look today and get them working reliably.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -52,8 +52,8 @@ def makeGangaListByRef(_list, preparable=False):
     result = GangaList()
     if len(_list) == 0:
         return result
-    temp_list = [stripProxy(element) for element in _list]
-    result._list = temp_list
+
+    result._list.extend([stripProxy(element) for element in _list])
```
rob-c: Erm, after some thought, does this correctly set the parent of the objects here. This subverts the setParent calls in the parent class of ```_list``` so we may want to add a step to fix this since you don't get it from a copy.

milliams: `makeGangaListByRef` shouldn't change anything about the reference objects as it is copying *by ref*. Now, I think that because of this `__deepcopy__` should be changed back to call `makeGangaList` instead. It was changed to copy by ref in https://github.com/ganga-devs/ganga/commit/72e674ebb6858074fd9de975537a5a45507ee278#diff-49914f23e50f4cbb9e935817edbb86a4R289 but I don't that this is what we want.

rob-c: In that case how does the GangaList know that something else hasn't changed the objects it's supposed to be responsible for? How can we prevent someone deleting an element elsewhere without this class still having a copy as you would expect.

drmarkwslater:
Given the discussions, I suspect we should definitely hold off on this until 6.1.17 unless we come to a quick conclusion in the next 24 hours.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: Can we do away with the ```__deepcopy__``` method actually? it doesn't do anything not handled within GangaObject.

milliams: Hmm, well historically it looked like:
```python
def makeGangaListByRef(_list):
    """Faster version of makeGangaList. Does not make a copy of _list but use it by reference."""
    result = GangaList()
    result._list = _list
    return result
```
which I feel has the correct sematics. Now, this doesn't work with the current codebase due to some funky interception of the assignment to `_list`. If possible I would like to get back to this working in this way.

milliams: Regarding `__deepcopy__`, I feel that yes, perhaps we can get rid of it. Will need some tests to confirm that everything is behaving as expected though.

alexanderrichards: @milliams you beat me to it :smile: I think that the point of this method is to short cut any expensive list iteration/comprehension/duplication routines and simply assign the underlying list from the one given in the args

rob-c: I'm happy to keep the changes to ListByRef, infact I support it, but we do need to intercept bad assignments to list as I've seen it abused in too many other places in the codebase in the past.

This may not be the case due to recent changes/fixes/developments but we need to prevent for people doing bad things to ```_list``` such as breaking ownership.

I'm weary that by changing this here we effect:
```
./Ganga/Core/GangaRepository/VStreamer.py
./Ganga/GPIDev/Lib/Job/Job.py
./GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```

We should check these files to be 1,000% sure that they should be copying by ref vs using this incorrectly and causing things to break.
If we verify all uses of this function are correct we can at least fix this small problem.

rob-c: OK, after grepping the codebase only LHCbDataset abuses ```._list``` now so we can consider removing the intercept and warn users never to go near this internal parameter when modifying lists. (Which should be hidden behind a proxy anyway).

drmarkwslater: So I think we're probably in agreement that makeGangaListByRef should go back to just being a ref copy and makeGangaList should be (essentially) a deepcopy and be used mostly unless the reference version is specifically needed. Is that the case?

If so, I'll start working on that but before then it would be very good to get a performance test in to make sure we don't cripple the performance improvements already in. Has anyone got a good example of this they can put in with a comparison on, say 6.0.44/6.1.x?

rob-c: @drmarkwslater This will not by design impact the write performance changes which turn off dirty cache flushing and turns it back on after a full ```__set__``` operation is complete.
Not unless you add something to the equivalent of:
```
for a in list:
  new_obj[a] = old_obj[a]
```

anything to the extent of:

```
newlist = oldlist
```

is fully atomic at the registry level (or really should be by now).


I'm very weary of something such as the code below failing:

```
## pseudo-code as this 'I think' works
job1.something = GangaList([elements])
job2.something.extend(job1.something)
p2 = job2.something[element_i].getJobObject()
p1 = job1.something[element_i].getJobObject()
assert p1 is not p2
```

I think we should be protected by the assignment operators now being correctly fixed to handle the parent but I'd like to have a test of what parent (and/or job object) each element has.

milliams: Indeed at the moment the last example you put would fail. This is because `extend` is simply:
```python
def extend(self, ittr):
    for i in ittr:
        self.append(i)
```
It all really depends on exactly what semantics we want things to have. At the GPI-level I think we want things to be copied gratuitously in order to avoid stale references etc. In this case, we would need to make sure that `GangaList._export_extend` correctly deep-copies the objects before appending them.

rob-c: Surely it won't as you're subverting:

GangaList.append where GangaList is a GangaObject
to
GangaList._list.append where GangaList._list is a list object which is not governed by the ```__set__``` and ```__get__``` methods I think.

rob-c: Apologies I think I lost context here in this thread with my last comment I think I mixed up the parent issue with other stuff.

egede:
@drmarkwslater Can you please comment on which Task related bugfixes this PR address. I am a bit lost in it,.

drmarkwslater:
@egede It fixes the problem of the status of a Transform not persisting after you restart Ganga. This makes the task revert to `new` status.

drmarkwslater:
This has now morphed into #356

PR #220, opened 2016-02-24
Tidy SessionLock count

milliams:
There are two things here. One is a bugfix so that 11d05d5322ddd will actually do what it meant to and the other is some cleanup of the same code to make it much simpler.


milliams:
Am I OK to go for merge here or are there comments still to be made?

drmarkwslater:
:+1:

rob-c:
Does this address any particular problem other than just cleaning up the code? (I'm all for cleaning up the code I was just wondering if there was an issue I missed to do with the SessionLock?)

milliams:
The majority of the changes are just clean-up but the central change is in b595e254b845c0 which makes it actually return the cached value if it exists. This doesn't fix an observable *bug* per se but it is incorrect behaviour which reduces performance to some (small) degree.

rob-c:
Am I correct in thinking this shouldn't cause the stored dictionary to explode unless there are some several thousand (master) jobs constructed and submitted?

milliams:
Which stored dictionary?

rob-c:
*tuple rather than dictionary

milliams:
Indeed, the `val` element will be as large as the number of locked master jobs so it shouldn't explode. Certainly will be small in comparison to other things in memory.

rob-c:
OK, if this fixes some things feel free to merge

PR #221, opened 2016-02-24
Add current release notes for 6.1.16

milliams:
This is what I could find looking through https://github.com/ganga-devs/ganga/issues?q=milestone%3A6.1.16+is%3Aclosed


egede:
@milliams Is this the text that the user gets printed when they start Ganga? If yes, it is overly technical and would be improved if there were a few lines of *normal* text on top that explains what difference the user should see (in this case, stability of monitoring, consistency of status changes, ...).

Issue #223, opened 2016-02-25
exportToGPI test bug


rob-c:
As part of #192 I've discovered a few problems with the startup/stop ganga methods.

Basically I've discovered the problem that exported 'Objects' in the GPI should be replaced during a new instance during the ganga startup, but unfortunately they are not.

The result of this is that tests are inconsistently addressing objects which in some cases should have been destroyed (or attempted to be destroyed between tests by the shutdown system).

This only effects objects exposed via the GPI as we expect functions and classes to remain the same.
However, the objects exposed this way includes the Repositories as well as the queues system.

I'm working on a fix for 6.1.18 (or later) which makes sure that on a fresh ganga startup these objects within the GPI are replaced with a new instance.

This directly effects the bug that we saw that the repository directory doesn't follow changes in the 'gangadir' between tests so would allow us to move toward making sure that each test is run on it's own clean repository in the future.

PR #224, opened 2016-02-25
Bugfix/tests fix for exportToGPI

rob-c:
This addresses #223 

Aside from a few lines of change in the shutdown of the threadpool and queues shutdown methods and the addition of an extra function into the Repository_runtime this PR is mainly focused on fixing a bug in ```GangaUnitTest``` where objects were referencing a ```queues``` and ```repository``` objects which should have been destroyed and recreated between tests.

It is explicitly assumed that the credentials objects are safe to exist across multiple tests, however should this not be the case the reference to them should probably be removed from the GPI at the end of the shutdown and the objects correctly recreated upon the startup of GangaUnitTest.

This is a relatively trivial fix with regards to changes in Core but effects the test system in quite a strong way so I'm keen to see if the tests pass correctly with this PR.

### DIFF
--- a/python/Ganga/Runtime/Repository_runtime.py
+++ b/python/Ganga/Runtime/Repository_runtime.py
@@ -93,6 +93,21 @@ def checkDiskQuota():
 
     return
 
+def bootstrap_reg_names():
+    # ALEX added this as need to ensure that prep registry is started up BEFORE job or template
+    # or even named templated registries as the _auto__init from job will require the prep registry to
+    # already be ready. This showed up when adding the named templates.
+    def prep_filter(x, y):
+        if x.name == 'prep':
+            return -1
+        return 1
+
+    names = []
+    for registry in sorted(getRegistries(), prep_filter):
+        names.append(registry.name)
+
+    return names
+
 def bootstrap():
     retval = []
 
--- a/python/Ganga/new_tests/GPI/GangaUnitTest.py
+++ b/python/Ganga/new_tests/GPI/GangaUnitTest.py
@@ -76,6 +76,23 @@ def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
         logger.info("Initializing")
         Ganga.Runtime._prog.initEnvironment(opt_rexec=False)
     else:
+        from Ganga.Runtime.GPIexport import exportToGPI
+
+        from Ganga.Runtime import Repository_runtime
+        # boostrap the repositories and connect to them
+        for n, k, d in Repository_runtime.bootstrap():
+            # make all repository proxies visible in GPI
+            exportToGPI(n, k, 'Objects', d)
+
+        # JobTree
+        from Ganga.Core.GangaRepository import getRegistry
+        jobtree = getRegistry("jobs").getJobTree()
+        exportToGPI('jobtree', jobtree, 'Objects', 'Logical tree view of the jobs')
+
+        # ShareRef
+        shareref = getRegistry("prep").getShareRef()
+        exportToGPI('shareref', shareref, 'Objects', 'Mechanism for tracking use of shared directory resources')
+
         # The queues are shut down by the atexit handlers so we need to start them here
         from Ganga.Core.GangaThread.WorkerThreads import startUpQueues
         startUpQueues()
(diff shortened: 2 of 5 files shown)


rob-c:
Does anyone have any objections to merging this branch early in 6.1.17?
(Assuming the integration tests pass as expected)

This mainly effects the unittests and seeks to make sure that each test is isolated within it's own Ganga repository on disk by addressing some issues of some global GPI objects persisting between each unittest.

rob-c:
Any objections to this if it passes the unittests and only seems to have a small change in the codecov?

python/Ganga/Runtime/Repository_runtime.py
```diff
@@ -93,6 +93,21 @@ def checkDiskQuota():
...
+    # or even named templated registries as the _auto__init from job will require the prep registry to
+    # already be ready. This showed up when adding the named templates.
+    def prep_filter(x, y):
+        if x.name == 'prep':
+            return -1
+        return 1
+
+    names = []
+    for registry in sorted(getRegistries(), prep_filter):
+        names.append(registry.name)
```
milliams: This code seems to be duplicated from the `bootstrap` function. Can that other function be made to call this one?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -76,6 +76,27 @@ def start_ganga(gangadir_for_test='$HOME/gangadir_testing', extra_opts=[]):
...
+        exportToGPI('jobtree', jobtree, 'Objects', 'Logical tree view of the jobs')
+
+        # ShareRef
+        shareref = getRegistry("prep").getShareRef()
+        exportToGPI('shareref', shareref, 'Objects', 'Mechanism for tracking use of shared directory resources')
+
+        # The queues are shut down by the atexit handlers so we need to start them here
+        from Ganga.Core.GangaThread.WorkerThreads import startUpQueues
+        startUpQueues()
+
```
milliams: This duplicates quite a lot of code from bootstrap.py. Is there any way to avoid this?

python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -179,7 +214,7 @@ def setUp(self, gangadir=None, wipe_repo=None, extra_opts=[]):
         # This is called before each unittest
         if gangadir is None:
             import os
-            gangadir = os.path.join('$HOME/gangadir_testing')  # TODO make separate dirs per test suite, `self.__class__.__name__`
+            gangadir = os.path.join('$HOME/gangadir_testing', self.__class__.__name__)
```
milliams: Are you sure this is working correctly now?

rob-c: I just ran a test and I noticed that there was some bleed of metadata knowledge. Fixing in a small addendum to the PR now :D

rob-c:
OK, waiting for the integration tests for this to re-run now and this should isolate the repositories correctly.
I'm going to open a separate PR as I've discovered a bad bug which needs addressing correctly in Ganga >6.1.19 or even Ganga >6.2.xx

rob-c:
OK, running the tests on one lxplus node and examining the test directory in another node I see that multiple different repositories are constructed/removed between tests which should help further isolate tests from one another as a 'just in case' improvement.
I'm keen to see this as I've discovered potentially a very nasty bug in the ```prep.metadata```... again

python/Ganga/new_tests/GPI/GangaUnitTest.py
rob-c: Yes, move the code out of bootstrap and into somewhere else but I don't want to touch bootstrap as this becomes invasive and I'm not wanting to risk something bad happening as part of 6.1.17.

python/Ganga/Runtime/Repository_runtime.py
rob-c: Probably, it's a minor change I can get in later on before merge

rob-c:
@milliams Any objections to merging this branch as it is now?

milliams:
This looks like it's correctly updating the registry and repository locations now so I'm happy for you to merge :+1:

PR #225, opened 2016-02-25
Fix for NodeIndexCache to avoid it getting out of sync, a small PR

rob-c:
Fixes the NodeIndexCache problems


rob-c:
I've pulled in #224 as I was having issues with the test suite on the jenkins box which didn't appear in local testing.
I'm debugging what is causing the lockup from this branch as the code looks simple and safe enough that I'm simply not expecting this.

The 'only' unique changes in this branch are in the Internal tests and in the ```Object.py``` ```getNodeIndexCache``` and ```setNodeIndexCache```.

rob-c:
Does anyone have any idea what could be leading to:

```

**Comment by rob-c** [comment]
What sort of loads is this testing machine with seeing because I've tested in the following:
```
clean Python 2.6 env @ Imperial with 500 threads
clean Python 2.7 env @ my laptop with >1000 threads
LHCb Python 2.7 env @ Imperial with 500 threads
LHCb Python 2.7 env @ lxplus with 500 threads
```
and I cannot reproduce the lockups I see with this machine.

I see plenty of evidence that Ganga is having to choke on these systems as it deals with thousands of output files but I don't see any lockups at all.

egede:
Is it a memory issue ? The default VMs at CERN have almost no memory.

rob-c:
I'm seeing usage of ~100-200Mb @ imperial with 500 threads and the tests are normally run with 20 threads and 100 jobs so I wouldn't have thought so.

There is more CPU usage due to this PR by design but I doubt this should generate a too high load.
(And I've no idea how we can check that too high load tasks aren't killed by the OS or something similar)

rob-c:
I've got a suspicion that this probably looks like we're trying to do something equivalent to fork bombing the machine at this point when we submit 5*100 or more jobs during testing from nosetests.

egede:
But remember that these are not forks in the Linux sense. It is all just a single python process.

rob-c:
We do submit 500 or more jobs to the local host in order for the jobs to progress through from created to completed.

I know doing 500 equivalents to 'hello world' isn't too intensive but it does suddenly spawn a lot of python commands for all of the wrapping around this.

rob-c:
Nope, after some messing around this looks like a lockup in GangaObject as per the final line of:
https://ganga-ci.cern.ch/jenkins/job/PR%20builder/341/consoleText

which failed to get the 'j.id' from a job object.

I'm a bit bugged about how this ends up locking up from changing the getIndexCache which shouldn't be being used in this area of code unless there are some problems.
I think I can write the ```has_loaded``` code to maybe avoid spending too much time in the repository and make it a static function, but aside from this it could well be that the code which is used to generate the index cache is causing lockups when accessed incorrectly.

I'll do some digging and make sure that the IndexCache is indeed only being used where it is sensible.

rob-c:
Finally superseded by #325. This was the main intention of this branch but finding the source of the bottlenecks has been fixed elsewhere and all of the fixes in here are already in develop in some form I think from looking at the code.

PR #226, opened 2016-02-25
Removing __setstate__ and __getstate__ which were onced used in copyi…

rob-c:
…ng which are now dead code and doing bad things


rob-c:
I'm keen to get this into 6.1.19 unless there are any objections. This is really just a code cleanup which is overdue imo

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -114,7 +93,7 @@ def __copy__(self, memo=None):
     def __deepcopy__(self, memo=None):
         cls = self.__class__
         obj = cls()
-        this_dict = self.__getstate__()
+        this_dict = self.__dict__
```
milliams: I realise you haven't changed existing behaviour here but should this be `this_dict = copy.copy(self.__dict__)`? Otherwise the following line are altering the `__dict__` of *this* object.

milliams:
Modulo my one comment above, I've very happy with this change. This code has been bit-rotting for years and it's so much nicer to remove it.

python/Ganga/GPIDev/Base/Objects.py
rob-c: I think this may be the source of the bug of breaking the _parent knowledge in the past as it modifies the object it shouldn't. Either way I'll gladly change

rob-c:
This isn't as invasive as removing ```__construct__``` but I'm still very keen to clean up calls to functions that are not maintained or that interfere with basic class operations such as initialization/copying etc...

rob-c:
Any objections to merging this if the tests are all still passing?

PR #227, opened 2016-02-25
Don't read GangaRoot dir as a config file

alexanderrichards:
This PR stops Ganga trying to read the GangaRoot dir as a config file.

Given that ATLAS/LHCb people don't seem to see this error anyway, their self.options.config_path is slightly different at this point. I think it's because they have overridden what this removed line is doing. I think therefore that it is safe to remove this but would appreciate someone with ATLAS/LHCb environment testing this.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -585,7 +585,9 @@ under certain conditions; type license() for details.
         import Ganga.Utility.util
         import inspect
         GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
-        self.options.config_path = Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath, self.options.config_path))
+
+        if self.options.config_path:
+            self.options.config_path = Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath, self.options.config_path))
 
         # check if the specified config options are different from the defaults
         # and set session values appropriately


rob-c:
Not sure a fix is this simple for LHCb I see:

config_path: GangaLHCb/LHCb.ini

which is expanded using this line of code to:

config_path: /afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_Dev/install/ganga/python/GangaLHCb/LHCb.ini

which loads the LHCb.ini based upon the ganga path.

how about:

```
if self.options.config_path != ''
    self.options.config_path = Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath, self.options.config_path))
```

or modify the behavior above to expand the filenames to keep the plugin .ini files loading.

Does this then try and read ```''``` as a config file later on I've not looked.

alexanderrichards:
ok thanks rob, updated to catch both None and ''

alexanderrichards:
This fixes the issue for me (Vanilla Ganga) and should preserve the same behaviour for ATLAS/LHCb. can people please verify this and I'll merge it.

rob-c:
OK, appears to be fine for LHCb :) I think we're waiting until 6.1.16 has been released for PR to go in unless we trust the release tools/process.

@milliams Any opinions?

egede:
@rob-c Let us wait until Monday before we introduce any new PRs.

alexanderrichards:
Indeed, I wasn't planning to merge before 6.1.16 was released. Also would be good for @milliams or @drmarkwslater to try it in the ATLAS environment to check that nothing has changed there.

alexanderrichards:
rebased to incorporate develop changes.

alexanderrichards:
I'd like to merge this, it's a small change that seems to fix the issue for Vanilla and LHCb (ATLAS TBC). What do people think?

rob-c:
Unless this has changed anything since I tested still go for lhcb

alexanderrichards:
For the benefit of record keeping, This is a partial fix for https://github.com/ganga-devs/ganga/pull/227

PR #228, opened 2016-02-25
Clean up Safe Save method

### LINKED ISSUE #155: Check XML data file integrity
We want to make sure that the ```data~``` file contains a backup of the actual ```data``` file and to make sure that the operations which move a new data file into place are as atomic as possible to prevent future data loss.

I'll look into this asap.

drmarkwslater:
Cleans up the safe save method but most importantly makes sure there is absolutely minimal amount of code between creating the backup (data~) and creating the main file (data). Fixes #155


milliams:
This looks good to me. I'm not even sure we still need the `try except` since all it's doing is re-raising. Re-raising can cause issues as it loses the original stacktrace (though there is a solution in Python 3).

rob-c:
<aside>Is there a clear policy for how we want to deal with exceptions in Ganga?</aside>

I'm in favour of a new type of XML being thrown here as we should expect the safe_save to be a very basic operation which shouldn't really fail.

egede:
@rob-c An exception that can arise from the normal use of Ganga should result in an exception that inherits from GangaException (and thus not create a traceback). an exception that could indicate a bug in Ganga should not inherit for GAngaException.

drmarkwslater:
So from this do I conclude that I should remove the exception handling and just propagate everything through? From the arguments here, that would make sense as you really don't want safe_save to fail in normal operations.

rob-c:
That sounds sensible, if we need to add in some exception handling it should be an easy task to add back in but I doubt it's needed at this level.

milliams:
I agree.

drmarkwslater:
Rightio - new commit added :smile:

rob-c:
Looks nice and readable now compared to the numerous try/except statements before :)

drmarkwslater:

PR #231, opened 2016-02-26
3rd stab at index cache fix

rob-c:
This is designed to supersede #225. The main differences between this and the code there is some additional logic to try and prevent lockups which only happened in Jenkins testing.


python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -15,7 +15,7 @@
 logger = getLogger()
 
 _reg_id_str = '_registry_id'
-_id_str = 'id'
+_id_str = '_id'
```
milliams: 1. `_id_str` doesn't seem to be used anywhere
2. I don't like the idea of an object being told what its registry id (`_registry_id`) is. The registry should know but apart from that it should be internal to the registry. Too much coupling in both directions between the registry and the objects within is part of what's causing a lot of the lockup problems. If an object wants to know what its id is then it should query its registry to find out. Duplicating this information is likely just increasing the chance of them getting out of sync.

rob-c: We have always stored the registry_id within the object. If this changes now it potentially represents a __HUGE__ rewrite in how the registry works imo.

If we want to, then great, but I'm just trying to reduce the total amount of code and the ```GangaObject._id``` and ```GangaObject._registry_id``` were both sources of obvious duplication of information.

I'm strongly in favor of keeping the registry id and the ```.id``` from within the Schema separate as it will just avoid headaches. This is also the way the registry has been designed to work for a __long__ time now.

The ```_id_str``` was used in the past and I'm now wanting to remove this as it was defined in the ```GangaRepository.py```. It likely was a historical mistake that the object ```.id``` became confused with the object ```._id``` I feel from looking at the code over time.

I'm working on reducing the information in the registry here as doing some 'sane _little_ changes' within GangaObject are causing lockups (which I still cannot produce elsewhere) in the Jenkins tests.

milliams: Ok, it looks like going back to 6.0.44, `_registry_id` indeed was used but the *only* thing it was used for was a small amount of caching to make `Registry.find` a little faster. I'm happy to keep them for now but we should consider in the future work towards decoupling them.

rob-c:
retest this please

rob-c:
OK,

This is failing right at the last stage of the testing now and I suspect this is due to:

```
test_c_QueueSubmits (GPI.Parallel.TestQueuedSubmit.TestQueuedSubmit) ... [0;31mGanga.Lib.Localhost                : CRITICAL problem during monitoring: JobStatusError:  ('forbidden status transition of job 40 from "completed" to "running"',)[0;0m
  File "/var/lib/jenkins/workspace/PR builder/python/Ganga/Lib/Localhost/Localhost.py", line 309, in updateMonitoringInformation
    j.updateStatus('running')  # bugfix: 12194
  File "/var/lib/jenkins/workspace/PR builder/python/Ganga/GPIDev/Lib/Job/Job.py", line 595, in updateStatus
    raise JobStatusError('forbidden status transition of job %s from "%s" to "%s"' % (fqid, self.status, newstatus))
JobStatusError: JobStatusError:  ('forbidden status transition of job 40 from "completed" to "running"',)
```

I think this may need looking at in more detail as the jobs appear to be in the ```running``` state when passed to the backend functions.

rob-c:
At last, progress (sort-of) I can now reproduce the test lockup through the use of py.test. Now to investigate and hopefully understand/fix.

rob-c:
Superseded by #325

Issue #237, opened 2016-02-29
Unittests now need a grid certificate for non gridified tests


rob-c:
When I try to run the unittests on the latest develop I'm seeing that I need to provide my password to my grid certificate if I haven't make a valid proxy.

This appears as a minor issue as I can generate a proxy and then this goes away but it means that the unitests aren't starting up and getting the LHCb environment correctly defined for them.

I have correctly run ```SetupProject ganga xxx``` and I can start ganga for LHCb correctly so this is a problems with the unittests.

This means I don't trust the test configuration if it's not correctly pulling in the LHCb environment as expected.

I'm going to start investigating with stack traces to see if I can work out how the unittests are getting into this state.

rob-c:
This is to do with ```cmd.append('python/Ganga/Runtime')``` in ```setup.py``` and concerns me that some form of environment is being setup and possibly associated with the tests which isn't what we want.

I'm going to do some further digging to work out what this code is doing.

alexanderrichards:
We should have a presence on Zenodo such that Ganga can be properly referenced

e.g.
https://zenodo.org/record/46721#.VyHT-LPSRhE

PR #238, opened 2016-02-29
Bugfix/fix unittests errors

rob-c:
This is to address #237 as it fixes collection bugs when examining the GPI/Internal tests in addition to some minor changes in the setup.py to avoid LHCb problems and loading of zero tests where there are zero tests.

I'm keen to get this in as it addresses issues I'm seeing from the develop atm.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,8 +58,10 @@ class RunTestsCommand(Command):
 
         if self.type in ['unit', 'all']:
             cmd.append('python/Ganga/new_tests/Unit')
-            cmd.append('python/Ganga/Core')
-            cmd.append('python/Ganga/Runtime')
+            ## Ganga/Core finds zero tests
+            #cmd.append('python/Ganga/Core')
+            ## Ganga/Runtime imports GangaProgram but causes a vanilla instance to be loaded
+            #cmd.append('python/Ganga/Runtime')
             cmd.append('python/Ganga/Utility')
         if self.type in ['integration', 'all']:
             cmd.append('python/Ganga/new_tests/GPI')
--- a/setup.py
+++ b/setup.py
@@ -58,10 +58,8 @@ class RunTestsCommand(Command):
 
         if self.type in ['unit', 'all']:
             cmd.append('python/Ganga/new_tests/Unit')
-            ## Ganga/Core finds zero tests
-            #cmd.append('python/Ganga/Core')
-            ## Ganga/Runtime imports GangaProgram but causes a vanilla instance to be loaded
-            #cmd.append('python/Ganga/Runtime')
+            cmd.append('python/Ganga/Core')
+            cmd.append('python/Ganga/Runtime')
             cmd.append('python/Ganga/Utility')
         if self.type in ['integration', 'all']:
             cmd.append('python/Ganga/new_tests/GPI')
--- a/python/Ganga/new_tests/GPI/Internals/TestRegistry.py
+++ b/python/Ganga/new_tests/GPI/Internals/TestRegistry.py
@@ -1,21 +1,12 @@
 from __future__ import absolute_import
 
 import random
+import threading
 
-import Ganga.Utility.Config
-from Ganga.Core.GangaRepository import RegistryLockError
-from Ganga.Core.GangaRepository.Registry import Registry
-from Ganga.Core.GangaThread.GangaThread import GangaThread
-from Ganga.GPIDev.Base.Proxy import stripProxy
-from Ganga.Runtime.Repository_runtime import getLocalRoot
-from Ganga.Utility.logging import getLogger
 from ..GangaUnitTest import GangaUnitTest
 
-config = Ganga.Utility.Config.getConfig('Configuration')
-logger = getLogger(modulename=True)
 
-
-class HammerThread(GangaThread):
+class HammerThread(threading.Thread):
     def __init__(self, _id, reg):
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -58,8 +58,10 @@ def run(self):
 
         if self.type in ['unit', 'all']:
             cmd.append('python/Ganga/new_tests/Unit')
-            cmd.append('python/Ganga/Core')
-            cmd.append('python/Ganga/Runtime')
+            ## Ganga/Core finds zero tests
+            #cmd.append('python/Ganga/Core')
```
milliams: I don't think it matters if there are zero tests in there at the moment. There likely will be tests at some point so it shouldn't hurt?

setup.py
```diff
@@ -58,8 +58,10 @@ def run(self):
...
+            ## Ganga/Runtime imports GangaProgram but causes a vanilla instance to be loaded
+            #cmd.append('python/Ganga/Runtime')
```
milliams: Where is this vanilla instance being created? The only time this object is instantiated should be in `bin/ganga` and in `GangaUnitTest`.

rob-c: There is  default config being created by the Ganga/Runtime line which keeps asking for a grid proxy to be created if one doesn't exist.
The LHCb.ini should cause this not to happen but loading of Ganga/Runtime causes the the Remote code to be loaded which asks for a grid password.

setup.py
rob-c: Why are we putting tests in Ganga/Core in this case?
I was under the impression that we were loading all tests from a dedicated directory with the tests isolated from the codebase.
Or have I misunderstood something?

rob-c:
I'm also wondering how/why ```python/Ganga/new_tests/GPI/Bugs/TestSavannah77962.py``` is failing now on this branch as in principle I've touched nothing which should be effecting this test.

Investigating now.

setup.py
rob-c: More specifically this is caused in the collection step of the py.test using the latest py.test in an LHCb environment on lxplus.

rob-c:
retest this please

setup.py
milliams: Some tests like doctests (for example `Ganga/Runtime/bootstrap.py:new_version_format_to_old`) will need to be located alongside their source code. In general we will be locating all tests in the `[new_]tests` directory but for some components it might make sense to put them alongside the code they're associated with.

setup.py
milliams: Do you have a stack trace or a line of code where the object is being instantiated?

rob-c: Not yet, this is a difficult thing to track down as the collection step does a lot of hiding what goes on behind the scenes.
I suspect there is an import related problem but I can't confirm anything. I'll try injecting a bug into the code which will generate a complete exit and see if that lends any help.

I'm much more of the opinion to not trust a test that causes failures but I'm happy to revert the changes here if these tests are trusted and required on the Jenkins box.
Although this should be looked into in more detail in the future.

setup.py
rob-c: My concern here is are there expected to be any tests in Ganga/Core in this case as when I run ```py.test python/Ganga/Core``` I get no tests run.

rob-c:
Strange, I observed some intermittent behavior in ```TestSavannah77962``` but other than that this branch appears to fix my problems.

I'll look into reverting the changes of the test paths, but I'm concerned about tests behaving differently in Jenkins to other environments which can come back to bite us.

milliams:
I think the difference is that you are running the tests in a (slightly unusual) LHCb environment which is not tested by Jenkins. I've started work on being able to test within an LHCb environment but SetupProject is a strange tool. I don't think in general in the future that we will need SetupProject for running Ganga, it should be possible to just run it directly from CVMFS.

setup.py
milliams: Indeed there are currently no tests in Core. If the line is causing problems then remove it and we can add it in if and when it is needed.

rob-c:
In future we can use the ```lb-dev``` commands but None of the LHCb quirks above excuse the fact that some tests are causing a badly defined config to be setup.

This is not an environment issue as much as it is a badly defined test which either doesn't need the config to be setup (in which case needs to be fixed and isolated) or needs the config in which case should be based upon ```GangaUnitTest.py``` until we have something better.
Sure it looks ok in the short term. a single test is loaded and passes but in the future there often ends up being problems from the 'looks OK' sort of quirks like this.

rob-c:
The changes are now restricted to fixing the Internals tests for the time being. Any objections to merging once the tests have finished (assuming they pass of course)

milliams:
I'm happy to merge but I think I've tracked down the source of the proxy being needed. In [LCG.py:2337](https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/Lib/LCG/LCG.py#L2337) we create a `Grid` object which in turn makes sure there is a proxy available. I guess it all comes down to whether `GLITE_ENABLE` is `True` of `False`. Is this set by the LHCb environment or config file?

rob-c:
This is set to false when the LHCb environment is fully loaded (or should be).

As it stands I'm happy to forgo fixing the strange behavior of the test collection step in favor of just getting these 2 tests to work 'out of the box'.
It's an odd one as I can't reproduce this importing Ganga.Runtime objects into a vanilla python session so I'm willing to ignore this and move on.

PR #241, opened 2016-02-29
Fix list assign

### LINKED ISSUE #239: ArgSplitter breaks ordering
The following code in release 6.1.16

```
a = ArgSplitter()
a.args = [['Hello '+repr(i)] for i in range(20)]
for i in range(20): print a.args[i]
```
produces the output
```
['Hello 0']
['Hello 1']
['Hello 10']
['Hello 11']
['Hello 14']
['Hello 2']
['Hello 12']
['Hello 13']
['Hello 15']
['Hello 16']
['Hello 17']
['Hello 18']
['Hello 3']
['Hello 4']
['Hello 5']
['Hello 6']
['Hello 9']
['Hello 7']
['Hello 19']
['Hello 8']
```
where the numbers are not ordered. As it is supposed to behave like a list, I regards this as a bug.

rob-c:
Fixes #239 .

I'll add a more detailed description tomorrow.


rob-c:
A while ago as part of the performance problems it was observed there was a severe overhead with large list like objects in Ganga 6.1.
In order to try and combat this I tried to use a ```queues``` based system for initializing many objects and constructing a huge list with them which used to be done many times throughout the assignment process.
Unfortunately testing later showed that this code offered no performance improvement and list construction no longer appears to be the bottleneck.

However, the code remained.
As Ulrik pointed out this creates a bug that the assembled final list is not an ordered list so instead of trying to add complex logic to fix the multi-threaded case (which offered no performance improvement after-all) I've opted to drop to the single threaded case and remove the multi-threaded code which would now be inaccessible or commented out.

This PR also addresses the fact that there was object manipulation going on in the Proxy setter which was incorrect and was at least slowing down the assignment process and at worst potentially corrupting objects which were being externally passed to the setter as the new value.

aka, in some situations the following could potentially fail:
```
myObject = GangaObjectClass()
myObject2 = GangaObjectClass()
myObject.someAttribute = myObject2.someAttribute
assert isProxy(myObject2.someAttribute) != False
```
The following could lead to failure as the input value to the assignment ```myObject2.someAttribute``` was accidentally redefined as the proxy-stripped version of itself within the setter.

milliams:
I assume by `assert isProxy(myObject2.someAttribute) != False` you simply mean `assert isProxy(myObject2.someAttribute)`?

rob-c:
@milliams yes, I was just trying to be clear at a glance.
I'm working on fixing the things broken by this PR, looks like a harmless bug I introduced so there's an amendment incoming.

rob-c:
Any objections to merging this fix?

milliams:
I am in favour of merging this since it removes a lot of code which I didn't like. I'm not surprised that making the list creation multi-threaded didn't help as there was nothing there which was releasing the GIL to allow actual parallelism.

I slightly worry about the increased complexity in Proxy.py but it that's the simplest way to do it then so be it.

rob-c:
This wasn't helping even before the lock was introduced and I never tracked down where the bottleneck was coming from. This was strongly indicated in a profiling tool to be the source of the bottleneck.
Upon further inspection I think this was masking the fact that creating each of the 8k DiracFile objects was the main slow down in the piece of code I was profiling. The problem appeared to be that this was being performed within a list class constructor and the profiling tool hadn't descended quite far enough to work out the list class wasn't to blame for the performance problems.

PR #243, opened 2016-03-01
Move GangaList tests

milliams:
These were missed in the test reorganisation. They should be in `Unit`. No functional changes here, just moving some test files.


rob-c:
go for merge from me

PR #244, opened 2016-03-01
Minor bootstrap tidy

milliams:
These are some minor changes which should have no actual effect on any code. It is just removing code which is not needed in a long time.


rob-c:
Looks clean enough, if the tests pass feel free to merge from me

PR #248, opened 2016-03-01
GangaUnitTest teardown simplification

milliams:
This change enforces that all `GangaUnitTest` tests run in a separate gangadir and that the gangadir is always wiped after the whole test class has finished. Until we have a situation where we want to preserve the gangadir after the test run or run in a specific gangadir, I'd like to keep this code as simple and short as possible.

014e0ea07d9eb356027c9796f551eaf621dd626e is the important change, the other two commits are just minor cleanup.

### DIFF
--- a/python/Ganga/new_tests/GPI/GangaUnitTest.py
+++ b/python/Ganga/new_tests/GPI/GangaUnitTest.py
@@ -1,4 +1,8 @@
 from __future__ import print_function
+
+import sys
+import shutil
+import os.path
 try:
     import unittest2 as unittest
 except ImportError:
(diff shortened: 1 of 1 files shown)


python/Ganga/new_tests/GPI/GangaUnitTest.py
```diff
@@ -205,38 +203,30 @@ def stop_ganga():
...
         stop_ganga()
-        import sys
         sys.stdout.flush()
 
     @classmethod
     def tearDownClass(cls):
-        if cls.wipe_repo:
-            import shutil
-            shutil.rmtree(cls.gangadir, ignore_errors=True)
+        shutil.rmtree(cls.gangadir(), ignore_errors=True)
```
rob-c: Is cls.gangadir() still assigned in order for this to work?

rob-c:
Go for merge from me

alexanderrichards:
:+1:

PR #250, opened 2016-03-01
Moving startup/cleanup of Registries to a sane place

rob-c:
This PR cleans up the GangaUnitTest as well as making it clearer where the repositories are loaded from.

This also moves the removal of the registries to be managed by the registry shutdown which is more sensible.

### DIFF
--- a/python/Ganga/Runtime/Repository_runtime.py
+++ b/python/Ganga/Runtime/Repository_runtime.py
@@ -15,16 +15,19 @@ logger = getLogger()
 
 
 def requiresAfsToken():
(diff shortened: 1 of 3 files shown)


milliams:
I'm pretty happy with this. Docstrings on `startUpRegistries` and `removeRegistries` would be good to see so that it's documented exactly what behaviour these functions should have. Apart from that :+1:

rob-c:
I'll fix the smaller error I just committed in a bit, was due to refactoring some code in the UnitTest

rob-c:
retest this please

PR #251, opened 2016-03-01
Making warning of dead monitoring more verbose

rob-c:
Adding some more debug code to help with debugging any 'frozen' monitoring threads.

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -794,7 +794,7 @@ class JobRegistry_Monitor(GangaThread):
         _purge_actions_queue()
         stop_and_free_thread_pool(fail_cb, max_retries)
 
-        global_count = 0
+        cls.global_count = 0
         return True
 
     def stop(self, fail_cb=None, max_retries=5):
@@ -844,7 +844,7 @@ class JobRegistry_Monitor(GangaThread):
         #while self._runningNow is True:
         #    time.sleep(0.5)
 
-        global_count = 0
+        cls.global_count = 0
         return True
 
     def __cleanUp(self):
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -75,6 +75,7 @@ def checkHeartBeat():
 
             log.warning("Thread: %s Has not updated the heartbeat in %ss!! It's possibly dead" %(thread_name, str(dead_time)))
             log.warning("Thread is attempting to execute: %s" % this_thread._running_cmd)
+            log.warning("With arguments: (%s)" % str(this_thread._running_args))
 
             ## Add at least 5sec here to avoid spamming the user non-stop that a monitoring thread has locked up almost entirely
             ## You'll get a message at most once per 5sec or less if the Monitoring is busy/asleep
@@ -86,6 +87,7 @@ class MonitoringWorkerThread(GangaThread):
         GangaThread.__init__(self, name)
         self._currently_running_command = False
         self._running_cmd = None
+        self._running_args = None
         self._thread_name = name
 
     def run(self):
@@ -124,22 +126,31 @@ class MonitoringWorkerThread(GangaThread):
             try:
                 try:
                     self._running_cmd = action.function.__name__
+                    self._running_args = ""
+                    for arg in action.args:
+                        self._running_args.append("%s, " % arg)
+                    for k, v in action.kwargs:
+                        self._running_args.append("%s=%s, " % (str(k), str(v)))
                 except:
                     self._running_cmd = "unknown"
+                    self._running_args = ""
                 result = action.function(*action.args, **action.kwargs)
             except Exception as err:
                 log.debug("_execUpdateAction: %s" % str(err))
                 action.callback_Failure()
+            #finally:
+            #    pass
             else:
                 if result in action.success:
                     action.callback_Success()
                 else:
                     action.callback_Failure()
 
+            self._running_args = None
+            self._running_cmd = None
             self._currently_running_command = False
 
 # Create the thread pool
-
 
 def _makeThreadPool(threadPoolSize=THREAD_POOL_SIZE, daemonic=True):
     global ThreadPool, global_start_time, heartbeat_times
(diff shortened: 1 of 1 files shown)


egede:
While making the warnings about dead monitoring loops more verbose is good 
to debug the problem, it is the wrong direction for users who already in 
6.1.16 are drowned by spurious warnings. I think the solution is either to 
move this to debug level or to make sure that the warning is only ever 
issued once.

rob-c:
@egede Yes, you're ultimately correct.
But I still can't reproduce them in the same regularity that others can and I've no idea what is causing this. We also decided that until we're satisfied that this has gone away we were to keep 'warning' users to allow us to know if people are being effected by this at all.
The code is relatively simple so I doubt it's the heartbeat which is to blame I suspect this may be some 'spurious' behavior of the monitoring which it's failing to take into account.

This is a bit odd as there is an explicit test before displaying the warning that the monitoring thread be active and currently executing some command. Hence why I believe there is still something with either extremely poor performance or still locking up for some users.

rob-c:
If there is no objections I would like to push this to develop so that I can try and get it into some branches used by people having these problems to _try_ and see exactly what is causing the heartbeat errors to be thrown.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -124,23 +126,32 @@ def _execUpdateAction(self):
...
+                        self._running_args.append("%s=%s, " % (str(k), str(v)))
                 except:
                     self._running_cmd = "unknown"
+                    self._running_args = ""
                 result = action.function(*action.args, **action.kwargs)
             except Exception as err:
                 log.debug("_execUpdateAction: %s" % str(err))
                 action.callback_Failure()
+            #finally:
+            #    pass
```
alexanderrichards: do we need to add this commented out code?

rob-c: gone

egede:
@rob-c Could you make it display the warning only X times in a given Ganga session. I am worried that we will make an otherwise well functioning release very annoying for users.

rob-c:
OK, this should now only display 5 warnings per session. I'd still like to get to the bottom of this, so if there are no objections I'll merge this and propagate it through to LHCb Gang HEAD

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -70,15 +71,21 @@ def checkHeartBeat():
         last_time = heartbeat_times[thread_name]
 
         dead_time = config['HeartBeatTimeOut']
+        max_warnings = 5
 
-        if (latest_timeNow - last_time) > dead_time and this_thread.isAlive() and this_thread._currently_running_command is True:
+        global global_count
```
milliams: I don't think this is needed, you're not rebinding the name in this scope.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -33,6 +33,7 @@
 
 heartbeat_times = None
 global_start_time = None
+global_count = 0
```
milliams: Could we reset the `global_count` on shutdown so that it resets between tests etc.?

milliams:
We should think about in future the ability to make (LHCb) releases directly from a branch to test in situations like this without being forced to push it to develop.

rob-c:
agreed but I'm only advertising develop as it's supposed to be stable between 6.1.16 and 6.1.17.
Besides I'm not wanting to keep a parallel 'almost develop' branch as something will probably go wrong whenever I've done something like that and it's overhead time I'd rather put into bug fixing/development.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
rob-c: Frankly If this is showing up more than 5 times in tests (which it isn't at the moment) we will get swamped with errors making life difficult.
I'll amend but I'm not convinced it's needed.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -70,15 +71,20 @@ def checkHeartBeat():
...
+                and global_count < max_warnings:
 
             log.warning("Thread: %s Has not updated the heartbeat in %ss!! It's possibly dead" %(thread_name, str(dead_time)))
             log.warning("Thread is attempting to execute: %s" % this_thread._running_cmd)
+            log.warning("With arguments: (%s)" % str(this_thread._running_args))
 
             ## Add at least 5sec here to avoid spamming the user non-stop that a monitoring thread has locked up almost entirely
             ## You'll get a message at most once per 5sec or less if the Monitoring is busy/asleep
             heartbeat_times[thread_name] += 5.
+            global_count+=1
```
alexanderrichards: This line here is a rebinding so will probably need the
`global global_count`

right?

rob-c: This was my understanding as 1 is not 2 but I frankly give up on the use of globals, they give me a headache.

alexanderrichards: well I can't say I'm sorry :wink: I was getting a little developers shiver down my spine when I heard globals mentioned I think that they are rarely a good solution.

rob-c: True

milliams: Ahh, I'd forgotten that `+=` was a rebinding. In I agree about avoiding `global` wherever possible and I don't like adding even more. Could the count be a member of one of the classes?

rob-c: Not really, I'd have to mess around with propagating it through various layers and I'm adverse to this. I just wanted a global count to 5 to make sure nothing goes wrong.

alexanderrichards: it needn't be an instance variable. A static class variable is pretty close to a global but makes me shiver less :wink:

rob-c: And then we're passing around references to the monitoring class as a whole which I'd rather avoid even for debug statements.
I'll go work on the changes then since globals are causing such a controversy

rob-c:
Any objections to go for merge?

milliams:
I think I'm happy.

PR #259, opened 2016-03-03
Always flush a job every time its status changes

milliams:
In Ganga<6.1.16 the registry would, upon certain actions performed on it, do a flush to disk of all dirty jobs but only if it had been more than 30 seconds since the last flush. This wasn't a complete solution in that while there was a rate-limiter, there was no consistent prompt to flush. In 6.1.16 this logic was simplified with a look to adding a flushing thread in the near future which would attempt flush every 30 seconds ([see example](https://gist.github.com/milliams/2d95706c8f511f63d294)). This was deemed good enough since everything would be flushed on exit anyway.

This PR adds in an explicit flush each time a job's status changes. While this will slightly increase the total number of flushes it will ensure that important state changes like this are not lost.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -635,7 +635,7 @@ class Job(GangaObject):
         if self.status != saved_status and self.master is None:
             logger.info('job %s status changed to "%s"', self.getFQID('.'), self.status)
             self._setDirty()
-            # TODO try to force a flush here maybe?
+            self._getRegistry()._flush([self])
         if update_master and self.master is not None:
             self.master.updateMasterJobStatus()
 


rob-c:
Is this safe from a subjob perspective, does it walk up the chain to call for a flush?

milliams:
I believe that `updatestatus` does that automatically. It has a call to `self.master.updateMasterJobStatus()` at the end of it which should propagate upwards.

milliams:
This is causing some problem though due to calling the registry from within a `GangaObject` so I'm quickly putting together another PR which implements the auto-flushing thread.

rob-c:
Not unless were certain that the master is correctly transitioning. I'll check the logic again.

drmarkwslater:
I'm guessing this PR can probably be closed now because, after discussions and the application of the Auto Flusher (PR #267 ) it would be better handled using a task/request pool there rather than adding a specific flush in a different place.

rob-c:
I'd be a LOT happier if we also flushed on major state transitions of the job.
I'm concerned a little that for a large job with many subjobs (order 1,000 or so) this would introduce several thousand flush commands without care so maybe this would be best ignoring the 'submitting/submitted' state changes both ways, but I like the security offered by the fact that once a job has 'completed' this will trigger a flush and we know we're safe

drmarkwslater:
I agree - we certainly still want this but @milliams was thinking of adding a request to the Auto Flusher on a status transition to flush a particular object that the Auto Flusher can then handle ASAP. This means the flush will still happen in the Auto Flusher thread (hopefully reducing the chance of deadlocks) but we still flush very soon after that state transition.

You're also right about the number of flushes though - it may be worth avoiding subjobs or submitting -> submitted. Also, it might be worth having these flushes work on blocking locks? Not sure...

milliams:
I think we agreed that this was not the way we wanted to solve this problem. It would cause far too many flushes and the auto flusher has been working well for a while now.

alexanderrichards:
@milliams can the branch also be safely deleted or are we holding on unless someone thinks this issue should be re-opened?

milliams:
No, I can clear the branch too.

PR #261, opened 2016-03-03
Set the dirty flag to False after a flush

milliams:
As noticed by @rob-c, the dirty status of an object was not being reset after a successful flush. This does this so we should see far few flushes needed.

This may also help with the extraneous flushes on shutdown.


drmarkwslater:
:+1:

rob-c:
Go for merge so I can rebase

rob-c:
Hang on doesn't this go through the setter and then set itself dirty after modifying something about itself?

milliams:
What is setting it dirty?

However, you're right that it should use `_setFlushed()` instead although all that does is `self._dirty = False`.

PR #262, opened 2016-03-03
Some fixes which stop error messages from being thrown on shutdown du…

rob-c:
…e to daemon threads running whilst being destroyed by Python

I'd like to quickly push this into 6.1.17.

As part of fixing the flushing to disk at shutdown I discovered several other nasty errors which needed to be addressed to do with shutting down of threads.

If there is any long discussion we can hold off but this allows daemon-ed threads to shutdown more safely and gets rid of weird intermittent shutdown errors.

Any objections to merging as long as the tests are unaffected?

### DIFF
--- a/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
@@ -227,7 +227,7 @@ class WorkerThreadPool(object):
         for w in self.__worker_threads:
             w.stop()
             w.join()
-            w.setDaemon(False)
+            w.daemon = False
             # FIXME NEED TO CALL AN OPTIONAL CLEANUP FUCNTION HERE IF THREAD IS STOPPED
             # w.unregister()
             #del w
--- a/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
@@ -227,7 +227,6 @@ class WorkerThreadPool(object):
         for w in self.__worker_threads:
             w.stop()
             w.join()
-            w.daemon = False
             # FIXME NEED TO CALL AN OPTIONAL CLEANUP FUCNTION HERE IF THREAD IS STOPPED
             # w.unregister()
             #del w
--- a/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
@@ -230,7 +230,6 @@ class WorkerThreadPool(object):
             # FIXME NEED TO CALL AN OPTIONAL CLEANUP FUCNTION HERE IF THREAD IS STOPPED
             # w.unregister()
             #del w
-        del self.__worker_threads[:]
         self.__worker_threads = []
         return
 
--- a/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
@@ -225,9 +225,9 @@ class WorkerThreadPool(object):
     def _stop_worker_threads(self, shutdown=False):
         self._shutdown = shutdown
         for w in self.__worker_threads:
-            w.setDaemon(False)
             w.stop()
             w.join()
+            w.setDaemon(False)
             # FIXME NEED TO CALL AN OPTIONAL CLEANUP FUCNTION HERE IF THREAD IS STOPPED
             # w.unregister()
             #del w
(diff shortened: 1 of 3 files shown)


milliams:
In general I agree with with rethinking the use of daemonic threads in Ganga however, we must be careful to do it right. If a thread is not daemonic then it cannot be killed without hard-killing the process and this would break the current "there are N threads running, do you really want to quit" thing we have. Before we do this we need to make sure (for the user's sake) that any background threads are not performing long-running tasks which would cause long hangs on exit. I don't think that blindly switching to non-daemonic threads is a good idea just yet.

rob-c:
Fix incoming for the tests as it stands.

I forgot to change the order can't be set to non-daemonic unless the thread has stopped whatever it's doing anyway so I think this should address your concerns @milliams?

milliams:
I don't believe that you can change the daemonic status of a thread after `start()` has been called (https://docs.python.org/2.6/library/threading.html#threading.Thread.daemon).

rob-c:
ok, apparently can't unset the daemonic status of finished already-run thread but we should be joining the user threads (at least) on shutdown to correctly shutdown these threads either way.

python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
```diff
@@ -24,11 +24,14 @@ def shutDownQueues():
...
-    _global_queues._purge_all()
+    try:
+        _global_queues.lock()
+        _global_queues._purge_all()
+    except:
+        logger.warning("Error in shutting down queues thread. Likely harmless")
     del _global_queues
     _global_queues = None
     import Ganga.GPI
+    setattr(Ganga.GPI, 'queues', None)
```
milliams: Is this line needed if it's `delattr` on the next line?

rob-c:
PR is now much smaller and offers more reliability that the queues system has been shutdown correctly on the way out. Are there any objections to merging if the tests run OK?

python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
rob-c: I was trying to avoid possible double deletion of the _global_queues stored in the queues attribute

milliams: You can't double delete things in Python. You can't even single-delete them. All `delattr` does is remove the reference to it.

rob-c: In that case I'll remove this line and job done

python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
```diff
@@ -226,10 +226,11 @@ def _stop_worker_threads(self, shutdown=False):
         self._shutdown = shutdown
         for w in self.__worker_threads:
             w.stop()
+            w.join()
             # FIXME NEED TO CALL AN OPTIONAL CLEANUP FUCNTION HERE IF THREAD IS STOPPED
             # w.unregister()
             #del w
-        #del self.__worker_threads[:]
+        del self.__worker_threads[:]
```
milliams: There should be no need to `del` all the entries here as you're rebinding the list on the next line so any remaining reference to them should be gone.

alexanderrichards:
> In general I agree with with rethinking the use of daemonic threads in Ganga

What are the reasons for this? Surely you want to be able to terminate ganga without being stuck waiting for a blocking thread?

milliams:
@alexanderrichards Because killing a thread in the middle of what it's doing leaves everything in a very unknown state. File locks can be held, database transactions could be open and they will not be cleaned up.

rob-c:
It's to remove erroneous error messages I get from the queues system as Python starts to shutdown the monitoring threads before they're properly stopped.

alexanderrichards:
switching from daemonic to non-daemonic will not stop someone from screwing things with a kill. They will only stop people from exiting ganga and stopping the treads forcefully

rob-c:
This cannot be done and in the final version of this PR is NOT doing this.

The changes are now to protect the shutdown and to join the queues tasks which in some circumstances is not being done otherwise.

alexanderrichards:
since all you will do by this switch to non-daemonic is ensure that ganga dosn't shut down until their are complete you will probably get annoyed users saying why cant I forcibly shut down ganga and then just kill it so you haven't really saved anything.

milliams:
@alexanderrichards As @rob-c says, this PR no longer makes the daemonic change. My argument was that we should fix the stuck threads in the long term as the correct solution.

alexanderrichards:
agreed

rob-c:
These threads ARE NOT stuck threads, the threads I most commonly see fail are failing in random places with calls to basic python modules such as Queue which is due to the thread still running but python itself shutting down early.

To be clear.
This is not a change to address stuck threads and stuck threads are the same before and after this PR.

rob-c:
The exact error I'm dealing with here is that I see line: https://github.com/ganga-devs/ganga/pull/262/files#diff-b2907218abe232c832e5fb09ba8c40d1R78 throws a 'None object is not callable error' which makes no sense but is due to python shutting down with the thread still running.

rob-c:
Are there any objections to me merging this?

milliams:
It's a shame that it took 6 commits to make the 6 line change (which messes with my OCD for the commit history being messy) but the change overall looks good. I'd been seeing these messages being printed on shutdown too so thank you for solving them.

alexanderrichards:
FYI you can rebase and change some of the commits to fixups so that they get combined into one commit
`rebase -i`

PR #263, opened 2016-03-03
Use the GangaObject setter to mark as flushed

milliams:
As @rob-c pointed out, there is a setter function to mark an object as non-dirty.


rob-c:
Go for merge

Issue #257, opened 2016-03-02
Exiting ganga causes jobs to be loaded and flushed


rob-c:
Jobs are being loaded and flushed as part of the exit in 6.1.16 (and develop) this is caused by some logic being broken to do with the job indexes (I suspect in updating the master index).

The logic should be that the master index is updated using knowledge of only fully loaded jobs to update the information stored in various places.

I'll take a look at this and try to get a small fix in for 6.1.17 as I don't expect the logic to be too difficult to fix.

egede:
@rob-c Yes, this sounds like a serious issue. For somebody with 20 k jobs in the repository this could be a killer. It points to something else that we are not have tests for.

rob-c:
Actually writing a test for this shouldn't be too difficult. It's more difficult to understand/test the logic of the XMLRepository, there's a lot of code there not well understood and has some quirks.

Issue #256, opened 2016-03-02
Suppress repetitive warning messages


drmarkwslater:
We have had a couple of times recently (monitoring heartbeat and deprecation of SandboxFile, etc.) where Ganga can spew out many copies of the same warning/info message. It would be nice for repetitive messages can be removed, possibly through the basic python logger rather than GangaLogger as we want to get rid of this?

drmarkwslater:
A quick look at the python logger seems to indicate just creating our own filter that stops printing messages after a certain number or whatever should work well. We'd have to be careful about the implementation though to ensure we don't remove required messages.

rob-c:
How many features in the current logging system do we want to keep/drop? I'm very keen to replace the getLogger with 20-30 lines of code at max and not have any other code (if possible)

rmatev:
I think the suppression has to be requested explicitly. I am not sure if / how this can be done, but from a quick look
`logger.warning("something is deprecated", extra={'maxprint':10})`
might be feasible.

alexanderrichards:
@drmarkwslater I agree, using a standard python log Filter looks like a possible solution. This in combination with what @rmatev proposes would even allow different max numbr of prints for different log messages. As I understand it the `extra` keyword arg will populate the `__dict__` of the actual logrecord object which I guess we could then filter on.

drmarkwslater:
The question would then just be do we filter and limit **all** messages or just ones that send the appropriate `maxprint` parameter? Anyone got any thoughts either way?

alexanderrichards:
No strong feelings either way. Using the `maxprint` give your more flexibility but you will have to remember to include it for all log calls where you want it so either way is fine by me.

egede:
@drmarkwslater @alexanderrichards I do not like the solution with `maxprint`. I think it is better if we allow this to be controlled in the same way as we do for the logging levels. Something along the lines of

```
[Logging]
Ganga = INFO
Ganga.GPIDev = WARNING

[LogSuppression]
Ganga = 10
Ganga.GPIDev = 100
GangaLHCb = 1000000
```

rmatev:
I am not a big fan of suppressing messages globally, rather the problem giving rise to them should be fixed. For instance, you always want to get the INFO messages from job submission, etc.

However, for deprecation warnings, it is fine and desired not to repeat it to the user hundreds of times. It sounds that what is really needed there is `logger.deprecated("something is deprecated")`

[...]

PR #264, opened 2016-03-03
Bugfix/stop shutdown insanity

rob-c:
This addresses #257 and at some level #256.

This PR restores the old default behaviour that calling flush with no arguments flushes all dirty objects.

We now explicitly ignore non loaded and non dirty jobs when flushing.
If you want to flush an object mark it as dirty.
If you don't mark it as dirty and want it written you should.

Unless we explicitly tie dirty into ```_getWriteAccess``` I don't expect this policy on what is dirty and what should be flushed to change.

Repository unlock is not yet used consistently and I've some concerns over the safety of releasing locks on every flush against erroneous ganga use.
I'm not objecting to Ganga releasing an object but releasing an object we have in memory (and therefore will __NOT__ re-load from disk) is asking for very serious problems.

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -682,6 +682,8 @@ class Registry(object):
             #NO NO NO NO NO NO NO FOR THE SAKE OF MY SANITY NO, NEVER, EVER, EVER ITERATE OVER ALL OBJECTS BLINDLY FOR A FLUSH!!!
             #self._flush(self.values())
             self._flush()
+        if self.metadata and self.metadata.hasStarted():
+            self.metadata._flush()
 
     def _read_access(self, _obj, sub_obj=None):
         """Obtain read access on a given object.
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -216,6 +216,10 @@ class Registry(object):
     def hasStarted(self):
         return self._hasStarted
 
+    def _start_check(self, message):
+        if self.hasStarted() is not True:
+            raise RegistryAccessError(message)
+
     def lock_transaction(self, this_id, action):
         while this_id in self._inprogressDict.keys():
             logger.debug("Getting item being operated on: %s" % this_id)
@@ -529,8 +533,7 @@ class Registry(object):
         logger.debug("_add")
         obj = stripProxy(_obj)
 
-        if self.hasStarted() is not True:
-            raise RepositoryError("Cannot add objects to a disconnected repository!")
+        self._start_check("Cannot add objects to a disconnected repository!")
 
         this_id = None
         returnable_id = None
@@ -577,6 +580,7 @@ class Registry(object):
 
     def __reg_remove(self, obj, auto_removed=0):
 
+        self._start_check("Cannot remove objects from a disconnected repository!")
         logger.debug("_reg_remove")
         u_id = self.find(obj)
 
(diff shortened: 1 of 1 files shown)


drmarkwslater:
I'm afraid this still triggers a load on a completed job:

[13:04:03]
Ganga In [1]: 
Do you really want to exit ([y]/n)? y

Ganga.Core.InternalServices        : INFO     Stopping Job processing before shutting down Repositories
Ganga.Core.InternalServices        : INFO     Shutting Down Ganga Repositories
Ganga.Runtime                      : INFO     Registry Shutdown
load: [0]
  File "/home/mws/Ganga/install/ganga/python/Ganga/Core/InternalServices/ShutdownManager.py", line 153, in _ganga_run_exitfuncs
    Repository_runtime.shutdown()
  File "/home/mws/Ganga/install/ganga/python/Ganga/Runtime/Repository_runtime.py", line 171, in shutdown
    registry.shutdown()
  File "/home/mws/Ganga/install/ganga/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py", line 55, in shutdown
    stripProxy(self.shareref).closedown()  ## Commenting out a potentially EXTREMELY heavy operation from shutdown after ganga dev meeting - rcurrie
  File "/home/mws/Ganga/install/ganga/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py", line 414, in closedown
    all_dirs = copy.deepcopy(self.__getName().keys())
  File "/home/mws/Ganga/install/ganga/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py", line 134, in __getName
    if not hasattr(self, 'name') or self.name is None:
  File "/home/mws/Ganga/install/ganga/python/Ganga/GPIDev/Base/Objects.py", line 389, in decorated
    return get_or_set(self, obj, type_or_value)
  File "/home/mws/Ganga/install/ganga/python/Ganga/GPIDev/Base/Objects.py", line 443, in __get__
    obj._getReadAccess()
  File "/home/mws/Ganga/install/ganga/python/Ganga/GPIDev/Base/Objects.py", line 936, in _getReadAccess
    reg._read_access(root, self)
  File "/home/mws/Ganga/install/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 697, in _read_access
    self.__safe_read_access(_obj, sub_obj)
  File "/home/mws/Ganga/install/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 742, in __safe_read_access
    self._load([this_id])
  File "/home/mws/Ganga/install/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 166, in decorated
  File "/home/mws/Ganga/install/ganga/python/Ganga/Core/GangaRepository/Registry.py", line 719, in _load
    self.repository.load([obj_id])
  File "/home/mws/Ganga/install/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 879, in load
    traceback.print_stack()

This is from uncommenting the code in GangaRepositoryXML.py at the top of the load function, submitting a basic job, letting it complete, exiting, reloading and finally exiting again.

rob-c:
@drmarkwslater  This is the prep registry and is by design! (not a design I agree with but by design as far as I can tell)
This doesn't trigger the loading of a job only the loading of the prep repo which would then (used to but is disabled NOW) would go through and run a sanity/consistency check on shutdown

rob-c:
OK, There's an update due anyway as I need to rebase but how this is causing an IOError in the subjobs is beyond me.
I'll have a look but this makes me think even more that we need a decent test for the XML being generated and existing as expected.
(I'll start on that today hopefully)

rob-c:
OK, the subjob test actually appears to be a genuine bug as it's cropping up again. This will now get some attention.

drmarkwslater:
Ah - OK that makes sense then. However, I was actually having a look at this myself and I think we only need to make a minor change to get this working. See PR #266. You do this in your code as well but I personally would favour a lighter touch whenever possible.

rob-c:
retest this please

PR #253, opened 2016-03-01
Fix for job.subjobs.select(something...)

rob-c:
Fixes #246. The fix is done entirely within the RegistrySlice layer.

Just a side note the RegistrySlice is always created with filters for subjobs so needs care if we ever get rid of that system...

This represents a relatively isolated change so I'm looking to merge tomorrow if possible.

PR #265, opened 2016-03-03
Adding jobs.select test

rob-c:
A new test to check for #253. Accidentally based this on #179 so will rebase after that has finished.

This replaces #255 where I forgot to add the test anyway... ooops

### DIFF
--- a/python/Ganga/new_tests/GPI/Parser/TestSelect.py
+++ b/python/Ganga/new_tests/GPI/Parser/TestSelect.py
@@ -1,0 +1,48 @@
+from __future__ import absolute_import
+
+from ..GangaUnitTest import GangaUnitTest
+
+class TestSelect(GangaUnitTest):
+
+    def test_SelectJob(self):
+        
+        from Ganga.GPI import Job, jobs
+        j=Job()
+
+        job_slice =  jobs.select(status='new')
+
+        assert(len(job_slice) == 1)
+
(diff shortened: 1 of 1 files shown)


python/Ganga/new_tests/GPI/Parser/TestSelect.py
```diff
@@ -0,0 +1,48 @@
...
+class TestSelect(GangaUnitTest):
+
+    def test_SelectJob(self):
+        
+        from Ganga.GPI import Job, jobs
+        j=Job()
+
+        job_slice =  jobs.select(status='new')
+
+        assert(len(job_slice) == 1)
```
milliams: Don't put brackets around the arguments for `assert`. It is a statement, not a function and does not take arguments in the traditional way. This can cause serious problems if you add a second argument as a message like:
```python
assert(len(job_slice) == 1, 'should find 1 job')
```
as it will treat that as a single argument of a 2-tuple which will always evaluate to `True`

rob-c: fine, change incoming

milliams:
Once the CI tests pass it's a :+1: from me. I wonder if it's possible to test this in the unit tests system as well on a more basic level?

rob-c:
Not if it makes use of the subjobs. This is specifically to test a high level bug

PR #266, opened 2016-03-03
Bugfix/stop loading on exit

### LINKED ISSUE #257: Exiting ganga causes jobs to be loaded and flushed

drmarkwslater:
This should fix #257 and is alternative to PR #264. Basically, it just reorders the `obj._getWriteAccess` function to after the check on the `_dirty` flag as within the write access call (shouldn't this be read access?) the job is loaded which makes sense if you're requesting write access.

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -645,7 +645,6 @@ class Registry(object):
                 obj_id = getattr(obj, _reg_id_str)
                 self.repository.flush([obj_id])
                 obj._setFlushed()
-                self.repository.unlock([obj_id])
 
     @synchronised
     def flush_all(self):
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -636,12 +636,11 @@ class Registry(object):
         for obj in objs:
             with obj.lock:
                 # check if the object is dirty, if not do nothing
-                print obj._dirty
                 if not obj._dirty:
                     continue
 
                 # flush the object. Need to call _getWriteAccess for consistency reasons
-                # TODO: getWriteAccess should only 'get write access' and as that's not needed shouldn't be called here?
+                # TODO: getWriteAccess should only 'get write access', as that's not needed should it be called here?
                 obj._getWriteAccess()
                 obj_id = getattr(obj, _reg_id_str)
                 self.repository.flush([obj_id])
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -635,10 +635,15 @@ class Registry(object):
 
         for obj in objs:
             with obj.lock:
-                obj._getWriteAccess()
-                obj_id = getattr(obj, _reg_id_str)
+                # check if the object is dirty, if not do nothing
+                print obj._dirty
                 if not obj._dirty:
                     continue
+
+                # flush the object. Need to call _getWriteAccess for consistency reasons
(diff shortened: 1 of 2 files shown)


milliams:
This PR is a lot lighter and simpler than #264 and it seems to fix the problem in testing. If this passes the CI, I'd be in favour of merging it.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -635,10 +635,14 @@ def _flush(self, objs):
...
                 if not obj._dirty:
                     continue
+
+                # flush the object. Need to call _getWriteAccess for consistency reasons
+                # TODO: getWriteAccess should only 'get write access', as that's not needed should it be called here?
+                obj._getWriteAccess()
+                obj_id = getattr(obj, _reg_id_str)
                 self.repository.flush([obj_id])
                 obj._dirty = False
                 self.repository.unlock([obj_id])
```
rob-c: I am complaining about this being broken. It does bad things and leads to a bad situation within Ganga.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -635,10 +635,14 @@ def _flush(self, objs):
...
+                # check if the object is dirty, if not do nothing
                 if not obj._dirty:
                     continue
+
+                # flush the object. Need to call _getWriteAccess for consistency reasons
+                # TODO: getWriteAccess should only 'get write access', as that's not needed should it be called here?
+                obj._getWriteAccess()
+                obj_id = getattr(obj, _reg_id_str)
                 self.repository.flush([obj_id])
                 obj._dirty = False
```
rob-c: Please update this to match develop

rob-c:
Also I'd like to see the metadata flushed at the same time that the data is in flush_all.

At this stage save for some code reformatting and a few checks we arrive back at #264 but I'm happy to merge this as long as everything is correctly addressed.

python/Ganga/Core/GangaRepository/Registry.py
milliams: It's only in there because it used to be. I'm happy to remove the call to `unlock` if it's not needed. What is it *supposed* to do?

rob-c: This leaves an object loaded in memory in ganga and releases the file locks.
If someone/something/some-event causes the file on disk to change we lose all knowledge of it as we make no attempt to re-load the object.

Worst case scenario this allows for another instance of ganga to grab the job, manipulate it and then this instance to ignore the changes and obliterate them upon a future write action.

If we remove this lock we should fully un-load the job and we currently have no mechanism for this.

This just allows for 2 instances of ganga (launched on purpose or by accident by a user) to get into a very bad state which can be avoided by only releasing objects on exit at the moment.

drmarkwslater:
Anyone have problems with merging this? If so, speak now :smile:

milliams:
:+1:

rob-c:
Go for merge

drmarkwslater:
Slight update to prevent deadlocks in @milliams new auto flusher code (see PR #267 ). Shout if there are any problems!

drmarkwslater:
Another minor update to give some more info on the test that fails very occasionally.

drmarkwslater:
If there are no issues, I'll merge this.

milliams:
:+1:

PR #267, opened 2016-03-03
Add auto-flusher to registries

milliams:
This adds an auto-flushing background thread to all registries which flushes all dirty objects with a 30 second gap in between.

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -209,8 +209,15 @@ class RegistryFlusher(threading.Thread):
                 time.sleep(1/sleeps_per_second)
                 if self.stopped:
                     return
-            logger.debug('auto-flushing', self.registry.name)
-            self.registry.flush_all()
+            # We want this to be a non-blocking lock to avoid this
+            # interfering with interactive work or monitoring. It
+            # will try again in a while anyway.
+            if self.registry._lock.acquire(blocking=False):
+                try:
+                    logger.debug('Auto-flushing', self.registry.name)
+                    self.registry.flush_all()
+                finally:
+                    self.registry._lock.release()
 
 
 class Registry(object):
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -183,18 +183,24 @@ class RegistryFlusher(threading.Thread):
     def stop(self):
         self._stop.set()
 
+    @property
     def stopped(self):
         return self._stop.isSet()
 
     def run(self):
+        """
+        This will run an indefinite loop which periodically checks
+        whether it should stop. In between calls to ``flush_all`` it
+        will wait for a fixed period of time.
+        """
         sleep_period = 30
         sleeps_per_second = 10  # This changes the granularity of the sleep.
-        while not self.stopped():
+        while not self.stopped:
             for i in range(sleep_period*sleeps_per_second):
                 time.sleep(1/sleeps_per_second)
-                if self.stopped():
+                if self.stopped:
                     return
-            print 'auto-flushing', self.registry.name
+            logger.debug('auto-flushing', self.registry.name)
             self.registry.flush_all()
 
 
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -932,8 +932,6 @@ class Registry(object):
             self.repository.startup()
             # All Ids could have changed
             self.changed_ids = {}
-            self.flush_thread = RegistryFlusher(self)
-            self.flush_thread.start()
             t1 = time.time()
             logger.debug("Registry '%s' [%s] startup time: %s sec" % (self.name, self.type, t1 - t0))
         except Exception as err:
@@ -951,7 +949,6 @@ class Registry(object):
         logger.debug("Shutting Down Registry")
         logger.debug("shutdown")
         try:
-            self.flush_thread.join()
             self._hasStarted = True
             try:
                 if not self.metadata is None:
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -167,6 +167,33 @@ def synchronised(f):
     return decorated
 
 
+class RegistryFlusher(threading.Thread):
+    """
+    This class is intended to be used by the registry to perfom
+    automatic flushes on a fixed schedule so that information is not
+    lost if Ganga is shut down abruptly.
+    """
+    def __init__(self, registry, *args, **kwargs):
+        super(RegistryFlusher, self).__init__(*args, **kwargs)
+        self.registry = registry
+        self._stop = threading.Event()
+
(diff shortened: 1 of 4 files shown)


python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -883,6 +931,8 @@ def startup(self):
             self.repository.startup()
             # All Ids could have changed
             self.changed_ids = {}
+            self.flush_thread = RegistryFlusher(self)
+            self.flush_thread.start()
```
rob-c: Do all Registries call the baseclass for ```startup```?
Is it worth having this configurable within each registry?

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -167,6 +169,49 @@ def decorated(self, *args, **kwargs):
     return decorated
 
 
+class RegistryFlusher(threading.Thread):
```
drmarkwslater: Not sure - should this inherit from GangaThread?

drmarkwslater:
Minor point but would it be good to shift RegistryFlusher to a separate file? Not really bothered either way but I personally prefer one class per file :smile:

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -167,6 +169,49 @@ def decorated(self, *args, **kwargs):
...
+        """
+        This will run an indefinite loop which periodically checks
+        whether it should stop. In between calls to ``flush_all`` it
+        will wait for a fixed period of time.
+        """
+        sleep_period = 30
+        sleeps_per_second = 10  # This changes the granularity of the sleep.
+        while not self.stopped:
+            for i in range(sleep_period*sleeps_per_second):
+                time.sleep(1/sleeps_per_second)
```
rob-c: Doesn't this need to be 1./something?

python/Ganga/Core/GangaRepository/Registry.py
milliams: Every subclass calls (or doesn't override) this method and so every one will have a flusher. About per-registry configuration, I'm not sure. What are the reasons for and against?

python/Ganga/Core/GangaRepository/Registry.py
rob-c: It would be nice to see it do this, but I suspect (for safety at least) we want to handle this entirely within the registries to make sure it is not mistakenly shutdown out of order with the registry in the shutdown of Ganga.

python/Ganga/Core/GangaRepository/Registry.py
milliams: I have `from __future__ import division` at the top of the file so this is not needed.

milliams:
@drmarkwslater Perhaps, although this is an internal detail class. I would almost be tempted to go the other way and make this class a member of Registry so that the association is even clearer.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I would prefer to turn this off for tests in the future at least but I can make this change then if needed.

milliams: That's understandable. However, I'd like to avoid making changes to the code base for the purpose of running tests.

rob-c: Which is why you would write a test with and without the background thread manipulating things on you whilst you're working or trying to work.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -167,6 +169,49 @@ def decorated(self, *args, **kwargs):
...
+        self.stop()
+        super(RegistryFlusher, self).join(*args, **kwargs)
+
+    def run(self):
+        """
+        This will run an indefinite loop which periodically checks
+        whether it should stop. In between calls to ``flush_all`` it
+        will wait for a fixed period of time.
+        """
+        sleep_period = 30
```
rob-c: Can we have this configurable at all?
This would be quite useful to adjust when/if there are problems and I would like to access it for a unittest I'm currently writing

rob-c:
@milliams I found it helpful to skip trying to perform a flush from within the monitoring thread if the lock was already acquired (or at least to add a delay before attempting a flush from a background thread) in a previous iteration of something similar to this. It appears it's causing the tests to lockup so this may help.

python/Ganga/Core/GangaRepository/Registry.py
milliams: Ok, I will add a configuration variable to change the sleep time between flushes as well as turn it off completely.

python/Ganga/Core/GangaRepository/Registry.py
milliams: Yes, I will add this in for sure.

milliams:
@rob-c I considered this in the implementation phase of the registry locks. We can perform a non-blocking call to `acquire()` which, if another thread is holding the lock, will just carry on. @drmarkwslater suggests an option for the auto-flusher where it does a non-blocking acquire for flush every 30 seconds (which will just skip the flush if it can't get the lock) but if it reaches some maximum time since the last flush then it does a blocking acquire and really waits until its chance. I'll leave this out of this PR though but will consider it for another.

rob-c:
@milliams I meant a call to the registry lock rather than the object lock.
If the registry was locked I found strange things happened when a thread like this was queued to immediately grab access to the repo once it becomes free. I suspected the repo lock was being released early and then something was trying to re-acquire it but maybe you're not seeing the same thing here.

milliams:
@rob-c Yes, that is what I meant too, sorry if I wasn't clear. To avoid the sort of potential lockups that you describe I am planning changing the auto-flusher to to try to acquire a non-clocking lock on  the registry before attempting a flush so that if it's busy then it will not wait and grab it the second it becomes free. This means that if the auto-flusher tries to flush at the same moment that someone else is holding a registry lock (probably the monitoring thread but perhaps a parallel submission thread) it will just have to some back again in 30 seconds.

drmarkwslater:
:+1:

milliams:
Ok, this has now been updated with the changes from #266 and #272 which avoid the previously-seen deadlocks. The other change to this PR is that the auto-flusher now only tries to acquire a non-blocking lock on the registry so that if the registry is busy doing other things at the time the auto-flusher comes along, it will skip that run.

If people could look at the latest version of the changes in the PR and give their thoughts?

drmarkwslater:
:+1:

rob-c:
Am I correct in thinking the only way to deactivate this is to raise the timeout config to something very high?

milliams:
@rob-c At the moment, yes. I'm planning on adding option to turn off the flushing. I'm unsure though whether this should be a config variable or an attribute on the registry.

rob-c:
I would have thought a per-registry set of settings would be nice to see if we're thinking of adjusting these things. I can imagine reasons why different registries would want a feature like this on/off.

Also if this is working as expected now, we can begin to think about a background process to unload jobs which have been in memory but not accessed for some time. This would require a lot more work but it's nice to see registry monitoring finally starting to be feasible.

Issue #258, opened 2016-03-03
New test needed for XML Repository


rob-c:
I keep getting bitten time and time again trying to load corrupt information in some form from the XML repo.

Yes the situation is _better_ but it's far from perfect.

It would be extremely useful to have a test which constructs known bad jobs (job configurations) including but not limited to:

 1. Empty Index file
 2. Missing Index file
 3. Corrupt ```data``` file
 4. Missing ```data``` file 

This should be possible and I'm giving some thought as to how to do this.

rob-c:
From a unittest like this it should be possible to build one that checks for consistency of the ```data``` file at a very high level to make sure changes are flushed to disk.

rob-c:
In addition we __will__ need a test to be sure that a flush operation has happened whilst ganga is live

rob-c:
I think I'm testing all but a corrupt index file as part of #268. As soon as I add a final test for this I'll mark this as closed.

rob-c:
I think this has all been addressed by various fixes and tests around the writing of the XML file recently.

PR #268, opened 2016-03-03
Adding new XML tests

rob-c:
OK, this PR adds some new tests which explicitly test the status of the XML files for a job.

These are already known to fail as we're not auto-saving at any point at the moment due to changes 6.1.16 for locking.

These tests (when fixed and 1,000% working) will form the basis for a future set of tests which will break the XML files and indexes and monitor how well ganga copes with loading of XML from disk.

This future test will likely require light modification of the GangaUnitTest to provide an option to startup a ganga environment without the Repositories being fully loaded.
This should be a now trivial change but would allow for this section of code to be required to execute correctly and cleanup gracefully when dealing with file corruption or things like the disk becoming full.

Hopefully these tests should be working and debugged before 6.1.17 is out I would hope as a start.


rob-c:
This sets the groundwork for, but does NOT yet address #258 .

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,233 @@
...
+
+        j.submit()
+
+        newest_update = stat(getXMLFile(j))
+
+        assert len(j.subjobs) == len(testArgs)
+
+        assert newest_update.st_mtime > last_update.st_mtime
+
+    def test_b_JobXMLExists(self):
```
milliams: This function name is repeated twice in the class.

python/Ganga/new_tests/GPI/newXMLTest/TestXMLGenAndLoad.py
```diff
@@ -0,0 +1,159 @@
+from __future__ import absolute_import
+
+from ..GangaUnitTest import GangaUnitTest
+
+from os import path, stat
+
+import time
+
+def getJobsPath():
```
milliams: This function is repeated from the other file. Perhaps move them to a common place or into _init__.py?

python/Ganga/new_tests/GPI/newXMLTest/TestXMLGenAndLoad.py
```diff
@@ -0,0 +1,159 @@
...
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
+
+    def test_a_JobConstruction(self):
+        """ First construct the Job object (singular)"""
+        from Ganga.Utility.Config import getConfig
+        self.assertFalse(getConfig('TestingFramework')['AutoCleanup'])
+
+        from Ganga.GPI import Job, jobs
+        j=Job()
+        self.assertEqual(len(jobs), 1) # Don't really gain anything from assertEqual...
```
milliams: I don't understand this comment?

milliams:
Perhaps the directory name for these tests should be `XMLTest` rather than `newXMLTest`?

python/Ganga/new_tests/GPI/newXMLTest/TestXMLGenAndLoad.py
rob-c: It's copy/paste from a legacy test, I'm not expecting these tests to be fully debugged yet, this is a bit WIP but I was interested in feedback.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
rob-c: Ooops, wll fix

python/Ganga/new_tests/GPI/newXMLTest/TestXMLGenAndLoad.py
rob-c: WIP, I'm thinking about it in order to develop additional tests

milliams:
Overall I like the idea of these tests. I will need to check through them thoroughly to make sure they represent our expected behaviour but to first order, I'd like to have them in.

python/Ganga/new_tests/GPI/newXMLTest/TestXMLGenAndLoad.py
milliams: Great.

rob-c:
We already have some XML related tests we may want to port over I wanted to distinguish between the, so I'm leaving them in newXMLTest for now. Still very WIP anyway (read these tests still have their own bugs to be debugged rather than the code they're testing)

They represent nothing other than checking functions do what they should. The tricky part comes in testing the subjob master index properly as well as a few other quirks with the file comparisons not working due to a mistake.

After this I plan to make loading of the registries optional in GangaUnitTest so I cam test it properly but this will be added in the near future.

drmarkwslater:
I'm guessing an merge with develop should let these tests pass now #267 is in?

rob-c:
I need to do some private testing but I'll do that now to try and get these in.

rob-c:
These tests are now much closer to passing but I'm seeing that the job XML is not being updated after the job is submitted.
After the meeting yesterday I thought we agreed that at the very least the job should trigger a flush after being submitted but preferably it should trigger a flush before submitting and after splitting so that the result of the splitting (potentially a _very_ expensive operation) is saved.

milliams:
@rob-c Yes, I have a PR for that incoming.

rob-c:
OK, I'll wait until all PR are in to try and fix that bug. I'm seeing a deadlock I think in these tests which shouldn't be happening which may be cause for concern as it's unlikely the test (unless I'm obviously doing something insane which is possible as there are 2 jobs in memory with the same 'id')

rob-c:
OK, I've just added a basic test to see that Ganga can load correctly a job from backup with a missing or corrupt ```data``` file.
Ganga seems to support this well (although I've currently no test for if an appropriate warning is thrown).

I'll take a look at why any of the other tests are failing as it's worth understanding what the expected behaviour of the XML parsing is.

Once this all works I'm keen to get this into 6.1.17 so that some further testing can be done.

rob-c:
@milliams  Any idea what ```FATAL: invalid interpreter: /opt/rh/python27/root/``` could mean?

rob-c:
retest this please

rob-c:
Well I can't confirm or deny if the tests run on the Jenkins box as it appears to be down for PR 'builds' atm.

milliams:
The newly added tests seem to be failing (and taking an incredibly long time to do so). I'm not sure it's worth holding up the release for these new tests. In my private tests, jobs are being flushed to disk correctly so we've probably resolved that issue.

rob-c:
@milliams They're almost all passing on my machine eta 20min for a fix.

If I get these in and they pass I'll push for merge. If I get these in and they have problems in Jenkins then I say we delay.

ATM, they're simply not running on Jenkins due to a Jenkins problem, or has this been fixed?

rob-c:
OK, upon inspection Jenkins is now magically working for me...

I'll update the tests in a few minutes now that I've debugged some errors

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
+
+    def test_a_JobConstruction(self):
+        """ First construct the Job object (singular)"""
+        from Ganga.Utility.Config import getConfig
+        self.assertFalse(getConfig('TestingFramework')['AutoCleanup'])
+
+        from Ganga.GPI import Job, jobs, ArgSplitter
+        j=Job(splitter=ArgSplitter(args=testArgs))
+        self.assertEqual(len(jobs), 1) # Don't really gain anything from assertEqual...
```
milliams: I think this comment needs removing?

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def test_b_JobXMLExists(self):
+        """ Second get the job and check that getting it via jobs doesn't cause it to be loaded"""
+        from Ganga.GPI import jobs
+
+        assert len(jobs) == 1
+
+        print("len: %s" % str(len(jobs)))
```
milliams: You don't need to pass the arguments to `%` though `str()`.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+
+        XMLFileName = getXMLFile(j)
+
+        last_update = stat(XMLFileName)
+
+        j.name = testStr
+
+        from Ganga.Utility.Config import getConfig
+        flush_timeout = getConfig('Registry')['AutoFlusherWaitTime']
+        time.sleep(2.*flush_timeout)
```
milliams: Could we make the `AutoFlusherWaitTime` shorter for this test so that we're not waiting for 60 seconds for no reason.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+
+        from Ganga.GPI import jobs, Job
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+
+        from tempfile import NamedTemporaryFile
+
+        j=jobs(0)
+        XMLFileName = getXMLFile(j)
+        assert path.isfile(XMLFileName)
+        handler = open(XMLFileName)
```
milliams: Does this need closing? Or maybe using a context manager?

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+        to_file(stripProxy(j), new_temp_file, ignore_subs)
+
+        new_temp_file2 = NamedTemporaryFile()
+
+        j2=Job()
+        j2.name=testStr
+
+        to_file(stripProxy(j2), new_temp_file2, ignore_subs)
+
+        import filecmp
```
milliams: Put all imports at the top of the file

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+    def test_g_testSJXMLContent(self):
+
+        from Ganga.Core.GangaRepository.VStreamer import to_file, from_file
+
+        from Ganga.GPI import jobs
+        from tempfile import NamedTemporaryFile
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+
+        counter = 0
+        for sj in jobs(0).subjobs:
```
milliams: This could probably use `enumerate()` instead of carrying a counter variable.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+
+        j=jobs(0)
+
+        IndexFileName = getIndexFile(j)
+
+        assert path.isfile(IndexFileName)
+
+        handler = open(IndexFileName)
+        _obj, errs = from_file(handler)
+        obj=_obj[0]
```
milliams: How about:
```python
obj = from_file(handler)[0][0]
```
?

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+        from Ganga.Core.GangaRepository.PickleStreamer import to_file, from_file
+
+        from Ganga.GPI import jobs
+
+        assert len(jobs) == 1
+
+        j=jobs(0)
+
+        handler = open(getSJXMLIndex(j))
+        handler = open(getIndexFile(j))
```
milliams: This is overwriting the same variable.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,229 @@
...
+        _obj, errs = from_file(handler)
+        obj=_obj[0]
+
+        assert isinstance(obj, dict)
+
+        from Ganga.GPIDev.Lib.Registry.JobRegistry import getIndexCache
```
milliams: Is there a reason not to have this at the top of the file?

python/Ganga/new_tests/GPI/newXMLTest/TestXMLCorruption.py
```diff
@@ -0,0 +1,105 @@
...
+        """ First construct the Job object (singular)"""
+        from Ganga.Utility.Config import getConfig
+        self.assertFalse(getConfig('TestingFramework')['AutoCleanup'])
+
+        from Ganga.GPI import Job, jobs
+        j=Job()
+        assert len(jobs) == 1
+
+        global global_AutoStartReg
+        global_AutoStartReg = False
```
milliams: I'd prefer this to be a class variable rather than a global.

python/Ganga/new_tests/GPI/newXMLTest/TestXMLCorruption.py
```diff
@@ -0,0 +1,105 @@
...
+
+        XMLFileName = getXMLFile(0)
+
+        unlink(XMLFileName)
+        assert not path.isfile(XMLFileName)
+
+        handler = open(XMLFileName, 'w')
+        handler.write(badStr)
+        handler.flush()
+        handler.close()
```
milliams: Context manager again.

milliams:
Apart from my comments directly on the diff, I think the tests need some comments to explain what the tests overall are checking for and specifically what each assert exists for. I've found, going through some of the legacy tests, that it's impossible to understand an old failing test without an explanation for its existence.

rob-c:
The meaningless commits like this are because I'm working trying to debug something. If you want to merge them into something more sensible feel free but editing the git commit history is not something I want to become an expert in

rob-c:
OK, I'm having real trouble comparing the XML for subjobs. This aside I've tested the whole chain and aside from being able to test the content of the XML in the case of a split job everything appears to be working as expected,

Given this has dragged on much longer than the 20min I allotted I'm happy to push this back to 6.1.18 as it's a test and 6.1.17 is probably working as expected given the testing that it's gone through.

rob-c:
All tests should be working now after a small change to the logic in the VStreamer. (I'm keen to see the impact of this on other tests but it's a small change)

I need to add a proper nested XML object test now to make sure that this is unchanged as well as potentially a more modern take on the Schema test which just in isolation tests the creation, flush and load of all objects in the GPI.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
rob-c: but this doesn't cause any bad behaviour either

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
rob-c: I was not wanting to change from the defaults in Ganga. I'm trying to avoid un-needed setting of options and I don't want to introduce lockups by hammering the repo excessively with it's internal thread.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
rob-c: Probably, but I planned to add a assert test on the ```counter``` parameter and forgot to add it

python/Ganga/new_tests/GPI/newXMLTest/TestXMLCorruption.py
rob-c: Given the object shouldn't ever exist out of the scope of the test I'm leaving it as a global

python/Ganga/new_tests/GPI/newXMLTest/TestXMLCorruption.py
rob-c: I prefer to be explicit in the use of flush and open/close. I've seen too many strange things with context managers under the old test system to trust them

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -0,0 +1,249 @@
...
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
+
+    def test_a_JobConstruction(self):
+        """ First construct the Job object (singular)"""
+        from Ganga.Utility.Config import getConfig
+        self.assertFalse(getConfig('TestingFramework')['AutoCleanup'])
+
+        from Ganga.GPI import Job, jobs, ArgSplitter
+        j=Job(splitter=ArgSplitter(args=testArgs))
+        lssert len(jobs) == 1
```
milliams: Typo. How did this pass the test suite when you ran it locally before pushing?

rob-c: Broke after testing when I was editing the file

rob-c:
Interesting, both XML content tests are failing on these servers compared to my private testing. I've a suspicion what could be leading to this but aside from this a lot of these tests are in a state I think I'm happy to call working.

rob-c:
OK I've added a new test to verify that the nested XML works as expected and gives back the same nested list that was put in.

I'm now going to investigate what could be causing the few outstanding test errors ahead of pushing for merge.

rob-c:
retest this please

rob-c:
@milliams  I'm having the problem of 2 files on the Jenkins machine are not equal in the tests (they are when I run locally) can you suggest a way to return/get/inspect these files during testing.
I'm thinking of dumping them through the warning logger but I'm curious if there is a magic recipe that is better than inspecting the stdout?

milliams:
@rob-c pytest will provide a diff view of two long strings when they are compared. You should be able to do:
```python
assert open(file1).read() == open(file2).read()
```
and you will get lines like:
```
E       AssertionError: assert '\t\t    GNU ...is License.\n' == '\t\t    GNU G...is License.\n'
...
E           have the freedom to distribute copies of free software (and charge for
E           this service if you wish), that you receive source code or can get it
E         - if you want it, that you can change the software or use pieces of it
E         + if you want it, thatdf you can change the software or use pieces of it
E         ?                     ++
E           in new free programs; and that you know you can do these things.
E           
E             To protect your rights, we need to make restrictions that forbid
...
```
(note the `+`, `-` and `?`) if pytest is in verbose mode (which we do by default).

rob-c:
@milliams Thanks, shame it doesn't do this descending into filecmp... I'll update the comparisons now

rob-c:
OK I've finally finished debugging these tests and fixing a few smaller bugs which have been highlighted in the testing. I'm keen to push for merge if there aren't any objections tomorrow.

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -44,18 +44,19 @@ def __str__(self):
...
     vstreamer.begin_root()
     stripProxy(j).accept(vstreamer)
     vstreamer.end_root()
 
-def to_file(j, fobj=None, ignore_subs=''):
+def to_file(j, fobj=None, ignore_subs=[]):
     #used to debug write problems - rcurrie
     #_raw_to_file(j, fobj, ignore_subs)
     #return
+    _ignore_subs = [ignore_subs] if not isinstance(ignore_subs, list) else ignore_subs
```
milliams: Are there any places where this is called with a string or something else not a list?

python/Ganga/new_tests/GPI/newXMLTest/TestNestedXMLWorking.py
```diff
@@ -0,0 +1,121 @@
...
+
+        assert j.splitter.args == getNestedList()
+
+    def test_b_JobNotLoaded(self):
+        """ Second get the job and check that getting it via jobs doesn't cause it to be loaded"""
+        from Ganga.GPI import jobs
+
+        assert len(jobs) == 1
+
+        print("len: %s" % str(len(jobs)))
```
milliams: There's no need for `str()` call here.

python/Ganga/new_tests/GPI/newXMLTest/TestNestedXMLWorking.py
```diff
@@ -0,0 +1,121 @@
...
+        from Ganga.Core.GangaRepository.VStreamer import to_file, from_file
+
+        from Ganga.GPI import jobs, Job, ArgSplitter
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+
+        from tempfile import NamedTemporaryFile
+
+        j = jobs(0)
+        assert path.isfile(getXMLFile(j))
+        handler = open(getXMLFile(j))
```
milliams: Could this be done with a context manager? I don't see any corresponding `close` as it stands.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
milliams: It does make the code noisier than necessary though. It's much cleaner to do without.

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: Yes ```GangaRepositoryXML``` to (among other things I think) not write all of the subjobs into the master job ```data```

python/Ganga/new_tests/GPI/newXMLTest/TestNestedXMLWorking.py
rob-c: I'm really not in favor of context managers here as I just want to keep the files open and not deal with indentation all over the place. I'll add a corresponding ```close```.
This comes down to personal preference and maybe in a re-iteration of these tests we'll go that way but for now I'm not a fan of it.

python/Ganga/new_tests/GPI/newXMLTest/TestNestedXMLWorking.py
rob-c: Yes, but it does no harm :P

python/Ganga/new_tests/GPI/newXMLTest/TestXMLCorruption.py
milliams: Context managers should always (excepting interpreter crashes) be safe. If we can't rely on them then we have much bigger problems.

setup.py
```diff
@@ -54,7 +54,7 @@ def _get_test_env():
 
     def run(self):
 
-        cmd = ['py.test']
+        cmd = ['py.test -vv']
```
milliams: We shouldn't merge this change in. For testing you can always put these things in `setup.cfg`.

python/Ganga/new_tests/GPI/newXMLTest/TestXMLCorruption.py
rob-c: The old unittest system had bigger problems. I discovered it bleeding files everywhere when mis-using contexts.It also abused logging redirection enough that there were problems with reliably printing string to the terminal.

setup.py
rob-c: Is there ever a reason why more verbosity would be bad?
If a test ever fails it's likely one of the first things we're going to do is working out how the ```ganga-ci``` differs from some systems we're developing on. This likely requires some more feedback from assert errors.

milliams: It's not that verbosity is bad but rather that this configuration should be done in `setup.cfg` so that it affects all runs of pytest, not just those through `setup.py`.

python/Ganga/new_tests/GPI/newXMLTest/TestNestedXMLWorking.py
milliams: If we're dealing with resources we should *always* be opening and closing them either with context managers or with `try/finally` (if the former isn't possible). It's more explicit and it's the only way to be safe from leaking resources (as recently seen in `SessionLock.py`).

setup.py
rob-c: In that case is it possible to potentially modify this ```setup.cfg``` on the Jenkins machine at all?
Otherwise we're going to keep putting this in/taking it out and we'll likely end up missing it once and it'll go in. Or we have to keep checking for this which is making work for work's sake.

python/Ganga/new_tests/GPI/newXMLTest/TestNestedXMLWorking.py
rob-c: If resources aren't cleared between tests isn't this an issue for the test framework to worry about?
Afterall the test may have failed and caused a bug to leave objects lying around.

milliams: Files will be deleted but file handles will be kept open until the whole test process finishes. Nothing's going to explode but we should be treating our test code with the same care and attention that we treat core code and that means writing in a clean and Pythonic way.

rob-c: If this is going to be a pressed issue I'll change this but this will push back the PR (yet again) until I get the time to edit/test/commit the code.

milliams: I won't press the issue further. I can always make the changes myself in a future PR.

rob-c:
OK, move to contexts and yet more indentation. On top of this reverted the ```-vv``` which I would still prefer to see used with Jenkins.
If this passes and everything is good then I'm hoping to merge tomorrow.

milliams:
I'm pretty happy for this to be merged. My only hesitation is that two of the new tests take a minute to run when most of that time is spent simply waiting for the auto-flusher to do it thing. Since tests are most useful if they are fast and run often, I would be inclined to tweak those tests to make the auto-flusher run more frequently.

drmarkwslater:
Isn't there a config option that can be set to speed the auto flusher up? I forget what it is but that should speed things up...

rob-c:
@milliams They only take ~30sec the test keep re-checking for the XML file to be modified and once it is it moves onto the comparison. If the XML file isn't modified it will wait a total of 2*auto_flush time length.
I'm more concerned with reducing the length of time taken to submit the non-split parallel jobs which hasn't highlighted a proper bug in a while.

rob-c:
Is there a strong call to reduce this timeout from 30sec? I would have thought this was an acceptable pause time for a test.

rob-c:
OK, auto flush time reduced. If the tests pass and nothing lockups I plan to merge unless there are any objections

rob-c:
OK, there's something broken with the configuration of this parameter, either it's not being respected, not working or the auto-flushing loop can't run through an instance of a loop in 10sec.
I'll reluctantly drop this until 6.1.19 as this is not understood (yes I've tried setting the parameter through the extra_opts in to GangaUnitTest as well. This is confusing me.

milliams:
I think it's not respecting the config parameter since it is currently only reading it on first start. Probably the easiest solution is to change `RegistryFlusher.run` from:
```python
...
sleep_period = getConfig('Registry')['AutoFlusherWaitTime']  # Move this line...
sleeps_per_second = 10  # This changes the granularity of the sleep.
while not self.stopped:
    for i in range(sleep_period*sleeps_per_second):
        time.sleep(1/sleeps_per_second)
        ...
```
to
```python
...
sleeps_per_second = 10  # This changes the granularity of the sleep.
while not self.stopped:
    sleep_period = getConfig('Registry')['AutoFlusherWaitTime']  # ...to here
    for i in range(sleep_period*sleeps_per_second):
        time.sleep(1/sleeps_per_second)
        ...
```
so that it updates it every time it loops.

rob-c:
I've changed the config to change the value before the test starts and changed my test logic to fix a comparison error. This should be working so if it is I'll just merge into develop after 6.1.18 is out

rob-c:

milliams:
My main objection is the commit messages. There's about 10 "Update TestSJXMLGenAndLoad.py"-type messages and a whole bunch of "fixing tests". Could you squash these together in some way to make the commit history more readable and useful?

rob-c:
I would if I knew how, is there a good dummies tutorial for doing these sort of things?

alexanderrichards:
with your branch checked out, issue a:
```bash
git rebase -I develop
```

then change the p for pick to either s for squash or f for fixup

milliams:
https://robots.thoughtbot.com/git-interactive-rebase-squash-amend-rewriting-history is a decent tutorial about these sorts of things. Also https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History and http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

alexanderrichards:
> then change the p for pick to either s for squash or f for fixup

squash will combine the commit message for that commit with the previous ones while fixup will discard the message all together

alexanderrichards:
**NOTE:** This will rewrite the git history and your commits will get new hashes (or at least those below any changes you make). This will mean that when you try to git push it will fail as you are trying to push a new history. you will need to add the --force flag e.g.
```bash
git push --force
```

rob-c:
@alexanderrichards  I'm assuming you meant ```-i``` either way that was an unpleasant 5 minutes with nano I'll have to erase from my mind...

There should really be a click-button gui interface for this sort of thing I appreciate managing it through text files if probably more powerful but I'm amazed there's not a simple button to squash 2 commits in the github gui.

alexanderrichards:
yes sorry I mistyped and put a capital -I, I have corrected the original post

PR #269, opened 2016-03-03
Disabling HeartBeat checking in Monitoring system by default

rob-c:
Disabling heartbeat by increasing default timeout to an obscene number.

I want to keep these as warnings so that we can much more easily debug what's going on but for users who think they're experiencing lockups can change this to be something more sensible like 300.


milliams:
:+1:

rob-c:
I'll wait to merge due to not wanting to force this.

drmarkwslater:
:+1:

PR #272, opened 2016-03-04
Make GangaObjects only lock their root object

milliams:
Rather than having a lock on every single GangaObject which then propagates up the parent chain, instead just acquire the lock in the object's root object. This is sensitive to altering the parent of an object so extra logic is need in _setParent to lock both the old and new root.

This is a requirement for getting #267 working properly.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -141,8 +141,8 @@ class Node(object):
     @contextmanager
     def lock(self):
         """
-        This is a context manager which acquires the lock on the object and all it's ancestors.
-        When the context manager exits, all the locks are released in the reverse order to that which they were acquired.
+        This is a context manager which acquires the lock on the
+        object's root object.
         """
         root = self._getRoot()
         root._lock.acquire()
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -125,13 +125,17 @@ class Node(object):
         setattr(obj, '_registry', self._registry)
         return obj
 
-    @synchronised
     def _getParent(self):
         return self._parent
 
-    @synchronised
+    @synchronised  # This will lock the _current_ (soon to be _old_) root object
     def _setParent(self, parent):
-        setattr(self, '_parent', parent)
+        if parent is None:
+            setattr(self, '_parent', parent)
+        else:
+            with parent.lock:  # This will lock the _new_ root object
+                setattr(self, '_parent', parent)
+            # Finally the new and then old root objects will be unlocked
 
     @property
     @contextmanager
@@ -140,19 +144,12 @@ class Node(object):
         This is a context manager which acquires the lock on the object and all it's ancestors.
         When the context manager exits, all the locks are released in the reverse order to that which they were acquired.
         """
-        self._lock.acquire()
-        ancestors = []
-        p = self._parent
-        while p is not None:
-            p._lock.acquire()
-            ancestors.append(p)
-            p = p._parent
+        root = self._getRoot()
+        root._lock.acquire()
         try:
             yield
         finally:
-            for p in reversed(ancestors):
-                p._lock.release()
-            self._lock.release()
+            root._lock.release()
 
     # get the root of the object tree
     # if parent does not exist then the root is the 'self' object


rob-c:
This gets us back to the problem of potentially not being able to 'get' an attribute whilst 'set'ing another one doesn't it?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -140,19 +144,12 @@ def lock(self):
         This is a context manager which acquires the lock on the object and all it's ancestors.
```
rob-c: Can we update the text if this isn't being done

milliams:
That situation is already present in 6.1.16, it locks the object tree so that only one thread can interact with its schema items at once. This PR doesn't make any functional changes to the external expected behaviour, it is intended to simplify the acquisition of object locks to reduce the chances of deadlocks.

Presently, acquiring a child lock in one thread while another thread is printing the root object will cause a deadlock (I think I've seen this during submit for example). This PR solves this problem by only having a single lock per object tree.

python/Ganga/GPIDev/Base/Objects.py
milliams: Indeed. Will do that now.

rob-c:
Hmm, I'm suspicious, wouldn't a top down approach to locking fix this issue, lock the Root then lock everything along the path to the Root to be sure nothing bad happens?

milliams:
It would, but if the root object is locked then there's no need to lock all the children as well.

rob-c:
I'd be much happier if we were correctly locking the children as well. Otherwise what's the point in having numerous threading Locks in every GangaObject?

milliams:
Every `GangaObject` can potentially be a root object (if it's unparented) so they all need to have ability to be locked. By default none of them will be ever activated unless they are at the root of a hierarchy.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -125,13 +125,17 @@ def __deepcopy__(self, memo=None):
...
         return self._parent
 
-    @synchronised
+    @synchronised  # This will lock the _current_ (soon to be _old_) root object
     def _setParent(self, parent):
-        setattr(self, '_parent', parent)
+        if parent is None:
+            setattr(self, '_parent', parent)
+        else:
+            with parent.lock:  # This will lock the _new_ root object
```
rob-c: Should this be parent._lock?

rob-c: nevermind I see it, but isn't .lock vs ._lock confusing a little?

milliams: No, it's asking the parent for its lock which will in-turn get the lock on the root object of `parent`. For consistency, the only place that `_lock` should ever be accessed should be inside the `lock` method since that method knows how to use it.

rob-c:
I agree one lock per tree is simpler but then I'd like to see some form of getLock method to avoid assigning several thousand locks when we construct objects.

python/Ganga/GPIDev/Base/Objects.py
milliams: Well, `lock` is the external interface what API users can use. `_lock`, given its leading underscore, in an implementation detail.

milliams:
We have `with obj.lock:` so there's no use for (and would be dangerous to have) a `getLock()` method. On top of that `getFoo` methods in Python are an anti-pattern which should be instead replaced with `@property`s (which is what `obj.lock` is).

rob-c:
I'm not thinking of literally calling it this I'm just wanting to cut down on the amount of objects constructed at ```__init__```

milliams:
According to my tests, it takes about 900 ns to create a `threading.RLock` object which means in one second we can create more than a million of them. I don't imagine that this is contributing to any noticeable slowdown, especially given everything else that is done in `GangaObject.__init__`.

rob-c:
I'm seriously wanting to reduce to a bare minimum the amount of stuff lying around in GangaObject and Node but if this works then go with it, I just think we could have avoided passing extra stuff around everywhere.

python/Ganga/GPIDev/Base/Objects.py
rob-c: I understand the naming convention I just missed that ```def lock``` existed :P

rob-c:
I'm keen to see if this fixes my issues elsewhere with deadlocks so go for merge

milliams:
I can understand that, however the alternative is to have `hasattr` checks or similar every time a lock acquisition is attempted which over the course of things will be much slower. I don't like the idea of object maybe having or maybe not having attributes where it can be avoided.

PR #278, opened 2016-03-04
Fixing use of flush commands...

rob-c:
I'm assuming we want to standardize on flush_all in this case, but either way this line is broken

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -712,7 +712,7 @@ class Registry(object):
             self._flush(self.values())
 
         if self.metadata and self.metadata.hasStarted():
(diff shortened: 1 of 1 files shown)


rob-c:
I'm just going to merge as this is obviously little more than a typo

milliams:
Yes, I have made this change to mine locally but hadn't pushed it yet. I think that happened because @drmarkwslater copied those two lines from your PR to his.

rob-c:
I think this is because the default behavior at some point have repo._dict() not do what it was expected to do (flush all dirty). Given we're standardizing on flush_all being a wrapper for this I'm happy enough to go with what is being changed here.

drmarkwslater:
Oops - Good catch @rob-c ! I did just blindly copy this :unamused:

PR #279, opened 2016-03-04
Add explicit flush calls at strategic points throughout Job submit

milliams:
As discussed in the meeting, this will explicitly call flush at the beginning and end of submission. The performance impact of this remains to be seen but it is probably worth it.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -1491,7 +1491,7 @@ class Job(GangaObject):
 
         For split jobs: consult https://twiki.cern.ch/twiki/bin/view/ArdaGrid/GangaSplitters#Subjob_submission
         """
-
+        self._getRegistry()._flush([self])
         logger.debug("Submitting Job %s" % str(self.getFQID('.')))
 
         gpiconfig = getConfig('GPI_Semantics')
@@ -1625,7 +1625,7 @@ class Job(GangaObject):
             # make sure that the status change goes to the repository, NOTE:
             # this commit is redundant if updateStatus() is used on the line
             # above
-            self._commit()
+            self._getRegistry()._flush([self])
 
             # send job submission message
             if len(self.subjobs) == 0:
@@ -1684,9 +1684,7 @@ class Job(GangaObject):
 
             ganga_job_submitted(getName(self.application), getName(self.backend), "0", "1", str(submitted_count))
 
-
-        stripProxy(self)._setDirty()
-        stripProxy(self)._getRegistry()._flush([stripProxy(self)])
+        self._getRegistry()._flush([self])
 
         return 1
 
@@ -2151,7 +2149,7 @@ class Job(GangaObject):
     def _commit(self, objects=None):
         """ Helper method to unconditionally commit to the repository. The 'objects' list specifies objects
         to be commited (for example the subjobs). If objects are not specified then just the self is commited """
-
+        # TODO This function should force a flush. However it is currently used everywhere which causes too many flushes.
         self._getRoot()._setDirty()
 
 


python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -2151,7 +2149,7 @@ def check_changability(obj1, obj2):
     def _commit(self, objects=None):
         """ Helper method to unconditionally commit to the repository. The 'objects' list specifies objects
         to be commited (for example the subjobs). If objects are not specified then just the self is commited """
-
+        # TODO This function should force a flush. However it is currently used everywhere which causes too many flushes.
         self._getRoot()._setDirty()
```
rob-c: Surely the calls to this should be wrapped behind something sensible.
I think _commit should only be being called when the status changes and this can easily be set to exclude all changes to which aren't ```'submitting'``` or ```'submitted'```?

drmarkwslater:
Am I right in saying this is now not needed thanks to the Auto flusher and will be superseded by a queue/task request to the flusher instead?

rob-c:
I would __strongly__ prefer to have something like this. Yes we shouldn't have a need for it as it's unlikely the user will get a broken connection/hit Ctrl+C within 30sec of a job transition but I would just sleep easier knowing that upon a job transition to any of the states: 'submitted', 'running' and 'completed'/'failed' just means that we're 100% sure the job has been flushed as expected.
We've gone from several thousand flush ops per minute submit to < 100 in the whole submit for most cases so I think we can afford the latency we have from flushing a job which has changed significantly.

drmarkwslater:
For the new functionality of the auto flusher that @milliams is going to add, a request to flush will be picked up and carried out (locks permitting) within a second so you won't have to wait 30s for it. The main reason I would very much prefer the queue idea is that we restrict flushes to just one place in the code and as the flush was the main source of deadlocks, this makes everything a lot cleaner. This also has the added benefit of pushing the (possibly very expensive) flush to the background and not holding up the user, i.e. j.submit goes even quicker :smile:

There is an additional question of do we actually want to flush half way through submission though. I would argue this wouldn't help and could do harm because you would saving the job in an inconsistent state that you couldn't do much with afterwards. Requesting a flush at the start and at the end should really be enough I feel.

Having said all that, with the code as we have it, direct flushing in submit is definitely a good idea and I withdraw my previous comment as the PR passes the tests and thinking about it I don't **think** it could cause deadlocks as it stands :smile:

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: The purpose of `_commit` historically was to *force* a flush whenever something significant changed. This is why it use ubiquitously throughout `Job`. As you found and fixed, this was flushing far too often so the first thing we need to fix with `_commit` is reducing the places that it is called.

rob-c:
I'm still very weary that the result of a crash due to some bug which could break a lot of things and cause the repo to be closed down (this still happens a bit!) could leave us in a state where we have 'just lost' the result of an expensive operation which we could have saved to disk with a strategic flush.

On a serious Repo error the repo is locked and all transactions and future flushes are forbidden so I don't like the over-reliance of 'oh the dirty monitoring thread will take care of it'.

I would argue we __need__ to flush during large job submission as it's very expensive and much easier to recover by launching a second session and just continue to submit a few un-submitted/bad subjobs.

milliams:
As @drmarkwslater said, my plan for the next release (6.1.18 or after) is to add a `flush_soon` queue to the registry flusher which will schedule a flush of that object to happen in the next second or so. This allows us to centralise the calls to `_flush` in one place (Registry.py) and also desynchronises the call so that it doesn't slow down submission. Whether we would then want a "really flush right now and I will wait until it's finished" call during `Job.submit` is up for discussion.

rob-c:
Why isn't the Job in this case calling ```_flush``` I'm not exactly won over by passing the equivalent of 'flush now' to a queue which may be busy with things like parallel submit and so will _not_ process this for much, much longer than 30sec.
Either way if this is a 6.1.18 development I'm keen to restore the expected behavior we talked about in the meeting in 6.1.17 to have some level of opportunistic flushing on job status changes.

I'm much, _much_ more in favor of making the user wait (at most 30sec-1min on AFS) to flush a very large job and have some level of safety in the job vs relying on a 'try and flush around now' system. It's a big change to what we've previously discussed and I'd be happier letting 6.1.17 get out with some level of belt and braces on the safety of the users data.
Yes this is heavy-handed but I'd rather be heavy-handed as then I can sleep at night rather than worrying about the next time users manage to configure things such that they lose data.

drmarkwslater:
I think we're actually in agreement then - for 6.1.17 it's :+1: for this PR (assuming no other deadlocks are seen) and then what we do for 6.1.18 is up for discussion :smile:

milliams:
When I say "add a `flush_soon` queue to the registry flusher" I don't mean using the Ganga queue system. I mean a completely separate queue which is internal to the flusher which processes flush requests.

rob-c:
@milliams In this case I now worry that this is a __very__ complex solution compared to asking the user to wait for some 30sec for their job to save. I'm not convinced we need this.

milliams:
It comes down to how cautious we want to be. You seemed to be arguing above that waiting 30 seconds is too dangerous as there could be crashes in the meantime which would lose data. My future proposed solution is intended to allow us to prioritise important status transitions so that they are processed as soon as possible. If and when we think this might be a possible solution, I will open a PR with the changes and we can discuss them in detail there. For now, let's focus on the changes in this PR.

egede:
@rob-c @drmarkwslater @milliams Seems like we have agreement here for 6.1.17. I suggest we merge this asap and then return to this discussion later.

rob-c:
@milliams  This is discussion for elsewhere yes.
As for this PR, I suspect that ```_flush``` changes commands are correct as we discussed in the meeting.
I'd prefer to merge this and finish my tests for 6.1.17

Issue #204, opened 2016-02-18
Issue with gaudi options parsing


rmatev:
An issue was noticed by a few users that the newest DaVinci (v39r0p1) does not work with ganga. 

To reproduce it is enough to do:
```python
app = DaVinci(version='v39r0p1')
app.extraopts = 'from Configurables import DaVinci; DaVinci(DataType="2015")'
app.prepare()
```

I think the issue is with the way ganga tries to read the pickled option file at
https://github.com/ganga-devs/ganga/blob/0b796886d15f843d38e1a85bbb6078f0677dca93/python/GangaLHCb/Lib/Applications/PythonOptionsParser.py#L66

The problem is that gaudirun.py is not called in the correct environment (DaVinci), thus reading of the pickle fails since various classes are not available. See this little demo outside ganga:

```sh
lb-run DaVinci v39r0p1 gaudirun.py --option "from Configurables import DaVinci; DaVinci(DataType='2015')" --output test.pkl -n
# the following fails as in ganga
python -c "import cPickle; cPickle.load(file('test.pkl'))"
# this one works
lb-run DaVinci v39r0p1 python -c "import cPickle; cPickle.load(file('test.pkl'))"
```

rmatev:
Okay, what I concluded is a bit premature, as the demo at the end fails with DaVinci v38 too. There must be more to it.

rmatev:
For the record, here is the interesting part of the failure output:
```
Ganga.GangaGaudi.Lib.Applications  : INFO     Preparing DaVinci application.
Ganga.GangaGaudi.Lib.Applications  : WARNING  The 'optsfile' is not set. I hope this is OK!
Ganga.GangaLHCb.Lib.Applications   : ERROR    Cannot eval() the options file. Exception: name 'DataObjectDescriptorCollection' is not defined
Traceback (most recent call last):
  File "/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v601r13/install/ganga/python/GangaLHCb/Lib/Applications/PythonOptionsParser.py", line 74, in _get_opts_dict_and_pkl_string
    options = eval(opts_str)
  File "<string>", line 1, in <module>
NameError: name 'DataObjectDescriptorCollection' is not defined
Traceback (most recent call last):
  File "/afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc48-opt/lib/python2.7/logging/handlers.py", line 76, in emit
[...]

PR #280, opened 2016-03-04
Adding workaround for dictionary bug

rob-c:
Addressing #204


milliams:
I would favour a very obvious comment in the code pointing to the issue on GitHub so that when we forget to remove this in a few months if someone comes across it in the far future they know that it's safe to delete.

rob-c:
I'll add it though the edit function on github

PR #282, opened 2016-03-07
Changed little ganga to use getProxyClass(cls) to avoid error

drmarkwslater:
Very minor change to avoid errors to do with objects not having the '_proxyClass' member.


milliams:
:+1:

rob-c:
Can we drop this to 6.1.18 if we're not advertising little ganga use yet?

drmarkwslater:
We can I suppose but it doesn't touch any of the main code base and would make my life easier with creating the tutorial tests.

milliams:
I think we should go ahead and merge this. It won't affect anything negatively but will at least allow `import ganga` to succeed.

rob-c:
In this I say go for merge so we can get 6.1.17 out

Issue #285, opened 2016-03-07
Exceptions thrown from threads on shutdown with Python 2.6 and Python 2.7


egede:
Working on the 6.1.17 release candidate, I see the errors below, each time I exit when using Python 2.6. With Python 2.7 there are no exceptions.

```
[20:15:07]
Ganga In [9]: 
Do you really want to exit ([y]/n)? 

Ganga.Core.MonitoringComponent     : INFO     Some tasks are still running on Monitoring Loop
Ganga.Core.MonitoringComponent     : INFO     Please wait for them to finish to avoid data corruption
Ganga.Core.InternalServices        : INFO     Stopping Job processing before shutting down Repositories
Ganga.Core.InternalServices        : INFO     Shutting Down Ganga Repositories
Ganga.Runtime                      : INFO     Registry Shutdown
Exception in thread GANGA_Update_Thread_Ganga_Worker_0 (most likely raised during interpreter shutdown):
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
  File "/usr/lib64/python2.6/threading.py", line 484, in run
  File "/home/egede/Ganga/install/ganga/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 78, in __worker_thread
  File "/usr/lib64/python2.6/Queue.py", line 177, in get
  File "/usr/lib64/python2.6/threading.py", line 254, in wait
<type 'exceptions.TypeError'>: 'NoneType' object is not callable
Exception in thread GANGA_Update_Thread_Ganga_Worker_1 (most likely raised during interpreter shutdown):
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
  File "/usr/lib64/python2.6/threading.py", line 484, in run
  File "/home/egede/Ganga/install/ganga/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 78, in __worker_thread
  File "/usr/lib64/python2.6/Queue.py", line 177, in get
  File "/usr/lib64/python2.6/threading.py", line 254, in wait
<type 'exceptions.TypeError'>: 'NoneType' object is not callable
Exception in thread GANGA_Update_Thread_Ganga_Worker_2 (most likely raised during interpreter shutdown):
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
[...]

PR #289, opened 2016-03-08
Making sure 'join' is called on all stopped queues before the end

rob-c:
OK, 3rd time lucky, (Apologies for the spam)
I'm new to cross-branch pull-requests and would have preferred to stick this in develop myself and resync the release branch.

This should address #285

If there are questions feel free to ask.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepository.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepository.py
@@ -61,6 +61,8 @@ class RepositoryError(GangaException):
         try:
             from Ganga.Core.InternalServices.Coordinator import disableInternalServices
             disableInternalServices()
+            from Ganga.Core.GangaThread.WorkerThreads import shutdownQueues
+            shutdownQueues()
             logger.error("Shutting Down Repository_runtime")
             from Ganga.Runtime import Repository_runtime
             Repository_runtime.shutdown()
--- a/python/Ganga/Core/GangaThread/WorkerThreads/ThreadPoolQueueMonitor.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/ThreadPoolQueueMonitor.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import collections
 from Ganga.Core.GangaThread.WorkerThreads.WorkerThreadPool import WorkerThreadPool
 from Ganga.Utility.Config import getConfig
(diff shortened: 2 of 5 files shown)


milliams:
This looks overall sensible to me. My only question is why `_stop_all_threads` wasn't already being called from elsewhere? It should be called in both `disableMonitoringService()` as well as by `GangaThreadPool.shutdown()` (though in the latter it's called in another thread for some reason).

rob-c:
No, it should not be being called from the ```disableMonitoringServices``` as this is a strictly unrelated feature to the users queued threads I would argue. If I queue job submission and disable Monitoring I expect the queues to _still_ work.
As to why this isn't called from GangaThreadPool... This has/had issues with daemon threads in the queue system and whether they're registered.
The Queues should be shutdown first before we look into shutting down the rest of the ThreadPool imo as it's possible that the Queues could spawn more threads.

drmarkwslater:
If this fixes the crash then it's a :+1: from me though it sounds like the queues should be looked at in the future.

milliams:
It *is* called by `disableMonitoringService` (Coordinator.py line 169) but I guess that's happening at the wrong point at the moment. Maybe that's not the best idea but it seems to currently be the case. It looks like ti was added by @rob-c in 0539f4cc972112d2473333d6fe1d

rob-c:
@milliams Thanks for the heads up, I think this is clearly a bug. I'll amend the PR

rob-c:
@milliams Ah, this is the wonderfully strange territory. I was thinking we were talking about ```disableMonitoring``` so apologies for the confusion.

```disableMonitoring``` is called on shutdown this with ```enableMonitoring``` currently start and stop the whole monitoring loop.

The ```disableMonitoringServices``` is the 'emergency-stop' called when a server error has occurred within the Registry and nowhere else (I think).
This function used to be called on shutdown but left some things in a strange state. The shutdown manager now no longer calls this due to the fact it's much better at shutting down everything by itself.

The call to stop the threads here was due to if a user has running queues and something bad happened they would get spammed on the screen about not being able to access a disabled repository by all of the queued threads as they try and execute.

**Comment by milliams** [comment]
@rob-c It looks like `disableMonitoringService` and `disableMonitoring` are actually the same thing. In Coordinator.py line 306 we have:
```python
exportToGPI('disableMonitoring', disableMonitoringService, 'Functions')
```

It's called by `disableInternalServices` and this looks like it's only called automatically by the disk space, grid proxy checker or `RegistryError`.

In this case I agree with you that we don't expect this function to be the one to shutdown the queues threads.

The other route by which `_stop_all_threads` is called is via `GangaThreadPool`'s `shutdown`→`_really_shutdown`→launch a thread to call `__do_shutdown__`→`_stop_all_threads`. Should we expect that this route is doing the right thing at the right time?

rob-c:
@milliams Ah so they are the same thing (I missed the wood for the trees).

I'll upload a proposed change in a few min

rob-c:
OK, I've removed all of queues related behavior from ```disableInternalServices``` so that it should be much clearer where the queues are shutdown from. I've also added the ability to shutdown the queues to the RepositoryError which will need this due to the problems I mentioned earlier

milliams:
:+1: This is looking really good. If this is fixing the observed errors on shutdown then I'm happy for it to be merged.

drmarkwslater:
:+1:

PR #290, opened 2016-03-08
Move GangaUnitTest to a lib directory

milliams:
I'm working on a `Dirac` backend test and this is prerequisite for that so that tests in other plugins can access `GangaUnitTest` via `from Ganga.new_tests.lib.GangaUnitTest import GangaUnitTest`.

This only affects the tests and I believe I've (well, PyCharm) caught all the places where the import path needs to change

### DIFF
--- a/python/Ganga/new_tests/GPI/Bugs/TestJIRA1961.py
+++ b/python/Ganga/new_tests/GPI/Bugs/TestJIRA1961.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from ..GangaUnitTest import GangaUnitTest
+from Ganga.new_tests.lib.GangaUnitTest import GangaUnitTest
 
 
 class TestJIRA1961(GangaUnitTest):
--- a/python/Ganga/new_tests/GPI/Bugs/TestSavannah10064.py
+++ b/python/Ganga/new_tests/GPI/Bugs/TestSavannah10064.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from ..GangaUnitTest import GangaUnitTest
+from Ganga.new_tests.lib.GangaUnitTest import GangaUnitTest
 
 
 class TestSavannah10064(GangaUnitTest):
--- a/python/Ganga/new_tests/GPI/Bugs/TestSavannah13404.py
+++ b/python/Ganga/new_tests/GPI/Bugs/TestSavannah13404.py
@@ -1,8 +1,8 @@
 from __future__ import absolute_import
 
-from ..GangaUnitTest import GangaUnitTest
-
 import os
+
+from Ganga.new_tests.lib.GangaUnitTest import GangaUnitTest
 
 
 class TestSavannah13404(GangaUnitTest):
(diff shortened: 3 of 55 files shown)


rob-c:
Is it worth looking at elevating the new_tests out of ```Ganga``` finally?

Maybe move ```Ganga\new_tests``` to ```GangaTesting``` or somewhere similar akin to a top-level Plugin (even though I'm aware it's not currently a top-level plugin (and probably shouldn't be)

rob-c:
I'm wondering why the motivation to move ```GangaUnitTest``` to the ```lib``` folder? This feels like it's not needed imo

milliams:
My logic is that we want things like `GangaUnitTest` to be accessible from anywhere in any Ganga plugin with a simple `import`. Putting these things into a `GangaTesting` plugin means that that plugin must be loaded explicitly before the tests are runnable.

There are two separate things in the test system: 

1. Test framework stuff like `GangaUnitTest`, `run_until_state` etc. I'd like these to be globally accessible so I think in the long term having them in `Ganga.tests.lib` is sensible (we could call it `Ganga.tests.framework` or whatever).
2. The tests themselves. We should make sure that tests for a particular plugin live in that plugin so that they are as near to the relevant code as possible. This means having core tests in `Ganga/tests` (currently `Ganga/new_tests`) and Dirac tests in `GangaDirac/tests`.

In general I don't like putting things in `GangaTest` or similar since I feel it is abusing the plugin system somewhat to just add things to `sys.path`.

As for renaming `Ganga/new_tests` to `Ganga/tests`, I think that will have to wait until we are ready to switch off the legacy tests (unless we can change them to look in `Ganga/old_tests` or similar?)

rob-c:
I don't mean to make a plugin, (in fact I think a lot of us want to see ```GangaTest``` go away with the old tests asap).

I'm of the opinion that ```Ganga\new_tests``` was a temporary name/place for the tests (hence why I liked using the relative import paths).
I'm a little put off by now hard-coding this into all of the tests now as it will mean we will probably stay with this directory structure in the future. (if it's not broken don't touch it...)

The old test system doesn't strictly rely on the tests being in ````Ganga\test``` to run, this is an argument passed in a script.

I'm tempted to suggest that we now make the following change:

```Ganga\test``` -> ```Ganga\old_tests```
```Ganga\new_tests``` -> ```Ganga\tests```

I'd be a bit happier to hard-code ```Ganga.test``` into the newer Jenkins tests

I wasn't sure what the plan for the lib folder was. If we're collecting static external functions pulled in by testing then I very much approve of this and yes it probably makes sense to move the baseclass here.

egede:
@milliams I do not like that the framework code for the testing (what you have in lib) resides inside the same directory as the tests for the core package. I would much prefer that the Ganga*/tests have the identical structure in all packages. I siggest to put the framework stuff in some other directory (Ganga/testframework or whatever).

milliams:
@rob-c Sorry I wasn't clear. It sounds like we're mostly on the same page and I agree that probably before something like this branch gets merged we should make the `new_tests`→`tests` change you suggest.

@egede I've been thinking the same thing. Given that I will likely close and remake this PR anyway I will likely do as you suggest and put the framework code in `Ganga/testlib` or `Ganga/testframework`

milliams:
Moving the tests is open in #291 and I will close this PR and open another one with a more sensible directory layout.

PR #291, opened 2016-03-09
Move legacy tests to `old_tests`

milliams:
As discussed in #290, this moves all the old tests to an `old_tests` directory and moves all the new tests to a `test` directory.

I think I've caught all the references to the test directory name in the scripts.


milliams:
Is this still something that we want to do? I feel like doing this at the beginning of a release cycle makes the most sense since it's most likely to conflict with other PRs.

drmarkwslater:
Sounds good to me - as an aside: Should I/we delete tests from legacy that have been converted to Jenkins?

milliams:
@drmarkwslater Yes, in the PRs that I've done so far for this I've been deleting the old tests in the same commit that I add the new ones. You'll see between [6.1.10](http://ganga.web.cern.ch/ganga/release/6.1.10/reports/latest/html/index.html) and [6.1.18](http://ganga.web.cern.ch/ganga/release/6.1.18-pre/reports/latest/html/index.html) there's been a decrease in the total number of tests.

rob-c:
OK, are we planning an equivalent PR to move the ```new_tests``` to ```tests``` before the start of the 6.1.19 release?

milliams:
Sorry, the title wasn't exhaustive. This PR does indeed also copy the `new_tests` folder to its new location at the same time.

rob-c:
In that case can we have the move of the new_tests to tests either wait until just before the testing starts or until all other test related PR are in? I think there are a few outstanding new/modified tests against 6.1.19 already.
If you want to get this in now for technical reasons go for that but it's going to mean fixing conflicts in a few other branches I suspect if we do.

milliams:
Indeed, that will be a problem. Unfortunately, it will be a problem either way round we do it. Either we:

1. merge this branch now and then have to fix each 6.1.19 PR individually
2. or we merge those 6.1.19 PRs and then have to edit this PR to accommodate those changes. We will then still have to fix a number of >=6.1.20 PRs anyway

It's kind of lose-lose but if I had to choose I would go for 1.

rob-c:
Given there are few PR for 6.1.20 which have been written/finalized I think they accept that there may be breaks in 6.1.19 development. Many smaller PR in 6.1.19 have already reached a point of 'about to be merged' and so I'd say this change is a bit more invasive than it need be.

milliams:
This is the last PR before 6.1.19. I am planning on updating it and merging soon. Are there any objections/changes etc.?

rob-c:
go from me as soon as it works

PR #292, opened 2016-03-09
Don't cache the grid proxy command set

milliams:
This allows the proxy type to be changed during a session. For example when running tests, it is necessary to change proxy type per-test. There is still some higher-level caching so this shouldn't affect performance too much.

Obviously this is only a short-term thing which will be replaced with the new credentials system but in the meantime it will be needed for running Dirac tests etc.


python/Ganga/GPIDev/Credentials/GridProxy.py
```diff
@@ -206,12 +204,10 @@ def chooseCommandSet(self):
         Return value: None
         """
         from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
-        if "ICommandSet" == self.command._name or isType(self.command, GangaList):
-            if self.voms:
-                self.command = self.vomsCommand
-            else:
-                self.command = self.gridCommand
-        return None
+        if self.voms:
```
rob-c: Take care with this one. I touched this line once and we spent a week arguing over it and decided to never touch this file again until the credentials branch is ready.

Are we sure we never want to cache this? it somehow feels like a waste.

rob-c:
In this case can you add some logic to make sure this is removed and re-added to the GPI if applicable as the remaining objects that aren't removed on the shutdown of a test and re-created are the credentials objects.
If it's safe to persist these then ok, but I'm a little weary of this.

python/Ganga/GPIDev/Credentials/GridProxy.py
milliams: The problem at the moment is that it's cached the first time the object is created and then it can never be changed again. The problem here is that we have one interface (`GridProxy`) which can act in any number of ways depending on the config variables when Ganga was started. Admittedly the change here is only needed for running the tests but I feel that caching this is unnecessarily restrictive.  

Information about the path, voname and info() are all cached still.

milliams:
Ok, I will look into that. It's possible that this will solve the problem I was seeing in my upcoming tests anyway in which case this PR will not be needed.

python/Ganga/GPIDev/Credentials/GridProxy.py
egede: Are tests of different top level packages (Ganga,  GangaDirac, ...) not 
done in different processes? If that is the case, I fail to see why this is 
required. In the old tests we had some backend matrix test that I do not 
see any reason to migrate. Better to have the tests locally with they backends.

milliams: @egede Currently we only have 'new' tests for `Ganga` not any other modules. I wasn't planning on necessarily running each plugin's tests in a separate process but it would be possible. The problem that this PR is attempting to fix is that some things in Ganga read the config on initial startup (i.e. config variables in `defaults_SomeClass`-type sections) and never read them again, even if they are edited by the user at runtime. If we want to change these on a test-by-test basis then they need to be able to re-read the config if there's no other mechanism to override them.

The problem in the (soon-to-be incoming) Dirac test isn't that it's being polluted by values set in previous tests but rather that without editing the config file and restarting ganga, there's no way to change what command `gridProxy.create()` calls.

Indeed I want the Dirac tests to be in the GangaDirac folder etc.

rob-c: @milliams Isn't it possible to have a .gangarc which is used in the tests for GangaDirac etc such that this is the config known to work on this machine?
It should be possible to add an argument to the ```GangaUnitTest``` to use the config file from wherever we put it, so we could potentially just wget a file from github at the start of the test and tell GangaUnitTest to use that config for that test surely?

milliams: That would be technically possible but this PR is also fixing other problems in that it's currently impossible to change VO at runtime. This PR allows a script to set which VO it is working with which is useful even outside of the tests and makes the function `chooseCommandSet` actually do what it is supposed to.

There'll be no need for the tests to do any wgetting, we will simply provide the config file in a known location.

Issue #293, opened 2016-03-09
Aggressive locking causing some lockups in ganga


rob-c:
Hi all,

I suspect what I'm seeing is the result of slightly aggressively locking objects.

I have a session open where a job with ~100 subjobs is being monitored and typing ```myJob.subjobs``` caused a lockup.

I suspect this is likely due to being slightly aggressive in how we're locking objects.

I don't think this was a flush to disk keeping the object locked as it lasted ~5min before I manually exited, cleaned up and restarted ganga and in that time the monitoring loop had failed to progress to the next subjob it should have been updating.
(Edit python was also using <1% CPU throughout this 5min on an lxplus node with load < 3)

I have an idea of how to try and avoid this a bit more in the future and it only requires some light modifications to the current locking code.

I think that in the few remaining cases where the user can cause a lockup from the prompt the advice should be wait 60sec and then to force close and restart for the time being.

Regards,

Rob

rob-c:
This seems to be readily reproducible for me, I have 2  jobs with ~100 subjobs all on the Dirac backend and if I sit at the interactive prompt typing ```jobs``` and ```jobs(1227).subjobs``` eventually Ganga seems to lockup on one of the subjobs calls.
This could be a subjob related bug so I'll see if I can make any progress in this.
I'm going to try leaving a monitoring session open without issuing commands at the interactive prompt to rule out that ganga is still locking up by itself.

rob-c:
OK, leaving Ganga alone for 10min causes it to hard-lockup. The session locks are still being updated but I can't run ```jobs``` and the flushing of objects stopped about 10min ago suggesting no activity is taking place behind the scenes.

milliams:
Do you have a stacktrace of all the threads for this?

rob-c:
Not yet, do you have a usefully modified form of this or should I use the vanilla tool?

I'm going to (not in order):
 1. Try a fix I have in mind for some possible lockups
 2. Try to re-create this with a test submitting many subjobs to the Local backend
 3. Try and get an stacktrace of the threads
 4. Try to turn off the Registry auto-flushing and try again.

milliams:
I've been using https://gist.github.com/milliams/18f442009db59e92d8a2 which is a modified version to tidy and enhance the syntax highlighting.

rob-c:
OK, I think I may have a fix, give me about 30min to check a few things then I'll put up a PR for testing

rob-c:
I think this has been fixed

PR #294, opened 2016-03-09
Making Registry and __set__ excluively lock and __set__ and __get__ lock but not registry and __get__

rob-c:
This PR makes the Descriptor  ```__set__``` and registry access exclusively lock to the same thread and have the ```__get__``` and ```__set__``` of the Root of an object exclusively lock on a thread without requiring that the Registry and Descriptor ```__get__``` methods exclusively lock access.

This appears to alleviate the problems experienced when submitting man LHCbDirac subjobs that causes Ganga to lockup after a few minutes of use.

I'm keen to see how this performs on the Jenkins system but this alleviates problems on lxplus for LHCb.

I can't motivate a strong reason why the Registry and Descriptor ```__get__``` methods should mutually lock the root of an object.

If there are strong objections then another solution to #293 is going to have to be found ahead of 6.1.17

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -41,7 +41,7 @@ logger = Ganga.Utility.logging.getLogger(modulename=1)
 
 _imported_GangaList = None
 
-do_not_copy = ['_index_cache', '_parent', '_registry', '_data', '_lock', '_reg_lock', '_proxyObject']
+do_not_copy = ['_index_cache', '_parent', '_registry', '_data', '_read_lock', '_write_lock', '_proxyObject']
 
 def _getGangaList():
     global _imported_GangaList
@@ -58,7 +58,7 @@ def synchronised(f):
     """
     @functools.wraps(f)
     def decorated(self, *args, **kwargs):
-        with self.lock:
+        with self._internal_lock:
             return f(self, *args, **kwargs)
     return decorated
 
@@ -70,8 +70,8 @@ class Node(object):
         self._data = {}
         self._parent = parent
         self._index_cache = {}
-        self._lock = threading.RLock()
-        self._reg_lock = threading.RLock()
+        self._read_lock = threading.RLock() # Don't read out of thread whilst we're making a change
+        self._write_lock = threading.RLock() # Don't write from out of thread when modifying an object
         super(Node, self).__init__()
         #logger.info("Node __init__")
 
@@ -134,32 +134,38 @@ class Node(object):
         if parent is None:
             setattr(self, '_parent', parent)
         else:
-            with parent.lock:  # This will lock the _new_ root object
-                setattr(self, '_parent', parent)
+            with parent.const_lock:
+                with parent._internal_lock:  # This will lock the _new_ root object
+                    setattr(self, '_parent', parent)
             # Finally the new and then old root objects will be unlocked
 
     @property
     @contextmanager
-    def lock(self):
+    def _internal_lock(self):
         """
-        This is a context manager which acquires the lock on the
+        This is a context manager which acquires the internal read lock on the
         object's root object.
         """
         root = self._getRoot()
-        root._lock.acquire()
+        root._read_lock.acquire()
         try:
             yield
         finally:
-            root._lock.release()
+            root._read_lock.release()
 
     @property
     @contextmanager
-    def reg_lock(self):
-        self._reg_lock.acquire()
+    def const_lock(self):
+        """
+        This is a context manager which acquires the const write lock on the
+        object's root object.
+        """
+        root = self._getRoot()
+        root._write_lock.acquire()
         try:
             yield
         finally:
-            self._reg_lock.release()
+            root._write_lock.release()
 
     # get the root of the object tree
     # if parent does not exist then the root is the 'self' object
@@ -384,21 +390,31 @@ class Node(object):
 ##########################################################################
 
 
-def synchronised_descriptor(get_or_set):
+def synchronised_get_descriptor(get_function):
     """
-    This decorator should only be used on ``__get__`` or ``__set__`` methods of a ``Descriptor``.
+    This decorator should only be used on ``__get__`` method of the ``Descriptor``.
     """
-    @functools.wraps(get_or_set)
+    @functools.wraps(get_function)
     def decorated(self, obj, type_or_value):
         if obj is None:
-            return get_or_set(self, obj, type_or_value)
-        if get_or_set.__name__ == "__get__":
-            with obj.lock:
-                return get_or_set(self, obj, type_or_value)
-        elif get_or_set.__name__ == "__set__":
-            with obj._getRoot().reg_lock:
-                with obj.lock:
-                    return get_or_set(self, obj, type_or_value)
+            return get_function(self, obj, type_or_value)
+
+        with obj._internal_lock:
+            return get_function(self, obj, type_or_value)
+
+    return decorated
+
+def synchronised_set_descriptor(set_function):
+    """
+    This decorator should only be used on ``__set__`` method of the ``Descriptor``.
+    """
+    def decorated(self, obj, type_or_value):
+        if obj is None:
+            return set_function(self, obj, type_or_value)
+
+        with obj.const_lock:
+            with obj._internal_lock:
+                return set_function(self, obj, type_or_value)
     return decorated
 
 
@@ -425,7 +441,7 @@ class Descriptor(object):
         if self._getter_name:
             raise AttributeError('cannot modify or delete "%s" property (declared as "getter")' % _getName(self))
 
-    @synchronised_descriptor
+    @synchronised_get_descriptor
     def __get__(self, obj, cls):
         name = _getName(self)
 
@@ -537,7 +553,7 @@ class Descriptor(object):
 
         return v_copy
 
-    @synchronised_descriptor
+    @synchronised_set_descriptor
     def __set__(self, _obj, _val):
         ## self: attribute being changed or Ganga.GPIDev.Base.Objects.Descriptor in which case _getName(self) gives the name of the attribute being changed
         ## _obj: parent class which 'owns' the attribute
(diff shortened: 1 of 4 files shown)


rob-c:
@milliams I've no idea how ```TestThreadSafeGangaObject``` is supposed to work would you be able to look at how/why this is failing in this PR?

rob-c:
Can we lighten the amount of stuff in ```TestGangaObject.py``` there are a load of very complex unittests here to do with thread safety that imo would be better placed in a separate file testing the thread-safeness of the objects rather than alongside the basic tests of 'does everything run without crashing'.

milliams:
Mark and I were just discussing a solution to the deadlocks you saw and were honing on something similar to this PR. I think this is in general the correct approach and I like its direction. My main thoughts are around clarity and naming.

> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

Really, what the registry wants to acquire when flushing is a 'const lock'; that is a lock which prevents anyone *changing* anything in the schema for the duration of the lock but allows others to *read* it. This is basically your `reg_lock`. The other sort of lock that could exist in the system would be an 'exclusive lock' which would prevent any *read*s or *write*.

My suggestions are:

1. Make `reg_lock` acquire the *root* object's `_reg_lock` for consistency (also edit the `__name__ == "__set__"` part to use this lock directly).
2. Rename `reg_lock` to `const_lock` to signify its intent and document it.
3. Rename `lock` to `_exclusive_lock` since that is what it does; stop anyone doing anything. Also with prepended `_` to signify it's internal.
4. Rename the two `_lock`→`_reading_lock` and `_reg_lock`→`_writing_lock`.
5. Change the tests to acquire `const_lock` instead.

I think the documentation for `const_lock` should be something like: "This lock acquires exclusive access over an object tree to prevent it changing. Reading schema attributes on the object is still allowed but changing them is not. Only one thread can hold this lock at once."

milliams:
As for lightening `TestGangaObject.py` we can do, yes but for now it's not a priority. A chunk of the code in there will be going to `Ganga.testlib` once that exists and at that point we will be able to slim down the amount of code in here.

rob-c:
This seems to be similar to the approach I was attempting to take in my first draft. Good to know there'a a formal methodology around this.

I'll rename the parameters to follow this convention. As for lightening the ```TestGangaObject``` these tests are too complex for me to debug and failing atm so it's either we ditch them for the release or fix them here. I'm opting for breaking them out into a new file in a new PR if we don't fix them here.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -151,6 +152,15 @@ def lock(self):
...
 
+    @property
+    @contextmanager
+    def reg_lock(self):
+        self._reg_lock.acquire()
+        try:
+            yield
+        finally:
+            self._reg_lock.release()
+
```
alexanderrichards: This contextmanager is not necessary as the threading.RLock will already work as a context:
[threading locks as contexts](https://docs.python.org/2/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement)

this merely needs to return the lock, i.e.

```python
@property
def reg_lock(self):
    return self._reg_lock
```

rob-c:
As for the ```__name__ == "__set__"``` aside from having different wrappers for the get and set I've no easy way to do this ```get_or_set.__name__ is self.__set__``` fails so I'll have to do object inspection which starts to get a bit heavy handed.

python/Ganga/GPIDev/Base/Objects.py
rob-c: I'm modifying this to acquire the root anyway so this will stay but change

rob-c:
For the sake for full disclosure although the test have passed running the test #295 on this PR is known to cause issues. However I'd prefer to issues relating to that test there.

milliams:
I checked out 0ad281b and made the following changes locally:

- Removed `with parent._internal_lock` from `_setParent`
- Changed `synchronised()` in Objects.py to use the `const_lock`
- Made the tests in `TestThreadSafeGangaObject` use the `const_lock` as I said in point 5 above.

And the tests run perfectly well.

Deleting a failing test is absolutely not what we want to do. The test exists for a reason. We should fix tests, not ignore them.

milliams:
This was my first time using `git rebase -i` (I followed [this excellent tutorial](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)) so forgive me if I've messed something up. We'll see how the tests run.

rob-c:
@milliams I'm seeing real problems with changing the ```_setParent``` lock in this way.

Using the const lock means I see more deadlocks. I can't resubmit a subjob when the monitoring is running no for instance.

The const lock isn't needed here is it, it just maintains that parent is consistent which we don't strictly need to hang something off it with inheritance do we?
Adding a read-lock means that we don't hang a new object off a parent when the parent is in a state of flux which probably prevents (some) problems.

wrt to the test, I wasn't wanting to remove the test, however fixing the test looked to be a slightly overwhelming task given its several hundred lines long this was a case of the test having issues.

rob-c:
If a job is locked because we're modifying an application then we still want to be able to know that an object (eg a inputfile) is assigned as a child of the parent and it's only _later_ when we want to modify the child object that we should in future have to wait until the new top level parent is locked.

milliams:
`_setParent` should not be being called at all by anything during either resubmit or the monitoring. Unless this is related to the over-zealous reparenting @drmarkwslater found in `SubJobXMLList`? I used a `const_lock` here so that even while an object is being reparented, other threads can read information about the object. Using an `_internal_lock` will prevent any reads or writes.

Potentially, in the short term we could remove the locks from `_setParent` entirely? It's not strictly 'safe' but as we discussed there's not much in the way of parent-changing happening anyway.

rob-c:
I can't even run ```jobs``` atm without modifying this so it's not just this I fear

rob-c:
The parent has no knowledge of it's children through inheritance so I don't think the locking within setParent applies in the same way it does elsewhere.
It's only when attempting to modify the attributes (or prevent this from happening) that the locks should ever be used so I'm getting more of the opinion that we should drop the locks from this method.

milliams:
Before we try dropping that lock could you try, in your local checkout, editing `SubJobXMLList` on lines 371 and 372 to indent them one step so that they are inside that `if not index...` block? I suspect that at least partly it's the reparenting being called too often that it causing some issue.

I will have a think about the `_setParent` lock...

rob-c:
I've tried this with no luck. I added a test for is ```_getParent() is not self._definedParent``` before attribute assignment and it doesn't appear to alleviate the deadlock at all.

Before changing ```_setParent``` I see memory usage at 280Mb and after I see it at 350Mb indicating some potential problem with loading of objects caused by this but it doesn't appear to be the assignment in SJXMLList, I'll check again now by reverting to the head of the branch and changing SJXMLList again

In the future do we want to consider advertising children to parents?
This would at least allow for a top-down query of ```getChildren``` or equivalent within an object which would be very nice to have for debugging at least.

milliams:
`getChildren` should be equivalent to the items in a objects schema. At any time getting any object B from an object A's schema and then asking B for its parent should always return 'A'. The slight exception to this is sequences where you have to look inside the sequence and ask for a parent for it to point to A.

rob-c:
Yes I expect it should be the same but it would allow for a cross-check to see that everything has been assigned directly rather than running around in future looking for problems with ```_setParent``` imo.

I've a fix incoming for the SJXMLClass which appears to fix these new lockups.

milliams:
This may be a stupid question but can I just check what `git rev-parse HEAD` returns for you? I've been submitting `Local` jobs with hundreds of subjobs which sleep for a minute each and I have been unable to reproduce the lockups. @drmarkwslater is trying a similar thing now as well.

rob-c:
Rats, I'd committed before I could run that, either way I suspect I was running on https://github.com/ganga-devs/ganga/commit/17fc5fe777e596dd3ed61736e098f9458458f565 from memory.
The minor changes are just to reduce the amount of setParent calls in SJXMLList which I think fixes the problems from there as I can't get it to hang now.

(It's difficult to diagnose hard-lockups with something that will recover when running on lxplus as other factors can make ganga begin to run very slowly.)

rob-c:
@milliams is it a problem that this is a PR against develop and that there are changes already in release/6.1.17 ?

milliams:
Don't worry about the `rev-parse` thing. It was just to check that you weren't still on your old detached-head from before I did the rebase. Given that your commit 30af618 is parented correctly I think it was all fine.

Your changes to `SubJobXMLList` look sensible, especially if they are reducing the number of calls to `_setParent`. @drmarkwslater Do they make sense given what you've been seeing in `SubJobXMLList`?

milliams:
Ok, so this PR is working for me and is passing the tests. Are there any outstanding issues with it?

rob-c:
I'm not aware of any issues. Shall I go for merge?

milliams:
:+1:

rob-c:
Here's looking forward to 6.1.17 :D

milliams:
I'll start the legacy tests again so let's hope for no surprises!

alexanderrichards:
> This was my first time using git rebase -i (I followed this excellent tutorial) so forgive me if I've messed something up. We'll see how the tests run.

FYI a quick glance at the tutorial that you post here seems to exclude the fixup and reword options. In the tutorial you post I see:
```
# Commands:
#  p, pick = use commit
#  e, edit = use commit, but stop for amending
#  s, squash = use commit, but meld into previous commit
```

There are two more commands (I guess depending on your git version):

```
#  r, reword = use commit, but edit the commit message
# f, fixup = like "squash", but discard this commit's log message
```

These are actually quite useful for rewording the commit message or just ignoring it if it's a simple "more fixes" type message. Arguably fixup is more useful than squash as you don't always want to keep every commit message.

milliams:
@alexanderrichards Indeed, I've since been making use of `fixup` a lot in my local branches. `git rebase -i` is really nice and I'm glad you've convinced me to start using it :smile:

PR #295, opened 2016-03-09
Adding thrading test into the Unit test folder

rob-c:
I don't know if this unittest is valid or broken but I've broken it out from the ```TestGangaObject.py``` so that it can be looked at.

I don't know if people feel that it should strongly go in for 6.1.17 or not.


rob-c:
Superseded by work back in #294

PR #296, opened 2016-03-10
Modifying get/set cache to be more reliable

rob-c:
Minimal implementation of #231 on top of #294 . I'm wanting to test on the Jenkins server.


rob-c:
I think this PR made it into develop (mostly) in a few later bits of work, it can be removed.

PR #298, opened 2016-03-10
Simplify Grid

milliams:
This is a clean-up of the code in `Grid.py` to change it from being a stateful class being passed around into a module of free functions which work independently of each other. Historically, the main reason for this being a stateful class was they a user might want to have a `Grid` object for 'EDG'  and one for 'GLITE'. This is now not needed as EDG is no more (since lots of years).

Of course this necessitates a number of changes across the board but overall it is removing code.

The reason I am making these changes now is that in my [credentials branch](https://github.com/ganga-devs/ganga/compare/feature/credentials) I ended having to make a number of these changes in order to have proxy prompts happen at the right times. My idea here is to start breaking out some of the larger changes in that branch which might not be directly related to the new credential system to make the whole review process easier.

While this is a large and scary PR overall, the changes are all documented in the commit messages along the way so hopefully you can follow it.

### DIFF
--- a/templates/CONFIG_TEMPLATE.INI
+++ b/templates/CONFIG_TEMPLATE.INI
@@ -341,10 +341,7 @@
 
 #  sets the generic LCG-UI configuration script for the GLITE workload
 #  management system
-#Config = 
-
-#  sets the VO-specific LCG-UI configuration script for the EDG resource broker
-#ConfigVO = 
+#Config =
 
 #  sets the baseURI for getting the input sandboxes for the job
 #CreamInputSandboxBaseURI = 
@@ -360,13 +357,7 @@
 
 #  sets the space token for storing temporary files (e.g. oversized input
 #  sandbox)
-#DefaultSRMToken = 
-
-#  enables/disables the support of the EDG middleware
-#EDG_ENABLE = True
-
-#  sets the LCG-UI environment setup script for the EDG middleware
-#EDG_SETUP = /afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh
+#DefaultSRMToken =
 
 #  sets excluded computing elements by a regular expression
 #ExcludedCEs = 
--- a/python/Ganga/GPIDev/Credentials/AfsToken.py
+++ b/python/Ganga/GPIDev/Credentials/AfsToken.py
@@ -144,7 +144,7 @@ class AfsToken (ICredential):
     _hidden = 1
     _enable_config = 1
 
-    def __init__(self, middleware=""):
+    def __init__(self):
         super(AfsToken, self).__init__()
         from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
         if "ICommandSet" == self.command._name or isType(self.command, GangaList):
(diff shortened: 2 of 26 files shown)


milliams:
Since a lot of this PR is whitespace changes, you can simplify viewing them by [appending `?w=1` to the end of the compare URL](https://github.com/ganga-devs/ganga/pull/298/files?w=1). You can't make comments in that view though. Also, a lot of the changes are in `Grid.py` which has changed too much for GitGub to see so you will have to run it locally with
```bash
git diff develop..feature/credentials -w python/Ganga/Lib/LCG/Grid.py
```

milliams:
Ok, I've updated this branch to the latest `develop` and on running some local tests it seems to be working well. I'd like to merge this soon as it's been waiting for two months now.

Comments?

drmarkwslater:
I can't claim I've gone over everything here but it certainly looks OK 👍 

One thing I would say: would it be possible to add a couple of LCG tests with this just to be sure or is that not possible at the moment?

milliams:
I've had a go at setting up LCG tests for this purpose in the past but getting a working LCG environment is rather tricky. In the end I decided to focus on the DIRAC tests rather than LCG. I'd welcome help in this direction if anyone has an easy setup script for getting from base SL6+CVMFS to a working submission and monitoring.

rob-c:
I don't use the LCG stuff but as long as DIRAC is unaffected by touching the credentials code I say go for merge

milliams:
Ok, so I've [tested this](https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/40/) against the recently added LCG tests and it's working well. In my local tests against CREAM it works well too (at least for the simple tasks of submitting, monitoring, finalising and killing a job). I'm happy to merge this in its current state. Any other comments?

templates/CONFIG_TEMPLATE.INI
```diff
@@ -360,13 +357,7 @@
...
 #  sets the space token for storing temporary files (e.g. oversized input
 #  sandbox)
-#DefaultSRMToken = 
-
-#  enables/disables the support of the EDG middleware
-#EDG_ENABLE = True
-
-#  sets the LCG-UI environment setup script for the EDG middleware
-#EDG_SETUP = /afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh
+#DefaultSRMToken =
```
alexanderrichards: Minor point not about the changes, actually more like a comment...I'm pretty sure that we don't use these templates any more as the gangarc is generated from the config object itself. Do we want to be keeping them up to date or should we have a separate issue for removing them.

alexanderrichards:
Looks ok from the glance I had :+1:

templates/CONFIG_TEMPLATE.INI
milliams: I would be in favour of removing them in another PR. There's a bunch of related scripts in `release/tools` which could also just be burned now. It would probably be quite cathartic to whoever wants to give that a go.

alexanderrichards: :wink:

alexanderrichards: In that case there is no hard updating them here and we can burn them in a separate PR

rob-c:
Well if we're holding off on this I'll revert the recent changes and keep with the broken prompt.
I wonder if I can hook into this through my .ipythonrc and .gangarc to get a fix from my local configs, this is quite an annoying bug when trying to test, debug and work with lots of subjobs

milliams:
@rob-c I assume you posted that in this PR by mistake?

PR #299, opened 2016-03-10
Cleanup start up

rob-c:
This PR moves a lot of code out of Runtime.bootstrap.

A lot of it was mixed up, messy and confusing.

I've made a few basic functions to collect together common tasks and this has the advantage of reducing ```ganga/__init__.py``` significantly as well as making the code a bit more readable.

I plan to add some code to load the config in ```ganga/__init__.py``` in this branch before merging but I'm curious about any feedback people have on this?


milliams:
I certainly like the idea behind it. However, I find the whole `additional_interface` API a bit messy. I think that we should be planning for a near-future where we have only `ganga` and not `Ganga.GPI`. In the interim, I'm happy if we just hard-code the two of these in `exportToGPI` (and for simplicity add a `removeFromGPI` for clean-up perhaps). As it is, adding in the logic for `additional_interface`, only to remove it in the near-future is probably not necessary.

rob-c:
Is the plan to populate both ```Ganga.GPI``` and ```ganga``` interfaces when performing ```import ganga```?
(I suspect not but I'm copying what was there for now)

I'm happy to make the change in this branch to have something like ```exportToInterface``` and by default point this to ```Ganga.GPI``` unless we overload it.
I could keep ```exportToGPI``` then as a wrapper around this.

The ```removeFromGPI``` is not really needed as you'd only ever run a delattr against the GPI. I could add it but every time I do it feels like several lines of overkill...

milliams:
It's not intended that importing `ganga` will populate `Ganga.GPI` but I don't think it's a problem if it does so in the short-term.

As an aside: If you're inside `ganga/__init__.py` and you define a function then there's no need to then also *export* it to `ganga` since it's already there.

rob-c:
OK, moved to only populate ```ganga``` when doing ```import ganga```. I'm now using ```exportToInterface``` which does a similar thing to ```exportToGPI``` with the latter now wrapping to the former for everywhere else it's used in the codebase.

This bring ```ganga/__init__.py``` down to about 80 lines and makes it much easier to read :)

I'm hoping to open a 2nd PR to address the loading of the config into ```ganga``` once we merge the fixes for resolving the directories of all of the .ini files.

Hopefully this should only add a few lines of code to ```ganga```.

rob-c:
It would probably make sense to move the outstanding ```exportToGPI``` calls to somewhere within the bootstrap implicitly as these are not duplicating the correct objects for ```ganga``` now and should probably be addressed in this PR.

rob-c:
I've just addressed some issues with 'ganga' not making use the proxy correctly if Ganga.GPI isn't populated.
This is encouraging as I can load my 'ganga' repo correctly and create/submit a simple job with ```ganga\__init__.py``` now being ~ 70 lines of code :)

rob-c:
Interesting this PR is also locking up on Jenkins yet I cannot reproduce the problems locally. I'll try and re-run the whole set of tests to see if this is caused by something bleeding between tests still.

milliams:
@rob-c I've seen a few PR test runs hanging recently. I think this might me due to stale files in the registry from previously-stalled test runs. I've now added a `rm -rf ~/gangadir_testing` to the beginning of every test run so we'll see how that affects things.

@ganga-ci retest this please

rob-c:
This PR is causing problems with the monitoring which need to be understood. I'm looking into this when I get a chance but this PR by itself isn't effecting how the services are running so 'shouldn't' be causing the lockups, this may be fixed elsewhere

rob-c:
OK, this simplification of ```ganga``` and ```Runtime\bootstrap``` now works. I've had to slightly change the Registries to behave a little more sanely hence the new test ahead of running the monitoring which now tests that the objects are what are expected from a few different function calls.

rob-c:
I'm keen to get this into 6.1.19 if possible as it refactors a lot of the code to do with startup and passes all of the tests, how do others feel about this PR?

python/Ganga/Core/GangaRepository/__init__.py
```diff
@@ -20,3 +20,10 @@ def getRegistries():
 
 def getRegistry(name):
     return allRegistries[name]
+
+def getRegistrySlice(name):
+    return allRegistries[name].getSlice()
+
+def getRegistryProxy(name):
+    return allRegistries[name].getProxy()
+
```
milliams: Is it worth adding these functions if they're only used for testing? We should aim to keep the API as simple as possible.

python/Ganga/Runtime/GPIexport.py
```diff
@@ -18,6 +18,31 @@
...
+def _addToInterface(interface, name, _object):
+
+    if isType(_object, GangaObject):
+        setattr(interface, name, addProxy(_object))
+    elif isclass(_object) and issubclass(_object, GangaObject):
+        setattr(interface, name, getProxyClass(_object))
+    else:
+        setattr(interface, name, _object)
+
+def exportToInterface(myInterface, name, _object, doc_section, docstring=None):
```
milliams: Do we need this `myInterface` option here? Why don't we just always export to both `ganga` and `Ganga.GPI`?

python/Ganga/new_tests/GPI/Monitoring/TestMonitoring.py
```diff
@@ -23,10 +23,26 @@ def setUp(self):
 
     def tearDown(self):
         from Ganga.Utility.Config import getConfig
-        #getConfig('PollThread').getOption('autostart').revertToDefault()
-        #getConfig('PollThread').getOption('base_poll_rate').revertToDefault()
         super(TestMonitoring, self).tearDown()
 
+    def teat_0_testRegistryAsserts(self):
```
milliams: typo in `teat`

python/Ganga/Core/GangaRepository/__init__.py
rob-c: I prefer having ```getRegistryProxy('jobs')``` to ```from Ganga.GPI import jobs``` as it allows us to drop some bad uses of ```GPI.jobs``` in places like Tasks in future. I know there are other solutions but this is hopefully a drop-in replacement.
Yes the programmer could just use ```getRegistry('...').blah()``` but this starts to get tiresome if it occurs multiple times in some section of code imo.

Maybe an argument could be made that the un-proxied slice ```getRegistrySlice``` isn't used in too many places and you could use a stripProxy with the Proxied method but I'd prefer to keep it.

python/Ganga/new_tests/GPI/Monitoring/TestMonitoring.py
```diff
@@ -23,10 +23,26 @@ def setUp(self):
...
+        assert getRegistryProxy('prep') is prep
+
+        assert getRegistrySlice('jobs') is stripProxy(jobs)
+        assert getRegistrySlice('box') is stripProxy(box)
+        assert getRegistrySlice('tasks') is stripProxy(tasks)
+        assert getRegistrySlice('prep') is stripProxy(prep)
+
+        assert getRegistry('jobs') is stripProxy(jobs).objects
+        assert getRegistry('tasks') is stripProxy(tasks).objects
+        assert getRegistry('box') is stripProxy(box).objects
```
milliams: Should these be in `TestMonitoring`? Would a `TestGPIExports` or similar make more sense?

python/Ganga/Runtime/GPIexport.py
rob-c: I moved to this explicitly after the conversation where we agreed we should only ever export to one or the other not both...

python/Ganga/new_tests/GPI/Monitoring/TestMonitoring.py
rob-c: Maybe/possibly I'm not so sure at the moment. I'm debating this but the problem I was having with the Monitoring not working was that some of these asserts would fail. I can branch out into another test if needed it's only a small check.

python/Ganga/Runtime/GPIexport.py
milliams: I said
> "It's not intended that importing `ganga` will populate `Ganga.GPI` but I don't think it's a problem if it does so in the short-term.

to say that in the future, everything should be `ganga` but for now it's not a problem if both are populated.

Once we switch to just `ganga` we will just have to go through and clean up this code.

rob-c: OK. To set in stone in this case are we populating both ```Ganga.GPI``` and ```ganga``` for the moment?

This does NOT expose all of the problems that __will__ come from 'turning ```Ganga.GPI``` off when we're ready to transition'.
There are a lot of subtle places (most of which I think I've fixed) where code is relying on old ```Ganga.GPI``` calls.

Why does the code need to be changed once we're using ```ganga``` the fact you can pass an interface isn't such a bad thing really is it?

milliams: I'm happy to keep the code as you've written it. About problems from people using `import Ganga.GPI` inside the codebase, those are things that must be fixed as they are violating the proxy layer. They should be accessed through a global variable or through the plugin system.

About the extra argument to this function: My reason for wanting to remove it (in the future) is that every single function, argument, class or module that you add increases the maintenance cost of the system. We must be thoughtful and careful about what we add and how it works. This is a classic example of technical debt.

rob-c: There are many things 'in-front' of the proxy which aren't available from the Plugin system ```jobs``` is just one of these.
There are also situations where you don't a-prioiri know what you're looking for other than it's in front of the proxy.

milliams: You can get `jobs` with:
```python
from Ganga.Core.GangaRepository import getRegistry
jobs = getRegistry('jobs')
```
and for the rest of them we *can* know. Perhaps this is something which we will have to tackle on a case-by-case basis but it's certainly possible (even if it requires changes to the bootstrap like `queues` did) and it's also correct.

rob-c: No the example you give should be:
```
from Ganga.Core.GangaRepository import getRegistry
imported_jobs = getRegistry('jobs').getProxy()
assert jobs is imported_jobs
```
(This ```assert``` will only work on this PR as the proxy gets regenerated on every call before the changes I made here!)


The ```jobs``` exposed to the user at Runtime is NOT a registry object as you'd naively expect it's a ```JobRegistrySliceProxy``` around a ```JobRegistrySlice``` which has it's internal ```.objects``` store (list) replaced with the _jobs_ repository.

rob-c: If you're parsing the strings passed from the Proxy you can't know without inspecting the object and comparing to the proxy what is needed. (Or at least this is the example of you can't know in advance of parsing the code whether to get the object from in-front of the proxy or behind).

milliams: Are there any situations where we want access to the `JobRegistrySliceProxy` object in *our* code? I assume not.

In what cases do we really want access to a `JobRegistrySlice` in our code? Is access to the `JobRegistry` not good enough?

milliams: Surely the only time we're parsing strings are when we're able to look them up in the plugin system? Are there times when this is not true?

rob-c: To access the list of jobs currently available I'd say Tasks as you want to do things such as get RegistrySlices which are useful for other tasks in core, otherwise you've got to make your own from the registry which is more complex than just calling the ```JobRegistrySliceProxy```.

The ```RegistrySlice``` is used heavily in monitoring (but I suspect it can be swapped out for a plan dictionary there or directly use the registry itself. It is also used/returned when performing a ```select``` operation and I'd say having a ```Task``` perform a ```jobs.select``` is quite a powerful thing to allow for job management.

Looking through the plugins gets difficult if you're just parsing ```"LocalFile"``` into ```LocalFile``` it's much easier to just look for ```"LocalFile"``` within the interface dictionary I would have thought otherwise you've got to iterate over the nested dictionaries in the Plugins to determine if the string can be parsed in this way.
Maybe a better solution would be a ```pluginLookUp``` function of some sort which does this but for now there are areas where I'm parsing the interface within the Proxy code to keep old behavior.

milliams: I understand the need for using a `RegistrySlice` in some contexts. I just think that getting it via the GPI is not the best way to do it. This is somewhere we need a case-by-case process to check them. For example, I'm almost certain the monitoring could use a plain dictionary for `JobRegistry_Monitor.registry`.

As for looking up `"LocalFile"`. In what situations do we need to do this without any context? Usually we're loading a job (or similar) from disk and find a `class="LocalFile"` in which case we know what plugin category to look in since we know what category there is on that schema item (because it's a `ComponentItem`).

In the situation where we're given a contextless `"LocalFile"` (I wonder where this happens?) then in reality there's never a plugin objecte in more than one category with the same name (since that would break when exporting to the GPI) in which case adding a function to `PluginManager` to find a name by searching all categories would be a good idea.

I really dislike loading a namespace and then searching it's module `__dict__` for a match like we do in a few places at the moment.

rob-c: I'm not promoting we _should_ get it via the GPI. I'm promoting that we get it via the methods in ```python/Ganga/Core/GangaRepository/__init__.py```
For the moment we abuse the GPI because it's not clear what's going on. I would prefer to have a simple way of accessing GPI objects rather than doing something complex to regenerate them as we need them (at the least this is wasteful).

I agree searching the interface ```__dict__``` is very lazy and to be avoided but it's done in the Proxy layer atm to avoid having to make changes to the Plugins and the Proxy in the same PR.
I'm happy to put together a pull request which removes the lookups against the interface ```__dict__``` in fact I'll open an issue for this exact thing.

rob-c:
Does anyone object to merging this branch as it stands?

python/Ganga/Core/GangaRepository/__init__.py
milliams: If we're introducing new functions then they need full docstrings (preferably in the Google style see [here](http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html) and [here](http://www.sphinx-doc.org/en/stable/ext/napoleon.html#module-sphinx.ext.napoleon)).

python/Ganga/GPIDev/Lib/Registry/BoxRegistry.py
```diff
@@ -155,14 +161,11 @@ def proxy_remove(self, obj_id):
...
+
     def getProxy(self):
-        slice = BoxRegistrySlice(self.name)
-        slice.objects = self
-        proxy = BoxRegistrySliceProxy(slice)
-        proxy.add = self.proxy_add
-        proxy.rename = self.proxy_rename
-        proxy.remove = self.proxy_remove
-        return proxy
+        return self.stored_proxy
```
milliams: We don't need 'getters' for these. It's more Pythonic to just access the names directly or at least to provide access to them via `@property`s

python/Ganga/Core/GangaRepository/__init__.py
rob-c: Frankly they're 1 line wrappers and other pre-existing functions don't... but there's an edit button on github :)

python/Ganga/GPIDev/Lib/Registry/BoxRegistry.py
rob-c: This is only because I've re-written it, I'm not about to go in and change the design as the getters return different things (or don't) for certain repos anyway. This is better in a future PR imo.

rob-c:
I'm planning to merge this once #329 and #328 are in unless there are any objections.

PR #305, opened 2016-03-11
Close file handles in SessionLock

milliams:
I've been working on my new laptop today and when running the test suite I was finding that it was failing towards the end with a `Too many open files` error. Looking into the number of open file handles Ganga had, I was finding that it was increasing steadily with time. Looking in `/proc/<PID>/fd/` I saw lots and lots of `/home/matt/gangadir_testing/TestJob/repository/testframework/LocalXML/6.0/sessions/global_lock (deleted)`.

This PR makes sure firstly that all the calls to `close()` are either guarded by a context manager or by a `finally`. Secondly, it actually closes the `global_lock` file when the session closes down.

### DIFF
--- a/python/Ganga/Core/GangaRepository/SessionLock.py
+++ b/python/Ganga/Core/GangaRepository/SessionLock.py
@@ -394,8 +394,10 @@ class SessionLockManager(object):
             if not os.path.exists(self.cntfn):
                 try:
                     fd = self.delay_init_open(self.cntfn)
-                    os.write(fd, "0")
-                    os.close(fd)
+                    try:
+                        os.write(fd, "0")
+                    finally:
+                        os.close(fd)
                     #registerGlobalSessionFile( self.cntfn )
                 except OSError as x:
                     if x.errno != errno.EEXIST:
@@ -413,8 +415,10 @@ class SessionLockManager(object):
             # Setup session file
             try:
                 fd = self.delay_init_open(self.fn)
-                os.write(fd, pickle.dumps(set()))
-                os.close(fd)
+                try:
+                    os.write(fd, pickle.dumps(set()))
+                finally:
+                    os.close(fd)
                 registerGlobalSessionFile(self.fn)
             except OSError as err:
                 logger.debug("Startup Session Exception: %s" % str(err))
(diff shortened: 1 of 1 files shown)


python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -484,17 +489,17 @@ def global_lock_setup(self):
                 if not os.path.exists(lock_path):
                     os.makedirs(lock_path)
                 if not os.path.isfile(lock_file):
-                    lock_file_hand = open(lock_file, "w")
-                    lock_file_hand.close()
+                    with open(lock_file, "w"):
+                        pass
```
rob-c: Are you certain that this creates a new lockfile if one doesn't exit?

milliams:
retest this please

egede:
For your information you can have a look at the files Ganga will open and close over two minutes. In this case the session was quiet and there was no monitoring to be done. Why are we writing so many files to /tmp? And why do we open `/home/egede/gangadir/repository/egede/LocalXML/6.0/sessions` 10 times with 0.02 seconds - and then repeat that pattern every 15 seconds?

[strace.txt](https://github.com/ganga-devs/ganga/files/169368/strace.txt)

rob-c:
@egede We make a lot of use of ```/tmp``` to execute scripts to check things like there is no work to be done on jobs and to interact with the command line.
Was this definitely a quiet ganga session launched with ```--no-mon``` vs having nothing but completed jobs in your repo?

As for what's going on in the session lock files could this be because the system is iterating over all of the lockfiles once every 15 seconds to keep any items in the various repositories locked? This sounds like expected behaviour.

python/Ganga/Core/GangaRepository/SessionLock.py
milliams: It does indeed. I just tested it in an interactive Python prompt.

rob-c: OK, I was just wanting to be certain I think it used to be this way but there was a problem with contexts in the old tests ages ago which required us to open/close a lot more

egede:
@rob-c This was with monitoring on, but only completed jobs. As long as the behaviour is expected, this is all fine.

milliams:
I've now seen this error crop up on the ganga-ci machine so I think it's worth fixing ASAP. Are there any objects to merging this in for 6.1.18?

rob-c:
go for merge from me

PR #306, opened 2016-03-11
Making import GangaList work from within Proxy layer

rob-c:
This change is to primarily allow for ```from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList``` to work at the top of ```Proxy.py```

This moves around a load of imports but not much else. I find this nicer than the previous solution where it wasn't clear from reading the code if it has always been loaded or not.

### DIFF
--- a/python/Ganga/GPIDev/Schema/Schema.py
+++ b/python/Ganga/GPIDev/Schema/Schema.py
@@ -400,16 +400,7 @@ class Item(object):
     # compare the kind of item:
     # all calls are equivalent:
     # item.isA(SimpleItem)
-    def isA(self, _what):
-
-        try:
-            from Ganga.GPIDev.Base.Proxy import stripProxy
-        except ImportError:
-            def dummyfunc(arg):
-                return arg
-            stripProxy = dummyfunc
-
-        what = stripProxy(_what)
+    def isA(self, what):
 
         if isinstance(what, types.InstanceType):
             what = what.__class__
--- a/python/Ganga/GPIDev/Schema/Schema.py
+++ b/python/Ganga/GPIDev/Schema/Schema.py
@@ -111,8 +111,8 @@ class Schema(object):
 
     @property
     def name(self):
-        from Ganga.GPIDev.Base.Proxy import getName
-        return getName(self._pluginclass)
+        from Ganga.GPIDev.Base.Objects import _getName
+        return _getName(self._pluginclass)
 
     def allItems(self):
         if self.datadict is None: return zip()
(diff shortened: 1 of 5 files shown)


rob-c:
Is there any opinion as to whether this is a PR people agree with or do people not mind juggling the import of ```GangaList``` into ```Proxy.py```?

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -402,7 +402,12 @@ def __len__(self):
     # item.isA(SimpleItem)
     def isA(self, _what):
 
-        from Ganga.GPIDev.Base.Proxy import stripProxy
+        try:
+            from Ganga.GPIDev.Base.Proxy import stripProxy
+        except ImportError:
+            def dummyfunc(arg):
+                return arg
+            stripProxy = dummyfunc
```
milliams: Under what situation does this `ImportError` occur? I don't like complicated logic like this in such a central bit of code.

milliams:
It's a difficult thing to solve since `Proxy.py` and `GangaObject.py` are on logically similar levels and they, by construction, use each other. For now I can't see a better way to solve this that you've done here except maybe splitting on or the other file into two separate modules to get the layering correct.

python/Ganga/GPIDev/Schema/Schema.py
rob-c: If you take this out launch ganga. This fails due to a circular dependency. Proxy->GangaList->Schema->Proxy I think which emphasizes that we should just disallow proxies objects getting into the Schema by design imo.
I'd be happy to remove the stripProxy as long as we have a test which throws an exception if someone tries to stick a proxied object into the schema.

milliams: If this is to protect against putting proxied objects as default values in schemas then I'm not sure it's needed indeed. The only way it can happen is if one of us explicitly puts one in there in the code. The place we will have to be careful is to not put proxy objects that exist in physical Python code in there like `JobRegistrySliceProxy` or similar. There's already a check in `Schema._getDefaultValueInternal` (added by you I believe) to protect against it. If we end up with proxy objects in `_data` then we *want* it fail so that we can fix it.

rob-c: In that case this can be simplified by removing the stripProxy entirely but the ```isA``` is used in many places in the source code so I'm weary of this suddenly causing things to re-break again

rob-c: Or rather this _was_ used in many places I think I swapped many of these out for better type ```isType``` checks.

rob-c:
I see this being irrelevant with some planned future work, I can regenerate this if it becomes needed again.

PR #307, opened 2016-03-11
Ensure parent set on subjobs

drmarkwslater:
This is a minor bit of cleanup to make absolutely certain the parent of a loaded subjob has been set before it is accessed anywhere else as this was seen to cause lock ups when flushing and monitoring jobs that hadn't had all their subjobs loaded.

In addition, it adds another 'hammering' test that exposes the above problem by submitting a job with 100 subjobs, 'quitting out' and then firing off a load of flushing threads while at the same time accessing the subjobs info thus forcing a load.

This test failed/hung on the 6.1.16 release but passes on 6.1.17. However, I'd still like to add the change to the setParent code just to be sure.


python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -438,6 +441,9 @@ def flush(self):
                 subjob_data = self.__get_dataFile(str(index))
                 subjob_obj = self._cachedJobs[index]
 
+                if subjob_obj is subjob_obj._getRoot():
+                    raise GangaException(self, "Subjob parent not set correctly in flush.")
```
rob-c: How does this come about? do we ever see this exception?

python/Ganga/new_tests/GPI/TestSubjob.py
```diff
@@ -0,0 +1,56 @@
...
+    def testSetParentOnLoad(self):
+        """
+        Test that the parents are set correctly on load
+        """
+        from Ganga.GPI import jobs, queues, Executable, Local
+        from Ganga.GPIDev.Base.Proxy import isType
+
+        def flush_full_job():
+            j = jobs(0)
+            j._impl._setDirty()
```
rob-c: I'd prefer to avoid using ```._impl``` but I suppose this is personal preference.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
drmarkwslater: This was the problem with the lock that your PR fixes. Basically, the flusher was trying to flush a subjob that was in _cachedObjs but didn't have it's parent set. This can happen (though very rarely) with the previous code as there was a small gap between adding the loaded subjob to the list and setting the parent. This exception should now NEVER happen but would at least be a very obvious red flag as to the problem if it was ever seen. If the flush continues with the object parent not set correctly you will probably get lockups which would be a lot harder to debug.

python/Ganga/new_tests/GPI/TestSubjob.py
drmarkwslater: What would be the preference? I'm not wedded to this, just wanted make sure the job was considered dirty as otherwise the flush wouldn't do anything :smile:

milliams: Maybe just set a parameter on it. Perhaps the `comment`?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: This sounds like we should be adding some locks into this file as well for safety to do with loading of jobs.
I see how your change now means that we don't get conflicts from the job being in the cache before having it's parent set so this should avoid any future problems. I just worry about double-loading now which could cause issues.

drmarkwslater: I see the point but I'd be wary of putting more locks in at present, at least until we see an issue. Also, I would expect the worse that can happen is that two threads load the same job at slightly different times and the second overwrites the first but with the same info. In which case, I don't think it would matter.

python/Ganga/new_tests/GPI/TestSubjob.py
drmarkwslater: Changed!

rob-c:
@drmarkwslater Would you object to me adding a thread lock around the getitem method on this branch?
Hopefully this will avoid any double loading I've mentioned.

drmarkwslater:
@rob-c I have to say I'm very wary of doing that because the call to `setParent` will be grabbing other locks and nested locks -> deadlocks if we're not very careful. Let me have a think about other possible solutions. I've got to go and move some computers right now :)

rob-c:
@drmarkwslater I've had a stab at doing this for #308 and can't see any real lockups so I'm not so worried about putting this method behind a lock. It should be to prevent double loading imo as it's the only method which loads a subjob from disk but is potentially called from different places.

drmarkwslater:
@rob-c Had a good think and I think we are safe to put a lock here as you suggest because the loading won't affect anything and the setParent call afterwards is going to be setting it from 'None' and so will be safe as well. I've made a few changes from your branch:

1. changed the name of the lock to `_load_lock`. Seemed to make a bit more sense :smile: 
2. Put the acquire of the lock just when we're sure we want to load so it's not acquired when not needed and we can also test to see if it was loaded while waiting for the lock.

What do you think? If you're happy, please add in your additional queued tests and then I'll merge :smile:

rob-c:
@drmarkwslater I'm concerned this does still __not__ prevent double loading.

If a non-loaded subjob is requested by 2 threads at the same time they will both load the job and at best one thread will always point the wrong job for a short period of time and at worst we have one thread always running with stale/bad information.
The lock needs to be in front of the ```if index in cache``` check which checks to see if the job has yet been loaded. Once the job has been loaded by 1 thread it only makes sense to check it again here.

Would it be possible for you to have a look at the ```Ganga/new_tests/GPI/Parallel/TestQueuedSJSubmit.py``` test in my branch and merge it back at all?

This should hopefully provide a complimentary test to your construct/load/submit test which hammers the code from a slightly different direction.

drmarkwslater:
This **does** prevent double loading as, in your example, if 2 threads attempt to load, one will block on acquiring the lock while the other loads the subjob. After the lock is released the first thread then checks if it has been loaded while it was waiting and return if so:

```
            # obtain a lock to make sure multiple loads of the same object don't happen
            with self._load_lock:

                # just make sure we haven't loaded this object already while waiting on the lock
                if index in self._cachedJobs:
                    return self._cachedJobs[index]
```

The benefit of doing it here is that you are only acquiring a lock when it's actually needed, not on every `getItem` call. Considering the previous issues we've had with deadlocks, I really want to minimise the chances of being hit by them again.

I'll add your test from #308 in ASAP.

rob-c:
OK, it appears your correct reading pulls can be difficult.

I suppose your method is slightly quicker if double-loading isn't going on, but I wasn't expecting to perform the same check again within the locked code.

OK, once your happy go for merge from me and then I'll remove #308.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -326,52 +331,60 @@ def __call__(self, index):
 
     def __getitem__(self, index):
         """Return a subjob based upon index"""
-        return self._getItem(index)
         try:
             return self._getItem(index)
         except Exception as err:
-            logger.error("CANNOT LOAD SUBJOB INDEX: %s" % str(index))
+            logger.error("CANNOT LOAD SUBJOB INDEX: %s. Reason: %s" % (str(index), str(err)))
```
milliams: This doesn't need the explicit conversion with `str()`. `%s` will *always* do this call for you.

drmarkwslater: You are entirely correct - doh! Fixed.

drmarkwslater:
Any other comments from anyone? If not, I'll merge this afternoon.

rob-c:
Go for me if the tests are all working

rob-c:
This test was working on my branch even though it took about 5min so I'm concerned it's not passing here.

milliams:
The ganga-ci machine does seem much more sensitive to thread-based lock ups than any other. I'm not sure exactly what's going on but it seems that the VM struggles when there are even 20 threads running in some conditions.

retest this please

rob-c:
But this test worked on ganga-ci on my branch as well so I'm more concerned it's the implementation

drmarkwslater:
Can we beef up the CI machine at all? This could make tracking down legitimate deadlocks rather tricky. Though saying that, I'm guessing the deadlock tests that have been added (mine certainly) are **way** overboard for forcing failures...

rob-c:
I'm not a fan of running the deadlock tests on every change and commit, can we leave them as something which has to be manually run right before each PR is committed?
(I don't know if this breaks the current model which is why I ask)

milliams:
The CI machine (at least ganga-ci-worker-01) is a quad-code 2.2GHz with 8GB of RAM so it's not nominally a slow machine. However, we're seeing problems there which none of us can reproduce on our real computers which gives me pause.

As for these stress-tests I would recommend trying to find the minimal point (number of threads, concurrent jobs, whatever) which allows you to reproduce the original problem every time on your local machine and use that as the test.

drmarkwslater:
Hmm... That's not a bad idea. I'll start going through and trying to find the sweet spot from the ones that are causing problems. A variation of what @rob-c suggests though: Could we have less heavy tests on every commit but safe some really serious ones for the nightlies? I'm a bit wary of getting tests to pass by just relaxing them :smile:

drmarkwslater:
Actually, just having a look at the tests on Jenkins, the tests in this PR are making the tests take ~3 times as long so I certainly need to address that at least...

rob-c:
Would it be possible to have the thread tracing on within ganga-ci so the trace of a lockup can be dumped at least for a failed job. We're losing all information on the killed jobs atm which is difficult as this machine is the only place where we're seeing most of them.
And I've got a feeling as soon as we release with a known failing test we get 5 users who can reproduce this doing the simplest of tasks.

milliams:
I'm looking into incorporating the stacktracer at the moment. We should be able to run it for all tests for exactly the reason you give. I will hopefully have something to show tomorrow.

rob-c:
Aside from the tests in this PR taking a long time are there any other reasons not to merge?

drmarkwslater:
I don't believe so, especially seeing as users are possibly hitting this bug. However, I'm going to do a bit more investigating before pushing for anything as I don't want to push code into develop in a rush that either doesn't fix the problem or makes things worse!

rob-c:
I think this fixes a separate problem that I'd be keen to rule out the branch is potentially leading to this. Given the lines fixed here are in the trace against issue #317 I am keen to merge this for testing __ASAP__.

drmarkwslater:
I'm ~95% sure this will fix the lockups that we've seen today so I'm just going to tweak the tests a little and then we can merge.

drmarkwslater:
I'm now having great difficulty getting either of the tests in this branch to fail. Can someone else have a go? I've tried with both 6.1.16 and 6.1.14 and they all pass. Because they seem to take a long time and for no (apparent) benefit, I'll probably drop them at least from this PR so we can get it merged.

Any thoughts/comments?

milliams:
Given that `TestQueuedSJSubmit` is taking 15 minutes to run at the moment I feel it's far too heavyweight at the moment to be worth it. `TestSubjobs` is still quite heavy at 50-60 seconds but it's perhaps worth having it in anyway. The real problem is if they're not even exposing the expected bug before the fix. I will have a go at running them myself to see if they show anything.

rob-c:
We can reduce the number of jobs/threads if needed but I'd prefer to have the logic tested at least lightly in parallel

drmarkwslater:
@milliams This is my thought as well - if none of these tests actually reveal any bugs (at least reliably) then I'm in favour of dropping them.

rob-c:
I think the parallel tests can be pushed to a nightly script as it would make life easier. Having to wait 10+min to run _all_ tests on every commit to a PR is a little time consuming I feel.

drmarkwslater:
Actually, looking at the Queued SJ test, I agree with @rob-c in that this should be tested somewhere (i.e. queued submission of subjobs) but 100 jobs is probably too much. I'll reduce it to 20 for the moment and see how long that takes. I'll also remove the other set of tests.

milliams:
Running `TestSubjobs` on the `develop` branch is revealing a lockup in the call to `_setParent` in `SubJobXMLList._getItem` and it's doing so reliably. I imagine that this deadlock is something that we introduced in the last release or so and so we're seeing it manifest here.

I haven't tested the other one since it takes so long but this branch definitely seems to have fixed at least one lockup.

drmarkwslater:
I **knew** I'd seen those tests fail somewhere. Good to know I'm not going mad :smile:

drmarkwslater:
Decision time: Am I OK to merge this branch? (assuming tests work of course!)

python/Ganga/new_tests/GPI/TestSubjobs.py
```diff
@@ -0,0 +1,59 @@
...
+            queues.add( flush_full_job )
+
+        # Now loop over and force the load of all the subjobs
+        for sj in j.subjobs:
+            self.assertEqual( sj.splitter, None )
+            self.assertTrue( isType(sj.application, Executable) )
+            self.assertTrue( isType(sj.backend, Local) )
+            self.assertEqual( sj.application.exe, "sleep" )
+            self.assertEqual( sj.application.args, ['400'] )
+            self.assertIs( sj._impl._getRoot(), j._impl)
```
milliams: The formatting and style of a lot of these files is a little off. There's inconsistent spacing around parentheses, `import`s inside functions and similar. Maybe we could use the standard Python/pytest asserts too? Perhaps it's worth fixing those and doing a `rebase -i` to squash them together?

rob-c:
go for merge from me if the tests are running in a sane amount of time :+1:

python/Ganga/new_tests/GPI/TestSubjobs.py
drmarkwslater: I can certainly move the system imports but the Ganga ones I regard as part of the test so would keep them as is. 

As regards spacing, should there be space or not in parentheses? I generally prefer it but I notice that PyCharm isn't a fan. I don't really mind, just happy to be consistent :)

I had thought it was better to use `self.assert...(...)` as that's used a lot elsewhere. Is that not the case? If so, I'll switch back to just `assert ...`.

milliams:
`TestSubjobs` only took 11 seconds (:tada:) and while `TestQueuedSJSubmit` took a minute and a half it's worth it if it's fixing a large bug and testing something useful.

python/Ganga/new_tests/GPI/TestSubjobs.py
alexanderrichards: > As regards spacing, should there be space or not in parentheses? I generally prefer it but I notice that PyCharm isn't a fan. I don't really mind, just happy to be consistent :)

pep8 says no

milliams: re. `self.assert...()`: when we first started porting over tests we were using `unittest` and `nose` to provide feedback. They use a style ported from `JUnit` (i.e. Java) for their asserts. They provide a lot more information than plain `assert`s in that situation and so I was encouraging their usage. Now that we are using pytest it has a system to interpret `assert` statements and extract the information in a much cleverer way.

The upshot is that before we *had* to use `self.assert...()` to get good messages but now we're using pytest we don't *have* to any more. I'm not very fussed either way but if we can get as good feedback from a simpler and cleaner `assert foo is bar` then that *slightly* wins out in my mind. There's no imperative to change anything but sometimes changing them if you're doing other changes anyway might be worth it.

drmarkwslater:
I found the sweet spot for TestSubjobs as below this the tests passed on the develop branch. I couldn't get the TestQueuedSJSubmit tests to fail but I still think they test something sensible so just scaled them back quite a bit.

I've just put in a cleanup of the tests so I think that's addressed all concerns and if no one says otherwise, I'll merge tonight.

Tomorrow we'll have to decide whether to just release with what we have or decide to put in more from the 6.1.18 milestone.

egede:
I am pretty keen on a straight release and then have a discussion about the 
rest on Thursday.

PR #308, opened 2016-03-13
Ensure parent set on subjobs plus lock

rob-c:
Adding a lock around #307 and introducing a test to see if submitting jobs with subjobs is thread-safe. Passes on my test machines, interested to test on Jenkins.


rob-c:
Superceded by #307

Issue #311, opened 2016-03-14
Objects incorrectly being set dirty


rob-c:
LHCb users are seeing what I saw that some objects are marked as dirty when they should never be loaded.

The API for the ```_setDirty``` also doesn't respect ```_setDirty(False)``` which is a slight annoyance. If we're changing to use ```_setDirty()``` and ```_setFlushed()``` we should not respect the old API.

I'm tracking down what is causing some jobs to be loaded on shutdown which we thought was fixed at some level. (It appears this may only be fixed for simple jobs)

milliams:
`_setDirty(False)` has not worked since at least 2009 so we really shouldn't be using it anywhere. In the future I'd much prefer either using `setDirty`/`setFlushed` or a `@property` so we can just do `obj.dirty = True`/`obj.dirty = False`.

It's possible that there's an issue at the moment that if you set a schema attribute to the same value it already contains that it will still set it dirty which it probably should not be.

rob-c:
I'm seeing that empty objects are being marked dirty which needs addressing.

rob-c:
wrt the ```_setDirty(False)``` I plan to remove this in that case, promising, or allowing this is a bad idea if it doesn't work. I'm preparing a short PR which doesn't load jobs it shouldn't.

drmarkwslater:
Just to say that I think the mail that sparked this on the DAST list may have been a red herring as I suspect this may have been due to missing index cache files instead. The warnings given by the user are actually on **startup** not shutdown and at the top, you can see that it's complaining about a missing `data` file. It would certainly try to load at this point if the index cache file wasn't there and I suspect that's what's causing the problem.

drmarkwslater:
Out of interest, what's the problem about marking an Empty Object dirty? Doesn't it only become a problem if the object is attached to a Registry/Repo which an Empty Object isn't? Therefore, you can happily change stuff on an Empty Object, it gets marked dirty but that doesn't matter as it will never be picked up for flushing (or loaded for that matter). 

There would be a problem if you have an object attached to a registry that is lazy loaded (so I guess just Jobs) that has been marked Dirty BUT hasn't been fully loaded yet. I'm assuming this can't happen for jobs though I haven't gone through the code to specifically check.

rob-c:
Dirty empty objects will get picked up for flushing because the registry is populated with empty objects capable of triggering a load from disk when interrogated for a schema attribute they don't have.

OK, with regard to indexes...
[...]

PR #313, opened 2016-03-15
Fix bad flushes

rob-c:
Trying to address #311 where I'm seeing some unloaded 'dirty' jobs written to disk. This in some circumstances is dangerous and concerns me that some data could be overwritten.

I'm unaware of any mechanism to allow for an object in memory in a registry and not be fully loaded so this appears safe.


rob-c:
If there aren't any complaints I'd like to push for merge today

python/Ganga/Core/GangaRepository/GangaRepository.py
```diff
@@ -205,7 +205,7 @@ def _make_empty_object_(self, this_id, category, classname):
         #obj.__init__()
         obj.setNodeData({})
         obj.setNodeAttribute('id', this_id)
-
+        obj._setFlushed()
```
milliams: Do you know what was setting this object dirty here?

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -691,7 +691,8 @@ def _flush(self, objs):
 
         for obj in objs:
             # check if the object is dirty, if not do nothing
-            if not obj._dirty:
+            this_id = self.find(obj)
+            if not obj._dirty or this_id not in self._loaded_ids:
```
milliams: If it's `not loaded` then is it not also always clean?

python/Ganga/new_tests/GPI/Loading/TestLazyLoadingManyJobs.py
```diff
@@ -0,0 +1,95 @@
...
+        """ Second get the job and check that getting it via jobs doesn't cause it to be loaded"""
+        from Ganga.GPI import jobs
+
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+
+        from Ganga.GPIDev.Lib.Job.Job import lazyLoadJobApplication, lazyLoadJobBackend, lazyLoadJobStatus, lazyLoadJobFQID
+
+        self.assertEqual(len(jobs), numJobs)
+
+        print("len: %s" % str(len(jobs)))
```
milliams: Don't need `str()`

python/Ganga/new_tests/GPI/Loading/TestLazyLoadingManyJobs.py
```diff
@@ -0,0 +1,95 @@
...
+
+        for j in jobs:
+
+            if j.id is not 9:
```
milliams: This should use `!=` for portability.

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: Indeed - I'd be very worried if we can have jobs that can be marked dirty but **not** loaded! Do we have an example of this somewhere?

drmarkwslater:
I'm a bit concerned that we don't know what's causing this to load and then flush. I feel that if we don't fix the underlying problem that could really come back to bite us. I've been trying all morning and I can't recreate this bug at all so I'm going to try to get some more info from the user to see if I can get any more insight...

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I at least once observed that an EmptyObject had been marked dirty which is a bad thing obviously.
I don't know if this goes away with fixes elsewhere but I want to be belt and braces against loading potentially a large amount of very old jobs into memory on shutdown.

python/Ganga/Core/GangaRepository/GangaRepository.py
rob-c: I never tracked it down but I'm being explicit in-case we chose that getting an object from the Plugins performs an operation in the class ```__init__``` that can mark an object dirty. aka, setting the default backend in ```__init__``` for a Job will mark an empty job object in memory dirty and appear asif it's safe to write to disk (potentially killing whatever is on disk)

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I'm concerned by it but I've not been able to track down a solid example so I'm in favor of always having the correct behavior here at least. I think we can end up with a partly-loaded job being dirty but not marked as correctly loaded (this I'm not 100% sure about but I suspect).
This is more of an attempt to stave off future problems.

rob-c:
I'm just glad we have the ```_getWriteAccess``` which loads the job before flushing it vs torching the job with the default data which has at some level been modified.

I think it's probably best to add some exceptions in the EmptyGangaObject class to catch setting it's dirty state which would highlight this error quite loudly.

python/Ganga/Core/GangaRepository/GangaRepository.py
milliams: Ah, so this function naming is confusing. It doesn't create an `EmptyGangaObject` but rather simple an object of the correct type with its data emptied out. What is the purpose of the `_make_empty_object_` function in this case?

rob-c: looking up the object type used to be much more difficult.
(frankly you can always use a lookup in the GPI for the proxied class and strip that to create an instance so all the dancing around in plugins I dislike)

Also the object used to be something distinctly different but caused more difficulty it used to be a new object which wasn't fully initialized (and this caused many, many errors).
As it stands the Object is made _empty_ as there is no dictionary associated with schema attributes which will trigger a load in the ```__get__``` method. Returning a vanilla object here will just make the object look like a new GangaObject and never trigger a load from disk until disk access is requested.

I'm much more keen to move to a framework where each object explicitly knows that it's been loaded or not based upon an internal boolean. This avoids all of this conditional object inspection and allows us to simply the question of if the thing ever loaded from the disk or not.

drmarkwslater:
Ah - just suggested the same on #311 :smile:

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: I'd personally prefer to check for the `obj._dirty` flag first to avoid a mildly expensive `find` operation on every object. Other than that, I'm happy with this I think given the reservations about `EmptyGangaObjects`

rob-c:
OK, I'd like to merge this PR as it fixes some breaks due to the API change (another PR) and adds another test.

milliams:
I'm happy for this to be merged once the commits from today are squashed down to something sensible.

rob-c:
Frankly squashing the commits is a thankless task imo, who is actually watching the minor white-space changes on develop?

milliams:
When I'm trying to do a `git blame` to work out when or why a line of code was changed it's absolute hell when the code is being changed, changed back, changed to something else and then changed back again due to someone trying to get a test to pass which could have been done locally or been squashed down. When I finally get to the change and the commit message is 'fixes' it just makes me want to quit. As a group we need to take care over our commit history or it becomes pointless having version control at all.

rob-c:
Well why not squash the commits on merge to develop which would speed up bisecting the problem?

milliams:
I assume you're referring to the new GitHub feature to choose whether to use a merge commit or a single in-line squashed commit when merging a PR? For small PRs I think that's a good idea (and this one fits it) but any PR that is doing more than one logical thing I think it's worth having the commits separated out.

rob-c:
when using ```git blame``` for 1k commits (I don't think we're reaching this) it still takes only about 10 steps to work out where a testable break happens doesn't it?
Or, are we getting much more than 1k commits per release?

I'm just going to squash this into 1 commit in a bit

milliams:
I think you're confusing `git bisect` and `git blame`? The latter is a more forensic tool to track manually though the history to see who changed which line and when. When a line of code is edited and then reverted, both those changes are opaque to `git blame` and it takes work to track around them.

rob-c:
I'm probably getting mixed up but frankly would never use ```git blame``` in this case due to it's shortcomings imo. I feel this is less of a coding/work-style problem.

PR #318, opened 2016-03-15
Add a thread stacktracer

milliams:
This adds a module to `Ganga.Utility` to provide the dumping of thread stacktraces to a file. It can be used with:
```python
from Ganga.Utility import stacktracer
stacktracer.trace_start('trace.html')
```
That code that can, for example, be added to a user's `~/.ganga.py` or just run live.

It may be useful in the future to incorporate this with `report()` or maybe the monitoring thread heartbeat monitor. I think we should also think about running this alongside the test suite so that locked-up tests can be debugged even on Jenkins.

### DIFF
--- a/python/Ganga/Utility/feedback_report.py
+++ b/python/Ganga/Utility/feedback_report.py
@@ -186,6 +186,7 @@ def report(job=None):
         gangaLogFileName = "gangalog.txt"
         jobsListFileName = "jobslist.txt"
         tasksListFileName = "taskslist.txt"
+        thread_trace_file_name = 'thread_trace.html'
         from Ganga.Utility import Config
         uploadFileServer = Config.getConfig('Feedback')['uploadServer']
         #uploadFileServer= "http://gangamon.cern.ch/django/errorreports/"
@@ -651,6 +652,15 @@ def report(job=None):
             except Exception as err:
                 logger.debug("Err %s" % str(err))
                 writeErrorLog(str(sys.exc_info()[1]))
+
+        # Copy thread stack trace file
+        try:
+            thread_trace_source_path = os.path.join(getConfig('Configuration')['gangadir'], thread_trace_file_name)
+            thread_trace_target_path = os.path.join(fullLogDirName, thread_trace_file_name)
+            shutil.copyfile(thread_trace_source_path, thread_trace_target_path)
+        except (OSError, IOError) as err:
+            logger.debug('Err %s', err)
+            writeErrorLog(str(sys.exc_info()[1]))
 
         resultArchive = '%s.tar.gz' % folderToArchive
 
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -26,9 +26,11 @@ import string
 import sys
 import time
 import re
+import atexit
 
 from Ganga import _gangaVersion, _gangaPythonPath
 from Ganga.Utility.Config.Config import getConfig
+from Ganga.Utility import stacktracer
 import Ganga.Runtime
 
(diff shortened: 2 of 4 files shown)


egede:
At what point in time does it give the stack trace?

milliams:
@egede If you run the code I mentioned above then it will refresh the stacktrace page in the background every 5 seconds. This is configurable though by passing `interval=1` (for 1 second for example). The way it works overall is by starting a separate background thread which takes snapshots of all the threads every `interval` seconds.

rob-c:
Can we modify this to save the last and last-1 traces? This would allow for any threads which haven't completely stalled to be identified and ignored as part of any post-mortem.

I'd be keen to ship this with the next release if possible.

milliams:
It would be technically possible to store the n-1 set of traces, compare them and only print those which haven't changed. However, any monitoring threads or queues which are just waiting for work will also fit this criterion so it wouldn't necessarily provide the most useful feedback.

egede:
This is a very good idea, but it needs to be separate from 'report' as we 
was to make sure that it doesn't in itself lock up while uploading. Maybe 
just export a function to the GPI called stacktrace.

rob-c:
@milliams I'd rather not make the comparison automatic, I'd strongly prefer to keep this manual for the time being, I just think it would be useful to get a sense of if anything is changing in time or not. (we _could_ be catching infinite loops with this!).
@egede I'd like to see this integrated in some form of report and I think a review of whats in this payload is probably due

egede:
A few things were agreed at today's developer meeting. 

- The stacktrace code should run by default
- It should fail silently if it can't write the file
- The time should be written at top of stacktrace file
- It should be included in the files that `report` is uploading.

milliams:
Ok, I think I've made all the relevant changes. I'm looking to merge whenever people are happy.

python/Ganga/Utility/feedback_report.py
```diff
@@ -652,6 +653,15 @@ def renameDataFiles(directory):
                 logger.debug("Err %s" % str(err))
                 writeErrorLog(str(sys.exc_info()[1]))
 
+        # Copy thread stack trace file
+        try:
+            thread_trace_source_path = os.path.join(getConfig('Configuration')['gangadir'], thread_trace_file_name)
+            thread_trace_target_path = os.path.join(fullLogDirName, thread_trace_file_name)
+            shutil.copyfile(thread_trace_source_path, thread_trace_target_path)
```
rob-c: Is there a reason we're copying the file from one destination to the other? Why not just write directly to the new location?

milliams: I did it this way so that it reuses the existing `thread_trace.html` file that's on disk rather than creating a new one. I'm not averse to changing this to create a new file though if that's what people prefer?

rob-c: I'm confused why it's writing it in 2 places in this case?

milliams: It's not. It's:

1. Setting `thread_trace_source_path` to the source of the file (`~/.gangadir/thread_trace.html`) where the background thread is writing it.
2. Setting `thread_trace_target_path` to where the file should be copied to.
3. Copying the file at `thread_trace_source_path` to `thread_trace_target_path`

Unless I misunderstand you or have made a mistake with this code?

rob-c: Ah I think I see what's going on here. Apologies was a little confused on my side. Good to see this go into the report I suppose, it's just a shame we can't test it for a few days.

rob-c:
OK go for merge from me, sorry for the noise

PR #319, opened 2016-03-16
Python 3 compatibility for unit tests

milliams:
This updates the code so that the unit tests can run under Python 3. This does not make the whole of Ganga Python 3-compatible but it's a solid start.

Most of these changes are simply syntax compatibilities but some require the use of the `future` package to provide compatibility between Python 2 and 3.

While we can't yet say "Ganga runs on Python 3", having these tests turned on from now on will hopefully at least prevent us back-sliding any further.


rob-c:
?!? I'm concerned.

Are we running the tests for Python 2.7 yet? We still don't officially support that do we?

milliams:
Of course we officially support Python 2.7 as all LHCb users are using it. All our testing at the moment is done on 2.6 (at minimum) and up until now all our 'unit' tests have been run on 2.6 and 2.7. All this PR does is allow them to run on Python 3 as well.

drmarkwslater:
There's quite a bit changed here so give me a bit of time to check! I'm sure it's fine, but want to make sure :smile:

Out of interest though, are we going to regularly run the tests on Python 3 **and** 2.7? I ask because otherwise we'll probably forget to write Python 3 compatible code (I know I will!) and so you'd have to do a similar thing again in a few months...

milliams:
@drmarkwslater If you check be commit it might be easier as I have grouped them by topic.

That's exactly why I've put this PR in. I wanted to run the Travis unit tests against Python 3 from the start but since they weren't working I thought it best to not have a known set of failing tests. One of the things that this PR does is enable running the Travis unit tests on Python 2.6, 2.7 (they were already enabled) as well as 3.3, 3.4 and 3.5 (b37ce75cf41503046d13b6348e6f3b84eb6157d8). This way we will be always test on all these versions (for the unit tests anyway) from now on so that we will avoid introducing too many new Python 3 incompatibilities. Of course, this is currently only the unit tests so there's probably loads of code which is Python 2-only but this makes a good start.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -406,9 +406,6 @@ def isA(self, _what):
 
         what = stripProxy(_what)
 
-        if isinstance(what, types.InstanceType):
```
drmarkwslater: Is this a problem removing this? I'm not sure I understand what this does anyway...

python/Ganga/Utility/Shell.py
```diff
@@ -109,7 +109,7 @@ def __init__(self, setup=None, setup_args=[]):
             command = 'source %s %s > /dev/null 2>&1; python -c "from __future__ import print_function; import os; print(os.environ)"' % (setup, " ".join(setup_args))
             logger.debug('Running:   %s' % command )
             pipe = subprocess.Popen('bash', env=env, cwd=this_cwd, shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
-            output = pipe.communicate(command)
+            output = pipe.communicate(command.encode())
```
drmarkwslater: Is this required?

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -7,7 +7,7 @@
 import collections
 
 from Ganga.GPIDev.Schema import ComponentItem
-from Ganga.Utility.external.OrderedDict import OrderedDict as oDict
```
drmarkwslater: I'm guessing this means that OrderedDict in Utility can be removed?

drmarkwslater: Ignore - just looked further down the page of changes :smile:

drmarkwslater:
Other than a couple of comments, looks good to me :smile: Quick query - the GangaList changes for slicing: should this be in a separate PR?

rob-c:
Can we consider doing ```try:  ... except: ...``` around import changes between 2 and 3 and some more code redesign so that we can have a 2 and 3 compatible codebase?
I'm slightly bugged at merging this if we're no longer Python 2.x compatible are we even using 2.7 as default yet?

python/Ganga/Utility/Shell.py
milliams: Yes, as `command` is a unicode string and `communicate()` expects a locale-encoded byte stream (almost always `utf-8`).

python/Ganga/GPIDev/Schema/Schema.py
milliams: It was intended to check whether a particular variable pointed to an *instance* or a *class*. Unfortunately it did not do what was expected anyway and the test always failed.

milliams:
@drmarkwslater The changes for `GangaList` slicing could indeed come in a separate PR but they're not *broken* on Python 2, just perhaps not ideal. It was the requirement of Python 3 compatibility which necessitated the change.

@rob-c I'm not sure why you think that this is making the code not Python 2 compatible. With this PR the codebase will work fully on Python 2 and the unit tests will pass on both Python 2 and Python 3. The whole point of this PR is to add Python 3 compatibility, not to remove Python 2 compatibility.

rob-c:
I'm slightly concerned by the fact that ordereddict is introduced in Python 2.7, does this explicitly work with Python 2.6 or does this inherently break Python 2.6 compatibility?
(I wonder if a ```pip install ordereddict``` is required on 2.6 systems? if not then I'm happy but has this been tested?)

I've no idea what we're running on __ganga-ci__ are we testing in Python 2.6+2.7 and/or Python 3.0 as a result of this?
I have a general aversion to most things being modified to be _future proof_ but that's just a gut reaction. If all tests are passing this should be OK.

milliams:
This PR makes `OrderedDict` be provided by the [`future`](https://pypi.python.org/pypi/future/) package which is included in the [list of dependencies](https://github.com/ganga-devs/ganga/pull/319/files#diff-2eeaed663bd0d25b7e608891384b7298R92). `future` works by providing the module on those versions where it is not available (2.6) but using the standard library version where it is (2.7, 3.x).

ganga-ci until recently was only testing on Python 2.6 though Travis was testing on both 2.6 and 2.7 (unit tests only). A few weeks ago I added two new nightly test jobs: [Nightly2](https://ganga-ci.cern.ch/jenkins/job/Nightly2/) which is running all the tests on both Python 2.6 and 2.7 and [Nightly Python 3](https://ganga-ci.cern.ch/jenkins/job/Nightly%20Python%203/) which is doing the same on Python 3.3 and 3.4. Obviously the latter is failing badly at the moment but with this PR it is at least able to get most of the way through starting Ganga.

*Nightly2* will be renamed to *Nightly* soon and will replace the existing one (i.e. with email notification of failure). The Python 3 tests will remain as they are without email notification. I am not currently planning on running *PR Builder* on both 2.6 and 2.7 as that will only serve to make the turn-around twice as slow.

This isn't making the code *future-proof* so much as making it compatible with both Python 2 and Python 3 with one code-base. This is the widely recommended way of migrating to Python 3 or even of maintaining long-running compatibility with both versions (as opposed to running `2to3` or similar).

drmarkwslater:
retest this please

drmarkwslater:
retest this please

drmarkwslater:
retest this please

drmarkwslater:
retest this please

egede:
Abandon this ancient PR.

Issue #39, opened 2015-11-11
Add a file type which writes to a shared filesystem


milliams:
It's common when running jobs on a local batch system to write the output files of the jobs manually to some shared file-system which is also accessible from the desktop machine of the user. We don't currently have a file type which supports this automatically. `LocalFile` will copy it over the sandbox so that is not what we want.

The closest we have is using a `MassStorageFile` with setting the `defaultProtocol` to `''`, the `uploadOptions.path` to the shared file-system and `uploadOptions.cp_cmd` etc. to `cp` etc.

Perhaps it would be worth making a new file type (e.g. `SharedFile`) which derives from `MassStorageFile` which sets these defaults to something sensible?

alexanderrichards:
This seems reasonable to me

egede:
In the past we said that this was just a special case of 
MassStorageFile. However, I agree that this is probably not clear and an 
explicit specialisation could be a good idea.

The exact configuration will depend a bit on if a local batch backend 
has access to this shared directory or not.

Ulrik.

milliams:
Ok, I'll go ahead with this. Does anyone have an opinion on the name? `SharedFile` makes sense to me but is there a better option?

rob-c:
Would it be more sensible to have this as the default for MassStorageFile and move the current 'defaults' to the LHCb and Atlas config files?

I think this is looking like a case of documenting what options people can/should use rather than adding more file objects. (just my gut reaction)

JacekHoleczek:
I agree with rob-c. No need for an additional option. Just several well described examples in the default .gangarc (don't remove anything, leave the "more complex" examples inside, just add the "simple ones", too -> they will anyhow be "commented out").

egede:
This is a good idea. Just on Monday we discussed how we would like to 
make the "default" in Ganga more generic and then specialise it for LHCb 
and ATLAS (rather than the other way around).

My suggestion after discussing with Rob is

- Create a new SharedFile type
- Create the MassStorageFille but in a dummy configuration that
   is not working (i.e. make it explicit that it needs a site
   specific implmentation
- Configure MassStorageFile in LHCb.ini and ATLAS.ini for using EOS.
   Preferably this should be based on "are you at CERN" instead but
   I am not sure how to do that in a safe way.

Ulrik.

rob-c:
I've fixed some long standing issues with the `MassStorageFile` object and have a working local configuration in #735 for the testing of the `MassStorageFile` class.
[...]

PR #321, opened 2016-03-16
Fixing the SJID issue

### LINKED ISSUE #120: Error when trying to submit a job using MassStorageFile and {'sjid'}
When trying to submit a new job in 6.1.14 which uses MassStorageFiles and the {'sjid'} flag, the following error is returned:

JobError: Error in MassStorageFile.outputfilenameformat field :  job is non-split, but {'sjid'} keyword found

This occurs even in the same Ganga session that was used to create the job.

rob-c:
This fixes #120 All I had to do was check against the correct subjobs when the job has subjobs in order to correctly support the sjid file structure.


egede:
Can you just make sure that you can resubmit a subjob as well using the sjid notation. In the past that has been a problem.

rob-c:
OK.

In this case I'll leave this issue open and add a test to the PR to make sure we can submit jobs with this notation as well as re-submit them.

rob-c:
Do we have to keep these codecov tests running against the PR? It's marking a lot of pulls as bad due to what look like they could be rounding errors.

This pull is working, in that there is a working job which submits.
I'd like to have a test which actually matches and creates a file using SJID but I suspect this may then be difficult if MassStorageFile doesn't work on ganga-ci?

milliams:
I've turned off codecov commit status reporting.

Does the new `TestSJIDName` test need to be broken into those many functions? I feel it would be simpler to either run it as one large function or as separate tests completely without the 'multipass'-style.

rob-c:
@milliams I'm also hoping to test that the entire chain of storing, re-loading as well as submit and resubmit all 'work' from a high-level api perspective.

Ultimately I'm wondering if we can see the {sjid} notation adopted by other (all) IGangaFile objects which would probably mean we have well defined parsing functions defined and broken out.
But that's a little beyond the scope of this for now I suspect.

milliams:
@rob-c That makes sense. I'm mostly happy for this to be merged once the release if finished. The only thing that's missing is some comments (or preferably docstrings) in the test class and functions to explain the purpose of the test and why it exists.

egede:
@rob-c You say `MassStorageFile` doesn't work on ganga-ci, but can't you just in the config redefine the command to be a standard copy to a temporary directory?

rob-c:
@egede I'm not asserting it doesn't work I just don't know if it does or is configured correctly, or if it has access to EOS at all.
I think we can come up for a proper solution for the tests in some way that tests the MassStorageFile though.

Should just doing a 'copy' command be the sane default for out of CERN users as well? In which case the change needs to be made in the config system not on ganga-ci.

milliams:
I guess this calls back to #39 which is still on my radar.

rob-c:
@milliams Thanks for finding the issue, I thought this sounded familiar. Do we want to just have a vanilla ```cp``` command for 'out of CERN' users, this sounds most sensible imo.

egede:
@milliams Yes indeed. So this will then solve the testing issue as we will test against the general `SharedFile`.

rob-c:
OK, as this stands this PR checks the submit mechanism but doesn't check the ```MassStorageFile```. Given we plan to have a ```SharedFile``` I'd say worrying about this class working should be done in another PR.
With this in mind are there any objections to merging this PR?

milliams:
Apart from docstrings on the tests (as I mentioned above) I'm happy.

milliams:
This has just failed on the nightly tests at https://ganga-ci.cern.ch/jenkins/job/Nightly/214/testReport/junit/python.Ganga.new_tests.GPI.FileTests.TestSJIDName/TestSJIDName/test_b_jobResubmit/

rob-c:
Hmm, this did pass, I didn't modify any code in https://github.com/ganga-devs/ganga/pull/321/commits/2163c3ca230f59262f31ab982cccc0b563d0e3d9 so now I'm concerned we don't understand this

rob-c:
Obviously a conflict between this and develop was introduced at some level which is a little surprising

rob-c:
This is working when I test it locally. Are you sure the first test submitted and split the jobs correctly. This would be much easier to debug if we had the __full__ stdout from failed tests as per the old testsuite to find strange problems that crop up only in testing.

milliams:
I'm seeing them fail locally on my machine. I'm looking into why we're not getting the full stdout+stderr from the tests.

rob-c:
Well this was working on ganga-ci (as expected), and is working my machine so I'm unable to do much to help with this one.
I'll check what branch things are on.

rob-c:
It's possible ```if len(self.subjobs) != 0:``` is broken on some systems I suppose although I'd struggle to motivate how.
What sort of fs/os is used for ganga-ci and your tests system?

milliams:
My local tests are run on an NFS file system under SL6 and the ganga-ci tests are run on a local virtual file system which equates to pretty much ext3 and is also on SL6.

rob-c:
I'm working on the assumption this was able to actually generate and split the job in the first step. Although maybe an assert on the subjob length wouldn't hurt at the end of the first test.

I'll try re-running on xfs or ext4 if I get a chance today as it could be the ```__len__``` of subjobs is tied to the parent folder modified time which is probably not correct for all systems.
(Could probably be tied to the parent ```data``` file modified time which would update more but would be more correct for the actual length of the subjobs.)

milliams:
The problem is being caused by the `MassStorageFile` having no parent. I'm looking into it now.

milliams:
...which in turn is being made very difficult to debug because of the inconsistent use of `GangaList`s for schema attributes. I expect that I can fix this but really the whole system needs looking at.

rob-c:
OK, now I'm extremely concerned.
Bugger this could be the 'fix' for reading in lists in a way which doesn't create problems with GangaList.
(As to why this works on my machine then I don't know)
I'll prepare a test PR to see

rob-c:
There isn't inconsistent use of it as an attribute (until I put in a PR it's specifically not allowed to be in the Schema) but all schema objects are ALWAYS set to be GangaLists through ```__set__``` and there's nothing magic there.
I think this is a bug I introduced into VStreamer not to do with GangaList as a whole which is another set of PR but I suspect this makes the point we __need__ the GangaList in order to have correct object inheritance unless we do some horrible gymnastics.

Issue #320, opened 2016-03-16
String interpolation warning in the AutoFlusher


rob-c:
I'm seeing too much of this:

```
[16:56:49]
Ganga In [1]: Traceback (most recent call last):
  File "/afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc49-opt/lib/python2.7/logging/handlers.py", line 76, in emit
    if self.shouldRollover(record):
  File "/afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc49-opt/lib/python2.7/logging/handlers.py", line 156, in shouldRollover
    msg = "%s\n" % self.format(record)
  File "/afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc49-opt/lib/python2.7/logging/__init__.py", line 732, in format
    return fmt.format(record)
  File "/afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc49-opt/lib/python2.7/logging/__init__.py", line 471, in format
    record.message = record.getMessage()
  File "/afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc49-opt/lib/python2.7/logging/__init__.py", line 335, in getMessage
    msg = msg % self.args
TypeError: not all arguments converted during string formatting
Logged from file Registry.py, line 217
<LogRecord: Ganga.Core.GangaRepository, 10, /afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_Dev/install/ganga/python/Ganga/Core/GangaRepository/Registry.py, 217, "Auto-flushing">
None
```
All I was doing was running ```ganga --no-mon --debug```

I'll prepare a PR to fix this.

rob-c:
Also is there a strong reason for the auto-flusher non-blocking?

If a system is under heavy load we run the risk that this never flushes for us (or waits a very long time).

I'm tempted to play with this being a blocking lock to see how the performance is effected.

milliams:
[...]

PR #322, opened 2016-03-16
Fixing AutoFlusher logger string and making it blocking

rob-c:
#320 and have made auto-flusher blocking to see if it causes lockups. There is always the option to revert if problems arise.


python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -180,6 +180,7 @@ def __init__(self, registry, *args, **kwargs):
         super(RegistryFlusher, self).__init__(*args, **kwargs)
         self.registry = registry
         self._stop = threading.Event()
+        self.logger = getLogger()
```
milliams: Is there not already a global logger?

rob-c: There is, I just wanted to have a separate one for the flusher. Can revert this if there's a strong reason to use the same one as the Registry.

milliams: They will be indistinguishable from each other as they will have the same name I believe and in fact might even just be references to the same object.

rob-c:
Any objections to merging this in it's current state? I've incorporated changes from #320 into this branch and all of the tests are passing.

milliams:
Assuming that you're still not seeing any LHCb lock ups then I'm happy for this to be merged.

rob-c:
I'm not seeing any from 'normal' use in testing so I'm keen to get this fix in to stop the debug logger getting spammed with stacktraces if nothing else.

Issue #315, opened 2016-03-15
Monitoring Error: cannot import name queues


egede:
The report from Mike Sokoloff, had an error in the shuitdown as well, that I think we missed in previous discussions. See printout below.  The error is coming from python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py

```
Ganga.GangaLHCb.Lib.Files          : WARNING  !!! Please update your scripts before LogicalFile is removed
Ganga.Core.MonitoringComponent     : INFO     Some tasks are still running on Monitoring Loop
Ganga.Core.MonitoringComponent     : INFO     Please wait for them to finish to avoid data corruption
Ganga.Core.InternalServices        : INFO     Stopping Job processing before shutting down Repositories
Job status update or output download still in progress (shutdown not completed after 6 seconds).
1 background thread(s) still running: ['MonitoringWorker_1_1457975480358'].
Do you want to force the exit (y/[n])? Ganga.GPIDev.Adapters              : ERROR    Monitoring Error: cannot import name queues
Ganga.Core.MonitoringComponent     : ERROR    Monitoring Error: cannot import name queues
Ganga.Core.MonitoringComponent     : ERROR    Monitoring Error: cannot import name queues
```

rob-c:
@egede Can you find out the likely job type to lead this? (or is the user running tasks)

I suspect it could be a backend/tasks problem of trying to load the queues system after it's already been shutdown. I'll move the queues shutdown into the shutdown handler manually I think to try and avoid this in the future.

egede:
@rob-c I have not seen this myself (the user is using tasks though).

rob-c:
I suspect this is due to some worker thread being shutdown out of the order that is expected so I can protect against this with a few minor changes.

rob-c:
Given the queues system has moved around quite a few releases ago I think this is safe to close as nobody seems to be complaining that the queues are causing problems.
And we not try to access the queues directly rather than via the GPI so I suspect the original error applies to code we've taken out.

PR #323, opened 2016-03-16
Freezing queues before shutdown methods start

rob-c:
This addresses #315 by re-organizing the shutdown methods a little.


python/Ganga/Core/InternalServices/ShutdownManager.py
```diff
@@ -114,6 +114,10 @@ def add_priority(x):
 
     logger.info("Stopping Job processing before shutting down Repositories")
 
+    from Ganga.Core.GangaThread.WorkerThreads import _global_queues
+    if _global_queues:
+        _global_queues.lock()
```
milliams: A comment explaining the reason for this would be very useful. Otherwise in the future it will be found and my code smell detector will trigger at the unbalanced `lock`/`unlock`.

rob-c: Frankly the function need renaming but I'll get around to this, probably in this PR, I suspect ```freeze``` more correctly matches what is going on here anyway.

rob-c:
OK, I've changed the queues system to freeze the queue rather than _lock_ it. The effect is that no more things can be added to the queue during shutdown to prevent a queued task from spawning more queued tasks when the user wants to quit.

rob-c:

milliams:
This PR (I'm guessing) is causing a small problem with the tests suite in that once it's finished, it is printing
```python
Error in sys.exitfunc:
  File "/disk/homedisk/home/mpw/ganga/ganga/python/Ganga/Core/InternalServices/ShutdownManager.py", line 150, in _ganga_run_exitfuncs
    shutDownQueues()
  File "/disk/homedisk/home/mpw/ganga/ganga/python/Ganga/Core/GangaThread/WorkerThreads/__init__.py", line 33, in shutDownQueues
    delattr(Ganga.GPI, 'queues')
AttributeError: queues
```
to the terminal. You can see it in the ganga-ci output at https://ganga-ci.cern.ch/jenkins/job/PR%20builder/570/console too.

rob-c:
I'll put in a quick PR which adds the hasattr check but this possibly means there's a GangaUnitTest issue

PR #325, opened 2016-03-17
Changing NodeIndexCache

rob-c:
A short PR with a minor change to the IndexCache which has shown lockups in the past on ganga-ci. I'm keen to test it now on ganga-ci hence the PR.

If this works I'll update with a full description of what is going on.


python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -380,15 +380,23 @@ def removeNodeAttribute(self, attrib_name):
...
+            ## Fully loaded so lets regenerate this on the fly to avoid losing data
+            return self._getRegistry().getIndexCache(self)
+        ## Not in registry or not loaded, so can't re-generate if requested
         return self._index_cache
+                                                                 
+    def fullyLoadedFromDisk(self):
+        if self._getRegistry():
+            if self._getRegistry().has_loaded(self):
+                return True
+        return False
```
milliams: If the object has no registry then it probably is 'fully-loaded'.

rob-c: Good catch, but that doesn't fix the lockup I see in removing jobs with this in which I'm keen to explore

rob-c:
OK this PR addresses a bug which has annoyed me that the NodeIndexCache isn't dynamically regenerated during runtime to properly avoid problems.
This change used to trigger a lockup elsewhere in Ganga but this now appears to be fixed.

This fix includes some minor improvements in the use of getNodeIndexCache.

People should prefer the use of the ```LazyLoadJob<...>``` methods in the Job class when they expressly want to try and load things from the cache without breaking lazy-loading.
(The exceptions here being for the display methods and getting ```subjobs:status```. The former being the main reason for the cache, the latter should be used sparingly as loading the subjobs cache will trigger a load of the master job from disk. There are times when this is to be avoided.)

I've made some very minor tweaks to avoid constantly re-generating the cache when calling a function which lazy-loads an object, this doesn't really work without the rest of this PR.
This should give a slight performance boost as some areas of Ganga were accidentally abusing a runtime interpretation of a string. This was a minor bug, at worst this was a minor performance bottleneck in strange places in the monitoring (I think).
Although I expect the performance boost is probably lost in correctly re-generating the cache correctly when it's called now.

As well as dynamically recreating the cache on calling ```getNodeIndexCache``` (for objects already in memory) this adds some minor fixes to guarantee that an index file per loaded-job is written to disk on shutdown as well as adding some additional mocked up functionalist to the Internal tests.

Are there any objections to merging?
(modulo tests passing of course)

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -279,22 +281,22 @@ def index_load(self, this_id):
             logger.debug("Just silently continuing")
         return False
 
-    def index_write(self, this_id):
+    def index_write(self, this_id, shutdown=False):
```
milliams: What does this `shutdown` argument mean?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -354,15 +354,29 @@ def removeNodeAttribute(self, attrib_name):
         if attrib_name in self._data.keys():
             del self._data[attrib_name]
 
-    def removeNodeIndexCacheAttribute(self, attrib_name):
-        if self._index_cache and attrib_name in self._index_cache.keys():
-            del self._index_cache[attrib_name]
-
     def setNodeIndexCache(self, new_index_cache):
+        if self.fullyLoadedFromDisk():
+            logger.debug("Warning: Setting IndexCache data on live object, please avoid!")
```
milliams: Do we want to be more severe with this and raise an exception?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -354,15 +354,29 @@ def removeNodeAttribute(self, attrib_name):
...
+        ## Not in registry or not loaded, so can't re-generate if requested
         return self._index_cache
+                                                                 
+    def fullyLoadedFromDisk(self):
+        if self._getRegistry():
+            if self._getRegistry().has_loaded(self):
+                return True
+            else:
+                return False
+        return True
```
milliams: Could be:
```python
def fullyLoadedFromDisk(self):
    if self._getRegistry():
        return self._getRegistry().has_loaded(self)
    return True
```

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -354,15 +354,29 @@ def removeNodeAttribute(self, attrib_name):
...
 
-    def getNodeIndexCache(self):
+    def getNodeIndexCache(self, force_cache=False):
+        if self.fullyLoadedFromDisk() and not force_cache:
+            if self._getRegistry():
+                ## Fully loaded so lets regenerate this on the fly to avoid losing data
+                return self._getRegistry().getIndexCache(self)
+            else:
+                ## No registry therefore can't work out the Cache, probably empty, lets return that
+                return self._index_cache
```
milliams: Does an object which is not in a registry ever have an index cache? I can't think of a time where that would be true, especially given that the index cache is stored on disk *in* the registry.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -354,15 +354,29 @@ def removeNodeAttribute(self, attrib_name):
...
-        if self._index_cache and attrib_name in self._index_cache.keys():
-            del self._index_cache[attrib_name]
-
     def setNodeIndexCache(self, new_index_cache):
+        if self.fullyLoadedFromDisk():
+            logger.debug("Warning: Setting IndexCache data on live object, please avoid!")
         setattr(self, '_index_cache', new_index_cache)
 
-    def getNodeIndexCache(self):
+    def getNodeIndexCache(self, force_cache=False):
```
milliams: This needs a docstring describing the arguments.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -52,6 +52,10 @@ def lazyLoadJobApplication(this_job):
 
 def lazyLoadJobObject(this_job, this_attr):
 
+    if stripProxy(this_job)._getRegistry():
+        if stripProxy(this_job)._getRegistry().has_loaded(stripProxy(this_job)):
+            return getattr(this_job, this_attr)
```
milliams: This function should never be called from the proxy layer so the `stripProxy` should not be needed.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
```diff
@@ -241,6 +241,9 @@ def test_h_testXMLIndex(self):
             index_cat = raw_j._category
             this_index_cache = (index_cat, index_cls, index_cache)
 
+            print("just-built index: %s" % str(this_index_cache))
+            print("from disk: %s" % str(obj))
```
milliams: No need for `str()`.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: that it's run during shutdown as per convention elsewhere in the registry/repo where this has been used

python/Ganga/GPIDev/Base/Objects.py
rob-c: No, not for now, I don't think we're doing this anywhere out of reading in the index anyway when it's not fully loaded. Once we know we're safe we can consider this but I'm not comfortable doing this for now.

python/Ganga/GPIDev/Base/Objects.py
rob-c: As do a lot of the functions elsewhere

python/Ganga/GPIDev/Base/Objects.py
rob-c: Yes it has an index cache of ```{}```. We'd rather return whatever is there just in case as this makes sense :)

python/Ganga/GPIDev/Base/Objects.py
rob-c: Probably, I'll see about reducing it

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: but my intention is that this may be called from the proxy layer in the future and for testing. I don't see anything intrinsically bad about this?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
milliams: Ok, a docstring explaining the argument is needed then.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
rob-c: It's a test, and yes there is isn't there as this is a dictionary which for me didn't directly translate without this as I got an error

python/Ganga/GPIDev/Base/Objects.py
milliams: Indeed, so let's add them as we edit functions, especially if they are adding new arguments.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: as per elsewhere in the file yes

python/Ganga/GPIDev/Base/Objects.py
rob-c: Wouldn't it be far more productive to add them in one go, one file at a time and not end up with numerous commits and recommits fixing typos around the code

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: Not intrinsically bad, no, but a layering violation of sorts. Keeping a clear separation of proxy and non-proxy is a good idea. Maybe in this case it would be served better by a simple `job = stripProxy(this_job)` on the first line of the function to reduce the number of overall calls?

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
milliams: It is defined as part of the Python language that `"%s" % a` is equivalent to `str(a)`. I would be *very* interested in any example where this does not work.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Yes the number of overall calls can be reduced, a minor change at the start of the function then

python/Ganga/GPIDev/Base/Objects.py
milliams: No, I think it makes the most sense to add them as we go. No one is going to sit down and take the time to understand the inner working of every function simply to write a docstring, however if someone is working on the function then they best understand it and writing a docstring at that time is most useful.

On top of this, we should not be adding any *new* arguments to functions or adding new functions without docstrings.

python/Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
rob-c: Yes this is needed:
```
52rcurrie@lxplus047:->py.test Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py
============================================================================== test session starts ==============================================================================
platform linux2 -- Python 2.7.9, pytest-2.8.7, py-1.4.31, pluggy-0.3.1 -- /afs/cern.ch/sw/lcg/releases/LCG_79/Python/2.7.9.p1/x86_64-slc6-gcc49-opt/bin/python
cachedir: ../.cache
rootdir: /afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_Dev/install/ganga, inifile: setup.cfg
collected 9 items

Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_a_JobConstruction PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_b_JobXMLExists PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_c_XMLAutoUpdated PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_d_XMLUpdated PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_e_SubJobXMLExists PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_f_testXMLContent PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_g_testSJXMLContent PASSED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_h_testXMLIndex FAILED
Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py::TestSJXMLGenAndLoad::test_i_testSJXMLIndex PASSED

=================================================================================== FAILURES ====================================================================================
____________________________________________________________________ TestSJXMLGenAndLoad.test_h_testXMLIndex ____________________________________________________________________

self = <GPI.newXMLTest.TestSJXMLGenAndLoad.TestSJXMLGenAndLoad testMethod=test_h_testXMLIndex>

    def test_h_testXMLIndex(self):
        # Check index of job
        from Ganga.Core.GangaRepository.PickleStreamer import to_file, from_file

        from Ganga.GPI import jobs

        j = jobs(0)

        assert path.isfile(getIndexFile(j))

        with open(getIndexFile(j)) as handler:
            obj, errs = from_file(handler)

            assert isinstance(obj, tuple)

            from Ganga.GPIDev.Base.Proxy import stripProxy, getName
            raw_j = stripProxy(j)
            index_cache = raw_j._getRegistry().getIndexCache(raw_j)
            assert isinstance(index_cache, dict)

            index_cls = getName(raw_j)
            index_cat = raw_j._category
            this_index_cache = (index_cat, index_cls, index_cache)

>           print("just-built index: %s" % this_index_cache)
E           TypeError: not all arguments converted during string formatting

Ganga/new_tests/GPI/newXMLTest/TestSJXMLGenAndLoad.py:244: TypeError
```

Either we submit a bug report to py.test or wherever else is breaking part of the language or we translate this into a string. Otherwise the test __will__ fail.

milliams: This is happening because the object on the right-hand side is a tuple. Strictly, old-style Python string formatting works with `string % tuple` where the elements of the tuple are inserted into the string. Python is seeing *three* things that you want to inject into the string but only seeing *one* `"%s"`. For full safety with this doing `"%s" % (this_index_cache,)` will ensure the correct behaviour, otherwise using `.format()` is the modern recommended method.

See, for example, http://stackoverflow.com/a/1455623/96704

rob-c: Given nothing changes by having this as ```str()``` over changing pace I'm leaving it as it is so the test passes

milliams:
:+1: and thanks for adding all that documentation!

rob-c:
I strongly feel that adding documentation is much easier done in bursts, I don't like adding things as I go it ruins my workflow, but I can cope with adding it on a per-file basis, so now  a lot of it is out of the way.

rob-c:
retest this please

rob-c:
Any objections to merging?

milliams:
Go for it

PR #326, opened 2016-03-18
Fix object equality

### LINKED ISSUE #229: Job equality broken
There is a test in the legacy suite (*JobCopyConstructor.gpi*) which looks like:
```python
j = Job()
j2 = j.copy()
j3 = Job(j)

assert j == j2
assert j2 == j3
```
Naïvely I expected this to fail as things like the `id` would have changed but I checked the `Job`'s schema and it has `comparable=0` set on the `id`  (and some other items). 

Looking back at previous results I see that this test was failing in 6.0.44, passing in 6.1.14, still passing in 6.1.13 and failing again in 6.1.14.

The code as written should have this test passing as `GangaObject` equivalence ignores attributes with `comparable=0`. The failure for 6.1.16 is due to `postprocessors` not equating which is a bug which needs looking into.

The larger question is *what do we want object equality to look like?* In my opinion two `GangaObject`s should only be equal if *all* their schema attributes are the same. `__eq__` at the moment seems to be performing an "are these two objects a bit similar" test. If we really want this latter comparison then an `is_similar_to()` function might make sense.

Thoughts?

milliams:
This fixes #229 by slightly altering the definition of equality of `GangaObject`s. Previously when doing
```python
a == b
```
they would always be unequal if `b` was 'falsy' (i.e. things like `[]`, `0` and, in our case, an empty `MultiPostProcessor`).

This removes that check completely and (along with some slight tidying to `__eq__`) migrates over another (previously-failing) test.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -325,10 +325,7 @@ class Node(object):
         from Ganga.GPIDev.Base.VPrinter import VSummaryPrinter
         self.accept(VSummaryPrinter(level, verbosity_level, whitespace_marker, out, selection, interactive))
 
-    def __eq__(self, _node):
-
-        node = _node
-
+    def __eq__(self, node):
         if self is node:
             return 1
 
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -327,18 +327,18 @@ class Node(object):
 
     def __eq__(self, node):
         if self is node:
-            return 1
+            return True
 
         if not isinstance(node, type(self)):
-            return 0
+            return False
 
         # Compare the schemas against each other
         if (hasattr(self, '_schema') and self._schema is None) and (hasattr(node, '_schema') and node._schema is None):
-            return 1  # If they're both `None`
+            return True  # If they're both `None`
         elif (hasattr(self, '_schema') and self._schema is None) or (hasattr(node, '_schema') and node._schema is None):
-            return 0  # If just one of them is `None`
+            return False  # If just one of them is `None`
         elif not self._schema.isEqual(node._schema):
-            return 0  # Both have _schema but do not match
+            return False  # Both have _schema but do not match
 
         # Check each schema item in turn and check for equality
         for (name, item) in self._schema.allItems():
@@ -346,9 +346,9 @@ class Node(object):
                 #logger.info("testing: %s::%s" % (str(_getName(self)), str(name)))
                 if getattr(self, name) != getattr(node, name):
                     #logger.info( "diff: %s::%s" % (str(_getName(self)), str(name)))
-                    return 0
+                    return False
 
-        return 1
+        return True
 
     def __ne__(self, node):
         return not self.__eq__(node)
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -331,8 +331,6 @@ class Node(object):
 
         if self is node:
             return 1
-        if not node:  # or not self._schema.isEqual(node._schema):
-            return 0
 
         if not isinstance(node, type(self)):
             return 0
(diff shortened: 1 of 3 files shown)


python/Ganga/new_tests/GPI/TestJob.py
```diff
@@ -100,4 +103,12 @@ def testJobCopy(self):
...
\ No newline at end of file
+        self.assertEqual( j2.splitter.values, ['arg 1', 'arg 2', 'arg 3'])
+
+    def test_job_equality(self):
+        from Ganga.GPI import Job
+        j = Job()
+        j2 = j.copy()
+        j3 = Job(j)
+        assert j == j2
+        assert j2 == j3
```
rob-c: Can you check the behavior of un-proxied copies here or in another short test as well

milliams: Will do.

python/Ganga/new_tests/GPI/TestJob.py
```diff
@@ -100,4 +105,15 @@ def testJobCopy(self):
...
+
+    def test_job_equality(self):
+        """Check that copies of Jobs are equal to each other"""
+        from Ganga.GPI import Job
+        j = Job()
+        j2 = j.copy()
+        j3 = Job(j)
+        assert j == j2
+        assert j2 == j3
+        assert stripProxy(j) == stripProxy(j2)
```
rob-c: I was hoping to see ```assert j2 is j1.copy()``` for stripped jobs works or doesn't. Can we add this?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -325,25 +325,20 @@ def printSummaryTree(self, level=0, verbosity_level=0, whitespace_marker='', out
...
-    def __eq__(self, _node):
-
-        node = _node
-
+    def __eq__(self, node):
         if self is node:
-            return 1
-        if not node:  # or not self._schema.isEqual(node._schema):
-            return 0
+            return True
```
rob-c: This could cause problems with the Schema which makes a lot of use of ```0``` and ```1``` rather than ```True``` and ```False```

python/Ganga/new_tests/GPI/TestJob.py
milliams: I'm not sure I understand. `assert j2 is j1.copy()` will *never* be true for any formulation of how this works. These tests are only checking object equality, not object identity.

python/Ganga/GPIDev/Base/Objects.py
milliams: It should cause no problems as `1 == True` and `0 == False`. We should be safe to make changes like this incrementally across the system.

python/Ganga/new_tests/GPI/TestJob.py
rob-c: Apologies I was thinking of something else when I typed this I think. Ignore the noise

python/Ganga/GPIDev/Base/Objects.py
rob-c: Yes as long as there are no ```is True``` and ```is ``` checks I agree, but I'm a little concerned that this may occur around the Schema as it appears a valid check to make.

milliams: True, but there should not be any `(a == b) is 1` checks. Since all this is doing is changing `__eq__` so that it conforms to the [Python Data Model](https://docs.python.org/2.7/reference/datamodel.html?highlight=__eq__#object.__eq__) ("By convention, `False` and `True` are returned for a successful comparison.") there should be no conflict.

In the future I would like to change the Schema so that it used `True` and `False` rather than `0` and `1` but I agree that it will need some care. It would be nice to finally start using a feature that's been around for 13 years now.

rob-c: I'm absolutely all for the change I'm just weary that this could effect something not covered in the current unittests.

rob-c:
If everything works after the update to develop has been tested then go for merge from me

PR #328, opened 2016-03-18
Fix queue shutdown

rob-c:
The 1 line to fix the annoyance from shutting down queues For sake of waiting I'll wait for the tests to finish then push.


rob-c:
I'll merge this once the tests are passing and #329 is in

PR #329, opened 2016-03-18
Trying to fix issue with lists and parents

rob-c:
Try for fix for the parent not being set in SJID test. If this works I'll update this.


rob-c:
OK this fixes the tests so once they're fixed I'll merge and then rebase my other PR to use develop

PR #333, opened 2016-03-22
j.inputdata is a 'datasets', not a sequence

milliams:
Resetting the value of the `inputdata` parameter here to a list is incorrect as `inputdata` is not a sequence but rather a `ComponentItem` with a `defvalue` of `None`.


rob-c:
Looks ok to me :)

PR #334, opened 2016-03-22
Tidy up logging module

milliams:
Primarily this is removing the `_get_logging` and `_get_handlers` functions which were used before to get access to the standard library `logging` module. This is replaceable with `absolute_import`.

Additionally it rearranges some imports, deletes unused code and removes unused arguments from some functions.

I realise we are moving towards deprecating this module but in the meantime it's worth reducing the amount of cruft.

### DIFF
--- a/python/Ganga/Utility/logging/__init__.py
+++ b/python/Ganga/Utility/logging/__init__.py
@@ -1,4 +1,4 @@
-from __future__ import print_function
+from __future__ import print_function, absolute_import
 
 ##########################################################################
 # Ganga Project. http://cern.ch/ganga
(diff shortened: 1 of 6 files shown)


python/Ganga/Utility/logging/__init__.py
```diff
@@ -80,7 +59,7 @@ def _get_handlers():
 private_logger = None  # private logger of this module
 
 # main logger corresponds to the root of the hierarchy
-main_logger = _get_logging().getLogger()
+main_logger = logging.getLogger()
```
rob-c: Are we certain that the ```main_logger``` is never multiply defined when the class is imported multiple times?

I had a lot of grief getting the logger to shut down correctly as the final object in the Shutdown manager a while ago (hence a lot of the changes for ```_get_logging``` and I'd like to make sure it gracefully shuts down now)

rob-c:
I'd be tempted to remove https://github.com/ganga-devs/ganga/pull/334/files#diff-cafdab496baeebb8d12db8dfdd0c8e8eR313 While we're here making changes. I often end up having to do this to work out where an actual error comes from.
(I run my logging formatted with line numbers and the 'filename' as parsed by ganga)

python/Ganga/Utility/logging/__init__.py
milliams: A module will not be imported multiple times as it is cached in `sys.modules`. Unless we have a very messed up `PYTHONPATH` we should be safe.

milliams:
Yes, I've often had to do the same to be able to track down the source of errors but we need to ensure that:

1. It doesn't make the printout too verbose for users
2. it doesn't break anything where users have set custom logging levels for various modules.

I agree that in general `_guess_module_logger_name` could use some simplification but I probably don't have a good enough understanding of it to do so just yet.

rob-c:
I'll add a config option to control that line if we don't want to inflict it upon users, but I strongly think it would help speed up understanding what has gone on when users post their terminal capture asking for help.

rob-c:
I'd prefer to see the 'remove filename' from the logger removed unless there are strong objections?
Other than this I'd say go for merge :+1:

milliams:
I don't have an objection in principle but I'll leave it to another PR to avoid holding this one up. It might cause some discussion so I'd rather keep that separate.

rob-c:
In the interest of moving forward OK, I'll open another PR for that

Issue #52, opened 2015-11-23
Remove the reexec machinery


milliams:
We've discussed several times about getting rid of the ability for Ganga to re-exec itself. Historically this was needed in order to tweak `LD_LIBRARY_PATH` for linking against specific version of numpy but this is not relevant any more.

At the very least I would like to change the default to not re-exec but if possible I would like to completely remove the code from `bootstrap.py`. Given that LHCb has been running for years with `GANGA_NEVER_REEXEC` set without problems, I don't think there should be any trouble.

Any complaints?

drmarkwslater:
Though I'm completely in favour of this I believe that in Atlas it's still needed. I could be wrong but I would need time to test it to be sure. Could we postpone this for 6.1.15 please?

alexanderrichards:
nothing but support here :wink:
 I'm actually surprised that LHCb is running without this as I though that they needed this for some obscure reason.
One question, what was the reason that we needed the specific version of numpy? and why is that no longer an issue?

milliams:
@alexanderrichards I think the reason was we were packaging our own externals which were compiled for a particular libc ABI and compiler version etc. When we do an 'import numpy' it needs to be able to look in the right place for support libraries otherwise you would get a crash. That is how it was explained to me anyway and I wasn't that convinced. Either way, I can't see a good reason for it any more. We no longer use numpy anywhere as it was only a dependency of matplotlib which was used in GangaPlotter.
@drmarkwslater Sure, we can postpone this if necessary.

alexanderrichards:
excellent, get rid of it as soon as then

milliams:
It looks like ATLAS, in its `PACKAGE.py` is giving access to external tools by actually appending the environment variable `PATH` so that when Ganga is re-execd it is running in an environment where doing a `subprocess.Popen` will find the tools.

This is a very heavy-handed way of running an external tool, especially when we can just pass the necessary environment to `Popen` with the `env` argument.

This goes hand-in-hand with my assertion that `PACKAGE.py` should only be used for Python packages, not for adding command-line tools to the `PATH` etc. Paths to tools should be applied at-use and not rely on `os.environ['PATH']` being set correctly.

rob-c:
Is it possible to disable this by default in the config yet or do we still need some environment variabele to be passed by default?
If so can we change the cvmfs scripts to do this in future. Having this code always running concerns me as it causes horrible things like cProfile to give up the ghost.

milliams:
[...]

PR #335, opened 2016-03-22
Don't re-exec Ganga when running tests

milliams:
Re-execing is now not needed for most of Ganga and this was breaking the logging while running tests.

This change allows all logging messages to be correctly captured by the test framework.

@rob-c I believe this addresses the complaint you had last week about logging within tests.

### DIFF
--- a/python/Ganga/new_tests/GPI/GangaUnitTest.py
+++ b/python/Ganga/new_tests/GPI/GangaUnitTest.py
@@ -34,6 +34,7 @@ def start_ganga(gangadir_for_test, extra_opts=[]):
     logger.info("Parsing Command Line options")
     this_argv = [
         'ganga',  # `argv[0]` is usually the name of the program so fake that here
+        '--no-rexec',  # Don't re-exec Ganga when running tests
     ]
 
     # These are the default options for all test instances


rob-c:
@milliams My complaint was that for failed tests the whole stdout/stderr isn't printed anywhere (or wasn't for some reason in some test I was looking at)

Either way, yes, keen to see this go, forgot this was enabled by default. Are we still looking to drop the re-exec code in the future?

milliams:
For a failed test, the stdout is printed in a `-------- Captured stdout call -------` section by pytest. The problem was that all the logging messages emitted during the test were being filtered away and never directed to stdout.

I'm hoping to drop the re-exec as soon as possible. The thing stopping it happening already is that the ATLAS code relies on it for setting the PATH for running external commands. Of course technically this is fixable by passing an appropriate environment to `subprocess` but someone needs to do the actual work. See #52 for discussion on this.

rob-c:
Can we look to disable this by default actually and change the Atlas config to still use it?

milliams:
That's a good idea. I will take a look into this.

rob-c:
For the time being I think we can safely merge this PR (It's only for the tests), disabling the re-exec by default sounds like another PR probably.

PR #336, opened 2016-03-23
Improve stacktracer

milliams:
Due to an oversight on my part, every single test was waiting up to 5 seconds for the stacktracer to finish its loop before exiting. It's a daemon thread and we're about to delete its file anyway so there's no point in waiting.

This error was causing the integration tests to take 17 minutes on my local machine. With this PR, they now only take 7 and a bit.

### DIFF
--- a/python/Ganga/Core/InternalServices/ShutdownManager.py
+++ b/python/Ganga/Core/InternalServices/ShutdownManager.py
@@ -40,6 +40,8 @@ Extend the behaviour of the default *atexit* module to support:
 
 import atexit
 
+from Ganga.Utility import stacktracer
+
 
 def _ganga_run_exitfuncs():
     """run any registered exit functions
@@ -159,6 +161,9 @@ def _ganga_run_exitfuncs():
     from Ganga.Core.GangaRepository.SessionLock import removeGlobalSessionFiles, removeGlobalSessionFileHandlers
     removeGlobalSessionFileHandlers()
     removeGlobalSessionFiles()
+
+    if stacktracer._tracer:
+        stacktracer.trace_stop()
 
     from Ganga.Utility.logging import requires_shutdown, final_shutdown
     if requires_shutdown is True:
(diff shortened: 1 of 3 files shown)


rob-c:
I'm slightly worried by the potential bug this might have exposed, could it be transient or is this a symptom of the XML not being flushed between 'completed' and the next test?

I've just realised you're shutting down this tracer before the repositories and queues are shutdown in the Shutdown manager. Would it be better to explicitly call this before the shutting down of the logging there?

milliams:
Are you referring to the failing `TestSJXMLGenAndLoad` test? It looks to me like the test is (slightly) incorrect. It's comparing the index cache dictionary of a running job to the index cache loaded directly from disk. What's happening here is it's loading it from disk (using `from_file()`) and then in the meantime the job progresses to running. Thus, when the comparison is made, they are different.

This probably wasn't showing before because the 5 second delay on shutdown meant the job was already in `running` or `completed` before this test even started.

rob-c:
I think a pause to wait for the job to complete in this test would help, I'll make a PR for that in a bit.

milliams:
That makes sense. I'm happy to wait for that before merging this PR?

rob-c:
I would prefer to see the shutdown of this stack-trace moved to after the Repositories and Queues have both shutdown as these systems have shown lock-up behaviour numerous times before during shutdown.

milliams:
So, do you think that moving the call to `stacktracer.trace_stop()` from being managed by an exit handler to being explicitly called in `_ganga_run_exitfuncs()` near the end after all the queues etc., have finished makes sense?

rob-c:
Yes please, I think it makes sense to do it just before the logging is shutdown to catch any Ganga modules which may have problems.

I think we should probably try and register everything and examine the order in which shutdown tasks are registered/run but unfortunately hard-coding the solution in ```_ganga_run_exitfuncs()``` has been the quickest solution for now.

PR #337, opened 2016-03-23
Allowing auto-extraction and creation of bz2 repositories

rob-c:
This PR adds the ability to create and extract bz2 compressed archives.

I was hoping to expand this to make use of xz, however this appears to only be supported in Python3.

This does clean up a few minor bits of code around extracting archives and adds a small amount of safety when trying to extract a file to make sure sure we don't try to parse archives we don't understand.


python/Ganga/Lib/Interactive/InteractiveScriptTemplate.py
```diff
@@ -37,7 +37,8 @@
     os.makedirs('###WORKDIR###')
 os.chdir( '###WORKDIR###' )
 for inFile in ###IN_BOX###:
-    getPackedInputSandbox( inFile )
+    if str(inFile).endswith('.gz') or str(inFile).endswith('.bz2') or str(inFile).endswith('.tgz'):
```
milliams: Are the files not always `str`s? I can't see a route by which they are not.

python/Ganga/Core/Sandbox/Sandbox.py
```diff
@@ -56,7 +56,15 @@ def createPackedInputSandbox(sandbox_files, inws, name):
...
 #
 # Future release with tarball module
-    with open(tgzfile, 'w') as this_tarfile:
+
+    if str(tgzfile).endswith('.gz') or  str(tgzfile).endswith('.tgz'):
+        file_format = 'gz'
+    elif  str(tgzfile).endswith('.bz2'):
+        file_format = 'bz2'
+    else:
+        file_format = ''
```
milliams: It would be nice to be able to use the [`mimetypes`](https://docs.python.org/2.6/library/mimetypes.html) module to do this.

rob-c: File doesn't exist yet though, would ```mimetypes``` be of any help?

python/Ganga/Lib/Interactive/InteractiveScriptTemplate.py
rob-c: I'll look into changing it in a bit

python/Ganga/Core/Sandbox/Sandbox.py
milliams: `mimetypes.guess_type()` works only on the file name (it also works on remote URLs) so it would be usable. If it doesn't make the code any more readable or short then I'm happy to stick with the manual file extension checking.

rob-c: And annoyingly doesn't return ```gz``` or ```bz2```... I'll do some digging but unless I can just reduce this to a single line I'm not seeing any advantages.

rob-c:
OK, I've updated this PR to use mimetypes and it's a little neater as a result. This also wraps the extraction of compressed sandboxes around effectively a type check which should 'get around' the stale AFS lockfiles reported a while ago when submitting some jobs.

milliams:
I'm happy for this to be merged whenever. Cheers.

PR #338, opened 2016-03-23
Remove, removal of filename

rob-c:
As per discussion on #334 I'm keen to get rid of this line of code. This adds a small amount of verbosity back to the user but would make fixing things much quicker easier in the long term for us devs I suspect.


milliams:
So that everyone is clear: we currently have logging messages printed like:
```
Ganga.Core.InternalServices        : INFO     Internal services reactivated successfuly
```
and this PR would change them to be:
```
Ganga.Core.InternalServices.Coordinator : INFO     Internal services reactivated successfuly
```
where the last segment of the `a.b.c` is the name of the Python file (with the `.py` removed). Until now all logging from a directory of files were printed under the same name.

@rob-c We will also need to change the `_formats` dict in the logging module. At the moment it adds pads the name column to 35 spaces but with these longer names it will need to be significantly more. I'm not sure what the longest module name is but some are quite long, e.g. `Ganga.Core.GangaThread.WorkerThreads.ThreadPoolQueueMonitor` (59 characters).

rob-c:
@milliams If we're agreed to remove this we should decide on the 'typical' width of the logger.
(I don't have a strong bias on this as I'm using and have been using a custom format for quite some time to track thread-name and timestamps.)
I'm also quite keen to drop the leading ```Ganga.``` but this is likely something for another PR.

alexanderrichards:
I guess I have two points to input here. Firstly, this new proposed way would more closely match the prescription recommended in the python logging docs of using a module level `getLogger(__name__)` so I think in that sence (especially as it would make debugging easier) it sounds like a good idea. 

I would be weary of making the logger name column too long in the logging output however. It's worth considering @rob-c suggestion of keeping the most significant part (i.e. the latter part) and dropping the `Ganga` prefix. But then again knowing which package the logging message comes from might also be important.

In short I'm rambling because I don't really know what the right answer is, open to discussion I guess.

rob-c:
I'm willing to try and remove ```Ganga.``` but I suspect this could cause some problems with the config system changing on people so I don't know if it's worth it

egede:
I find this long initial part of logger messages annoying already. To make it even longer would not be helpful. Can we just let this go to the `.ganga.log` file and then try to change people to post extracts from that one instead?

milliams:
Ignoring test modules, the longest module names we have are (based on `find . -name "*.py" | sed -e "s/^\.\///" -e "s/\//\./g" -e "s/\.py$//" | awk '{print length, $0}' | sort -n -s | grep -Ev "\.(new_)?tests?\." | tail`):
```
52 GangaND280.ND280RecoValidation.oaReconPlusoaAnalysis
53 Ganga.Core.GangaThread.WorkerThreads.WorkerThreadPool
53 Ganga.GPIDev.Lib.File.scripts.MassStorageFileWNScript
53 GangaLHCb.Lib.Applications.options.GaudiPythonExample
53 GangaLHCb.Lib.RTHandlers.LHCbGaudiDiracRunTimeHandler
53 GangaND280.ND280TPCGasInteractions.TRExPlusOAAnalysis
56 GangaAtlas.Lib.AtlasLCGRequirements.AtlasLCGRequirements
57 GangaTest.Lib.TestMonitoringService.TestMonitoringService
59 Ganga.Core.GangaThread.WorkerThreads.ThreadPoolQueueMonitor
60 Ganga.Core.GangaRepository.GangaRepositoryImmutableTransient
```

I worry that removing `Ganga.` will both confuse the logging system hierarchy and also break people's custom logging config. It will also mean that we can't tell between `GangaLHCb` and `GangaDirac`.

alexanderrichards:
how about a solution whereby we leave the current behaviour for the screen output but we can add the full name to the log file? This probably gives you the best of both worlds if as @egede says we then ask people for their log output

rob-c:
Yes but due to our design GangaLHCb gets ```Ganga.GangaLHCb``` (which I find quite a pain)...

OK this is clearly looking like personal taste here as there are too many conflicts. I'll add the option to turn this on/off in the logger from the config.

Is there a reason we ever really want the path to the file rather than the file-name anyway?

egede:
@rob-c You want the path to the file when you have to work out what to put into your `.gangarc` file to make some certain type of message less verbose.

rob-c:
@egede I meant on the interactive logger if we're considering making a change to between what is on the screen and in the log file do we want to put the full path on the screen if it's often (too) long?

egede:
@rob-c For the screen, I think that everything before the colon could probably be omitted. It will take a little bit of education to tell users that they can find the details in the log file on disk, but it will reduce bloat on the screen.

rob-c:
OK, I think the conclusion is that there is some level of noise we want to reduce from reaching the screen whilst still going to the disk logger. Whether we're using the same format for both the file logger and screen I don't know without checking but I can modify this to be different I should have thought.

milliams:
Having the log file use a more verbose format and the screen logger use a terse one is relatively simple. In this case it's important to distinguish the 'name' of the logger from what is printed. The name of a logger in python should follow a pseudo module layout with dot separated components. To this end I think that @rob-c's proposed change in this PR is a good idea.

The file logger (see `_make_file_handler`) already uses the `VERBOSE` logger and the screen loggers (we have one for stderr and one for stdout) use the `NORMAL` logger. All we should need to do is change the `NORMAL` logger to just be `'%(levelname)-8s %(message)s'` and we're where we want to be.

rob-c:
I'm playing around with this atm. I'm think I'm going to propose that the ```file_handler``` be set to use the ```pathname``` in the logger format. (https://docs.python.org/2/library/logging.html)
I think it's probably perfectly acceptable to 'hard-code' the file logger format as it's not (easily) currently editable. I think I'll bump the file logger to have some extra information over the ```VERBOSE``` format atm.

I'm playing around with making the ```name``` of the logger something different to just the path of the file. I think I'm adding some minor checks to make sure that the name is never longer than 35 characters whilst keeping the filename intact. e.g:
```GangaTest.Lib.TestMonitoringService.TestMonitoringService : INFO     SomeMessage```
becomes:
```GangaTest.L---TestMonitoringService : INFO     SomeMessage```
this means that users would change the level of this logger by having:
```GangaTest.L---TestMonitoringService = DEBUG```
in their ```.gangarc```

The name of the logger is what is setup by the config system so there will have to be some modification from some users if they've heavily modified their logging format, although this is a minor modification imo.

I'll sort out an update to the PR at some point today.

**Comment by rob-c** [comment]
OK, I've put in some changes if people are interested in reviewing the logger changes.

**Comment by milliams** [comment]
I'm not sure that we want to be altering the `name` of the logger objects like this. The `name` should be almost an internal implementation detail of the logger which is independent of what is printed to the screen or file. 

As Ulrik said, for the stdout logger we can probably just not print the name at all and for the file logger, print the whole name:
```python
_formats = {
    'DEBUG': '%(asctime)s "%(filename)s":%(funcName)-10s at %(lineno)d, %(threadName)s: %(levelname)-8s %(message)s',
    'VERBOSE': '%(asctime)s %(name)-35s: %(levelname)-8s %(message)s',
    'NORMAL': '%(levelname)-8s %(message)s',
    'TERSE': 'Ganga: %(levelname)-8s %(message)s'
}
```
should be all we need.

If we really want to print some elided form of the module name in the screen log then a [custom formatter](https://docs.python.org/2.6/library/logging.html#formatter) would be a much cleaner idea.

rob-c:
@milliams Yes but then I have worries that working out where the errors/warnings come from becomes very difficult

milliams:
True, but we as developers can run with any format we like, we just have to set it to print the name in our output.

rob-c:
@milliams yes, but good luck in dealing with the DAST list when you can't remember where a message comes from, it just seems like a bad route to go down imo.

I'd maybe move to use ```filename``` over ```name``` rather than not having the name of the object.

I can't think what to to control the logging level on a per-file basis other than 1 logger per file, maybe we have a CustomFormatter which compares the path on the message before displaying it to have a custom logging level. This seems like an evolution but somehow a separate full PR.

milliams:
I was suggesting it based on what Ulrik had described.

Using `module` rather than `name` (for the screen logger) could also work. It looks quite neat and gives us most of the information we need.

rob-c:
Isn't ```module``` effectively what we're inventing for the ```name``` now anyway.
And this can still end up _too long_ imo.

OK, maybe we drop any reference to filename in the default this case then since it's contested?

milliams:
I tested it and `module` seems to just be the last part of the `name`, i.e. the filename without the `.py`. So you get messages like:
```
ShutdownManager      : INFO     Stopping Job processing before shutting down Repositories
ShutdownManager      : INFO     Shutting Down Ganga Repositories
Repository_runtime   : INFO     Registry Shutdown
```
We can then allow `name` to be set to long things like `Ganga.Core.GangaThread.WorkerThreads.ThreadPoolQueueMonitor` as that will only display as `ThreadPoolQueueMonitor`.

rob-c:
OK, I'm happy to move to ```module``` in this case.
At some point later today I'll open an issue to describe the requirements for a custom formatter which can replace most of this and allow us to move to use one global logger instance with any luck.

drmarkwslater:
retest this please

drmarkwslater:
retest this please

drmarkwslater:
retest this please

drmarkwslater:
retest this please

rob-c:
Just looking at this again.

How do people feel about the logger being formatted such as:
```
Job                                :submit               INFO     submitting job 25
Job                                :updateStatus         INFO     job 25 status changed to "submitting"
Executable                         :prepare              INFO     Preparing Executable application.
Executable                         :prepare              INFO     Created shared directory: conf-4dee5242-8e10-4784-afb3-6f3559bdbea3
Job                                :_getJobSubConfig     INFO     Preparing subjobs
IBackend                           :master_submit        INFO     submitting job 25 to Local backend
Job                                :updateStatus         INFO     job 25 status changed to "submitted"
Job                                :updateStatus         INFO     job 25 status changed to "running"
Job                                :postprocess_hook     INFO     Job 25 Running PostProcessor hook
Job                                :updateStatus         INFO     job 25 status changed to "completed"
```

(possibly adjusting column widths)

egede:
@rob-c 

The
> Job                                :submit               INFO     

part of the message I really find rather useless for the daily user. In every single line of the example you give, the information is redundant with the message itself. I think it would be much less cluttered if all of the pre-ample only went to the `.ganga.log` file.

milliams:
I do think it's useful to have at least the `INFO`, `WARNING`, `ERROR` parts to make it clearer to the user when we really want their attention. I don't think we want the function name part (`submit`) as that's very much an implementation detail. As for the module name (`Job`) I'm not sure how useful it is to the user but I do think it helps give some context for the more ambiguous messages.

I think the best approach is just to get in the habit for us developers of using a custom verbose string for our own work and not flooding the users with info which is only useful for us.

rob-c:
@egede What part of the message do you think should be left in, should we drop everything but the message itself?
i.e. drop all columns but the last one?

btw the branch currently writes out almost the equivalent of the debug to the log file anyway to try and help with potential problems.
If we want to strip everything but the message in the default mode this should be really easy to implement.

egede:
@rob-c I agree with @milliams that the `INFO, WARNING, ERROR` part should be kept. Only keeping this makes also makes the formatting easier, as we know the maximal width. Please also keep the colour coding of this part that I find really helpful.

rob-c:
I think the colour coding is provided to us 'for free' as in it's not in our codebase.

I think dropping to:
```
INFO     submitting job 25
INFO     job 25 status changed to "submitting"
INFO     Preparing Executable application.
INFO     Created shared directory: conf-4dee5242-8e10-4784-afb3-6f3559bdbea3
INFO     Preparing subjobs
INFO     submitting job 25 to Local backend
INFO     job 25 status changed to "submitted"
INFO     job 25 status changed to "running"
INFO     Job 25 Running PostProcessor hook
INFO     job 25 status changed to "completed"
```

By default is quite nice.

I'm wondering if we want to reduce the debug output so it's a little less painful for users and developers alike, or are people happy to keep the ```--debug``` mode the way that it currently is?

milliams:
I think we want to be careful about the amount of `logger.debug` statements we put into the code base to avoid it flooding but I'm not sure there's any technical change we need to make. Very verbose debug output if very useful for debugging and if it's not then we should tweak the messages that are being written.

rob-c:
OK, so talking about the debug logging for a second, should we consider something more verbose than what we have now?
I've spent a bit of time killing statements which appear multiple times around methods we trust work as expected.
We currently have:
```'%(asctime)s "%(filename)s":%(funcName)-10s at %(lineno)d, %(threadName)s: %(levelname)-8s %(message)s',```
I'd like to move to something more close to:
```%(asctime)s %(threadName)s %(module)-10s:%(funcName)-10s:%(lineno)d %(levelname)-8s: %(message)s```

Basically introduce the line number, replace ```filename``` with ```module``` and re-arrange the order of some of the parts of the thread, does this make sense?

**Comment by milliams** [comment]
Makes sense to me.

**Comment by rob-c** [comment]
OK now I have:
```
[11:11:12]
Ganga In [1]: j=Job(); j.submit()
INFO     submitting job 39
INFO     job 39 status changed to "submitting"
INFO     Preparing Executable application.
INFO     Created shared directory: conf-7a8dfaee-25ab-49d8-99be-5b0583c98d50
INFO     Preparing subjobs
INFO     submitting job 39 to Local backend
INFO     job 39 status changed to "submitted"
Ganga Out [1]: 1
```

This is now also the same for ```--debug``` though as I realise this changes the logging level not the format...
We may want to consider changing the format with the logger when we run with debug logging as the user/developer is probably trying to debug something.

**Comment by egede** [comment]
> We may want to consider changing the format with the logger when we run with debug logging as the user/developer is probably trying to debug something.

@rob-c Yes, that is a quite good idea. It makes for a different behaviour to if you just at the interactive prompt do

```
config.Logging.Ganga='DEBUG'
```

but I do not think that matters.

**Comment by rob-c** [comment]
I'm changing this such that launching Ganga with ```--debug``` uses the debug level and format, by default the user gets the normal logging level and format and if the logging level is ever changed post-bootstrap then the format it unchanged.
There is currently no hook to allow the user to change the format of the logger whilst working and I don't think there should be. Allowing people to change the level is perfectly sane but changing the formatting during a session is confusing I fear.

**Comment by rob-c** [comment]
Any objections to merging this branch?

File: python/Ganga/Utility/logging/__init__.py [hunk 1]
```diff
@@ -48,10 +48,14 @@
...
-    'NORMAL': '%(name)-35s: %(levelname)-8s %(message)s',
-    'TERSE': 'Ganga: %(levelname)-8s %(message)s'
+#    'DEBUG': '%(asctime)s "%(filename)s":%(funcName)-10s at %(lineno)d, %(threadName)s: %(levelname)-8s %(message)s',
+    'DEBUG': '%(asctime)s %(threadName)s %(module)-20s::%(funcName)-20s:%(lineno)d %(levelname)-8s: %(message)s',
+#    'VERBOSE': '%(asctime)s %(name)-35s: %(levelname)-8s %(message)s',
+    'VERBOSE': '%(asctime)s %(module)-25s::%(funcName)-10s: %(levelname)-8s %(message)s',
+#    'NORMAL': '%(name)-35s: %(levelname)-8s %(message)s',
+    'NORMAL': '%(levelname)-8s %(message)s',
+#    'TERSE': 'Ganga: %(levelname)-8s %(message)s'
+    'TERSE': '%(levelname)-8s %(message)s',
```
milliams: I'm not sure we want to keep the old ones commented out. They're in version control if we need them.

rob-c: True, they were left in for comparisons while working. I'll re-pull the changes and test everything works again locally

rob-c:
I like the codecov-io comment but will this appear on all sets of tests that are run?

milliams:
Previously I had disabled this feature as I wanted to avoid the noise. It seems that codecov updated their service yesterday and so there's now a new way of managing the settings. I'm planning on re-disabling it for now as I think that it will likely post a new comment on every commit to a PR which may flood discussion.

The coverage report will still run but it will be accessible via https://codecov.io/gh/ganga-devs/ganga

If we decide later that we want the comments or the commit status stuff then we can re-enable it.

rob-c:
@milliams Is there a wiki page with the tests that are being run but don't report? I'd be keen to review them every now and then but don't think they should be posting on every set of automatic tests :)

PR #339, opened 2016-03-23
Fixing test

rob-c:
Fixing test which fails occasionally due to transitioning state.

This may itself fail as a fix if the test is inconsistent but I hope to find out in testing now.


rob-c:
Any objections to merge?

PR #342, opened 2016-03-28
New Applications:

VladimirRomanovsky:
BenderScript and Ostap


rob-c:
@VladimirRomanovsky Sorry for the ignorance, glancing at the code I see a lot of old references to the Bender application. Is there a dramatic difference between the 2 applications or can this be loaded in the same way that the other LHCb Gaudi Apps are using the Bender application as a template? I only ask as the changes for that are about 5-10 lines of python vs adding 2 new files of code.

VladimirRomanovsky:
From my point of view these two applications (Bender and BenderScript) are different. In one case Bender used as module in other it is script. 
The main difference of Ostap is absence of input data for it.

egede:
@VladimirRomanovsky When reviewing patches, we decided not to merge this one in its present state. Rather a review of how the `Bender/GaudiPython` applications work should be made with an eye for being used with the new CMake style release areas.

PR #345, opened 2016-04-01
Fix makegangalist semantics

drmarkwslater:
This PR fixes issues relating to the semantics of GangaLists and copies the tests over to make sure these semantics are tested against. The specific things addressed are:

makeGangaList - This now creates a GangaList with copy of the given list (but with the same items in it)
makeGangaListByRef - This now creates a GangaList with reference of the given list
Copy of List - contents are references (i.e. `gl[0] is gl2[0]`)
Deep copy of List - contents are NOT references (i.e. `gl[0] is not gl2[0]`)
Append/Extend - Contents are added as refs (i.e. contents not copied)

To get the above working, the changes I've put in are:

- `setatttr` used on GangaObject `__init__` as otherwise default lists do not have their parent set correctly
 - change the `_list` item from Component to Simple as otherwise copies (technically clones) of objects are made when assigning/extending lists
 - Added specific check for GangaList `_list` item when writing XML to make sure nested lists still work
 - Made sure proxies are stripped when comparing lists

All of this will get superseded by the new List classes but this should give us a base line on the semantics and also should be a start on getting Tasks working again.


python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -784,8 +788,8 @@ def __init__(self):
             for attr, item in self._schema.allItems():
                 ## If an object is hidden behind a getter method we can't assign a parent or defvalue so don't bother - rcurrie
                 if item.getProperties()['getter'] is None:
-                    defVal = self._schema.getDefaultValue(attr)
-                    self.setNodeAttribute(attr, defVal)
+                    setattr(self, attr, self._schema.getDefaultValue(attr))
```
rob-c: setNodeAttribute changes the object parent https://github.com/ganga-devs/ganga/pull/345/files#diff-3a5f5ed88f274eb570dccbab85337abfR360 doesn't it? Isn't this still required here?

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -87,7 +83,7 @@ class GangaList(GangaObject):
     _hidden = 1
     _enable_plugin = 1
     _name = 'GangaList'
-    _schema = Schema(Version(1, 0), {'_list': ComponentItem(defvalue=[], doc='The raw list', hidden=1, category='internal'),
+    _schema = Schema(Version(1, 0), {'_list': SimpleItem(defvalue=[], doc='The raw list', hidden=1, category='internal'),
```
rob-c: I'm assuming that nested XML now still work here (I think we have some tests for this but I'm just curious as this _was_ needed a whie ago, but I didn't understand why)

python/Ganga/new_tests/GPI/GangaList/TestMutableMethods.py
```diff
@@ -1,19 +1,14 @@
...
-from Ganga.GPIDev.Adapters.IGangaFile import IGangaFile
-
-from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
-
 import random
 import string
 import time
 
+from ..GangaUnitTest import GangaUnitTest
+from Ganga.GPIDev.Base.Proxy import isProxy, isType, TypeMismatchError
```
rob-c: I'd prefer to see these imports made within the test class as py.test can't know where to find these on import and checking the class until the path has been manipulated by the GangaUnitTest constructor.

That or we should fix the GangaUnitTest to modify the sys.path on import which we're currently not doing.

python/GangaTest/__init__.py
```diff
@@ -5,6 +5,8 @@ def getEnvironment(c = None):
     return {}
 
 def loadPlugins( config = None ):
+    import Lib.GListApp
```
rob-c: Why have these been moved to be imported from here, Is the GListApp within the GangaTest package?

python/GangaTest/Lib/GListApp/GListApp.py
```diff
@@ -18,13 +20,14 @@ class GListApp(IApplication):
     _name = 'GListApp'
     _schema = Schema(Version(1, 0), {
         'bound_print_comp': ComponentItem('files', defvalue=[], sequence=1, summary_print='_print_summary_bound_comp', typelist=['str', 'Ganga.test.GPI.GangaList.TFile.TFile']),
-        'bound_print_simple': SimpleItem(defvalue=[], sequence=1, summary_print='_print_summary_bound_simple'),
+        'bound_print_simple': SimpleItem(defvalue=GangaList(), summary_print='_print_summary_bound_simple'),
```
rob-c: I had problems in the past with doing this in several places without addressing some problems in the Schema, Is this object constructable from the interactive prompt? It could well be whatever was causing the problems I was seeing to go away.

python/Ganga/new_tests/GPI/GangaList/TestCopy.py
```diff
@@ -43,7 +36,7 @@ def testCopy(self):
         gl2 = copy.copy(gl)
         assert len(gl2) == len(gl), 'lists must be equal'
         assert gl2 is not gl, 'list must be copies'
-        #assert gl[0] is gl2[0], 'the references must be copied' ## rcurrie 09/12/15 Not sure this is valid given we have to cope with GangaObjects here which may not be copyable!
+        assert gl[0] is gl2[0], 'the references must be copied'
```
rob-c: so what is the parent of gl[0] and what is the parent of gl2[0]?
I think we expect in this case that the parent to both be None.
How about in the case of the GangaList being a member of a class object:
i.e.:
```
j=Job(); j.someArg = GangaList()
gl2 = copy.copy(j.someArg)
assert stripProxy(j.someArg[0])._getParent() is stripProxy(j)
assert stripProxy(j.someArg[0])._getParent() is not stripProxy(gl2[0])._getParent()
assert stripProxy(gl2[0])._getParent() is None
```

python/Ganga/GPIDev/Base/Objects.py
drmarkwslater: setNodeAttribute only sets the parent for values that are GangaObjects which fails (at least) when you have a default value of `[]` for a schema item as through the setter, this would get changed to a GangaList and the parent set correctly, but through setNodeAttribute does not. This then causes the `print_summary` to be ignored as well as other possible issues. There may be other cases where not running through the setter causes problems but this was the reason I put this change in.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
drmarkwslater: I'm also not sure if we really need nested GangaLists (I know I used to think we needed it for Tasks but after talking with Matt I don't think that's the case anymore). In any case though, we do have tests for it and the change to `Objects.py` for the `accept` method above ensures this still works.

python/Ganga/new_tests/GPI/GangaList/TestCopy.py
drmarkwslater: This is up for debate but my personal feeling is that this test is actually for under-the-hood stuff and so keeping the references is the correct and expected behaviour. As regards the user, they should never knowingly manipulate GangaLists anyway and when the set do a copy through the GPI, the GangaList should (and does I believe) do clones of the contained objects thanks to the proxy layer. However, in the Ganga code, I think it makes more sense for GangaList to behave like a python list. In addition, this was the behaviour before in 6.0.44 and earlier.

python/GangaTest/__init__.py
drmarkwslater: After some problems on getting the GListApp to import properly I though I may as well just put it in GangaTest with the other testing based classes. It made sense and made sure it was imported properly (with for example, the TestSubmitter)

python/Ganga/GPIDev/Base/Objects.py
rob-c: I would __MUCH__ rather see the setNodeAttribute fixed to be safe, or a the parent to be correctly assigned after the call to setattr. I know it's probably correctly set higher up in the copy process somewhere but I don't want to re-introduce places where we're not setting the parent as it was a pain to get things as _stable_ as they are with regards to this atm.
I'm rather concerned you say this breaks things. Does this simply break with a bad call to ```_setParent``` or does something else break in a new and exiting way here?

python/Ganga/new_tests/GPI/GangaList/TestCopy.py
rob-c: In explicit terms I understand what you mean is that the following is true:
```
j=Job(); j.someArg = GangaList()
gl2 = copy.copy(j.someArg)
assert stripProxy(j.someArg[0])._getParent() is stripProxy(j)
assert stripProxy(j.someArg[0])._getParent() is stripProxy(gl2[0])._getParent()
assert stripProxy(gl2[0])._getParent() is stripProxy(j)
```

I don't mind what the expected behavior was/is/will-be I'm only bugged that when we were having problems with the parent of objects not being set correctly this was causing some levels of problems.

I'm just bugged by the following being allowed:
```
j=Job(); j.someArg = GangaList()
gl2 = copy.copy(j.someArg)
gl2[0].someActionToChangeItsValue()
assert j.someArg[0] == gl2[0]
```

Although this is close to expected python behavior so we should probably just live with that.

I'm not sure that the user does ever get a clone of an object through the Proxy layer (they do mostly when it matters but not from a ```copy.copy```) and that they get the proxy of each object to play with but the test here shows they're the same proxy hence are pointing to the same object.

python/GangaTest/__init__.py
rob-c: I thought the general move was such that we're looking to get rid of ```GangaTest``` once we've moved all tests away from it so will the tests run without these changes?

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -292,14 +288,21 @@ def __deepcopy__(self, memo):
...
+    def __getListToCompare(self, input_list):
+
+        # if the arg isn't a list, just give it back
+        if not self.is_list(self.strip_proxy(input_list)):
+            return input_list
+
+        # setup up the list correctly
+        tmp_list = input_list
         if isType(input_list, GangaList):
-            return stripProxy(input_list)._list
```
rob-c: This is more correct as ```_list``` should never contain proxied objects. Changing this here is at least wasteful and potentially hiding problems elsewhere which need to be addressed.

python/GangaTest/__init__.py
drmarkwslater: Quite possibly, but at present, this seemed like the best place. If we want to move the testing classes elsewhere that's fine but at least they're all in one place so they can be moved together,

python/GangaTest/Lib/GListApp/GListApp.py
drmarkwslater: My bad - from debugging. Fixed.

python/Ganga/new_tests/GPI/GangaList/TestMutableMethods.py
drmarkwslater: No trouble. done.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
drmarkwslater: Fair point. Fixed.

drmarkwslater:
Hopefully I've addressed any issues. If so, I'd like to get this merged sooner rather than later as it's quite a significant step to fixing Tasks :smile:

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -191,7 +191,11 @@ def accept(self, visitor):
         visitor.nodeBegin(self)
 
         for (name, item) in self._schema.simpleItems():
-            if item['visitable']:
+
+            # Special check for Ganga list to allow nested lists
+            if name == "_list" and self._name == "GangaList":
+                visitor.componentAttribute(self, "_list", self._getdata("_list"), 1)
```
milliams: I feel that instead of having a special check, we should use inheritance and just override `accept` in `GangaList` to something like:
```python
def accept(self, visitor):
    visitor.nodeBegin(self)

    for (name, item) in self._schema.simpleItems():
        if name == '_list':
            visitor.componentAttribute(self, '_list', self._getdata('_list'), 1)
        elif item['visitable']:
            visitor.simpleAttribute(self, name, self._getdata(name), item['sequence'])

    for (name, item) in self._schema.componentItems():
        if item['visitable']:
            visitor.componentAttribute(self, name, self._getdata(name), item['sequence'])

    visitor.nodeEnd(self)
```
so that the `GangaList`-specific logic is contained within `GangaList`

python/Ganga/new_tests/GPI/GangaList/TestCopy.py
milliams: I don't expect that any user will ever do a `copy.copy` of any of our objects. I agree it will cause problems with parenting and perhaps on the proxy-level it might be worth prohibiting it and only allowing `deepcopy`. Most users will simply use our `.copy()` method we provide on proxy objects.

python/Ganga/GPIDev/Base/Objects.py
rob-c: Yes I think any special things for ```GangaList``` should be moved there and not into ```GangaObject``` :+1:

python/Ganga/new_tests/GPI/GangaList/TestCopy.py
rob-c: Disallowing a ```copy``` would cause more problems imo. I think the above behavior is sensible. It's impossible (I hope) to assign the reference to another which can be a parent without a deepcopy being made within Ganga atm and many tests would at least need restructuring with copy being disabled in front of the proxy.
I think if the user does this and is surprised it's a probably good learning opportunity for them to learn some Python features/behaviour.

python/GangaTest/__init__.py
rob-c: In that case I think a separate PR isolating the current tests again from the GangaTest is needed, there is some bleed already happening between these 2 codebases and I'd rather not end up fixing one at the expense of the other until we have the majority of all test in the new test system.

python/Ganga/GPIDev/Base/Objects.py
drmarkwslater: Totally agree - fixed!

drmarkwslater:
So after fixing some stupid errors on my part, this is now all working. Am I OK to merge or are there any more issues?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -23,7 +23,7 @@
 from Ganga.Core.GangaRepository.VStreamer import from_file as xml_from_file
 from Ganga.Core.GangaRepository.VStreamer import XMLFileError
 
-from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList, makeGangaListByRef
+from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
```
milliams: In fact, `GangaList` isn't used in this file either.

drmarkwslater: Removed :smile:

milliams:
:+1: This looks good to me. Has this started helping with the tasks fixes yet?

drmarkwslater:
I don't know - I haven't tried yet :smile:

rob-c:
Go for merge from me if things are working :D :+1:

PR #346, opened 2016-04-02
Stop using gpi

rob-c:
This is the result of a branch I've been playing with recently to remove all references to ```import Ganga.GPI``` or any variants there of.

This appears to pass all of the basic tests on my machine and I'm keen to push this upstream as it should clear up a few bits of bad practice in the code. I haven't touched the ```Tasks``` in ```GangaAtlas``` as I'm not familiar enough with the code there. The same applies for the ```GangaRobot```, however in this case there is a __LOT__ of ```from Ganga.GPI import *``` which is very un-needed in 99% of cases I feel.

There are a few bits which need testing/fixing but I'm keen to show people what I'm keen to merge.


python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
```diff
@@ -2,6 +2,9 @@
 _global_queues = None
 _queues_interface = None
 
+def getQueues():
+    return _global_queues
+
```
milliams: I'm not sure we need a getter function here. I think that just renaming `_global_queues` to `global_queues` and importing that would be sufficient.

python/Ganga/Core/InternalServices/Coordinator.py
```diff
@@ -52,9 +52,10 @@ def isCredentialRequired(credObj):
         return Workspace_runtime.requiresAfsToken() or Repository_runtime.requiresAfsToken()
 
     if getName(credObj) == 'GridProxy':
-        from Ganga.GPI import jobs, typename
+        from Ganga.Core.GangaRepository import getRegistryProxy
+        from Ganga.Runtime.GPIFunctions import typename
         from Ganga.GPIDev.Base.Proxy import stripProxy
-        for j in jobs:
+        for j in getRegistryProxy('jobs'):
```
milliams: Does this need to be a proxy-wrapped object. Could we not just access the job registry object directly?

python/Ganga/GPIDev/Adapters/IBackend.py
```diff
@@ -142,9 +142,9 @@ def handleError(x):
 
         if parallel_submit:
 
-            from Ganga.GPI import queues
+            from Ganga.Core.GangaThread.WorkerThreads import getQueues
 
-            threads_before = queues.totalNumIntThreads()
+            threads_before = getQueues.totalNumIntThreads()
```
milliams: This should be `getQueues().` if using the getter function.

python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py
```diff
@@ -185,13 +185,13 @@ def lookup(self, sharedir, unprepare=False):
         The optional parameter 'unprepare=True' can be set to call the unprepare method 
         on the returned objects.
         """
-        from Ganga.GPI import jobs, box, tasks
+        from Ganga.Core.GangaRepository import getRegistryProxy
         objectlist = []
-        for thing in jobs.select():
+        for thing in getRegistryProxy('jobs').select():
```
milliams: Again, do these need to be proxy-wrapped? We want to avoid using proxy-level objects in our code as much as possible as it increases the chance of accidentally infecting our code with more proxy-wrapped objects.

python/Ganga/GPIDev/Lib/Tasks/IUnit.py
```diff
@@ -104,7 +104,8 @@ def checkForSubmission(self):
             return False
 
         # if we're using threads, check the max number
-        if self._getParent().submit_with_threads and GPI.queues.totalNumUserThreads() > self._getParent().max_active_threads:
+        from Ganga.Core.GangaThread.WorkerThreads import getQueues
+        if self._getParent().submit_with_threads and getQueues.totalNumUserThreads() > self._getParent().max_active_threads:
```
milliams: `getQueues()`

python/Ganga/Lib/Batch/Batch.py
```diff
@@ -164,8 +164,7 @@ def submit(self, jobconfig, master_input_sandbox):
         if jobnameopt and job.name != '':
             # PBS doesn't like names with spaces
             tmp_name = job.name
-            from Ganga.GPI import PBS
-            if isType(self, PBS):
+            if self._name == "PBS":
```
milliams: The old check probably never worked as `self` would never be a proxy type. Either way, this should be changed to just:
```python
 if isType(self, PBS):
```
since `PBS` is a valid type in this scope.

python/Ganga/new_tests/GPI/Bugs/TestSavannah1006.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import absolute_import
+
+from ..GangaUnitTest import GangaUnitTest
+
+class TestSavannah10006(GangaUnitTest):
```
milliams: The filename should be `TestSavannah10006.py` (three `0`s) and the old test should be removed at the same time.

python/Ganga/Core/GangaThread/WorkerThreads/__init__.py
rob-c: Yes but I am undecided if I want to check for if queues have started or not and issue a start command if they have not.
This can only be done from a getter I think.

python/Ganga/new_tests/GPI/TestFolders.py
```diff
@@ -30,10 +32,11 @@ def test_addremove(self):
         jobtree.rm('/*')
 
     def test_clean(self):  # asaroka
+        from Ganga.GPI import jobtree, Job
         jobtree.cd()
         jobtree.mkdir('testdir')
         jobtree.add(Job())
-        print jobtree
+        print("%s" % jobtree)
```
milliams: Does `print(jobtree)` not work?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -710,12 +710,12 @@ def updateMonitoringInformation(_jobs):
...
         logger.debug('Requeue jobs    : ' + repr([j.fqid for j in requeue_jobs]))
 
-        from Ganga.GPI import queues
+        from Ganga.Core.GangaThread.WorkerThreads import getQueues
 
         # requeue existing completed job
         for j in requeue_jobs:
             #            if j.backend.status in requeue_dirac_status:
-            queues._monitoring_threadpool.add_function(DiracBase.job_finalisation,
+            getQueues._monitoring_threadpool.add_function(DiracBase.job_finalisation,
```
milliams: `getQueues()`

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -787,7 +787,7 @@ def updateMonitoringInformation(_jobs):
                     if job.master:
                         job.master.updateMasterJobStatus()
 
-                queues._monitoring_threadpool.add_function(DiracBase.job_finalisation,
+                getQueues._monitoring_threadpool.add_function(DiracBase.job_finalisation,
```
milliams: `getQueues()`

python/Ganga/Core/InternalServices/Coordinator.py
rob-c: There is different behavior. Besides the code was previously expecting a proxy wrapped object so it would be wrong to replace it without being careful.
I want to remove the raw calls to the GPI here and delay asking if it was even correct until a future PR fixing these issues.

python/Ganga/GPIDev/Adapters/IBackend.py
rob-c: Thanks, typo :)

python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py
rob-c: See above, I don't want to fix/change the code behavior here I want to fix the bad GPI calls and change the behavior later.
Although in this case, yes an un-proxied repo is probably fine.

python/Ganga/GPIDev/Lib/Tasks/IUnit.py
rob-c: Oops same typo again, thanks :)

python/Ganga/Lib/Batch/Batch.py
rob-c: I thought it should be. I'll modify it back to simply not pull in the GPI :)

python/Ganga/new_tests/GPI/TestFolders.py
rob-c: Never tried tbh but I prefer this notation myself.

python/Ganga/new_tests/GPI/Bugs/TestSavannah1006.py
rob-c: Ah, I'll make a '''git mv''' which should work correctly.

rob-c:
Any objections to merging this? It reduces the dependency of the Core on the ```GPI``` which means a few less problems are now exposed by using ```import ganga``` in testing. I'm reluctant to change any code behavior here this is only to remove the use of bad ```GPI``` calls from core code which makes it clearer exactly what is going on. Any mistakes/problems should be handled in a future PR as this is not intending to change any behavior.

PR #348, opened 2016-04-04
More SJXML tests

### LINKED ISSUE #347: XML corruption issues for subjobs
Ganga is still having problems correctly falling back to a backup XML file or other solution when failing to load an XML for a subjob.

I've also spotted that the test for this is broken owing to accidentally badly named tests running out of order.

I've taken a first look at this and have expanded the standard corrupt XML test for subjobs but I've not had a chance yet to begin debugging issues with subjob XML files missing or becoming corrupt.

This is a problem that a few users @ CERN have already run into, I'm just confirming here that the integration test for this is broken and I'll be looking to fix this for 6.1.19.

rob-c:
This PR includes some early work for #347 and highlights some of the problems currently in the code.

This is currently failing due to the fact that I've not tracked down and fixed the problems associated with trying to load either a missing subjob XML or a corrupt XML for either a subjob or a master job.

The changes in ```Job.py``` are accidentally based on #346 . I'll rebase this file to remove the ```GPI``` changes but this shouldn't distract from the changes being made in this PR.

Fixes #347


rob-c:
I was hoping to fix this ahead of 6.1.19 the consequence of the errors highlighted here is that a corrupt XML will cause an exception to be thrown which isn't handled correctly.
If we want to get 6.1.19 out quickly then I'm happy to push this back but otherwise we are susceptible to some XML problems.

milliams:
You say "The changes in `Job.py` are accidentally based on #346". Is that still the case? I still see some changes to that file in this PR.

rob-c:
Yes it was, and now #346 is merged and there are still changes in this PR which are unique to this PR which were made on top of that branch.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -2181,15 +2183,31 @@ def _repr(self):
...
-            subjobs.objects = self.subjobs
-        else:
-            for sj in self.subjobs:
-                subjobs.objects[sj.id] = sj
-        #print('return slice: %s' % str(subjobs))
-        return _wrap(subjobs)
+        from Ganga.GPIDev.Base.Proxy import stripProxy
+        if not isType(self._stored_subjobs_proxy, JobRegistrySlice):
+            subjob_slice = JobRegistrySlice('jobs(%s).subjobs' % str(self.id))
+            self._stored_subjobs_proxy = _wrap(subjob_slice)
```
milliams: Why do we store the proxied version here? Why not just the `JobRegistrySlice` and then wrap it on return?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -2181,15 +2183,31 @@ def _repr(self):
...
+
+        if len(self._stored_subjobs_proxy) != len(self.subjobs):
+
+            if isType(self.subjobs, SubJobXMLList):
+                subjob_slice.objects = self.subjobs
+                #self._stored_subjobs_proxy = _wrap(stripProxy(self._stored_subjobs_proxy))
+            elif isType(self.subjobs, (list, GangaList)):
+                subjob_slice = stripProxy(self._stored_subjobs_proxy)
+                if len(subjob_slice.objects.keys()):
```
milliams: Could be just
```python
if subjob_slice.objects:
```

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -2181,15 +2183,31 @@ def _repr(self):
...
+                    removeable = subjob_slice.objects.keys()
+                    for sj_id in removeable:
```
milliams: These two line could be:
```python
for sj_id in subjob_slice.objects:
```

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Because you can't do:
```
sj = myJob.subjobs
assert sj is myJob.subjobs
```
otherwise. Also this ends up with some slight performance improvements in some situations. This is the situation we have will all objects getting proxied and I keep end up having to fix this over and over.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: How does python handle the list changing as it iterates over it, does it? I keep having problems with this.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Probably yes, but this has to play nice with ```SubJobXMLList``` so I doubt it in reality

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: I thought that *did* work. We made the proxy system so that it stores the proxy-wrapper of a `GangaObject` in an attribute on it so that doing `addProxy` multiple times to an object will always return the same proxy object.
```python
j = LocalFile()  # A non-proxied object
assert addProxy(j) is addProxy(j)
```
?

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: Ahh, you're right. You cant edit a list as you are iterating over it. However, I'm not sure that your version solves that as it's possible that `keys()` is simply a *view* onto the underlying dictionary internals. In fact, in Python 3 it *is* a view so this will break.

But,
```python
removeable = subjob_slice.objects.keys()
for sj_id in removeable:
    del subjob_slice.objects[sj_id]
```
could be simplified further to:
```python
 del subjob_slice.objects[:]
```

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: `if subjob_slice.objects:` will use `__len__` ([see this](https://docs.python.org/2/reference/datamodel.html#object.__nonzero__)) which `SubJobXMLList` implements correctly so it should indeed work.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: probably. This however is getting off topic from the point of the PR.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: No you're mixing objects with automatic proxies and objects which have their own custom proxies. See the fixes which were done to keep the job repository and equivalent having a sane single proxy wrapper.
If we're fixing this we should fix the proxy handling in ```Ganga.GPIDev.Lib.Registry```

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: Slightly, however my point is that in Python 3 the code is probably going to break in confusing ways. Also, 3 lines when 1 will do makes the code harder to read.

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: Why don't objects with their own custom proxies 'register' them in the same way as for automatic ones (e.g. settng `_proxyObject`)? That way, they're effectively indistinguishable and functions like `addProxy` and `stripProxy` will just work as expected.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Python 3 is fixed firmly off my radar, there are too many other things which need to be done before I'm worrying about that. Such as how we deal with broken XML files on disk due to a corruption.

milliams: I understand not wanting to worry about Python 3 too much but shorter, simpler, more idiomatic code which replaces something which *might* break on Python 2 and *will* break on Python 3 is surely a no-brainer?

rob-c: Again, no disagreement here. I agreed 'probably' and until there is significant development in the PR I won't get around to fixing this or deciding if there is a good reason to keep the code (probably not).

I still don't know if I'm pushing for this in 6.1.19 or 6.1.20 as I've asked: https://github.com/ganga-devs/ganga/pull/348#issuecomment-209895246

milliams:
I don't feel that we are in a rush to get 6.1.19 out. If there are bugs we need to fix then I think we should do so before release.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: They do, but the coding convention is such that this is not the case, I'd strongly prefer to go in and change it all with one future PR and not mix things up too much atm now that I've (almost) understood what's going on

rob-c:
OK, I may have fixed the outstanding problems with handling corrupt XML within Ganga.

Corrupt/missing XML are now skipped and the Object loaded from disk is marked as dirty so that it can/will be flushed again to generate a new XML file.

I've fixed a bug where crawling down the attributes ladder would end up in a circular loop due to the likes of job.master which took about 4 or 5 attempts to correctly spot. This fixes ```_setFlushed``` properly.

The only exception I've not tested at this point is both the XML file and the backup becoming missing but at this level of catastrophic data loss there's nothing we can do so I'm not inclined to spend to much time debugging this.

I've added an extra getter return to deal with returning variables such as ```_dirty``` without triggering a load of the Objects from XML which happened once or twice for me in testing.

As soon as the tests pass I'd be keen to merge this unless there are any objections?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -352,17 +351,28 @@ def _getItem(self, index):
...
-                    else:
-                        raise RepositoryError(self,"IOError on loading subobject %s: %s" % (index, x))
+                except Exception as x:
+                    logger.warning("Error loading XML file: %s" % str(x))
+                    try:
+                        logger.debug("Loading subjob #%s from disk, recent changes may be lost" % index)
+                        subjob_data = self.__get_dataFile(str(index), True)
+                        sj_file = self._loadSubJobFromDisk(subjob_data)
+                        has_loaded_backup = True
+                    except Exception as err:
```
milliams: Can this exception (and the one a few lines above) be made any more specific?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -373,18 +383,23 @@ def _getItem(self, index):
...
                         logger.debug("Err:\n%s" % str(err))
                         raise
 
-                # if load was successful then set parent and add to the _cachedJobs dict
-                if loaded_sj:
-                    loaded_sj._setParent( self._definedParent )
+                loaded_sj._setParent( self._definedParent )
+                if has_loaded_backup:
+                    loaded_sj._setDirty()
+                    loaded_sj._getParent()._setDirty()
```
milliams: `_setDirty()` should always set the parent dirty already so this second line is not needed.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -967,10 +970,16 @@ def _setFlushed(self):
         """Un-Set the dirty flag all of the way down the schema."""
         if self._schema:
             for k in self._schema.allItemNames():
+                ## Avoid attributes the likes of job.master which crawl back up the tree
+                if self._schema[k].hasProperty('visitable'):
+                    if (not self._schema[k].getProperties()['visitable'] or self._schema[k].getProperties()['visitable'] == 0):
+                        continue
+                if self._schema[k].hasProperty('transient'):
+                    if(self._schema[k].getProperties()['transient'] or self._schema[k].getProperties()['transient'] == 1):
```
milliams: There's no need here to do `if a or a == 1` or `if not b or b == 0`. Simply `if a` and `if not b` is sufficient.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Yes but at the cost of accepting an unknown number of different types or Exceptions. We need to sort out the GangaRepositoryXML's handling of exceptions first and make sure that the exceptions thrown from there are understood. Once I know this should only return 2 or 3 exceptions I will re-visit this.

python/Ganga/GPIDev/Base/Objects.py
rob-c: These really shouldn't be defined as 1 and 0 everywhere it requires an extra few seconds to parse the code and work out whats going on that's quite unpleasant...

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Typo, was left in from dealing with the job.master attribute which made this step look needed.

python/Ganga/GPIDev/Base/Objects.py
milliams: I agree that they shouldn't be defined as `0` and `1` but `1` and `True` are almost entirely compatible (likewise for `0` and `False`). We should write code which works with any truthy or falsey values so that when we replace all the integers with booleans it will keep working.

The `or foo == 0` check above is entirely unnecessary as if `foo` is ever `0` then it will also evaluate to `False` in a boolean context.

rob-c: Yes and it was changed. I was changed I was just irked that when I type ```is True``` it doesn't work as expected due to historical reasons which are no longer valid.

milliams: Yeah, that's annoying but it's a also a good reason to avoid `is True` when doing boolean checks and instead rely on Python's duck typing most of the time. http://rhodesmill.org/brandon/slides/2012-11-pyconca/ makes interesting reading on this part.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -971,12 +971,10 @@ def _setFlushed(self):
...
-                if self._schema[k].hasProperty('visitable'):
-                    if (not self._schema[k].getProperties()['visitable'] or self._schema[k].getProperties()['visitable'] == 0):
-                        continue
-                if self._schema[k].hasProperty('transient'):
-                    if(self._schema[k].getProperties()['transient'] or self._schema[k].getProperties()['transient'] == 1):
-                        continue
+                if self._schema[k].hasProperty('visitable') and not self._schema[k].getProperties()['visitable']:
+                    continue
+                if self._schema[k].hasProperty('transient') and self._schema[k].getProperties()['transient']:
+                    continue
```
milliams: This can be simplified further to:
```python
if not self._schema[k]['visitable']:
    continue
if self._schema[k]['transient']:
    continue
```
as there are no schema items without `visitable` or `transient` defined.

rob-c: I'll throw this into another commit

python/Ganga/GPIDev/Base/Objects.py
rob-c: I hold particular bias against not using ```is True``` and ```is not True```, on a few projects I've seen much more time lost to problems introduced by things being equivalent to ```True``` than time saved by duck-type programming and design issues but this is delving into the realms of personal preference here quite strongly I think

rob-c:
Right, after some code polish and cleaning up most of the catch-all exceptions, does anyone have any final comments before I hit the merge button?

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -760,7 +760,7 @@ def _load(self, obj_ids):
                 ## Track the objects we've loaded into memory
                 self._loaded_ids.append(obj_id)
         except Exception as err:
-            logger.error("Error Loading Jobs!")
+            logger.error("Error Loading Jobs! '%s'" % str(obj_ids))
```
milliams: No need for `str()`, just:
```python
logger.error("Error Loading Jobs! '%s'" % obj_ids)
```
is sufficient.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -179,7 +179,7 @@ def write_subJobIndex(self):
         """interface for writing the index which captures errors and alerts the user vs throwing uncaught exception"""
         try:
             self.__really_writeIndex()
-        except Exception as err:
+        except (IOError) as err:
```
milliams: There's no need for parentheses round this expression.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -444,6 +444,9 @@ def __get__(self, obj, cls):
 
         # Since we couldn't find the information in the cache, we will need to fully load the object
 
+        if name in obj.__dict__:
+            return obj.__dict__[name]
```
milliams: There should be no situation where a `Descriptor` getter is getting things from the object's `__dict__`. They are completely separate things. A given `Descriptor` should only look at its entry in `_data` and no where else (modulo lazy-loading).

rob-c: I've just written a lazy-loading test to make sure this isn't needed, there was a way that this was broken while I was doing some debugging. It shouldn't be in develop and it's not now but and I don't have time to go back and re-generate what I did to break this. I'm just going to add a unittest to check that the descriptor is never called when accessing a Node object attribute.
I'll upload the test and reversion in a few min

Issue #164, opened 2016-01-29
Corrupt jobs get totalled after ganga exit.


rob-c:
Ganga doesn't appear to correctly avoid re-writing 'corrupt' or 'inaccessible' job XML to disk at shutdown.

This is causing bad behaviour of corrupt jobs which will lose data. (i.e. the half-loaded job object appears to be wiping the data which is corrupt from disk rather than leaving the job as untouched).

I'm investigating if this is a trivial fix or not now.

egede:
@rob-c Is this fixed by other PRs?

milliams:
I believe that this was fixed with @drmarkwslater's lock on `safe_save()`.

rob-c:
Nope, this was another issue which is that Ganga will half-load a job and save the half-loaded job over the bad XML which failed to load. This I don't believe has been addressed yet at all.

rob-c:
This is still happening and needs looking into asap. I have a repo which I am switching between branches, one which has a new application object.
I accidentally left the monitoring on and now all objects with the new application not in my repo have just been wiped from disk (ok, maybe I can restore some of them from backups).

This may get annoying until we have ```ganga``` loading all default plugins, but even still this is bad behavior and needs fixing. I suspect not writing a job which is not fully loaded is the correct thing to do regardless of it's dirty status.

rob-c:
This just bit me again during development. I'm going to see if I can find a fix for this now.

rob-c:
Fixed by #351

PR #351, opened 2016-04-04
Fixing destroying of job data when corrupt job is loaded

rob-c:
This addresses #164 

Specifically I've fixed:
 - Fixed a bug where exceptions in the repr were not being caught for non-loadable jobs.
 - Stop throwing exceptions in Schema where a simple check will suffice. It makes things faster and reduces the amount of noise when debugging.
 - Fixed the XMLRepository to handle not correctly loading of an object from disk correctly
 - Fixed a few minor bugs in handling of errors around objects which can't be loaded
 - Fixed some import statements which were causing exceptions to be thrown at startup behind the scenes.

The impact of this is that constructing a Job is now down from 0.02s to about 0.005s on a very unscientific test of testing how fast Ganga now starts. This has been an annoying performance bug which is associated with the Schema which impacts (among other things) The time taken to initialize many, many GangaObjects from scratch.

As a result of this I don't get ```EmptyGangaObject```s written to my repository when a Job object fails to load correctly during development.

Unless there are serious objections I would like to get this in asap (obviously modulo tests still passing).


python/Ganga/Core/GangaRepository/GangaRepository.py
```diff
@@ -36,7 +36,7 @@ def __str__(self):
 class InaccessibleObjectError(GangaException):
 
     def __init__(self, repo=None, id='', orig=None):
-        GangaException.__init__(self, "Inaccessible Object")
+        super(InaccessibleObjectError, self).__init__("Inaccessible Object: %s" % str(id))
```
milliams: No need for `str()` here.

rob-c: Given that this object is not at times a string is that or is that not a problem in python 2.6?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -811,10 +813,14 @@ def _actually_load_xml(self, fobj, fn, this_id, load_backup):
             a4=time.time()
             logger.debug("Loading XML file for ID: %s took %s sec" % (this_id, str(a4-b4)))
 
+            if errs is not None:
+                if len(errs) > 0:
```
milliams: Do we need `not None` as well as `len() > 0`? On no error, surely it either returns an empty list *or* a `None`? It looks like it should be the former.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -811,10 +813,14 @@ def _actually_load_xml(self, fobj, fn, this_id, load_backup):
...
+                    logger.error("#%s Error(s) Loading File: %s" % (str(len(errs)), fobj.name))
```
milliams: No need for `str()`

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: It should be the former but I wanted to avoid ```len(None)``` throwing an error if this ever changes. I can remove the check but I'm being paranoid.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -909,6 +915,8 @@ def load(self, ids, load_backup=False):
 
                 if should_continue is True:
                     continue
+                else:
+                    raise err
```
milliams: If this is just:
```python
raise
```
then it will conserve the stack trace.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -690,11 +689,17 @@ def _flush(self, objs):
             if not obj._dirty:
                 continue
 
+            if hasattr(obj, _reg_id_str):
+                obj_id = getattr(obj, _reg_id_str)
+                if obj_id not in self._loaded_ids:
+                    continue
+            else:
+                continue
```
milliams: This implies that we're still getting non-loaded ids with `_dirty=True`? That sounds bad.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -237,6 +237,8 @@ def _getDefaultValueInternal(self, attr, val=None, check=False):
         from Ganga.Utility.Config import Config
         is_finalized = Config._after_bootstrap
 
+        _useDefVal = False
```
milliams: This doesn't need to be prefaced with an `_` as it's not being exposed to any external interface.

python/Ganga/Core/GangaRepository/GangaRepository.py
milliams: It's always been the case that `"%s"` will explicitly call `str()` on the argument. This is defined by the language.

rob-c:
As I predicted this has caused a few problems to do with loading of jobs (or not loading them).

I'll investigate what has caused the failures and push the fixes

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Yes, but this needs doing everywhere else too

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
milliams: Unless we're expecting a `None` I see no reason to test for it. The variable is a `list` and we should be able to rely on that.

milliams:
Did it pass when you ran the tests locally?

python/Ganga/GPIDev/Schema/Schema.py
rob-c: I was playing with objects, this is what was left, I can re-name if necessary but yes it's never exposed or changed externally

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I haven't tested whether this is the case (I can if we're really paranoid about this) but the ```_dirty``` flag would be set if the object is modified at all regardless of if the 3rd modification causes an exception to be thrown (but the object to still exist in the list of objects known to the repository).
I plan to add a line which sets the dirty status of jobs freshly loaded from disk to be not dirty after the loading has finished but atm we currently don't have this either.

rob-c:
I didn't run them all locally, just most of them, the few I ran passed, the machines I am working on are under load so I just submitted the tests and went back to doing other work.

rob-c:
Aside from marking all loaded jobs as not dirty after loading I think this PR is taking shape. Thankfully it also seems to shave 5min off the integration tests which is nice to see.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: This is now just a belt and braces check. I'm keen to see it kept as I do not want to experience data loss due to a job failing to load!

rob-c:
OK I'm keen to push for this PR to get merged.

Unless there are any objections I'd like to get this merged first thing tomorrow so I can get on with other work without the risk that my jobs will get destroyed while I'm doing dev work.

rob-c:
OK, I'll resolve the conflicts with develop then look to merge

rob-c:
OK I'm seeing:
```
ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
Recording test results
ERROR: Step ‘Publish JUnit test result report’ failed: Test reports were found but none of them are new. Did tests run? 
```
I think this is probably related to the fact I'm seeing a lot of unicorns on github...

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -496,6 +496,8 @@ def update_index(self, this_id=None, verbose=False, firstRun=False):
                     ## we can't reasonably write all possible exceptions here!
                     logger.debug("update_index: Failed to load id %i: %s" % (this_id, str(x)))
                     summary.append((this_id, str(x)))
+                #finally:
+                #    pass
```
milliams: I'm guessing this was added accidentally?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -871,33 +878,35 @@ def load(self, ids, load_backup=False):
...
-        logger.debug("Could not load object #%i: %s", this_id, str(err))
+        logger.debug("Could not load object #%s: %s" % (str(this_id), str(err)))
 
         # try loading backup
         try:
-             self.load([this_id], load_backup=True)
-             logger.warning("Object '%s' #%i loaded from backup file - the last changes may be lost.", self.registry.name, this_id)
-             return True
+            self.load([this_id], load_backup=True)
+            logger.warning("Object '%s' #%s loaded from backup file - the last changes may be lost." % (str(self.registry.name), str(this_id)))
```
milliams: This line (and some above) don't need the `str()` added.

python/Ganga/Core/GangaRepository/Registry.py
milliams: If this is covering some some special case then a comment to that effect would be useful.

milliams:
@rob-c Yes, GitHub seems to be suffering some problems at the moment https://status.github.com/

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Frankly I end up adding/removing these when testing the exceptions and there is far too much exception handling in opening the xml files to follow atm imo

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Yes, but this never causes errors

milliams: Absence of errors does not make it correct. It adds unnecessary noise to the code and makes it harder to read.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
milliams: But `finally: pass` is a complete no-op so what was it used for?

milliams: But `finally: pass` is a complete no-op so what was it used for?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
milliams: Absence of errors does not make it correct. It adds unnecessary noise to the code and makes it harder to read.

rob-c:
retest this please

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: It's used when commenting out the exception handler because it's hiding problems rather than re-indent everything else for the sake of it

python/Ganga/Core/GangaRepository/Registry.py
rob-c: It's a belt and braces check and it's skipping any job which is not in the ```_loaded_ids``` list, this is what the code does so I fail to see how a comment helps?

rob-c:
This is working, merge-able and passes all tests, I'm merging this in a bit to work on other branches.

PR #357, opened 2016-04-06
Feature/optimisations

rob-c:
This is another branch I have been playing with on the side recently. I managed to get the time taken to generate 10,000 DiracFile objects down from ~30sec to ~20sec from some optimizations and basic caching.

As it stands my usual way of using cachegrind and cProfile is reaching a limit as a lot of time is spent doing some very nested things beyond the scope of the analysis tools in their default configuration.

This branch may turn into an ongoing WIP but I would like to see it go into 6.1.19 as LHCb users are still reporting some severe performance problems and whilst this doesn't go all the way to fixing them it does make things a little easier to work with.


python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -7,7 +6,8 @@
 from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList, makeGangaListByRef
 
 # config_scope is namespace used for evaluating simple objects (e.g. File)
-from Ganga.Utility.Config import config_scope
+#from Ganga.Utility.Config import config_scope
```
milliams: If we're not using this import, this line can be removed entirely.

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -18,9 +18,12 @@
 from .GangaRepository import SchemaVersionError
 
 import xml.sax.saxutils
+import copy
 
 logger = getLogger()
 
+_eval_strings = {}
```
milliams: We should avoid globals. This probably makes most sense as a class attribute of `Loader`. Also, perhaps rename to `_eval_string_cache` or similar?

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -361,7 +364,9 @@ def end_element(name):
                 # unescape the special characters
                 s = unescape(self.value_construct)
                 #logger.debug('string value: %s',s)
-                val = eval(s, config_scope)
+                if s not in _eval_strings:
+                    _eval_strings[s] = eval(s, getProxyInterface().__dict__)
+                val = copy.deepcopy(_eval_strings[s])
```
milliams: This could probably be simplified to one line with a `defaultdict`.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -130,9 +131,14 @@ def runtimeEvalString(this_obj, attr_name, val):
...
 
-
+_stored_evals = {}
 
 def raw_eval(val):
+    if val not in _stored_evals:
+        _stored_evals[val] = _really_eval(val)
+    return copy.deepcopy(_stored_evals[val])
+
+def _really_eval(val):
```
milliams: Can we avoid adding more function names like `_really_do_something`. The split is arbitrary and it makes tracking the code much more difficult.

Also, any new functions must have full docstrings.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -159,8 +160,8 @@ def inherit_copy(self):
     def createDefaultConfig(self):
         # create a configuration unit for default values of object properties
         # take the defaults from schema defaults
-        config = Ganga.Utility.Config.makeConfig(defaultConfigSectionName(self.name),
-                                                 "default attribute values for %s objects" % self.name)
+        _self_name = self.name
```
milliams: What is the purpose of this variable?

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -201,28 +200,24 @@ def prehook(name, x):
...
-            #    pass
-
             return x
 
         config.attachUserHandler(prehook, None)
         config.attachSessionHandler(prehook, None)
 
 
-    def getDefaultValue(self, attr):
+    def getDefaultValue(self, attr, make_copy=True):
```
milliams: If adding new arguments, they must be documented.

python/Ganga/Runtime/bootstrap.py
```diff
@@ -181,6 +181,9 @@ def manualExportToGPI(my_interface=None):
     from Ganga.GPIDev.Lib.Registry.JobRegistry import jobSlice
     exportToInterface(my_interface, "jobSlice", jobSlice, "Functions")
 
+    import datetime
+    exportToInterface(my_interface, "datetime", datetime, 'Classes')
```
milliams: Why are we exporting `datetime` to the GPI?

milliams:
It's strange that the unit tests are failing on Travis but must have passed on your local machine before you pushed. Is there some obvious difference?

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: Probably, this PR is still in dev, I'm examining the impact of some of the changes

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: yes, if the interface (```GPI``` or ```ganga```) didn't change... besides this has to eval an actual string not just find an object. We're doing some horrible, horrible things here fyi.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Sharing the decorators makes profiling more difficult, abstracting things helps me introduce another layer of logic rather than adding pointless indents. I can change the name but I think another function is called for here.

python/Ganga/GPIDev/Schema/Schema.py
rob-c: to avoid calling self.name twice which is more than just getting an attribute due to some changes.

rob-c:
I was using Travis to do the full set of tests for me while I checked something, but yes there are some problems a few quick checks didn't highlight. Some more work is needed when I get around to developing this branch some more.

python/Ganga/Runtime/bootstrap.py
rob-c: Because the JobTime does horrible things, see the comment in VStreamer if the VStreamer eval uses the GPI (which is more sensible) we need the datetime due to the JobTime being stored in a probably bad way. I should have added a line here to say why...

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: I don't see a problem with globals here (that doesn't suggest a larger problem elsewhere) a cache name is something likely to come in a revision :)

rob-c:
I'd like to push for this in 6.1.19 if possible as it addresses some performance issues. Are there any strong objections to this? It can wait until 6.1.20 but all of the tests should now be passing.

python/Ganga/Core/FileWorkspace.py
```diff
@@ -13,7 +13,7 @@
 
 from Ganga.Utility.logging import getLogger
 
-import os
+from os import path, makedirs, rename, mkdir
```
milliams: What was the reason for pulling everything out of the `os` namespace? It's so standard in Python code to do `os.path.join` or `os.rename`, when I'm seeing the code below it make me assume on first blush that `makedirs(path_to_build)` is some custom function rather than a standard library one.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -466,7 +466,8 @@ def flush(self):
 
     def _setFlushed(self):
         """ Like Node only descend into objects which aren't in the Schema"""
-        for index in self._cachedJobs.keys():
-            self._cachedJobs[index]._setFlushed()
+        def flushFunc(elem):
+            elem._setFlushed()
+        map(flushFunc, self._cachedJobs.items())
```
milliams: I'm not sure this is a good use of `map`. `map` should be used to *map* one sequence onto another, not as a replacement for a `for` loop. Why not:
```python
for elem in self._cachedJobs.items():
    elem._setFlushed()
```
?

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -192,8 +195,7 @@ def simpleAttribute(self, node, name, value, sequence):
                 self.level += 1
                 print(file=self.out)
                 print(self.indent(), '<sequence>', file=self.out)
-                for v in value:
-                    self.acceptOptional(v)
+                map(self.acceptOptional, value)
```
milliams: As above, `map` is a functional programming construct for creating new sequences out of existing ones (like a list comprehension). It is not just a replacement for a `for` loop.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -13,31 +13,31 @@
...
-import Ganga.Utility.logging
+from Ganga.Utility.logging import getLogger
 
 from copy import deepcopy, copy
-import inspect
+from inspect import isclass
+from functools import partial
 
-import Ganga.GPIDev.Schema as Schema
+from Ganga.GPIDev.Schema import Schema, Item, ComponentItem, SharedItem
```
milliams: As before, I think it's very useful to keep the namespace on these functions.

rob-c:
OK, given some of the discussion here and other work I'm pushing this back until 6.1.20 to better understand some of the nature of the changes and see how much may be due to other 6.1.19 work.

rob-c:
There are some parts of this I want to throw away and some parts that are needed/worth keeping. I'm dropping this from 6.1.20 as the only bottlenecks of dealing with lists may be fixed by new list objects which are incoming.

rob-c:
OK, so on this branch I'm using the following command as a 'benchmark' tool:
`%time dataset = BKQuery('/LHCb/Collision15/Beam6500GeV-VeloClosed-MagDown/Real Data/Reco15a/Stripping24/90000000/EW.DST').getDataset()`
On this branch this runs in about 50sec but running the same on develop takes between 3 and 4 min and has a higher memory footprint by a few hundred Mb.

I'm going to focus on other work for 6.1.21 so I would be very keen to get this in for 6.1.22 unless 6.1.21 gets delayed longer than expected.

There are still significant delays in loading/writing large XML files so I'm tempted to add some logging info when loading a large job (XML > 50Mb) from disk in this branch.

rob-c:
There's been a lot of private communication ahead of this next release from LHCb users in particular who I think are some of the worst effected by some of the performance issues.
Does anyone have any objections to moving this PR forward to 6.1.21 and going for review?

milliams:
As I mentioned before, I'm not sure this is a good use of `map`. `map` should be used to *map* one sequence onto another, not as a replacement for a `for` loop. Are you expecting to see a significant performance improvement from using `map`?

rob-c:
I can remove it and see what effect it does have now that this has settled, frankly it was to try and make the code more readable in the places where the for loop was adding yet another level of indentation which was starting to hurt my head

rob-c:
Well good news I see no difference as a result of taking the Map out and replacing it with explicit for loops.
I'll clean this up in this case and make a commit later/tomorrow.
Either way this branch is seeing consistently faster than before behaviour in performing large BK Queries and some other larger tasks.

rob-c:
OK, map functions gone, performance improvements still there and I think I'm looking to profile and merge this branch if there are any comments they're welcome.

I could have missed the wood for the trees in this one it's been an exercise in optimising code rather than changing overall design of things. If some of the code can be rewritten even better from a high level perspective please let me know.

I'll fix any bugs that creep in as a result of this but it works well for me in testing.

rob-c:
OK, removing the maps introduces bugs that don't show up in normal use... this is interesting... Why should this be the case?

I'll check I've not removed any that were in place before this PR and put them back if I've done this to see what happens.

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -361,7 +364,9 @@ def end_element(name):
                 # unescape the special characters
                 s = unescape(self.value_construct)
                 #logger.debug('string value: %s',s)
-                val = eval(s, config_scope)
+                if s not in _cached_eval_strings:
+                    _cached_eval_strings[s] = eval(s, getProxyInterface().__dict__)
+                val = copy.deepcopy(_cached_eval_strings[s])
```
milliams: Why don't we just use the plugin system here and look them up in it?

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -380,16 +385,17 @@ def end_element(name):
...
-                            obj.setSchemaAttribute(attr, makeGangaListByRef(obj._schema.getDefaultValue(attr)))
-                            #setattr(obj, attr, makeGangaListByRef(obj._schema.getDefaultValue(attr)))
+                            obj.setSchemaAttribute(attr, makeGangaListByRef(obj._schema.getDefaultValue(attr, make_copy=False)))
                         else:
-                            obj.setSchemaAttribute(attr, obj._schema.getDefaultValue(attr))
-                            #setattr(obj, attr, obj._schema.getDefaultValue(attr))
-
+                            obj.setSchemaAttribute(attr, obj._schema.getDefaultValue(attr, make_copy=False))
+                for item in obj._schema.allItems():
+                    schema_assign(item, obj=obj)
```
milliams: Why was this change made? It's a bit strange to define a local function only to then loop over a sequence with it.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -18,27 +18,26 @@
...
 
 from Ganga.Core.exceptions import GangaValueError, GangaException
 
 from Ganga.Utility.Plugin import allPlugins
 
 def _getName(obj):
     """ Return the name of an object based on what we prioritise"""
-    returnable = getattr(obj, '_name', getattr(obj, '__name__', None))
+    returnable = getattr(obj, '_name', getattr(obj, '__name__', getattr(getattr(obj, '__class__', None), '__name__', None)))
```
milliams: This is getting messier and messier. Why can't we just access `_name`?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -74,6 +73,7 @@ class Node(object):
     thread-safe usage.
     """
     __metaclass__ = abc.ABCMeta
+    __slots__ = ('_index_cache_dict', '_parent', '_read_lock', '_write_lock', '_dirty', '_registry', '_data_dict')
```
milliams: We should be very careful about defining `__slots__`. It can be a very good performance improvement but when I checked in the past it made no real difference. It also prevents us from ever dynamically adding attributes to objects of this type.

python/Ganga/GPIDev/Base/Objects.py
rob-c: This was used to replace situations where the object can have either. I can drop this back to multiple line statements if you're not a fan of this.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -86,9 +86,12 @@ def __deepcopy__(self, memo=None):
         cls = self.__class__
         obj = cls()
         this_dict = copy(self.__dict__)
-        for elem in this_dict.keys():
+        global do_not_copy
```
milliams: Why have you declared this as `global`?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -86,9 +86,12 @@ def __deepcopy__(self, memo=None):
...
+        def internal_deepcopyMethod(elem):
             if elem not in do_not_copy:
-                this_dict[elem] = deepcopy(this_dict[elem], memo)  # FIXED
+                this_dict[elem] = deepcopy(this_dict[elem], memo)
+        for k in this_dict.keys():
+            internal_deepcopyMethod(k)
```
milliams: Again, why define a function, only to call it over a sequence?

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: You need access to things like `datetime` to evaluate JobTime.
Which reminds me I planned to add a comment to this effect here.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -276,6 +279,7 @@ def __get__(self, obj, cls):
 
         # If obj is None then the getter was called on the class so return the Item
         if obj is None:
+            name = _getName(self)
```
milliams: 1. This is repeated from a few lines before
2. Why don't we just do `self._name`?

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: This was a used in a map and the scope was clearer but I moved back and this is the result.
I'll look into reverting this.

python/Ganga/GPIDev/Base/Objects.py
rob-c: That's why you add `__dict__` to the slots which allows for dynamic changes in the future, but should we ever be doing this in the Node class?

python/Ganga/GPIDev/Base/Objects.py
rob-c: Good question, unfortunately I've no answer. I plan to revert this method anyway so it'll drop then.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -529,15 +538,17 @@ def __init__(cls, name, bases, this_dict):
 
         # all Ganga classes must have (even empty) schema
         if cls._schema is None:
-            cls._schema = Schema.Schema(None, None)
+            cls._schema = Schema(None, None)
 
         this_schema = cls._schema
 
+        cls.__slots__ = ('__dict__', '_proxyObject')
```
milliams: We want to move away from coupling the proxy classes and non-proxy classes in `Objects.py`.

python/Ganga/GPIDev/Base/Objects.py
rob-c: code -> code with map -> code without map with bugs -> code without map working so I need to move back to the beginning of code in places where it makes sense.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -264,36 +274,39 @@ def full_print(obj, out=None, interactive=False):
...
-                    outString += result.rstrip()
+                    outStringList.append(result.rstrip())
                 else:
-                    result = str(x)
                     # remove trailing whitespace and newlines
-                    outString += result.rstrip()
-                count += 1
-                if count != obj_len:
-                    outString += ', '
+                    outStringList.append(str(x).rstrip())
```
milliams: The `else` is the same as the first `if`. I can't think of a reason to add in the special case at the beginning?

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -264,36 +274,39 @@ def full_print(obj, out=None, interactive=False):
...
+                if (len(outStringList)+1) != 2*obj_len:
```
milliams: What does this bit of arithmetic mean?

python/Ganga/GPIDev/Base/Objects.py
rob-c: Yes but we still have to have it in the do_not_copy... frankly we should advertise some knowledge in the Proxy of what it will dynamically create if we want to try and make the construction more sane.

I think I saw some performance improvement in memory allocation speeds when using slots and a large (Order 100k+) objects being created which is almost quite common.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -8,27 +8,32 @@
...
-        _list = map(mapfunction, _list)
+        if extra_args is None:
+            _list = [mapfunction(l) for l in _list]
+        else:
+            new_mapfunction = partial(mapfunction, extra_args=extra_args)
+            _list = [new_mapfunction(l) for l in _list]
 
     result = GangaList()
-    result._list = [ stripProxy(e) for e in _list ]
+    result._list.extend([stripProxy(l) for l in _list])
```
milliams: Why `.extend()` here? I can't see that this is an improvement. It makes the intent of the line less clear, implying that you want to preserve some existing elements.

python/Ganga/GPIDev/Base/VPrinter.py
rob-c: A test to see if we're at the end of the array or not very quickly. I'll add a comment

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -331,8 +340,9 @@ def __gt__(self, obj_list):
     def __hash__(self):
         logger.info("hash")
         result = 0
-        for element in self:
-            result ^= hash(element)
+        hashes = [hash(l) for l in self._list]
+        for element in hashes:
+            result ^= element
```
milliams: This seems to be making a temporary list for no reason?

python/Ganga/GPIDev/Base/VPrinter.py
rob-c: Good point, I'll make a change around this

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: This is intentionally subverting the checks around modifying the `_list` item in the Schema of the GangaList as we're stripping (any) proxy here, we want to make an instance as fast as possible and return it.

python/Ganga/Runtime/bootstrap.py
```diff
@@ -181,6 +181,10 @@ def manualExportToGPI(my_interface=None):
     from Ganga.GPIDev.Lib.Registry.JobRegistry import jobSlice
     exportToInterface(my_interface, "jobSlice", jobSlice, "Functions")
 
+    """Needed for the JobTime if we plan to use the interface to eval the XML as well"""
+    import datetime
+    exportToInterface(my_interface, "datetime", datetime, 'Classes')
```
milliams: I don't think that we should be using the GPI to evaluate strings in the XML. Looking up the names in the plugin system is how it should be done. The fact that we have to export `datetime` here is a string sign that there is something wrong.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: Can probably be reverted

python/Ganga/GPIDev/Base/Objects.py
milliams: I guess the question it prompts is "which objects?" At the time we're asking for the name of an object we should know what type it is. It's almost always because it's a `GangaObject` in which case `obj._name` is simpler than `_getName(obj)`. Trying to write some 'god function' which can ascertain the name of an arbitrary object is fraught with danger and causes this function to have to require far too much knowledge about the whole codebase.

There should never be a time where we don't know whether a particular variable is a `GangaObject`, a schema `Item` or some built-in type.

python/Ganga/GPIDev/Base/Objects.py
milliams: While you can indeed add `__dict__` to `__slots__` it rather undoes most (if not all) of the point of using `__slots__` in the first place and might in fact make it even worse. `__slots__` is something which should be fallen back on after careful memory profiling. What memory reduction did you see after this change?

rob-c: For the whole branch several hundred Mb, for slots I'll let you know once I've some more commits put in addressing some other stuff then I'll run with and without.

python/Ganga/Core/GangaRepository/VStreamer.py
milliams: I'm not sure that we should have to. What is the string that is being evaluated that requires `datetime` to be present in the namespace? When processing the name of a `ComponentItem` in a `<class>` entry we use the plugin system and when processing an value in a `<value>` entry we use a standard python namespace (and at *that* point, yes we require `datetime` to be present).

python/Ganga/Runtime/bootstrap.py
rob-c: Yes but way beyond the scope of this PR which was trying to address common bottlenecks which are showing up.
Fixing the classes which are exploiting this whilst being able to load older jobs is potentially extremely difficult.
I raised this a while ago but I don't think there is an issue.

python/Ganga/GPIDev/Base/Objects.py
milliams: The way that `__slots__` improves the memory usage is by disabling the automatic creation of a `__dict_-` per object and instead does them as descriptors. As soon as you allow `__dict__` to be in `__slots__` it undoes that and gives every object its dictionary back.

rob-c: I'm following http://stackoverflow.com/a/28059785 and of the opinion there isn't a good reason not to use? I'll benchmark in a bit, putting in other changes to clean up this PR to be minimal. It's evolved a lot

python/Ganga/Runtime/bootstrap.py
milliams: I am very much against adding more code (which is more complexity and maintenance effort) in a branch which is supposed to be improving performance. More code will almost always make things slower and constantly patching small problems across the code base without really understanding what's happening at the deeper levels is what has been causing problems for a long time. I guarantee that we could get equal performance improvements by removing and simplifying code rather than adding more for special cases etc.

rob-c: Is there any particular file where you feel this applies?
If we can get the performance improvements without adding more code I'm happy but I've specifically stripped out a lot of changes and gunk in this PR to get to the minimal changes which fix several performance issues which I've had a lot of contact from users over.

python/Ganga/GPIDev/Base/Objects.py
alexanderrichards: > It also prevents us from ever dynamically adding attributes to objects of this type.

Isn't this a good thing? Certainly this and the memory performance improvements that come with not having to define a `__dict__` for each instantiated object are the reasons that I used them in the stuff I've been doing.

My argument would be that only the schema attributes should be needed by the user for a given object and that they shouldn't be adding extras. Obviously as a developer if we want an attribute added to a derived class then we can add it to the `__slots__` for that class. I don't really like the idea of dynamic attributes as it's hard to say at any given time what an object contains.

python/Ganga/Runtime/bootstrap.py
milliams: Mostly because it feels like code smell that we're having to explicitly export a particular standard module into our GPI namespace so that another part of code can continue to work. `bootstrap.py` should not need to know about `JobTime`. What if some third-party plugin requires some other module to be present, would we add them all here? When you're working on a coding problem and you're finding that you're having to add strange hacks to make things work, that's very often a sign that there's something wrong. This is coupling unrelated parts of the software together, reducing encapsulation and destroying locality when reading the code. Ganga is built on these hacks and I've been working hard to make things maintainable in the future, the only way Ganga will survive with the amount of manpower we have is if it becomes simpler and smaller. Constantly patching problems on a half-a-million LOC project does not scale.

Python is a good enough language to allow us to act polymorphicly, at the point we need to and when we need to rather than maintaining a large and complicated global state which reduces clarity and makes testing harder.

python/Ganga/GPIDev/Base/Objects.py
alexanderrichards: @milliams is right though that we shouldn't add `__dict__` back into `__slots__` for a generic base class as this undoes most of the benefit. If you want the ability to have dynamic variables then just don't use `__slots__` or possibly add the `__dict__` back **only** for classes that you want to allow them. That way the other classes benefit from the `__slots__`.

I would be keen to know however where we are using dynamically assigned variables and would argue that we should perhaps review if they are necessary.

milliams: I agree in general that for our code, preventing dynamic attributes is a good idea. However in practice many `GangaObject` subclasses create `self.temp_foo` variables outside of the schema for tracking internal state. Also, further down, `__dict__` is added to `__slots__` anyway which re-allows dynamic attributes to be added. I'd be surprised if this gave more than a few percent improvement, especially given that most of the data in a `GangaObject` is in `_data` anyway and that gets quite large.

alexanderrichards: ah well... that's the advantage of forgetting the day to day code base and how it operates on the practical level, I can just pontificate about things that would be nice and a good idea. Looks like then it would be too much work and too intrusive to ban outright dynamic variables now :wink:

python/Ganga/Runtime/bootstrap.py
rob-c: OK, just turning this __off__ I see 324Mb consumed when the list from `%time dataset = BKQuery('/LHCb/Collision15/Beam6500GeV-VeloClosed-MagDown/Real Data/Reco15a/Stripping24/90000000/EW.DST').getDataset()` is loaded
but when I turn this __on__ I get 262Mb. Both times clean session with no objects loaded other than this and monitoring turned off.
This is a big enough change that I'm for keeping this in

python/Ganga/GPIDev/Base/Objects.py
rob-c: OK, just turning this __off__ I see 324Mb consumed when the list from `%time dataset = BKQuery('/LHCb/Collision15/Beam6500GeV-VeloClosed-MagDown/Real Data/Reco15a/Stripping24/90000000/EW.DST').getDataset()` is loaded
but when I turn this __on__ I get 262Mb. Both times clean session with no objects loaded other than this and monitoring turned off.
This is a big enough change that I'm for keeping this in

milliams: Ok. However, `_index_cache_dict`, `_registry` and  `_data_dict` are `GangaObject` properties, not `Node` properties so they should be moved to the `__slots__` attribute on that class.

python/Ganga/Runtime/bootstrap.py
rob-c: In this case I'll buffer the output from the flat eval which calls a mystery parameter which is badly defined aka the Config system. Using the GPI dictionary this way made life easier as we knew exactly what is misbehaving and means 1 less random dictionary we're looking up against in the future.

rob-c:
With a net change of ~+40 lines of code I think this branch is quite reasonable.
Unless the tests show that the latest fixes have introduced some additional bug does anyone have any objections before I merge this branch?

python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -17,10 +16,15 @@
 
 from .GangaRepository import SchemaVersionError
 
+from functools import partial
```
milliams: This import is not used as far as I can tell.

rob-c: ok, gone in next commit then

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -495,7 +503,7 @@ def __delete__(self, obj):
         del obj._data[_getName(self)]
 
     @staticmethod
-    def __createNewList(final_list, input_elements, action=None):
+    def __createNewList(_final_list, _input_elements, action=None, extra_args=None):
```
milliams: What does `extra_args` do?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -604,18 +616,21 @@ def __init__(self):
...
         #Node.__init__(self, None)
 
         if self._schema is not None and hasattr(self._schema, 'allItems'):
+            self._data_dict = dict.fromkeys(self._schema.datadict)
             for attr, item in self._schema.allItems():
                 ## If an object is hidden behind a getter method we can't assign a parent or defvalue so don't bother - rcurrie
                 if item.getProperties()['getter'] is None:
-                    setattr(self, attr, self._schema.getDefaultValue(attr))
-
+                    self._data_dict[attr] = Descriptor.cleanValue(self, self._schema.getDefaultValue(attr, make_copy=True), attr)
```
milliams: Can we not simply use the setters for these values?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -604,18 +616,21 @@ def __init__(self):
...
         super(GangaObject, self).__init__(None)
 
-        self._data_dict = {}
         self._index_cache_dict = {}
         self._registry = None
 
         #Node.__init__(self, None)
 
         if self._schema is not None and hasattr(self._schema, 'allItems'):
+            self._data_dict = dict.fromkeys(self._schema.datadict)
```
milliams: Why does this need to be initialised with the names from the schema>?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -731,13 +746,13 @@ def _actually_copyFrom(self, _srcobj, _ignore_atts):
             if not self._schema.hasAttribute(name):
                 #raise ValueError('copyFrom: incompatible schema: source=%s destination=%s'%(_getName(_srcobj), _getName(self)))
                 if not hasattr(self, name):
-                    setattr(self, name, self._schema.getDefaultValue(name))
+                    setattr(self, name, self._schema.getDefaultValue(name, make_copy=True))
```
milliams: There's no need to be explicit with `make_copy=True`. It's the default value of the argument isn't it?

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -214,7 +214,7 @@ def isType(_obj, type_or_seq):
     if isinstance(type_or_seq, getKnownLists()):
         clean_list = []
         for type_obj in type_or_seq:
-            str_type = type('')
+            str_type = str
```
milliams: There's no need for this temp variable. Just replace `str_type` with `str` below.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -1,15 +1,21 @@
...
 # $Id: VPrinter.py,v 1.1 2008-07-17 16:40:52 moscicki Exp $
 ##########################################################################
 from Ganga.GPIDev.Base.Proxy import isProxy, isType, runProxyMethod, stripProxy
 from Ganga.GPIDev.Base.Objects import GangaObject
+from cStringIO import StringIO
 
 from inspect import isclass
-from cStringIO import StringIO
+
+from functools import partial
```
milliams: This module is not used as far as I can tell.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -198,9 +204,11 @@ def _CallSummaryPrintMember(self, node, name, value):
 
     def _CallPrintSummaryTree(self, obj):
         sio = StringIO()
+        #logger.debug("_CallPrintSummaryTree")
```
milliams: This should be removed.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -198,9 +204,11 @@ def _CallSummaryPrintMember(self, node, name, value):
...
         if not hasattr(stripProxy(obj), 'printSummaryTree'):
             print("%s" % str(obj), file=self.out)
         else:
+            #logger.debug("calling printSummaryTree")
```
milliams: This should be removed.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -316,27 +324,31 @@ def summary_print(obj, out=None, interactive=False):
...
         else:
             outString = '['
-            count = 0
-            for x in _obj:
+            outStringList = []
+            for x in obj:
                 if isType(x, GangaObject):
-                    sio = StringIO()
+                    sio =StringIO()
+                    #logger.debug("summary_print: printSummaryTree")
```
milliams: This line should be removed.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -99,8 +108,10 @@ def __construct__(self, args):
 
         if len(args) == 1:
             if isType(args[0], (len, GangaList, tuple)):
-                for element_i in args[0]:
-                    self._list.expand(self.strip_proxy(element_i))
+                def myExpand(elem):
+                    self._list.expand(self.strip_proxy(elem))
+                for elem in args[0]:
+                    myExpand(elem)
```
milliams: This should be reverted to a simple `for` loop.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -118,18 +129,22 @@ def is_list(obj):
...
-                        if isType(elem, GangaObject):
-                            elem._setParent(self._getParent())
+                    returnable_list = [stripProxy(l) for l in value]
+                    my_parent = self._getParent()
+                    def safeSetParent(elem):
+                        if isinstance(elem, GangaObject):
+                            elem._setParent(my_parent)
+                        return
+                    for l in returnable_list:
+                        safeSetParent(l)
```
milliams: This should be reverted to a simple `for` loop.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -201,29 +217,24 @@ def strip_proxy_list(self, obj_list, filter=False):
...
-                category = o._category
-                if not category in result:
-                    result.append(category)
+        def return_cat(elem):
+            if hasattr(elem, 'category'):
+                return elem._category
             else:
-                result.append(type(o))
-        return result
+                return type(o)
```
milliams: `o` does not exist. Also, this is returning a single type, not a list of them.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -552,20 +562,24 @@ def printSummaryTree(self, level=0, verbosity_level=0, whitespace_marker='', out
...
+        def stringFunc(element, returnable_str):
+            if isType(element, GangaObject):
                 returnable_str += repr(stripProxy(element))
             else:
                 returnable_str += "'"
                 returnable_str += str(stripProxy(element))
                 returnable_str += "'"
             returnable_str += ", "
+        for l in self._list:
+            stringFunc(l, returnable_str=returnable_str)
```
milliams: This should be reverted to a simple `for` loop.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -577,29 +591,44 @@ def toString(self):
...
                 visitor.componentAttribute(self, name, getattr(self, name), item['sequence'])
 
+        for item in self._schema.simpleItems():
+            simpleFunc(item)
+
+        for item in self._schema.sharedItems():
+            sharedFunc(item)
+
+        for item in self._schema.componentItems():
+            componentFunc(item)
```
milliams: These should be reverted to simple `for` loops.

python/Ganga/Runtime/bootstrap.py
```diff
@@ -1248,7 +1248,7 @@ def launch_IPython(local_ns, args, error_handler, ganga_prompt):
         from Ganga.Utility.logging import enableCaching
         enableCaching()
 
-        ipshell.set_hook("pre_run_code_hook", ganga_prompt)
+        ipshell.set_hook("pre_prompt_hook", ganga_prompt)
```
milliams: Is this change intended to be a part of this PR?

python/Ganga/Utility/Config/Config.py
```diff
@@ -157,18 +157,19 @@ def f(*args, **kwds):
...
+            raise ValueError('config name %s is not a valid python identifier' % name)
         else:
             logger = getLogger()
-            logger.warning('obsolete config name found: replaced "%s" -> "%s"' % (name, name2))
+            logger.warning('obsolete config name found: replaced "%s" -> "%s"' % (name, name3))
             logger.warning('config names must be python identifiers, please correct your usage in the future ')
-            translated_names[name] = name2
+            translated_names[name] = name3
+    elif name in translated_names:
+        pass
```
milliams: This `elif` block should be removed as it does nothing.

python/Ganga/GPIDev/Base/Objects.py
rob-c: It's much more efficient than adding objects to a dictionary and when you're doing this a lot I thought it's worth the few cycles that we save per object.

python/Ganga/GPIDev/Base/Objects.py
rob-c: No but it's useful to see that there is a copy being performed in this call to explain why it's so heavy.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: It's in a sub-function which I'll revert

python/Ganga/Runtime/bootstrap.py
rob-c: not really waiting on #549

python/Ganga/GPIDev/Base/Objects.py
rob-c: Probably, I'll revert and check for problems.

python/Ganga/Utility/Config/Config.py
rob-c: This was to avoid doing the else replacement of assigning an object to itself which takes more cycles than checking if it's already in place as you still need to find an object in a dictionary to replace it don't you?

rob-c:
I think that's everything that's been addressed and if we're waiting on #549 Then I'm looking to merge this asap. Any additional comments?

python/Ganga/Utility/Config/Config.py
milliams: True, but that is a *very* micro optimisation being performed by a strangely written bit of code with no comment to explain it. Code should express intent first and only worry about optimisations like this is they are really going to be a problem. Over the course of a whole run of the program I guess this function might, at most, be run a few hundred times (~360 by my count) which will add up to milliseconds in efficiency gains at most. It is simply not worth it.

What is worth it is looking at the slightly bigger picture and realising that this function `_migrate_name()` is here to convert very old config section names from code going back to older than the repo history goes (>6 years). Not once does it actually seem to ever change any names. If you want to increase efficiency then simply removing this function entirely will do a much better job.

python/Ganga/GPIDev/Base/Objects.py
milliams: If this function is *always* going to be called with `make_copy` explicitly set (for the reasons you say) then why is it an optional argument?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -35,10 +35,11 @@ def _getName(obj):
     if returnable is None:
         returnable = getattr(getattr(obj, '__class__', None), '__name__', None)
     if returnable is None:
-        returnable = str(obj)
-    return returnable
+        return str(obj)
+    else:
+        return returnable
```
milliams: There is a general principle to prefer single exit points for functions. It's not a hard rule but it does make understanding the flow of a function easier. Given that this change is functionally equivalent to the previous version but is more complex, is it worth it?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -484,9 +481,20 @@ def cloneVal(v):
         if isinstance(new_val, Node):
             new_val._setParent(obj)
 
-        obj.setSchemaAttribute(_getName(self), new_val)
+        return new_val
 
-        obj._setDirty()
+    @staticmethod
+    def cloneVal(v, extra_args):
```
milliams: Why do we now have `__cloneVal` and `cloneVal`? Do we really need both?

python/Ganga/GPIDev/Base/Objects.py
rob-c: This comes about through mangling this function several times after changes weren't wanted, reverting

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -604,18 +615,17 @@ def __init__(self):
...
-
         if self._schema is not None and hasattr(self._schema, 'allItems'):
+            self._data_dict = dict.fromkeys(self._schema.datadict)
             for attr, item in self._schema.allItems():
                 ## If an object is hidden behind a getter method we can't assign a parent or defvalue so don't bother - rcurrie
                 if item.getProperties()['getter'] is None:
                     setattr(self, attr, self._schema.getDefaultValue(attr))
-
+        else:
+            self._data_dict = {}
```
milliams: I'm pretty sure that `self._schema` will never be `None` as it is always filled in by the metaclass. The whole `if` statement above is not needed as far as I can tell.

python/Ganga/GPIDev/Base/VPrinter.py
```diff
@@ -316,27 +320,28 @@ def summary_print(obj, out=None, interactive=False):
...
                     # remove trailing whitespace and newlines
-                    outString += result.rstrip()
-                count += 1
-                if count != obj_len:
-                    outString += ', '
+                    outStringList.append(str(x).rstrip())
+                # Check if we're at the last element or not and add commas
+                if len(outStringList)+1 != 2*obj_len:
+                    outStringList.append(', ')
+            outString += ''.join(outStringList)
```
milliams: Does this optimisation (appending a str → apending a list and then joining) speed things up? When I tried this out previously it made no difference and even made things slower in some cases.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -331,8 +338,8 @@ def __gt__(self, obj_list):
     def __hash__(self):
         logger.info("hash")
         result = 0
-        for element in self:
-            result ^= hash(element)
+        for element in [hash(l) for l in self._list]:
+            result ^= element
```
milliams: Why make a temp list, only to then iterate over it? The old code was simpler and probably faster.

python/Ganga/GPIDev/Schema/Schema.py
milliams: How much of a speed difference does it actually make? Simplifying code doesn't need justification but obfuscating it does.

python/Ganga/GPIDev/Base/Objects.py
rob-c: Renaming one of them

python/Ganga/GPIDev/Base/Objects.py
rob-c: Happily removing

python/Ganga/GPIDev/Base/VPrinter.py
rob-c: I managed to get some speed improvement from this when testing but it did depend on doing something lightweight to determine if the additional ',' character needs to be added. Would have to go looking for the benchmarks but this helped when generating a very large list I found

milliams: Ok, I agree that common Python wisdom says that this way should be faster so I'm happy to be convinced.

Would doing:
```python
outString += ', '.join(outStringList)
```
and removing the
```python
# Check if we're at the last element or not and add commas
 if len(outStringList)+1 != 2*obj_len:
    outStringList.append(', ')
```
not be a simpler way in comma separating the string though? Since that's basically what `join()` is for. Or am I missing something?

rob-c: Doesn't that give us an extra ',' at the start? Tests are running with the change now. If it causes problems I'm tempted to revert

milliams: Nope, the string is used as an interstitial element only:
```python
>>> l = ['one', 'two', 'three']
>>> ', '.join(l)
'one, two, three'
```

milliams: Wait, I think you misunderstood my suggestion. I was suggesting changing:
```python
outStringList = []
for x in _obj:
    if isType(x, GangaObject):
        sio = StringIO()
        stripProxy(x).printTree(sio, interactive)
        result = sio.getvalue()
        # remove trailing whitespace and newlines
        outStringList.append(result.rstrip())
    else:
        # remove trailing whitespace and newlines
        outStringList.append(str(x).rstrip())
    # Check if we're at the last element or not and add commas
    if (len(outStringList)+1) != 2*obj_len:
        outStringList.append(', ')
outString += ''.join(outStringList)
```
to
```python
outStringList = []
for x in _obj:
    if isType(x, GangaObject):
        sio = StringIO()
        stripProxy(x).printTree(sio, interactive)
        result = sio.getvalue()
        # remove trailing whitespace and newlines
        outStringList.append(result.rstrip())
    else:
        # remove trailing whitespace and newlines
        outStringList.append(str(x).rstrip())
outString += ', '.join(outStringList)
```

rob-c:
OK, any objections to merging?

rob-c:
OK, this should finally match develop again. I managed to get a very confused git history there so I'm not sure what has happened

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -552,14 +557,16 @@ def printSummaryTree(self, level=0, verbosity_level=0, whitespace_marker='', out
                 summary_print(self, out)
                 return
 
+        #logger.debug("self._list: %s" % self._list)
+
```
milliams: These added logging statements can be removed.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -527,6 +531,7 @@ def findSchemaParentSchemaEntry(self, parent):
         return result
 
     def printSummaryTree(self, level=0, verbosity_level=0, whitespace_marker='', out=sys.stdout, selection='', interactive=False):
+        #logger.debug("self._list: %s" % self._list)
```
milliams: These added logging statements can be removed.

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -27,6 +27,7 @@
 _found_components = {}
 _found_configs = {}
 _found_attrs = {}
+_found_eval = {}
```
milliams: This variable is not used anywhere.

milliams:
Once those few comments I just made are resolved, I'm happy to merge. Also, this PR is now removing more than it is adding :smile:

rob-c:
Just because a PR removes more than it adds doesn't directly effect what it's worth I'd strongly argue. I've worked with code where that has introduced hidden bugs

rob-c:
OK, I fixed a minor bug or 2 after walking through the code by eye to find any obvious problems. (The tests are still passing which seems to indicate we're not doing anything bad to impact calls to GangaObject which is good to see).
This has pushed the performance back slightly and now the benchmark function I quote above now finishes in 1min 15sec rather than ~50sec, however this is still well below the time it used to take which was running into several minutes.

I've synchronized this branch with develop and will look to merge at 12 today unless there are any objections.

I added some documentation over the weekend after deciding to go through a lot of code and make sure I knew what everything did.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -529,23 +607,34 @@ def __init__(cls, name, bases, this_dict):
...
         # sanity checks for schema...
-        if '_schema' not in this_dict.keys():
+        if '_schema' not in this_dict:
             s = "Class %s must _schema (it cannot be silently inherited)" % (name,)
             logger.error(s)
             raise ValueError(s)
 
+        attrs_to_add = []
+        for attr, item in this_schema.allItems():
+            attrs_to_add.append(attr)
```
milliams: This would be simpler as a list comprehension.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -577,7 +667,7 @@ def __init__(cls, name, bases, this_dict):
 
 
 class GangaObject(Node):
-    __metaclass__ = ObjectMetaclass
+    __metaclass__ = ObjectMetaclass # This overloads the default metaclass so we can do some clever stuff
```
milliams: This comment doesn't really explain anything. It's standard Python.

python/Ganga/GPIDev/Base/Objects.py
rob-c: changing in this case.

python/Ganga/GPIDev/Base/Objects.py
rob-c: yes but it was undocumented and I wanted to remind myself that the `__metaclass__` is defined within the python object not something inherited and constructed in Node or somewhere else.

rob-c:
OK, merging at 12:45

PR #362, opened 2016-04-08
Minor Little Ganga Bugfixes

drmarkwslater:
This PR fixes a couple of things in little ganga:

1. Set the gangadir properly with a username
2. Exports the runMonitoring command to the GPI


python/ganga/__init__.py
```diff
@@ -56,6 +56,8 @@ def ganga_license():
 interactive = False
 from Ganga.Core.GangaRepository import getRegistrySlice
 Ganga.Core.bootstrap(getRegistrySlice('jobs'), interactive)
+from Ganga.Runtime.GPIexport import exportToInterface
+exportToInterface(ganga, 'runMonitoring', Ganga.Core.monitoring_component.runMonitoring, 'Functions')
```
rob-c: You should just be able to pass ```ganga``` as an extra argument to ```bootstrap``` above or you end up exporting this to the ```GPI``` as well which is a little bad as this interface shouldn't be used when doing ```import ganga```

python/Ganga/__init__.py
```diff
@@ -26,6 +26,14 @@ def getLCGRootPath():
     else:
         return ''
 
+def getDefaultUserName():
+    # detect default user (equal to unix user name)
+    import getpass
+    try:
+        return getpass.getuser()
+    except Exception as x:
+        raise Ganga.Utility.Config.ConfigError('Cannot get default user name' + str(x))
```
milliams: If we never handle this exception (and I'm not sure under what circumstances `getuser()` would raise), perhaps we can make the line below be just:
```python
conf_config.addOption('user', getpass.getuser(), ...
```

rob-c:
Is it a trivial fix to get ```import ganga``` to use the .ini files defined in the bash env?
(is it understood what we used to be doing with the old system)

If I try and run with this I suspect unless I tell my ```.gangarc``` that I want to load a plugin it's not done according to any external config it won't pick up things like the Dirac backend for me.

python/ganga/__init__.py
drmarkwslater: I'm an idiot. Fixed :smile:

python/Ganga/__init__.py
drmarkwslater: That makes sense. Fixed.

rob-c:
Go for merge from me, I'm just curious if ```import ganga``` works safely if there is no ```.gangarc``` but I think this is a separate issue for another PR.

drmarkwslater:
I'm not sure and I would prefer to leave that for another PR as I suspect this could be a big(ish) job. I've just created an issue to track it actually (#363). These fixes were what I came across while getting the Tutorial Tests to work as I wanted that to use little ganga.

drmarkwslater:
Assuming everything works, any objections to merge?

milliams:
:+1:

PR #365, opened 2016-04-08
Changing logger catch behaviour

### LINKED ISSUE #364: logger is not buffering correctly
The logger isn't buffering until carriage return.

This is annoying as there are logging statements made every 30sec by default on the debug console which get displayed every time I press a character rather than waiting for a carriage return.

This gets annoying as I'm having to turn off more and more features in order to be able to test things with full debugging on.

rob-c:
Fixes #364 

I think we should only be flushing when the logging has come from the main thread and the user wants to see what is going on. I think all other threads, named "Thread-"..., "Ganga_Update_"... or "Dummy"... should buffer until it's more sensible to flush.


milliams:
:+1:

rob-c:
Hmmm I see
```
python/Ganga/new_tests/GPI/Parallel/TestQueuedSJSubmit.py::TestQueuedSJSubmit::test_d_Finished FAILED
```
on the test machine. I suspect this is just an oddity as this is a 1 line change in the logging.

drmarkwslater:
These errors may be due to the /tmp dir on the machine getting full up (there were ~16K directories!). @milliams is looking at cleaning it up and hopefully that will help...

rob-c:
hmmm, interesting, I wonder whether this is expected behaviour or not.

Any objections to merging the branch now?

rob-c:
retest this please

rob-c:
retest this please

Issue #360, opened 2016-04-07
Subjob marked as dirty makes all subjobs get flushed


drmarkwslater:
At present, according to `flush` in SubJobXMLList.py, if the parent job is marked as dirty (which it will be if any subjobs change) then ALL the subjobs get flushed. This can be an expensive operation if there are a lot of subjobs which could make the monitoring go very slowly.

Ideally, it should only flush dirty subjobs, not all of them.

egede:
Just to underline that it is suspected that this issue could be the cause of monitoring slowing down in LHCb.

rob-c:
I know we mark Job objects as not dirty after a flush but I don't think we correctly address marking all objects within the schema as clean afterwards, just the parent. Can we improve the ```_setFlushed``` method so we do things in future like what attributes are triggering most objects to become dirty.

egede:
> Can we improve the _setFlushed method so we do things in future like what attributes are triggering most objects to become dirty.

@rob-c I do not understand what you mean here? You want to make sure that we go in and mark the subjobs as not dirty after flushing or do you want to collect statistics on what is dirty in jobs?

rob-c:
@egede I want to be sure that we mark all dirty objects as clean after flushing. Atm an object and all of it's parents are marked dirty with a ```_setDirty```. We only mark the root object clean after flushing. I would like to make sure we mark all objects hanging off the root as clean.

In future we can think about looking at which objects are triggering this if we feel that objects are excessively dirty when they shouldn't be changing, but atm all objects are all dirty under the parent regardless of if they were dirty 5min ago or 5 days ago.

egede:
@rob-c OK. Essential thing is indeed to make sure that everything that is flushed is actually marked clean afterwards.

rob-c:
Fixed by #367

PR #367, opened 2016-04-08
Proposed flush fix

rob-c:
This is a proposed quick fix for #360. I've not had time to test it but it 'feels' faster from a very unscientific test of submitting 200 subjobs and monitoring them by hand, but this could be the hostname changes from recent changes.

I'm curious to see if the tests are effected by this.


rob-c:
Ouch, 'climbing down' the ladder appears to be a bit dangerous. I'm seeing some infinite loops when the ```_setFlushed``` trys to climb down the children within the schema.

I'll do some looking at this over the weekend if I get a chance. A bit disappointing as this worked in an isolated test by itself.

rob-c:
OK, I think I've found a fix, lets see if all tests pass with this.

rob-c:
retest this please

rob-c:
I just fixed a bug in the XML test. Basically the subjob XML was only written once even though the subjob went from new to completed in under 30sec (proof that this PR is working).
The fix was to mark all subjobs dirty and trigger a flush of the ```SubJobXMLList``` class to make sure at least 2 write events happened on a per-subjob basis.

This seems to save time (on subjob related tests) so I'm hoping to do some testing to see if monitoring a job with 1k subjobs is now manageable.

rob-c:
Any objections to merging this first thing Monday?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -957,12 +957,20 @@ def _getRegistryID(self):
...
+        """ Set the dirty flag all the way up to the parent"""
         self._dirty = True
         parent = self._getParent()
         if parent is not None:
             parent._setDirty()
 
     def _setFlushed(self):
+        """Un-Set the dirty flag all of the way down the schema."""
+        if self._schema:
+            for k in self._schema.allItemNames():
```
milliams: This should only need to be done over the component items.

rob-c: it _should_ but just in-case this hasn't been implemented correctly (and I know of a few places where this isn't implemented 100% correctly)

PR #368, opened 2016-04-08
Add append and extend functions to make GangaDataset more list like

drmarkwslater:
A quick addition to GangaDataset to make it behave more like a list. Just adds `extend` and `append` methods.


rob-c:
hmm, I'll explore if this effects the ```LHCbDataset``` and whether I'll need to mask the functions at that level.

drmarkwslater:

PR #369, opened 2016-04-11
Fix select

### LINKED ISSUE #350: jobs.select is broken again
I expect the following: ```jobs.select(id=1322)``` to return the job with id 1322, however when I run this command I get my whole repo returned as a jobs slice.

However: ```jobs.select(1322, 1322)``` works as expected and returns the one job.

The jobs.select method is used __heavily__ within the monitoring system and should not be doing this.
(I suspect fixing this would also fix #286.)

This just totaled my entire repo of custom jobs which are not all loadable at runtime due to a combination of this and #164.

Can we fix this for 6.1.19 please.

rob-c:
This fixes #350 and adds some more detailed tests to better check that ```jobs.select``` doesn't break again in the future without us noticing.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
+++ b/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
@@ -122,18 +122,18 @@ class RegistrySlice(object):
         logger = getLogger()
 
         this_repr = repr.Repr()
-        from Ganga.GPIDev.Base.Proxy import GPIProxyObjectFactory
+        from Ganga.GPIDev.Base.Proxy import addProxy
         attrs_str = ""
         ## Loop through all possible input combinations to constructa string representation of the attrs from possible inputs
-        ## Reuired to flatten the additional arguments into a flat string in attrs_str
+        ## Required to flatten the additional arguments into a flat string in attrs_str
         for a in attrs:
             from inspect import isclass
             if isclass(attrs[a]):
-                this_attr = GPIProxyObjectFactory(attrs[a]())
+                this_attr = addProxy(attrs[a]())
             else:
                 from Ganga.GPIDev.Base.Objects import GangaObject
                 if isType(attrs[a], GangaObject):
-                    this_attr = GPIProxyObjectFactory(attrs[a])
+                    this_attr = addProxy(attrs[a])
                 else:
                     if type(attrs[a]) is str:
                         from Ganga.GPIDev.Base.Proxy import getRuntimeGPIObject
@@ -236,6 +236,7 @@ class RegistrySlice(object):
                                 'undefined select attribute: %s' % str(a))
                     else:
 
+                        logger.debug("Testing: %s MOO!" % a)
                         if a == 'ids':
                             if int(this_id) not in attrs['ids']:
                                 selected = False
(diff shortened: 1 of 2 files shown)


rob-c:
I'll leave this up for a short while given GridPP36 any objections to merging this PR?

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -236,13 +236,18 @@ def select_by_range(this_id):
                                 'undefined select attribute: %s' % str(a))
                     else:
 
+                        logger.debug("Testing: %s MOO!" % a)
```
milliams: I'm guessing this can be removed?

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -236,13 +236,18 @@ def select_by_range(this_id):
...
                             if int(this_id) not in attrs['ids']:
                                 selected = False
                                 break
                         else:
                             try:
                                 item = obj._schema.getItem(a)
+                                logger.debug("obj: %s" % str(obj))
+                                logger.debug("Here: %s, item: %s" % (a, str(item)))
+                                logger.debug("%s = %s" % (str(a), str(attrs[a])))
+                                logger.debug("attrs[a] == item %s" % (attrs[a] == item))
```
milliams: I'm guessing that a lot (all?) of this can be removed?

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
rob-c: Yes, I was trying to determine what code paths where actually being called, this can at least be trimmed down

python/Ganga/new_tests/GPI/Parser/TestSelect.py
```diff
@@ -2,47 +2,68 @@
...
 
-    def test_SelectJob(self):
-        
-        from Ganga.GPI import Job, jobs
-        j=Job()
+    def setUp(self):
+        """Make sure that the Job object isn't destroyed between tests"""
+        super(TestSelect, self).setUp()
+        from Ganga.Utility.Config import setConfigOption
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
```
milliams: I'd like to avoid doing multi-pass style tests unless we really are testing the persistency features of Ganga. Tests really should be independent where possible unless there's a very good reason to couple them. We shouldn't be using tests functions in a class to group tests in order but rather to test separate things.

rob-c: Actually there is some level of this going on. The objects are either SubJobXML, Registry or a plain list in some cases depending on how the RegistrySlice has been defined. I don't strictly play-with/check this in the test but I felt having it change a bit is a healthy mix to avoid problems creeping in.

milliams: Ok, in that case I think we should make this more explicit. If one of the tests is checking the behaviour of a `SubJobXMLList` then it should at least say so and perhaps even do a type-check.

In general though, the jobs should be created in `setUp` and then tested in a test function. If we want to test whether `select` works on a lazy-loaded job then I guess that's another reason to look into the ability to manually 'unload' jobs from the registry.

rob-c: I would but the exact content of the objects varies at different stages depending on things like flush operations to disk. This is difficult and short of turning all of the auto-flushing off, manually putting in a flush op, testing things before and after the flush op that the select works as expected. And there we have PR #372 come into play.

This can be done, but is potentially overkill for just checking what is checked here.

I'm not sure that this is the best motivation to push for cleanly un-loading objects from the repo. (I think #374 is the best argument for having a clean load/unload mechanism) but yes having a load/unload would be nice, just getting far out of scope of the PR.

rob-c:
Are there any objections to merging this branch?
The test can be changes/expanded/fixed but essentially is a re-write and I'm quite keen to fix #350 ahead of 6.1.19

milliams:
:+1:

PR #372, opened 2016-04-12
Changing RegistrySlice to have ids better determined

rob-c:
Small cleanup of the use of keys within ```RegistrySlice``` This was 'fixed' as part of making sure ```SubjobXMLList``` worked with subjobs  properly from a user perspective but I've realised it should be used everywhere in the class as it's more correct.

I've checked ```SubjobXMLList``` and it now correctly supports the keys method in a way which doesn't cause problems.

I've tried to make sure that the ```keys()``` method isn't called in lists etc as this is a locked list which if it's called many times behind the scenes due to some poor code there will at least be a performance cost due to the number of calls to the function with the constant (un)locking.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
+++ b/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
@@ -41,7 +41,7 @@ class RegistrySlice(object):
         self._proxyClass = None
 
     def _getObjKeys(self):
-        return deepcopy(self.objects.keys() if not isType(self.objects, SubJobXMLList) else [_id for _id in range(len(self.objects))])
+        return deepcopy(self.objects.keys())
 
     def _getColour(self, obj):
         """ Override this function in derived slices to colorize your job/task/... list"""
--- a/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
+++ b/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
@@ -40,9 +40,6 @@ class RegistrySlice(object):
         self._colour_normal = Effects().normal
         self._proxyClass = None
 
-    def _getObjKeys(self):
-        return deepcopy(self.objects.keys())
-
     def _getColour(self, obj):
         """ Override this function in derived slices to colorize your job/task/... list"""
         return self._colour_normal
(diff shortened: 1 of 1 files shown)


rob-c:
retest this please

rob-c:
retest this please

rob-c:
Any objections to merging this, it's mainly a cleanup now that ```SubJobXMLList``` now supports keys correctly.

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -304,7 +301,8 @@ def copy(self, keep_going):
         return this_slice
 
     def __contains__(self, j):
-        return j.id in self.objects.keys()
+        allObjK = self.objects.keys()
+        return j.id in allObjK
```
milliams: Why was this split onto two lines? Likewise with the two similar changes below. I feel it makes the code slightly more verbose for no gain.

rob-c: This was to reduce the objects.keys potentially being called multiple times. I'm no expert on when python evaluates commands on inline stuff like this and I was going for clarity that this was called just once.

milliams: The code is exactly the same before and after the change. When doing:
```python
j.id in self.objects.keys()
```
it only calls `self.objects.keys()` once.

rob-c: Is this always the case in for loops in addition to the ```in``` test?

milliams: Yes, the thing `Y` in a `for X in Y:` will be assigned to a temporary internal variable and have `iter()` called on it which then has `next()` called on it until it raises a `StopIteration` exception. In this case, `keys()` returns a `list` which which, when passed to `iter()` returns a `listiterator` object. Something like:
```python
for X in Y:
    ... something using X ...

→

_Y = iter(Y)
while True:
    try:
        X = next(_Y)
        ... something using X ...
    except StopIteration:
        break
```
so `Y` is only accessed once.

rob-c: Yes but we can't do ```for x in y``` where y is a ```SubJobXMLList``` in which case ```.keys()``` allows for lazy loading

milliams: I'm afraid I don't understand.

rob-c: This was just a comment as to why we can't drop the use of some ```.keys``` calls, but I hope this is clear in the revision after this which touches less code.

rob-c:
Any objections to squashing and merging this if the tests pass?

milliams:
:+1:

PR #373, opened 2016-04-12
Stop and join job monitoring thread

milliams:
Currently, we are not joining the thread which is maintaining the job monitoring. This results in more than a hundred open threads persisting through the tests which is potentially causing a problem in the Jenkins test system.

Even if it's not causing a problem, it will at least clean up the thread tracer output.

### DIFF
--- a/python/Ganga/Core/InternalServices/ShutdownManager.py
+++ b/python/Ganga/Core/InternalServices/ShutdownManager.py
@@ -74,6 +74,8 @@ def _ganga_run_exitfuncs():
         from Ganga.Core.MonitoringComponent.Local_GangaMC_Service import getStackTrace
         getStackTrace()
         monitoring_component.disableMonitoring()
+        monitoring_component.stop()
+        monitoring_component.join()
 
     ## Stop the tasks system from running it's GangaThread before we get to the GangaThread shutdown section!
     from Ganga.GPIDev.Lib.Tasks import stopTasks


rob-c:
Looks good. Once the main tests are over please merge :)

drmarkwslater:
I'd give it a few retests first :smile:

python/Ganga/Core/InternalServices/ShutdownManager.py
```diff
@@ -74,6 +74,8 @@ def _ganga_run_exitfuncs():
         from Ganga.Core.MonitoringComponent.Local_GangaMC_Service import getStackTrace
         getStackTrace()
         monitoring_component.disableMonitoring()
+        monitoring_component.stop()
+        monitoring_component.join()
```
drmarkwslater: I had actually thought to put this lower down, just before the registries are shutdown (to be safer) but if it works here then that's OK as well :smile:

drmarkwslater:
Btw, I can certainly confirm that this stops the MANY thread problem. Let's see if it fixes the other issues as well :)

drmarkwslater:
retest this please

python/Ganga/Core/InternalServices/ShutdownManager.py
rob-c: The sooner the better in shutdown imo to stop extra threads starting up and getting in the way

rob-c:
retest this please

rob-c:
I think this had fixed a lot of our woes but it would still be work giving it a few runs though the testing :)

milliams:
The tests failed (after two successful runs) due to a tasks problem of which I have sent Mark the details. I'm running the tests once more as https://ganga-ci.cern.ch/jenkins/job/PR%20builder/702/ so we'll see how that does. If it passes I'm inclined to merge this?

rob-c:
I say go for merge as this I don't think causes any more problems but fixes some already.

milliams:
retest this please

Issue #374, opened 2016-04-14
Possible source of lockups with GangaObject._getWriteAccess


drmarkwslater:
I've just fixed a lockup in Tasks where the Task Monitor thread but I'm concerned this could cause problems in the future. The issue is this:

1. Object (Job or Task) is NOT loaded
2. thread calls `obj._getWriteAccess` which explicitly gets the `const_lock`
3. During this, it also starts to load the object which at various times will try to get the `internal_lock`
4. In another thread, the object is accessed which grabs the `internal_lock`
5. After this, it calls `obj._getReadAccess` attempt to load.
6. This then tries to grab the `const_lock`.

The order is not consistent in the two cases and so can lead to deadlocks.

This was a problem in Tasks because of the specific call to `_getWriteAccess` but I'm concerned this could happen elsewhere. I'm not sure how to proceed with this but I thought I'd put in an issue so we remember :smile:

rob-c:
I've been fighting with the Access locks elsewhere and the Registry simply stops doing them during Registry transactions which isn't ideal.

I think a _simple_ approach would be to abandon the read/write lock for explicitly calling a load of the job and disallowing multiple ganga sessions from accessing an object loaded by another ganga session? Do we really want/need the complexity that we have with the read/write access locks? (I really don't know here so if we do then this is fine)

milliams:
The read/write access locks (maintained as physical files on disk) are useful/necessary if we want to support two active Ganga session accessing the same repository. I'm not convinced that this is a feature that we want to support to this extent and without a good set of usecases/user-stories it's more of a burden than anything. I would be in favour of removing per-object session locks until we can reimplement it properly once we know what it's actually useful for. Someone needs to write down a set of situations/use-cases where it would be useful before we can really ascertain what level of complexity is warranted.

However, having a single global session lock is probably still useful to inform the user if there's still another Ganga session running.

rob-c:
Given that the locks at present prevent 2 sessions from modifying the same job I would rather not remove it without completely removing the ability to have 2 Ganga sessions.

Warning the user isn't enough as some(most?) don't pay attention to the warnings on startup until they run into more serious issues later on such as corrupting their job with 1k subjobs.

We still have to put in a mechanism to disable the monitoring if we detect another Ganga session is already running btw which was brought up in another PR a while ago.

egede:
[...]

PR #375, opened 2016-04-14
Make sure consistent id used to defer _getRead and _getWrite

rob-c:
Make sure the ```_inprogressDict``` keeps a consistent set of ```id()``` to allow for ```_getReadAccess``` ```_getWriteAccess``` calls at the repo level from causing problems. Possibly alleviates #374.

This was quicker to fix/demonstrate with the PR, if anyone has any further questions please ask.

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -733,7 +733,7 @@ class Registry(object):
     def _updateIndexCache(self, _obj):
         logger.debug("_updateIndexCache")
         obj = stripProxy(_obj)
-        if self.find(obj) in self._inprogressDict.keys():
+        if id(obj) in self._inprogressDict.keys():
             return
 
         self.repository.updateIndexCache(obj)
@@ -784,7 +784,7 @@ class Registry(object):
     def __safe_read_access(self,  _obj, sub_obj):
         logger.debug("_safe_read_access")
         obj = stripProxy(_obj)
-        if self.find(obj) in self._inprogressDict.keys():
+        if id(obj) in self._inprogressDict.keys():
             return
 
         if self.hasStarted() is not True:
@@ -834,7 +834,7 @@ class Registry(object):
     def __write_access(self, _obj):
         logger.debug("__write_acess")
         obj = stripProxy(_obj)
-        this_id = self.find(obj)
+        this_id = id(obj)
         if this_id in self._inprogressDict.keys():
             for this_d in self.changed_ids.itervalues():
                 this_d.add(this_id)
@@ -902,7 +902,7 @@ class Registry(object):
         logger.debug("_release_lock")
         obj = stripProxy(_obj)
 
-        if self.find(obj) in self._inprogressDict.keys():
+        if id(obj) in self._inprogressDict.keys():
             return
 
         if self.hasStarted() is not True:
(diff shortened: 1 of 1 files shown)


milliams:
So do I understand that with this change, `self.changed_ids` has the registry id of an object as its key but `self._inprogressDict` has the python `id()`?

I'm not sure why `find()` is not just:
```python
def find(self, _obj):
    return next(for obj in self._objects if obj is _obj)
```
(with maybe a smidgeon more error handling)?

rob-c:
a) because the find method will fail when adding the object to the registry. It also makes no sense to get a read/write lock on an object as you're adding it to the dictionary, as it makes no sense to get access while flushing or something similar.

b) aside from the error handling and assert statements this is what the find is? (I think)
The problem with doing literally what you suggest is that this becomes extremely expensive and requires correctly comparing the full content of a job or a partly-loaded job. Working out how to do this without violating lazy-loading is potentially very difficult imo.
If we get the object equivalency sorted out then I think we can make the find method simpler.

milliams:
a) Where are we calling `find` before adding a job to the registry? I agree that we shouldn't do that.

b) My problem with the current `find()` method is the chunk that looks like:
```python
...
if hasattr(obj, _reg_id_str):
    obj_reg_id = getattr(obj, _reg_id_str)
    objects_obj = self._objects[obj_reg_id]
    assert obj == objects_obj
    if hasattr(obj, _id_str):
        if hasattr(objects_obj, _id_str):
            assert getattr(obj, _id_str) == getattr(objects_obj, _id_str)
    assert obj_reg_id == getattr(objects_obj, _reg_id_str)
    return obj_reg_id
elif hasattr(obj, _id_str):
    obj_id = getattr(obj, _id_str)
    objects_obj = self._objects[obj_id]
    if hasattr(objects_obj, _reg_id_str):
        assert obj_id == getattr(objects_obj, _reg_id_str)
    assert obj_id == getattr(objects_obj, _id_str)
    return obj_id
...
```
It's messy and repetitive and it's not clear what it's doing. I don't think we should be storing on the object what its registry id is as that is duplicating information which then needs to be kept in line.

There's no issue with the object equality as my example uses `is` rather than `==`. It's saying 'is this object in the dict, if so give us the `id`'

rob-c:
And strictly __not__ the point of this PR.

If we want to clean this then OK, but all I'm saying is that we should be using the ```id()``` method with the dictionary rather than making the repo call as that returns the ```obj.id``` I'm making no statement about the find method which can be changed after this.

rob-c:
a) the inprogress stuff stops a call to _getReadAccess or _getWriteAccess which just confuses things during the add process and at worst can trigger loading of a half-written job from disk as the XML is being defined. It should use ```id()``` and not ```self.find()```

milliams:
Ok, I'm not arguing against this PR but simply thinking aloud about exactly why these things keep causing us problems.

I'm happy for this PR to go in once the declaration of `self._inprogressDict` in `__init__` has a comment making clear that the keys of the dictionary should be Python `id()`s rather than registry ids.

rob-c:
OK, I can add the comment but I'm just wanting to avoid touching anything here other than fixing a specific problem.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -252,7 +252,7 @@ def __init__(self, name, doc, dirty_flush_counter=10, update_index_time=30, dirt
         self._dirty_max_timeout = dirty_max_timeout
         self._dirty_min_timeout = dirty_min_timeout
 
-
+        ## Id's id(obj) of objects undergoing a transaction such as flush, remove, add, etc.
```
milliams: PEP8 specifies a single `#` for line comments. No need to change it here but for future...

milliams:
Once the tests pass :+1:

PR #376, opened 2016-04-14
Fix tasks lockup

drmarkwslater:
This should fix the Tasks lockup seen in the testing. However, I can't recreate this locally so I guess we'll have to wait and see...

### DIFF
--- a/python/Ganga/new_tests/GPI/Tasks/TestCoreTasks.py
+++ b/python/Ganga/new_tests/GPI/Tasks/TestCoreTasks.py
@@ -53,7 +53,6 @@ class TestCoreTasks(GangaUnitTest):
     def test_b_TaskPersistency(self):
         """Check the task is persisted properly"""
         from Ganga.GPIDev.Base.Proxy import isType
-        from Ganga.GPI import tasks
         from Ganga.GPI import tasks, CoreTask, CoreTransform, Executable, GenericSplitter
 
         assert len(tasks) == 1
@@ -102,7 +101,7 @@ class TestCoreTasks(GangaUnitTest):
         for _ in range(0, self._numTasks):
             t = self.createTask()
             t.run()
-            
+
         assert len(tasks) == self._numTasks + 1
 
     def test_f_CheckAccessToManyTasks(self):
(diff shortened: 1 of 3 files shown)


python/Ganga/new_tests/GPI/Tasks/TestCoreTasks.py
```diff
@@ -44,6 +53,7 @@ def test_a_TaskCreation(self):
     def test_b_TaskPersistency(self):
         """Check the task is persisted properly"""
         from Ganga.GPIDev.Base.Proxy import isType
+        from Ganga.GPI import tasks
```
milliams: This line duplicates the one below.

python/Ganga/new_tests/GPI/Tasks/TestCoreTasks.py
```diff
@@ -85,3 +95,30 @@ def test_d_TaskStillRunning(self):
...
+            
+        assert len(tasks) == self._numTasks + 1
+
+    def test_f_CheckAccessToManyTasks(self):
+        """Access the Tasks while monitoring"""
+        from Ganga.GPI import tasks, CoreTask, CoreTransform, Executable, GenericSplitter
+        from Ganga.GPIDev.Base.Proxy import isType
+
+        for _ in range(0, self._numTasks):
+            for t in tasks:
```
milliams: Why does this loop over the number of tasks and then again over each task?

python/Ganga/new_tests/GPI/Tasks/TestCoreTasks.py
drmarkwslater: Fixed!

python/Ganga/new_tests/GPI/Tasks/TestCoreTasks.py
drmarkwslater: Just to keep hammering the access to Tasks while the monitoring is loading. I guess I can just replace this with a random number but I wanted it to scale with the number of Tasks that are being generated.

As I say though, I can't get this to fail locally but should fail make the failure more obvious on the testing machines.

milliams: Ok, that makes sense but perhaps a comment to that effect would be useful for 'future us'.

drmarkwslater: Added!

drmarkwslater:
Any objections to a merge after the tests complete?

milliams:
:+1:

rob-c:
Looks good to me :+1:

PR #378, opened 2016-04-14
Fix 2.6 compatibility in logging module

drmarkwslater:
The `MemoryHandler` is an old style `classobj` not a `type` and so the `super` call breaks in python 2.6. This PR fixes that.


milliams:
:+1:

drmarkwslater:
retest this please

rob-c:
hmm, another 2.6 oddity,  go for merge from me :+1:

PR #380, opened 2016-04-15
Changing when the import path is manipulated

rob-c:

It changes the point at which the import path is changed to be the ```import GangaUnitTest``` or equivalent. This gets rid of a few collection errors for me when I run it on a test system.


milliams:
:+1:

PR #381, opened 2016-04-15
Fixing full and summary print for proxied objects

### LINKED ISSUE #301: The full_print(j) function is broken
The simple test 

```
j=Job()
full_print(j)
```
prints out a reference to the job  object rather than the full listing.

```
[12:32:10]
Ganga In [9]: full_print(j)
<Ganga.GPIDev.Lib.Job.Job.Job object at 0x1c52b90> 
[12:32:13]

```

rob-c:
Fixes #301.

Objects were incorrectly having their proxies stripped and then the string representation of the proxy was being called. This should now make more sense.

Any objections to merging?


milliams:
I'm a bit confused by the code. Some lines are acting on `obj` and some on `_obj`. It's not clear to me why there's a difference. I feel having one function which does different things depending on what type was passed in is asking for trouble.

rob-c:
There is a ```stripProxy``` at the start of the function. This is not something special to just this code we have different behavior depending on if it's the raw object or the potentially proxied object. I can change the names if it's difficult to follow the ```obj``` vs ```_obj``` naming throughout the functions here.

rob-c:
OK, this highlighted a small bug but I'm confused as to why a method named printingGPIObject was not printing a GPI object... Still it's all a good test of the API.

rob-c:
Any objections to merging this once the latest tests pass?

PR #386, opened 2016-04-18
Feature/move tutorial and add tests

drmarkwslater:
This moves all the tutorial docs from the Wiki to the repo and means that readthedocs is up to date. It also adds the `TutorialTests.py` file that includes all testable code. This is quoted directly in the docs but is also included in tests. These tests should be improved in the future as I haven't added (m)any asserts, etc. but at least it tests for general errors and functionality.


drmarkwslater:

milliams:
:+1:

PR #387, opened 2016-04-19
Feature/improve condor options

drmarkwslater:
This adds some more options to Condor:

j.backend.cdf_options = {"test2": 0}  # add this to the CDF file
j.backend.accounting_group = "mygroup"  # add an accounting group to the CDF file

This also zeroes the requirements so by default, submission should be trivial.


milliams:
:+1:

will-cern:
The generic cdf_options dictionary looks very useful. Thanks!

will-cern:
Can cdf_options and accounting_group please be made settable in the .gangarc file

At the moment they are unrecognised members of the CondorRequirements defaults

Also in the latest prerelease, the 'getenv' property of Condor defaults seems to be confused between bool and str

drmarkwslater:
The `cdf_options` and `accounting_group` options should be available under `defaults_Condor`. You may need to regenerate the .gangarc using 6.1.20 to see them though I though they should be added automatically when you try the new version...

The `getenv` thing is definitely an issue which I need to think carefully about. See #494.

PR #391, opened 2016-04-20
Atlas fixes and cleanup

drmarkwslater:
This fixes some issues with default input/output datatypes and local/grid backends as well as removing a load of old TAG stuff that isn't used anymore.


milliams:
:+1:

PR #392, opened 2016-04-20
Fix GangaList not-equal

milliams:
If an object `is` another then not-equal should be `False`, not `True`


rob-c:
I'm worried that this was passing before now if you've changed something which looks quite fundamental. Was this exhibiting errors which we should have been aware of?
If so can you provide an example?

milliams:
The reason this wasn't failing the tests until now was due to the fact that when doing, in `TestGangaList::testNq`:
```python
self.assertFalse(self.proxied1 != self.proxied1, 'Lists are the same')
```
the object on the right of the `!=` was being passed through as a proxy object but the object on the left was being stripped. So that if, inside `GangaList.__ne__`, you do:
```python
def __ne__(self, obj_list):
	print(type(self))      # Add these
	print(type(obj_list))  # two lines
	if obj_list is self:
		return True
	result = True
	if self.is_list(obj_list):
		result = self._list.__ne__(self.strip_proxy_list(obj_list))
	return result
```
you get:
```
<class 'Ganga.GPIDev.Lib.GangaList.GangaList.GangaList'>
<class 'Ganga.GPIDev.Base.Proxy.GangaList'>
```

Since the two arguments can *never* be the same object (since one is a proxy object) the if statements has never been true and so the incorrect short-circuit check has never some into play. I discovered the error when implementing #393 which correctly strips all proxies from incoming arguments.

rob-c:
I'm struggling to understand this because surely the GangaList is never equal to it's proxy so shouldn't this have turned up due to something similar to the following failing:
```
from Ganga.GPI import GangaList
gl = GangaList()
assert (gl != gl) is False
assert gl == gl
```
If not can we add some more tests that are fixed by this so they don't break in the future.
(Maybe also test some more objects other than the GangaList which may have implemented their own ```__ne__``` functions.)

milliams:
# The Bug

The simplest way to demonstrate this bug is:
```python
from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList as GL
gl = GL()
gl == gl  # This should be True and is
gl != gl  # This should be False but is True!
```
This is happening because `GangaList.__ne__` has a check for `LHS is RHS` which, if `True`, returns `True`. This is incorrect and a bug.

# The hiding of the bug

The reason we have not seen this bug in the test suite etc. is that we are usually dealing with proxied `GangaList`s. 
```python
glp = GangaList()  # A proxied GangaList
glp == glp  # Should be True and is
glp != glp  # Should be False and is
```
What is happening here is that `glp != glp` goes though a number of steps before it gets to `__ne__`:

1. `glp != glp`
2. `Ganga.GPIDev.Base.Proxy.GangaList.__ne__(glp, glp)` then via `ProxyMethodDescriptor` goes to
3. `Ganga.GPIDev.Lib.GangaList.GangaList.__ne__(stripProxy(glp), glp)`

Since `stripProxy(glp) is not glp`, the incorrect (with `self` = `stripProxy(glp)` and `obj_list` = `glp`):
```python
if obj_list is self:
    return True
```
is not used and so it defaults back to the usual (and correct):
```python
self._list.__ne__(self.strip_proxy_list(obj_list))
```
which correctly returns `False`.

rob-c:
Can we make this into an integration test which is only passed by this PR in this case?

milliams:
We already have a failing test on #393 ([seen here](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/783/testReport/junit/python.Ganga.new_tests.Unit.GangaList.TestGangaList/TestGangaList/testNq/)) which demonstrates this bug.

It's difficult to add an integration test for this since there's actually two bugs here which are working in concert. The first is the one described in this PR and the second is that fixed by #393. Any integration test I write would have to do some strange manual circumvention of the proxy layer to test the right things which will change/break/be unnecessary once #393 is in.

Instead a proper unit test of `GangaList` is needed, separate to the GPI tests it has currently (another example of `GangaList` straddling the interface in a strange way). This is all going to be replaced with my `List`/`ProxyList` some time soon anyway I expect so I don't think it's the time to be rejigging the test layout for this class.

rob-c:
If the 2 PR are difficult to separate with the bug I may suggest they be linked, but as long as it's understood and tested for I think it's good to go

milliams:
They are two separate bugs (#393 isn't strictly a bug though) and so I thought it best to separate out the discussion of them. I will merge this one and then we should move discussion to #393.

milliams:
retest this please

PR #393, opened 2016-04-20
Add and remove proxies on all proxy method args

milliams:
Until now it seems that while schema attributes were being correctly stripped and wrapped as they were passed through the proxy-interface, the same was not being done for arguments and return values of exported methods. 

All proxy-exported methods are accessed via `ProxyMethodDescriptor` which now correctly removes the proxies from all incoming arguments and adds them to any return value.

This depends on #392 to make one of the `GangaList` tests pass so I will rebase this on develop once that is merged.

### DIFF
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -13,6 +13,7 @@ from Ganga.GPIDev.Schema import ComponentItem
 from Ganga.GPIDev.Base.Objects import Node, GangaObject, ObjectMetaclass, _getName
 from Ganga.Core import GangaAttributeError, ProtectedAttributeError, ReadOnlyObjectError, TypeMismatchError
 
+import functools
 import os
 
 from inspect import isclass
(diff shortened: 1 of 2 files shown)


python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -626,6 +627,24 @@ def __set__(self, obj, _val):
...
+    # type: (Callable) -> Callable
+    """
+    A decorator to strip the proxy from all incoming arguments
+    (including ``self`` if it's a method) and add one to the return
+    value.
+    """
+
+    @functools.wraps(f)
+    def proxy_wrapped(*args, **kwargs):
+        s_args = [stripProxy(a) for a in args]
```
rob-c: Is this recursive?

milliams: No, and it needn't be since only the outer-most object in a hierarchy is proxy-wrapped.

alexanderrichards: indeed, you beat me to it @milliams. This is the beauty (IMO) of the system that I am using. The proxy is just an outermost layer that object cross. e.g imagine a list of objects, as you fill that list the objects are unwrapped and stored in the list without proxies. Their proxies are applied again when you get them out.

rob-c: As long as we're now stripping the proxies everywhere then OK I'm convinced, I'm just weary that users are usually good at finding corner cases such as the proxies which don't use the automatic format.

rob-c:
Well if the tests are passing then go for me, would be nice to see an auto-proxy wrapper to add proxies to objects returned automatically.

python/Ganga/GPIDev/Base/Proxy.py
milliams: Indeed, things with non-automatic proxies (like the registries etc.) are a potential leak in the proxy layer but they should be fixed by making them conform to the `_impl`/`_proxyObject` interface so that `assert a is addProxy(stripProxy(a))` always works.

The registries aren't usually passed through the proxy layer though so they're unlikely to be a problem in the short-term.

milliams:
What do you mean by an "auto-proxy wrapper"? `proxy_wrap` already wraps the return value of functions.

rob-c:
In that case I must have been looking at the wrong piece of code, I'm still happy to merge

PR #394, opened 2016-04-21
Fixed support for ProdTrans HC jobs with different output filenames

drmarkwslater:
Minor fix to sort out HammerCloud jobs and the naming of output files for ProdTrans.


python/GangaPanda/Lib/ProdTrans/ProdTransPandaRTHandler.py
```diff
@@ -180,17 +180,23 @@ def prepare(self, app, appsubconfig, appmasterconfig, jobmasterconfig):
...
             randomized_lfns.append(randomized_lfn)
             ofspec.destinationDBlock = jspec.destinationDBlock
             ofspec.destinationSE = jspec.destinationSE
             ofspec.dataset = jspec.destinationDBlock
             ofspec.type = 'output'
             jspec.addFile(ofspec)
+
+            # remove the first section of the file name if it matches the file type
+            if len(randomized_lfn.split('.')) > 0 and randomized_lfn.split('.')[0].find( lfntype ) != -1:
+                randomized_lfn = '.'.join( randomized_lfn.split('.')[1:] )
```
milliams: Calling `split` on a string will always create a list of at least one element:
```python
>>> 'test'.split('.')
['test']
>>> ''.split('.')
['']
```
so `len(randomized_lfn.split('.')) > 0` will always pass so this could be simplified to:
```python
if lfntype in randomized_lfn.split('.')[0]:
```
However, if `randomized_lfn` has no `.` in it then the call to `randomized_lfn.split('.')[1:]` could fail.

drmarkwslater: You are totally correct - fixed :)

drmarkwslater:
retest this please

Issue #379, opened 2016-04-15
Lots of (un-needed warnings) for new users on first launch.


rob-c:
In summary there are a few things we can do to improve the first launch experience for new users:

 - [ ] There are problems reading files at launch which look concerning even for us. Can we fix this or improve the error/warning or demote it to debug if it's not important.

 - [x] There is an IPython problem which will probably concern users of Python 2.6 (on first load, and maybe after this which should be looked at, I suspect we're using an old deprecated API call in ```bootstrap.py```).

 - [ ] Can we replace the GLITE stuff with a 1 line error saying that GLITE functions are not enabled. Users running (on purpose without this) shouldn't be constantly be being told about functionality they don't need/want.

- [ ] If we are suggesting people run from cvmfs should we provide a Python 2.7 binary to them rather than requiring that they have root access in order to install Python 2.7 by hand?

I think these are not too unreasonable to be looked into prior to 6.1.19,

Regards,

Rob

Original email below:

```
Hi Rob,

I tried a naive install of ganga following the instructions in your talk and
https://github.com/ganga-devs/ganga/wiki/Installation-and-Basic-Usage

It spews a lot of warnings, which are probably not relevant - so
should they really be there ?

The other major complaint I had for the 'user based' install is that
when I look at

WARNING  Ganga will soon be depending on Python 2.7. You have Python
2.6 installed. See https://github.com/ganga-devs/ganga/wiki/Python-2.7

this implies that you need root access to the machine to install ganga.

The dirac UI et all which rely on python2.7 all seem to install their
own version with it (as the user), is ganga not going to do this ?

Cheers,

Daniela




lx03:ganga :~] python ganga-install --extern=GangaDirac 6.1.14
Ganga Installer, $Revision: 1.24 $
--------------- NEWS -----------------

Mon, 11 Jan 2016 14:43:54 +0000: public release 6.1.14 is available !

--------------- NEWS -----------------

Downloading release 6.1.14- into /home/hep/dbauer/Ganga
Resolving additional external dependencies for:  GangaDirac
resolving dependencies for Ganga
[download] Checking pycrypto                   : REQUIRED
[download] Checking stomputil                  : REQUIRED
[download] Checking httplib2                   : REQUIRED
[download] Checking python-gflags              : REQUIRED
[download] Checking ipython                    : REQUIRED
[download] Checking paramiko                   : REQUIRED
[download] Checking google-api-python-client   : REQUIRED
resolving dependencies for GangaDirac

[...]

PR #395, opened 2016-04-21
Dropping use of global_ns

rob-c:
Should do a minor part of #379 as it removes the only (bad) reference to the ```global_ns``` which was left in due to the transition from IPython 0.6 to 1.x


milliams:
:+1:

PR #397, opened 2016-04-21
Fixing eval lookup fails

rob-c:
This PR sets out to fix one thing.

When running ```ganga --debug``` (to look at something else) I noticed an alarming amount of times that Ganga was attempting to evaluate strings within the GPI and failing.
These are mostly from within the Config where Ganga doesn't know if it's been given a 'str' or a classname as a 'str' or something else. But in these cases Ganga is left with little option but to try and evaluate the object and see if it maps to something it knows about.

I've added some very basic caching around the eval which means that the potentially expensive 'eval' calls are minimised. I've added some very basic changes at a slightly higher level to avoid evaluating some objects we know never to be anything other than a string. And I've added some minor changes to reduce the number of calls to locked registry functions such as ```__len__``` and ```__getitem__``` from some basic functions to that at least the debug output is more readable.

There are some slight performance improvements associated with this but I'm motivating this entirely from the perspective of it's bad to allow for ```eval``` of objects too much at runtime anyway so we should reduce our dependence on this as much as possible.


python/Ganga/GPIDev/Adapters/IPrepareApp.py
```diff
@@ -28,7 +28,7 @@ class IPrepareApp(IApplication):
     Base class for all applications which can be placed into a prepared\
     state. 
     """
-    _schema = Schema(Version(0, 0), {'hash': SimpleItem(defvalue=None, typelist=['type(None)', 'str'], hidden=1)})
+    _schema = Schema(Version(0, 0), {'hash': SimpleItem(defvalue=None, typelist=['None', 'str'], hidden=1)})
```
milliams: Really, I think that the items in `typelist` should not be strings, but rather the types themselves so it would be `typelist=[None, str]` which would avoid having to `eval` these at all. I can;t think of any example where this would be a problem. The code supports this way of doing it already.

rob-c: In that case I'll happily make the jump in this PR, was wanting to avoid changing anything too much as I simply don't want to break things that are otherwise working. I'll give it a shot in another commit

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -50,22 +50,28 @@ def lazyLoadJobApplication(this_job):
     return lazyLoadJobObject(this_job, 'application')
 
 
-def lazyLoadJobObject(raw_job, this_attr):
+def lazyLoadJobObject(raw_job, this_attr, dont_eval=False):
```
milliams: `dont_eval=False` is a potentially confusing double negative. Perhaps `do_eval=True` is better?

rob-c:
Do we know why the Jenkins server is giving 503's?

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: True, it's a quick hack though I was very confused by the code otherwise trying to evaluate 'completed' and 'new' into objects quite a lot

rob-c:
Yep, I've no idea why the tests are all failing, I'm guessing it's to do with the machine being updated. The ones I've run work fine for me locally, I'll run the full set by hand in a bit

milliams:
They're failing on Travis as well so it's nothing to do with the update of ganga-ci.

rob-c:
Well thery're running fine on my machine so this is going to get annoying I fear, they're all failing and I know for certain I can at least submit and run jobs on my account

rob-c:
OK, tests are working again and I think this was an LHCb plugin quirk

milliams:
We must take care with what we are doing here:

`SimpleItem`s have a `typelist` attribute which is used to check the incoming value against to make sure it matches. `ComponentItem`s do not have a `typelist` attribute and instead use the `category` to make sure that the value being passed in implements the correct plugin.

Both schema item types have a `defvalue` but they works very differently in each case. With a `SimpleItem` it works as you would expect where when the `GangaObject` instance is created, that schema value is set to a copy of the thing in `defvalue`. With `ComponentItem`s the `defvalue` can be either `None` (the default) in which case that schema value will be a new instance of the `category`'s default plugin or it is set to a string (i.e. `'Dirac'`) in which case the object instance will have its schema value set to a new instance of that type (found by looking in the plugin system, in the correct `category` and then for a plugin which matches `Dirac`).

Using `typelist` on a `ComponentItem` should never make sense and is probably ignored. Likewise, using `defvalue` on a will usually not do what you want. All you can do with a `ComponentItem` is set its `category` (which is used as sort of type-checker) and set its `defvalue` (to choose the default type).

rob-c:
I've not functionally changed the code to be anything other than replacing the contents of the ```typelist``` with what it should explicitly evaluate to.
If some code was using it, it still is, if some code wasn't using it, it still isn't this PR doesn't/shouldn't change from the intended design only fix some of the annoyance with eval functions taking a long time or failing due to oddities.

This PR now conflicts due to not waiting for the changes so I'll rebase and retest

python/Ganga/GPIDev/Lib/File/GoogleFile.py
```diff
@@ -48,11 +56,11 @@ class GoogleFile(IGangaFile):
                       'localDir': SimpleItem(defvalue="", copyable=1,
                                              doc='local dir where the file is stored, used from get and put methods'),
                       'subfiles': ComponentItem(category='gangafiles', defvalue=[], hidden=1,
-                                                typelist=['Ganga.GPIDev.Lib.File.LCGSEFile'], sequence=1, copyable=0,
+                                                typelist=[GoogleFileStub], sequence=1, copyable=0,
```
milliams: I'm unclear what these `FooStub` classes that have been introduced are for. There's no point in doing a type check against a different type.

rob-c: Well you can't do a type check against the same type. I would argue an instance check is best here anyway and I thought the code should have been doing this?

Is the typelist needed in the case of ComponentItems?

rob-c: Rather can't construct this without an eval unless we're doing an instance check. Can't construct nicely the typelist here without either an eval or a line in the ```__init__``` which wouldn't have worked up until recently I suspect.

milliams: This is my point. On any `ComponentItem`s we don't need `typelist` as it shouldn't be used anywhere. That attribute can be completely removed from all of them. There aren't many and can be found with some grepping.

rob-c:
OK, it looks like there is no file type checking when adding subfiles to a file object. Or at least I can do the following without problems:
```  lf=LocalFile(); lf._impl.subfiles=[MassStorageFile()._impl]; str(lf) ```

I think we'll need to open an issue to check for the types of component items and/or the types of lists of objects...

File: python/Ganga/GPIDev/Lib/File/GoogleFile.py (hunk 3 above)
Comment (rob-c) [inline python/Ganga/GPIDev/Lib/File/GoogleFile.py]: For now is there agreement to remove these typelist from the IGanagFile.sufiles objects?

File: python/Ganga/Core/GangaRepository/SubJobXMLList.py [hunk 4]
```diff
@@ -345,6 +345,7 @@ def __call__(self, index):
 
     def __getitem__(self, index):
         """Return a subjob based upon index"""
+        logger.debug("SJ:__getitem__")
```
milliams: I guess this can be removed?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -499,3 +500,28 @@ def _setFlushed(self):
             self._cachedJobs[index]._setFlushed()
         super(SubJobXMLList, self)._setFlushed()
 
+    def _private_display(self, reg_slice, this_format, default_width, markup):
+        """ This is a private display method which makes use of the display slice as well as knowlede of the wanted format, default_width and markup to be used"""
```
milliams: What is the purpose of this function? What does it return?

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -804,7 +804,9 @@ def stop(self, fail_cb=None, max_retries=5):
          fail_cb : if not None, this callback is called if a retry attempt is needed
         """
 
-        if not self.alive:
+        if not self.alive and ThreadPool != []:
+            import traceback
+            traceback.print_stack()
```
milliams: I guess this can be removed?

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -33,20 +33,31 @@
 
 _stored_Interface = None
 
+_eval_cache = {}
+
 def setProxyInterface(my_interface):
+    ## Set the proxy interface, not strictly needed for Ganga.GPI but good practice as we move to 'ganga'
```
milliams: If these are supposed to be docstrings then they should be """triple double quoted""".

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -133,22 +147,36 @@ def runtimeEvalString(this_obj, attr_name, val):
 
 
 def raw_eval(val):
+    ## Attempts to evaluate the val object and return the object it evaluates to if it is a Python object
+    ## Makes use of basic caching as we don't expect that things at this level should change.
+    from copy import deepcopy
```
milliams: We already have `copy` imported in this module.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -133,22 +147,36 @@ def runtimeEvalString(this_obj, attr_name, val):
...
+    if val in _eval_cache:
+        return deepcopy(_eval_cache[val])
+
     try:
+        ## Catch common strings which shouldn't be evaluated anyway and return them asif an exception occured
+        if val in ['', '.', 'new']:
+            _eval_cache[val] = val
+            return deepcopy(val)
```
milliams: Where are these strange strings being evaluated? The only things that should be eval'd are simple Python types from the config file or `<value></value>` blocks in the XML.

python/Ganga/GPIDev/Lib/Dataset/GangaDataset.py
```diff
@@ -17,8 +18,7 @@ class GangaDataset(Dataset):
     '''
     schema = {}
     docstr = 'List of File objects'
-    schema['files'] = GangaFileItem(defvalue=[], typelist=[
-                                    'str', 'Ganga.GPIDev.Adapters.IGangaFile.IGangaFile'], sequence=1, doc="list of file objects that will be the inputdata for the job")
+    schema['files'] = GangaFileItem(defvalue=[], typelist=[str, IGangaFile], sequence=1, doc="list of file objects that will be the inputdata for the job")
```
milliams: This `typelist` can be removed.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -188,22 +196,22 @@ class Job(GangaObject):
     Datasets are highly application and virtual organisation specific.
     """
 
-    _schema = Schema(Version(1, 6), {'inputsandbox': FileItem(defvalue=[], typelist=['str', 'Ganga.GPIDev.Lib.File.File.File'], sequence=1, doc="list of File objects shipped to the worker node "),
-                                     'outputsandbox': SimpleItem(defvalue=[], typelist=['str'], sequence=1, copyable=_outputfieldCopyable(), doc="list of filenames or patterns shipped from the worker node"),
+    _schema = Schema(Version(1, 6), {'inputsandbox': FileItem(defvalue=[], typelist=[str, File], sequence=1, doc="list of File objects shipped to the worker node "),
```
milliams: This `typelist` and others in this file (with type `FileItem`, ``GangaFileItem` and `ComponentItem`).

python/Ganga/GPIDev/Lib/Tasks/ITransform.py
```diff
@@ -14,41 +14,42 @@
...
 class ITransform(GangaObject):
     _schema = Schema(Version(1, 0), {
-        'status': SimpleItem(defvalue='new', protected=1, copyable=0, doc='Status - running, pause or completed', typelist=["str"]),
-        'name': SimpleItem(defvalue='Simple Transform', doc='Name of the transform (cosmetic)', typelist=["str"]),
+        'status': SimpleItem(defvalue='new', protected=1, copyable=0, doc='Status - running, pause or completed', typelist=[str]),
+        'name': SimpleItem(defvalue='Simple Transform', doc='Name of the transform (cosmetic)', typelist=[str]),
         'application': ComponentItem('applications', defvalue=None, optional=1, load_default=False, doc='Application of the Transform.'),
-        'inputsandbox': FileItem(defvalue=[], typelist=['str', 'Ganga.GPIDev.Lib.File.File.File'], sequence=1, doc="list of File objects shipped to the worker node "),
-        'outputsandbox': SimpleItem(defvalue=[], typelist=['str'], sequence=1, doc="list of filenames or patterns shipped from the worker node"),
+        'inputsandbox': FileItem(defvalue=[], typelist=[str, File], sequence=1, doc="list of File objects shipped to the worker node "),
```
milliams: This `typelist` and some others in this file can be removed.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: 'new' I know was from evall-ing the status which has been fixed elsewhere and was a bad idea from the start there. As for '' not a clue and for '.' even less of one tbh. I can track them down but I'd rather make them go away.

milliams: I agree with making them go away but I'm not sure that special-casing them and hiding them like this is the best way. If they're being caused by a bug somewhere then we should try to fix _that_.

rob-c: Frankly the only bug is '.'. (I suspect this has gone due to other work here tbh)
 '' will happen whenever you've an empty string being eval-ed because it wasn't populated (yet?) and the 'new' is effectively a typo as it should also have gone by now.

milliams: `eval('')` is not evaling an empty string, it's evaling _nothing_. `eval("''")` would be evaling an empty string and that would be fine. In this case, this function is turning a 'nothing' into an empty string which is probably not what is wanted. I would expect an exception to be raised here

rob-c: If the eval is evaluating '' it throws an exception and that's taken to be a string which can't be eval-led such as 'hello' so the returned value is the string in this case from this function, even if it did as it was having to deal with the likes of 'hello' until now it's difficult to know to distinguish between this and the special case of '' without a check

milliams: So where are strings like `'hello'` coming from? We should be being very careful about what we pass into a function like `raw_eval` so I'd like to be sure about where its inputs are coming from.

rob-c: without launching ```gannga --debug``` I don't know I've just  been reducing the total number of calls to this ass I want to avoid ever using it if possible

rob-c: OK so here's a gotcha:

```File.name``` defaults to ```''```. This is saved in the Schema as a SimpleItem.

Ganga attempts to work out what the 'type' of ```''``` is but ```''``` could also be ```'myClass'```, due to the way that we allowed strings to be parsed (at least in the past).
As a result Ganga has to try and evaluate the object and work out if it is the latter or the former.
If the ```defvalue``` is the latter we do one thing and if the ```defvalue``` is ```''``` then we do another.

If we intend to keep supporting things the way they are atm we don't throw an exception. If we do then we break a lot of stuff, and frankly that was not the intention of this PR. I'm happy to remove the ```''``` special case but fixing every last bit of the schema here isn't what I wanted to do, nor what the PR set out to accomplish.

**Comment by rob-c** [comment]

File: python/Ganga/Core/GangaRepository/SubJobXMLList.py [hunk 13]
```diff
@@ -345,6 +345,7 @@ def __call__(self, index):
 
     def __getitem__(self, index):
         """Return a subjob based upon index"""
+        logger.debug("__getitem__")
```
milliams: I guess we want to remove this?

python/Ganga/GPIDev/Base/Proxy.py
milliams: In the case of `File.name`, that's a `SimpleItem` and so it should only ever be simple Python objects, not `GangaObject`s. This means that `''` should only ever be a string.

There's a bunch of things I'd, like to fix in the schema handling, for example getting rid of the "typelist inferred from type of defvalue" thing. As you say, this PR isn't the place to fix it.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: I'd really rather not, this makes it clear in the debug when the job was potentially loaded from disk. There are similar debug messaged in the Registry for the same reason that should any of these crop up a lot in the ```--debug``` usage then this should be addressed, the message should show only a few times in normal use or on certain calls which trigger this behaviour.

milliams: In that case a more explanatory message would be useful? "Subjob being loaded" for example?

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I think I agree. I'm not a fan of the typelist inference as it causes problems like this and I'm looking forward to a rethink in some of the Schema as this would be very nice to see.

I'd like to consider dropping support for direct string translating when we move to Ganga 6.2. It's not used in too many places anymore (thankfully) and those can be made into special cases. i.e. the datetime/ShareRef objects in the XML. I think it's time we drop support for ```j.backend = "Dirac"``` and equivalent it's too fragile to keep patching/fixing this when it breaks. (conversation around this can happen in another PR as it stands)

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Given it's a call to the ```__getitem__``` which comes from this class I would have thought there is enough information in the logging to be able to get back here quite quickly. I can drop it seeing how there is a debug at the top of ```_getItem``` anyway which is explicitly called.

python/Ganga/GPIDev/Base/Proxy.py
alexanderrichards: I agree that type inferring is sub optimal. In the schema that I've been writing the typelist is exactly that, a list of types. 
:smile:

alexanderrichards: > I think it's time we drop support for j.backend = "Dirac"

:+1: 
Agreed, I didn't even realise that this worked! how horrible

rob-c: example: ```./Ganga/test/GPI/JobKill.gpi```

rob-c:
retest this please

rob-c:
retest this please

rob-c:
@milliams Removing that debug statement seems to expose some threading lockup, any ideas how to go about debugging this if we can't reproduce it anywhere other than on jenkins?

milliams:
From lxplus you should be able to do:
```bash
scp root@ganga-ci:~jenkins/gangadir_testing/TestQueuedSubmit/thread_trace.html .
```
You can see it at [thread_trace.html.zip](https://github.com/ganga-devs/ganga/files/234341/thread_trace.html.zip). Looking though, there seems to be no explicit thread lockup.

The job has just finished and it shows that the error is:
```
Monitoring Error: global name 'logging' is not defined
```
which is caused by
```python
logging.debug("Attempting to load subjob: #%s from disk" % index)
```
I'm very surprised that this didn't show up in your local tests though.

rob-c:
maybe I ran my local tests on the wrong branch... I'll edit and resubmit. Why this doesn't show more clearly before the parallel tests is worrying.

PR #399, opened 2016-04-22
Bugfix/fix slow atlas startup

drmarkwslater:
It was reported a while ago that GangaAtlas was taking a very (infinitely?) long time to start up. I had thought this was due to one of the lock ups that we were seeing but that apparently wasn't the case. The problem was that when checking to fill the active backends list, an empty backend object is created every time. In the `Jedi` case, this can actually be a bit slow because of the call to get the default cloud. This PR addresses this by:

1. only creating a backend object when a job is found to be active
2. caching the call to find the default cloud

This certainly fixes the problem in my tests!


milliams:
:+1:

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -947,11 +947,11 @@ def __defaultActiveBackendsFunc(self):
                 j = stripProxy(self.registry_slice(i))
 
                 job_status = lazyLoadJobStatus(j)
-                backend_obj = lazyLoadJobBackend(j)
-                backend_name = getName(backend_obj)
 
                 if job_status in ['submitted', 'running'] or (j.master and (job_status in ['submitting'])):
                     if self.enabled is True and self.alive is True:
+                        backend_obj = lazyLoadJobBackend(j)
```
rob-c: If the problem is that creating this (unwanted?) object takes so long would it be better to avoid creating it here at all, we could quite easily add a ```lazyLoadJobBackendName``` method which bypasses the constructor here or is the overhead in this case 'acceptible'?

drmarkwslater: @rob-c It would be better to not create the object as we only need a name but I didn't want to interfere too much as I didn't want to risk breaking lazy loading. This could be done in another PR I guess.

rob-c: I guess it can wait until another PR to address this as it's a relatively minimal overhead, unless there are many master jobs in a running state, does hammercloud get into this state?

drmarkwslater: I haven't heard any complaints yet so I guess not 😄

rob-c: OK, good :) then I'm happy to leave the fix like this tbh until anyone complains :+1:

rob-c:
:+1:

drmarkwslater:
retest this please

PR #400, opened 2016-04-22
Reorder terms in conditional

milliams:
Put the simpler term first so that if `force_cache` is `True`, it doesn't need to check `fullyLoadedFromDisk()`.

This will hopefully help with some lockups we've been seeing on ganga-ci which are being caused by this section of code.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -361,7 +361,7 @@ class Node(object):
 
     def getNodeIndexCache(self, force_cache=False):
         if not force_cache and self.fullyLoadedFromDisk():
-            if self._getRegistry():
+            if self._getRegistry() is not None:
                 ## Fully loaded so lets regenerate this on the fly to avoid losing data
                 return self._getRegistry().getIndexCache(self)
             else:
@@ -372,7 +372,7 @@ class Node(object):
                                                                  
     def fullyLoadedFromDisk(self):
         """This returns a boolean. and it's related to if self has_loaded in the Registry of this object"""
-        if self._getRegistry():
+        if self._getRegistry() is not None:
             return self._getRegistry().has_loaded(self)
         return True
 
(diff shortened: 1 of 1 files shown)


drmarkwslater:
👍

rob-c:
Can the ```fullyLoadedFromDisk``` method be made safer, should this maybe be locked itself?

rob-c:
:+1:  on the PR by the way.

I'm just curious if it's suspected we know what in the ```fullyLoadedFromDisk``` method is causing problems? We can always reduce the complexity of this by introducing a Boolean to check is loaded from disk in the Node object and have this correctly set in the ```Registry::load``` method

milliams:
To be honest @drmarkwslater have been scratching our heads at the stack trace:

```python
...
  File "python/Ganga/GPIDev/Base/Objects.py", line 465, in __get__
    obj_index = obj.getNodeIndexCache(force_cache=True)
  File "python/Ganga/GPIDev/Base/Objects.py", line 363, in getNodeIndexCache
    if self.fullyLoadedFromDisk() and not force_cache:
  File "python/Ganga/GPIDev/Base/Objects.py", line 375, in fullyLoadedFromDisk
    if self._getRegistry():
  File "python/Ganga/Core/GangaRepository/Registry.py", line 167, in decorated
    with self._lock:
  File "/usr/lib64/python2.6/threading.py", line 123, in acquire
    rc = self.__block.acquire(blocking)
```

We can't figure out how a call to `if self._getRegistry()` is ever managing to get to "Registry.py", line 167" :confused: The fix in this PR hopefully circumvents this but in general I think there is larger complexity problem with the index cache code which need looking at in the future.

retest this please

rob-c:
erm you want to change that line to: ```if self._getRegistry() is not None:``` at least as this calls the ```__len__``` method (and possibly others I'm not tracking) as well as just checking if it evaluates to True. 
I'm changing this in #397 as I was seeing a huge number of calls to ```__len__``` behind the scenes when running with ```ganga --debug```.

Whilst this doesn't 'fix' the problem it may at least make it a little happier.

rob-c:
rather the call of: ```if self._getRegistry()``` calls the ```__len__``` method which I'm assuming is in some complex way of working out if a class object should evaluate to True/False somewhere.

milliams:
Ahh, thanks @rob-c. You are correct, `if self._getRegistry()` is asking if the *registry* is true or false but when I was reading the code I was reading it as `if self._hasRegistry()`. I will update this PR with the `is not None`.

PR #407, opened 2016-04-25
Feature/dynamic init

rob-c:
This is an attempt to build back in a feature we effectively disabled until recently but kind-of/sort-of left in place.

When a ganga object is initialized in develop then all of the attributes are initialized at the same time (even if they're never used).

There is some effort in the ```__get__``` method which allows for the default schema value to be returned when the object hasn't been initialized but this should have never been called due to the code in ```GangaObject::__init__```

This PR attempts to remove the setting of the attributes in ```GangaObject::__init__```. This was breaking a lot of things before we moved to having everything thread-safe but now hopefully this will 'just-work'.

I don't want to propose when this will go in, it's a PR I've had sitting in my workspace for a while and doesn't show signs of having problems. It should reduce the memory usage of Ganga and potentially in some corner cases performance.

If this goes in for 6.1.20 then great but I suspect this will cause problems in testing as I've only run about 1/2 of the tests locally and it seems surprisingly OK.


rob-c:
This breaks a lot of things in a lot of strange ways. I don't see this making it in before 6.1.20 but the general idea I still like.

PR #409, opened 2016-04-26
Use _impl for setting is_prepared in GangaRobot

drmarkwslater:
`is_prepared` is protected in LHCb Apps so we need to use `_impl` to get around this. I'm sure this can be done better but I want a quick fix to get LHCb HammerCloud working.


alexanderrichards:
@milliams do I recall that we are going to get rid of these @codecov-io comments as they are very spammy and happen on every commit? We can always consult the tests if we need this information.

milliams:
@alexanderrichards codecov are having some teething problems with their new release at the moment. Once that is settled, the comments from them should cease. In the meantime I will just delete the comments.

alexanderrichards:
right o, thanks. Please delete these comments as well as they don't relate to Marks pull

rob-c:
Think I prefer ```stripProxy``` to ```._impl``` but if it's a quick fix to get something working then go from me :)

drmarkwslater:
Fair point. Change :)

PR #411, opened 2016-04-26
Bugfix/stop lcg annoyances

### LINKED ISSUE #410: Need a grid proxy just to start ganga
This keeps coming up time and time again and is causing real problems for me now in debugging is it known what causes the grid proxy to be initialized right before the bootstrap starts to run?

This should __not__ be happening I would like to see this behaviour gone asap.

It's running with the voms tools rather than with the dirac tools which I want it to run with and for some reason it's not finding a valid proxy in my environment and insists on constructing it's own.

From what I can understand this could be understood as a limitation in the Schema as it's not overloading the default proxy handling methods in the first instance that they're called but does so later on giving a false sense that the proxy is being correctly handled imo.

I've a PR incoming (eta 30min or so) which makes the LCG tools stop requesting a proxy on import, it's painful, it's bad and it causes no end of headaches for me on 2 different setups.

I've had to write this in order to submit a job to the correct VO with the GridPP Dirac.

I'd like to get this into 6.1.19 if it isn't too controversial but if there's discussion I'd _really_ like to see this make it into 6.1.20.

Thanks,

Rob

rob-c:
This PR stops the LCG from requesting a proxy to be constructed on startup.

It constructs them using the default ```voms-proxy``` tools which cause problems when a user is a member of multiple VOs and needs to override the functions with something a bit more correct.

This delays the creation of the proxy right until the last step. However it does come at the slight cost that the GridProxy tools need to be given full absolute paths. This isn't too much of a problem imo although I've not traced if it's possible to allow for the GridProxy tools to correctly make use of a given environment.
(I somehow feel this is a separate PR/credential problem)

This also replaces the LCG.Grid.active boolean with a property which should allow for a call to construct a proxy to be made right before some code which requires it.

As far as I can tell this shouldn't break too much but I'm not an LCG user and don't know anything about this codebase.
If this is going to break everything because LCG is coded up to expect a Grid instance to be constructed with a valid proxy there then I don't want to holdup the 6.1.19 release while this gets sorted out, if it doesn't I'm keen to get this in if possible else I'll have to keep working with this PR (or some equivalent) privately throughout 6.1.20.

Fixes #410 .
(OK it doesn't fix it I don't think but it addresses it)


python/Ganga/Lib/LCG/Grid.py
```diff
@@ -55,8 +53,15 @@ def __init__(self, middleware='EDG'):
...
-        self.active = self.check_proxy()
+
+    def get_active(self, create=True):
+        """ This checks if the proxy is valid and returns a True/False
+        If the create=True (default) it will attempt to create a new proxy.
+        If it is ceate=False then no calls which generate a new proxy on disk will be made."""
+        isValid = self.check_proxy(True)
+        return isValid
+    # Replace self.active with call to get_active with default args
+    active = property(get_active)
```
milliams: This should be
```python
@property
def active(self, create=True):
    ...
```

milliams: Also, is `get_active` ever called with `create=False`?

rob-c: I wanted to do this in https://github.com/ganga-devs/ganga/pull/411/files#diff-dcae5641b9610eca0e80152f406817edR2353 but I think it makes more sense to allow to possibly call get_active and check rather than forcibly recreate every single time which this PR likely causes to happen as there is no caching.

drmarkwslater:
So I've got a few questions about this:

1. Why is using the `voms-proxy` tools a problem for people in multiple VOs? At present, Ganga can't handle having multiple grid credentials anyway but I don't see why the GLITE tools are a particular cause of this.

2. If you don't want Ganga to request credentials at startup, can't you just disable GLITE in the config until you need it? It would certainly be nice for the system to only ask when it needs a proxy (though there may be issues about that with respect to submitting jobs in background threads) but I think the LCG/credential stuff as it stands **may** not be easily fixable to do this (though I would need to check) as I wouldn't be surprised that there aren't appropriate calls to check the LCG system is initialised every time it's needed. It would need @milliams credential branch to work properly I think.

3. It is a very big problem that you need to specify the full path to the GLITE tools as all `.gangarc`'s at the moment won't have the full path (and in fact, this could be in any number of places) and so will break a lot of things

rob-c:
I'm a member of ```gridpp_user``` and ```lsst_user``` when I submit a job using the proxy from the latter all of my jobs go to ```gridpp_user``` which is not what I want.
I changed my config to add the ```dirac-proxy-init```, destroyed my proxy and resent the job and again it want to ```gridpp_user```. I tried modifying the JDL by hand and discovered I couldn't _force_ the job into the ```lsst_user``` group.
I ended up running ganga with ```--debug``` to realise that my proxy commands were being ignored in favor of the default ```voms-proxy-init``` commands which generated a proxy which wasn't allowing me to do what I want.

I get that I can't change VO without restarting Ganga and doing some by-hand proxy handling but the ```voms-proxy``` tools are being used in such a way as to overwrite any proxy set externally to Ganga which caused me to be fooled into trusting it to do something sane.

All in all this was a big effort just to submit a hello world to the correct place which could be made simpler if the GLITE tool just waited until after the bootstrap to request it's proxy. There is a genuine bug there.

rob-c:
I'm happy to leave it enabled if there are user groups relying on this, but surely they don't need a proxy for every single time they startup ganga? Getting it right at the start 'just in-case' seems like a bit of a bad way of handling this.

milliams:
It is indeed a bad way of handling it, I doubt anyone here is disagreeing with you. This is why we have a new credentials system on the way.

rob-c:
Which brings me back to my strong feelings of can we at least get a working ```Dirac.ini``` that doesn't abuse GLITE in before 6.1.19 I'd be happier to move to that even if the LCG stuff causes errors on my personal system.

drmarkwslater:
@rob-c I can't seem to reproduce this bug. When I start Ganga using:

/cvmfs/ganga.cern.ch/runGanga-dirac.sh
(note this won't work out of the box right now as I needed to reenable the reexec because otherwise it couldn't find IPython)

It asks for a proxy (agreed - this isn't ideal) but using --debug it is using the correct `dirac-*` tools and the only time the `voms-*` tools are mentioned is when it creates the defaults for `VomsCommand`. If I then start again it doesn't ask for a proxy because it finds the valid one. 

I've also just tried submitting requesting an na62 proxy instead of a gridpp one and that worked fine as well (i.e. in the Dirac portal, the JDL only has references to na62.blah-user). 

Have you tried just using the above script and setting your .gangarc appropriately for the VO? That's all I've just done and it seemed to work...

drmarkwslater:
As I mentioned in another PR, **don't** set `voms` as that will trigger the generation of a VomsProxy as opposed to a Grid Proxy!

rob-c:
I'd like to revisit this at some point and fix the ```get_active``` function and add some basic caching of a few hours with a single valid proxy (or better integrate with the old credentials code just to check validity)

The fact we __still__ depend on a _blindly_ constructing proxy when the proxy tools are available by default is ugly imo. It confuses things and gets annoying when doing things such as work with ```ganga``` which doesn't fully load all of the config yet.

PR #413, opened 2016-04-26
Bugfix/fix default Dirac.ini

rob-c:
This at some level supersedes the need for #411 but the discussion about whether that is correct for what it tries to do is best left there.

This adds a default cvmfs env which works by with the modified Dirac.ini to support gridpp_user voms members on the GridPP Dirac instance.

This also turns off GLITE support by default as this _does_ ask for a proxy on every boot and is not the proxy which is either wanted or needed.

This works for multi-VO users now as it correctly waits to create a proxy until the Schema attributes have been overloaded with the correct functions to call.


python/Ganga/__init__.py
```diff
@@ -309,7 +309,7 @@ def getLCGRootPath():
                  filter=Ganga.Utility.Config.expandvars)
 
 lcg_config.addOption(
-    'GLITE_ENABLE', True, 'Enables/disables the support of the GLITE middleware')
+    'GLITE_ENABLE', False, 'Enables/disables the support of the GLITE middleware')
```
drmarkwslater: This is quite a significant change that will probably annoy a lot of people. Can't it be put in a the appropriate .ini file instead?

python/GangaDirac/Dirac.ini
```diff
@@ -4,6 +4,12 @@ RUNTIME_PATH = GangaDirac
 
 ReleaseNotes = False
 
+[defaults_GridCommand]
+info = /cvmfs/ganga.cern.ch/dirac_ui/scripts/dirac-proxy-info
+init = /cvmfs/ganga.cern.ch/dirac_ui/scripts/dirac-proxy-init -g gridpp_user -M
+[DIRAC]
+DiracEnvFile = GangaDirac/cvmfs_DIRAC_env
```
drmarkwslater: This should point at the matching Env file on CVMFS as that will be kept up-to-date with the Dirac client there

drmarkwslater:
I don't think we need to save the env file in the repo as it's very much linked to the dirac client on CVMFS and the EnvFile is there as well.

drmarkwslater:
I'm confused as to why you say that GLITE provides a proxy that's neither wanted or needed - if you've set the GLITE commands properly to point to `dirac-proxy-init` etc. it will just use those and create the Dirac proxy correctly (or at least it does for me). Where are you seeing a problem?

egede:
> I'm confused as to why you say that GLITE provides a proxy that's neither wanted or needed - if you've set the GLITE commands properly to point to dirac-proxy-init etc. it will just use those and create the Dirac proxy correctly (or at least it does for me). Where are you seeing a problem?

@drmarkwslater One of the nice features of using the Dirac backend is that it only asks for a proxy if it is actually required. Enabling GLITE at the moment blocks that feature.

drmarkwslater:
@egede In which case, this should be setup however LHCb do it I guess. I don't have a problem with that but when I did the settings I put in to the CVMFS Dirac.ini file (and recommend on the tutorial) I assumed that however the proxy was handled for LHCb was LHCb-specific so I just used the only other way I knew about to check for a proxy which was the GLITE method. If there's a better way of  doing this then great :smile:

python/Ganga/__init__.py
rob-c: Out of curiosity which groups are effected by this? LHCb and GridPP don't __need__ this feature as far as I'm aware. They can both be configured to use it but they can both be configured to run without it.
A nice message in the release notes should catch anyone else who uses this but doesn't explicitly request it.
As for ATLAS I'm assuming they're correctly sourcing the ```ATLAS.ini``` which explicitly has this enabled when I look at it.

python/GangaDirac/Dirac.ini
rob-c: I'm happy to do so, so where should we put it?
I don't know if the current envfile was enough I amalgamated the 2 files setting up env vars into 1 file just in-case.
If you give me a location I'll happily change it but we should probably have it by default here :)

python/Ganga/__init__.py
drmarkwslater: At the very least the Durham IPPP group as they submit to ARC. Beyond that I believe SNO+ are also still submitting to LCG/CREAM/ARC though I may be wrong in that.

python/GangaDirac/Dirac.ini
drmarkwslater: I'm confused - is the envfile in this PR different from:

/cvmfs/ganga.cern.ch/dirac_ui/envfile

? If so, I'll just update the CVMFS one and DiracEnvFile should point to that.

drmarkwslater:
So I've just tried running GangaDirac without having GLITE enabled and the Dirac backend doesn't ask for a proxy to be generated so there must need to be more stuff copied over from LHCB to get this to work as needed because as it stands (as far as I can tell), you need to use GLITE_ENABLE to get Ganga to ask for a proxy.

drmarkwslater:
Actually, ignore that previous comment. I tried again in a more clean way and it did seem to be asking but because the GLITE system wasn't setup, it couldn't find the command. There may be a nice config-y way around this which I'll have a look at...

drmarkwslater:
So as regards Dirac stuff, setting GLITE_ENABLE to False does stop it asking on startup and if you submit a job it's asks then. However, there are two drawbacks:

1. You need to either supply the full path to the `dirac-proxy-*` commands **or** source the bashrc script in the command beforehand (probably the latter). This isn't a big deal and is just a change in config for Dirac settings.

2. If you have active Dirac jobs and you start without a valid proxy, it **doesn't** ask, throws an error and then seems to leave the monitoring loop in an odd state.

The second of these is obviously more of a problem :smile:

rob-c:
1. configuring the cvmfs isn't too bad. If we're expecting a user not using cvmfs to use DIRAC or some form of grid this is off the recommended route so they know what they're doing.

 2. The Monitoring loop will and does resume after bailing out on the first error after gridProxy.renew() This is the way it has been for LHCb users for a long time now and the monitoring is robust against this.

 3. As for whether LCG is robust against this I've no idea how it would react to a proxy not being available on submission/monitoring. I can't comment as I've never used it.

Given the level of discussion here I'm leaning to get 6.1.19 out and deal with all of this during 6.1.20 unless people are optimistic this will be resolved within the next few hours. I don't think there are too many people belonging to multiple VOs on the GridPP DIRAC so this shouldn't effect too many people.

rob-c:
This caused far too much discussion and I think is better solved by #411 anyway as that removes my concerns here.

PR #414, opened 2016-04-27
Remove unused or unneeded code

milliams:
This is a simple cleanup PR to remove some small bits of old dead code which build up over time. Most of it is commented-out code but some is simply re-assigned variables.


milliams:
retest this please

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -126,10 +127,7 @@ def reload(self):
         self.registry._lock.acquire()
         try:
 
-            if self.registry.checkShouldFlush():
```
drmarkwslater: Just to check - I'm guessing that as it is an `IncompleteObject` this will always return `False` and that's why it is being removed?

milliams: It's being removed since the function `checkShouldFlush()` no longer exists. It was used previously to keep track of the dirtiness counter but now that it is time-based it is not used.

drmarkwslater: Ah - that's fair enough then 😄

milliams:
I'm going to go ahead and merge this PR, #415, #454 and #455 ASAP as there are no more comments.

rob-c:
I'm in favour of all bar some minor changes in #415 which will cause a notable change in Ganga's behaviour under some situations.

PR #415, opened 2016-04-27
Remove unused code from the Registry system

milliams:
In a similar vein to #414, this removes unused code in the registry system. This is largely due to the changes previously made in #209 and #267.


python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py
```diff
@@ -14,14 +14,9 @@
 
 class PrepRegistry(Registry):
 
-    def __init__(self, name, doc, dirty_flush_counter=10, update_index_time=30, dirty_max_timeout=60, dirty_min_timeout=30):
-
-        super(PrepRegistry, self).__init__(name, doc, dirty_flush_counter, update_index_time, dirty_max_timeout, dirty_min_timeout)
-
-        self.releaseThread = threading.Thread(target=self.trackandRelease, args=())
-        self.releaseThread.daemon = True
-        self.releaseThread.start()
```
rob-c: This causes us to end up back with problems with multiple sessions trying to submit jobs.
I'd rather keep the track and release code as there will be complaints from some users if this goes

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -316,92 +304,6 @@ def updateLocksNow(self):
...
-    def isAutoFlushEnabled(self):
-        return self._autoFlush
-
-    def checkDirtyFlushtimes(self, timeNow):
-        self._dirtyModTime = timeNow
-        if self._flushLastTime is None:
-            self._flushLastTime = timeNow
-
-    @synchronised
-    def checkShouldFlush(self):
```
rob-c: This can and should go :)

python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py
milliams: As far as I can tell, this code hasn't actually been run in several releases now due to `_dirtyModTime` and `_flushLastTime` not being set to anything any more. I don't think we've had any complaints about it have we?

rob-c: OK, if people have stopped complaining about this then I'm happy to drop the code, We still don't release the prep metadata lock during running which I regard as a bug since this stopped working but if users aren't needing to launch multiple sessions it probably says something for how things have improved.

PR #418, opened 2016-04-27
Fixing some bugs in LocalhostExec

rob-c:
Minor cherry-picking.

For some reason my Local jobs fail due to not picking up the script files I send along as an Executable.exe

Adding this line gives some clearer debugging and adding the ```abspath``` seems to alleviate my problems.

I don't know if this will cause issues and if there should be a check here, if someone knows a better fix ler me know else I'll push this in 6.1.20


rob-c:
Great so, this interferes with the standard 'echo' command I think. Maybe a test to see if the file exists in $PWD then to add the $PWD to the command before it executes should fix this

egede:
@rob-c That is exactly the issue I was referring to earlier today in that the `File` object try in some clever way to decide if you refer to a system command or if you refer to an executable. It will take some further discussion to clarify this.

rob-c:
@egede Yes I think you're correct. This doesn't turn up in testing because we don't submit a script to the local backend as part of our tests.

I'll try and write a short test to get in for 6.1.19 if I get the time but I'm just trying to some jobs to run atm so this may fall back to 6.1.20

rob-c:
@milliams some of the tests didn't run in the above testsuite have these been lost in moving to 6.1.19 release? Can we fix this in develop if the code is available in the release branch.

drmarkwslater:
This seems fine but I'm surprised it fixes anything as I thought (probably through the `File` mechanism as @egede mentioned) that the path was set correctly before being sent as a job. Could you post an example of how this was failing for you?

rob-c:
@drmarkwslater 
The following runs a .sh file which runs a ,py file which generates a selection of files containing randomised data with different extensions.
```
j=Job(application=Executable(exe='generateTest.sh'), inputfiles=[LocalFile('~/cmtuser/GANGA/GANGA_LSST/install/ganga/generateTest.py'), LocalFile('/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/generateTest.sh')]); j.outputfiles=[LocalFile('*.txt'), LocalFile('*.zip')]; j.submit()
```
This fails for me on develop but runs on my branches.

I don't think we test sending executable scripts as payload in an integration test atm but I could be wrong

rob-c:
Error from failing: ```PROBLEM STARTING THE APPLICATION SCRIPT: ['generateTest.sh', 'Hello World'] [Errno 2] No such file or directory```
which I don't get if I just call the executable with an absolute path

milliams:
The way it is supposed to work at the moment (I think) is that in `Executable.prepare`, the `exe` is copied into the input dir which is then bundled up and sent to the job. So the `exe` script should be present without doing the `inputfile` bit. Is the `PATH` not being set correctly?

rob-c:
It's possible now that I look at it that including it in the inputfiles it somehow loses an executable flag but that shouldn't happen. I'll retry

rob-c:
Nope this doesn't work when I remove the script from the inputfiles. Same error, and I can confirm that the file does exist on the WN and i don't think we test this properly because without the isfile test this causes errors.
If someone wants to look into the Executable feel free I'm just fixing the backend and moving on

drmarkwslater:
So the (admittedly, probably quite confusing) way the `exe` option works is that if it's a string, this is just passed and called as-is. However, if it's a `File`, then this file is sent with the job and made available on the worker which is then run directly. Your example above would work with:

```j=Job(application=Executable(exe=File('/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/generateTest.sh'))
j.inputfiles=[LocalFile('~/cmtuser/GANGA/GANGA_LSST/install/ganga/generateTest.py'
j.outputfiles=[LocalFile('*.txt'), LocalFile('*.zip')]
j.submit()```

The tutorial does have this actually (http://ganga.readthedocs.io/en/latest/UserGuide/RunningExecutables.html) and the code quoted there is tested.

Now, whether we can improve this to make it more obvious is up for debate :smile:

rob-c:
@drmarkwslater My point is that in the example above it does not work when the ```exe``` is a string and the executable has been added to the job via the ```inputfiles``` unless the path of the executable is corrected by the running code on the WN.
I have verified that the correct executable is sent when using the ```inputfiles``` but I'm keen to avoid using ```File``` wherever I can as I want to regard this as bad practice as of 6.1, and I just want it to work, and with 2 lines here the confusion goes away.

Alternatively I could add the WN path to ```PATH``` which is maybe slightly more correct here as it's not being added from what I can tell looking at the code.

egede:
@rob-c At the moment, there is nothing wrong in using `File` objects. We agree that we can in due course migrate away from them (and get some more flexibility), but at the moment they are fine. I feel that it would be a bad idea to make minor changes to the interface here now, and then in not too long time make another change. So I suggest to abandon this PR.

rob-c:
I'm sure this won't be the last time I'm getting mixed up between the use of ```File``` and ```LocalFile``` and wanting to avoid the mess but OK.
While I'm on the matter I'll amend the Dirac exe handler in this case...

PR #420, opened 2016-04-28
Remove Reexec machinery

### LINKED ISSUE #404: Stop GangaAtlas requiring Re-exec machinery
At present, if you turn off Re-Exec with GANGA_NEVER_REEXEC=1 GangAtlas fails to load because the python path isn't set correctly to pick up the externals. I'm not sure if this is something wrong with the bootstrap or Atlas but either way, this needs to be fixed so we can finally get rid of the Re-Exec stuff.

drmarkwslater:
This removes the re-exec machinery completely and makes Atlas work sensibly without it. Other than just the stopping of re-exec, this also checks for a `standardSetup` function in the `__init__.py` of any runtime packages and calls this during `initEnvironment`. This is essentially a name swap from `getEnvironment` as this is often used as a more general hook anyway. 

We could possibly split the runtime setup calls out completely from `initEnvionment` to make it obvious what's happening but I'll leave that for someone else to decide 😄 

Fixes #404, #52

### DIFF
--- a/python/Ganga/Utility/Setup.py
+++ b/python/Ganga/Utility/Setup.py
@@ -99,6 +99,7 @@ class PackageSetup(object):
             if not paths:
                 return True
 
+            # loop over the sys paths to be added
             for path in reversed(paths.split(':')):
                 if path not in sys.path:
 
@@ -107,7 +108,13 @@ class PackageSetup(object):
 
                 # add to PYTHONPATH?
                 if set_python_path:
-                    pp_toks = os.environ['PYTHONPATH'].split(':')
+                    if 'PYTHONPATH' in os.environ:
+                        pp_toks = os.environ['PYTHONPATH'].split(':')
+                    else:
+                        # PYTHONPATH is empty so add Ganga in first
+                        pp_toks = [sys.path[0]]
+
+                    # create a new PYTHONPATH
                     if path not in pp_toks:
                         pp_toks.insert(1, path)
 
--- a/python/Ganga/Utility/Setup.py
+++ b/python/Ganga/Utility/Setup.py
@@ -84,17 +84,34 @@ class PackageSetup(object):
                     pp = ''
                 os.environ[var] = path + pp
 
-    def setSysPath(self, name):
+    def setSysPath(self, name, set_python_path=True):
         """ Update the sys.path variable for a given package by adding a given package to the PYTHONPATH.
         It assumes that the first item in PYTHONPATH is Ganga, so the package path is added as a second item.
+        If set_python_path is True, it will update the PYTHONPATH as well
         """
         import sys
-        import os.path
+        import os
 
         if getExternalHome():
-            path = self.getPackagePath2(name, 'syspath')
-            if path:
-                sys.path.insert(1, path)
+
+            # get the path(s)
+            paths = self.getPackagePath2(name, 'syspath')
+            if not paths:
+                return True
+
+            for path in paths.split(':'):
+                if path not in sys.path:
+
+                    # add to sys path
+                    sys.path.insert(1, path)
+
+                # add to PYTHONPATH?
+                if set_python_path:
+                    pp_toks = os.environ['PYTHONPATH'].split(':')
+                    if path not in pp_toks:
+                        pp_toks.insert(1, path)
+
+                    os.environ['PYTHONPATH'] = ':'.join(pp_toks)
 
         return True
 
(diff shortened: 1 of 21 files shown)


milliams:
`GangaUnitTest.py` on line 41 and 80 needs updating.

milliams:
`GangaUnitTest.py` on line 41 needs updating. `'--no-rexec', # Don't re-exec Ganga when running tests` should be removed.

milliams:
I've rebased this and will merge it once the tests have completed.

python/Ganga/Utility/Setup.py
```diff
@@ -84,17 +84,41 @@ def prependPath(self, name, var):
...
+            paths = self.getPackagePath2(name, 'syspath')
+            if not paths:
+                return True
+
+            # loop over the sys paths to be added
+            for path in reversed(paths.split(':')):
+                if path not in sys.path:
+
+                    # add to sys path
+                    sys.path.insert(1, path)
```
rob-c: Why not insert this as 0? Is it because we expect to overload system defaults with Ganga functions?

milliams: I'm not sure. I expect because it was like that before. It may need looking at in the future but for now it will avoid breaking anything in a surprising way.

drmarkwslater: Basically, because it assumes that Ganga is the first thing in the the system path (see the comment). Whether or not that's a safe assumption and where it's set is up for debate but it seems correct.

milliams:
retest this please

python/Ganga/Utility/Setup.py
rob-c: OK, I was just curious if this was an implicit assumption or not :) thanks for confirming

drmarkwslater:
@milliams I've haven't removed the other mention of reexec yet from `GangaUnitTest.py`. I'll try to get that done (trying to finish off the atexit stuff at present!) and then get this merged if everyone's happy 😄

drmarkwslater:
@milliams Awesome - good service here 😄

milliams:
retest this please

milliams:
We're getting lockups on shutdown during the tasks tests here:

1. Main thread is shutting down and so has called flush_all. It has the reg lock and is waiting on an object lock (this is fine)
2. The auto flusher is waiting on the reg lock (this is also fine)
3. `GANGA_Update_Thread_GangaTasks` is calling `addUnitToTRF` which gets an object lock which prompts a call to `_getWriteAccess` which grabs the registry lock (this is bad)

Once again, a call to `_getWriteAccess` is causing lockups.

Strangely I've only seen it on this PR, and only after having updated it. However, this PR does seem to have run its tests on the same VM each time so there's a potential 'cause' there.

rob-c:
Why is the Tasks system still going stuff at this point. It should have been stopped already during the shutdown if we're now flushing the repos before shutting them down shouldn't we?
Doesn't stopTasks stop the tasks system from doing stuff? If not what do we need to add to it

rob-c:
Ah, we're flushing before stopping the tasks system. This could potentially cause us to try and double lock an object I suppose.
Should we move the call to ```flush_all``` in the exitfuncs to after ```stopTasks```.

milliams:
@rob-c (GitHub is being slow so I can't see your messages here. I'm getting the emails though).

Indeed, I just saw the same ordering problem. I think that indeed `Repository_runtime.flush_all()` should be after stopping the monitoring, tasks system and queues system. Maybe even moved into `Repository_runtime.shutdown()`?

rob-c:
I was wondering if there is a reason not to. I think this was originally added as high as possible out of fear that we could lose data when the shutdown has problems. I see no reason not to push this back to the repository shutdown now as we think we've understood most of what is going on.
I'm experiencing strange github issues as well, comments appear on 1 machine and not on the other

milliams:
If in doubt, check [status.github.com](https://status.github.com/).

milliams:
I'll update the PR with the reordering and see if it helps things.

@drmarkwslater I guess you should bear this in mind for your `cleanup-atexit-handling` branch too.

milliams:
Ok, the tests are looking good again. Are there any more comment on this?

drmarkwslater:
Will do - that PR is almost ready to go in actually. Good news: Coverage works properly again (56%!). Bad news: Takes sodding ages 😭

milliams:
@drmarkwslater Check the info in the [FAQ](http://coverage.readthedocs.io/en/latest/faq.html) about the C implementation. It might help.

drmarkwslater:
You're correct - though the version on my local desktop is using CTracer, the one on the testing machines are PyTracer. I'll get this fixed and put the PR in.

PR #424, opened 2016-04-28
Export the GPI into the global namespace of Ganga

milliams:
When reviewing #395 I didn't realise that the error message that spawned that PR was:
```
DeprecationWarning:  global_ns is deprecated, use module instead.
```
As such, simply removing `global_ns` has broken running scripts with `execfile` inside Ganga. This PR correctly *replaces* `global_ns` with `module` rather than removing it entirely.

What we're exporting to what probably needs a little looking at (i.e. not exporting the namespace of bootstrap.py into `locals()`) but this for now mostly maps to the old behaviour.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -1279,7 +1279,8 @@ under certain conditions; type license() for details.
         ipshell.confirm_exit = config['confirm_exit']
 
         # Launch embedded shell
-        ipshell(local_ns=local_ns)
+        from Ganga import GPI
+        ipshell(local_ns=local_ns, module=Ganga.GPI)
 
     @staticmethod
     def ganga_prompt(dummy=None):


rob-c:
That PR name gave me cause for concern but the actual patch looks good to me. :+1:

python/Ganga/Runtime/bootstrap.py
```diff
@@ -1279,7 +1279,8 @@ def launch_IPython(local_ns, args, error_handler, ganga_prompt):
         ipshell.confirm_exit = config['confirm_exit']
 
         # Launch embedded shell
-        ipshell(local_ns=local_ns)
+        from Ganga import GPI
+        ipshell(local_ns=local_ns, module=Ganga.GPI)
```
rob-c: the local_ns here is the GPI btw, or should be

milliams: It more-or-less is but not quite. It's seeded from bootstap.py's `__main__.__dict__` which, due to an import in `bin/ganga`, contains the GPI but also includes the local namespace of bootstrap.py including things like `new_version_format_to_old`, `GangaProgram` and `manualExportToGPI`. This is also the reason that doing `from __future__ import print_function` in bootstrap.py was affecting the running IPython session :confused:.

However, indeed `local_ns` *does* contain the GPI (even if there's some other things thrown in there as well). The problem we were seeing is `execfile`ing a script like:
```python
def test():
    j = Job()
test()
```
was giving `global name 'Job' is not defined`. This is because the GPI was being put into the *local* namespace of the `execfile` which is not then available inside the function `test`. This PR puts the GPI inside the *global* namespace as well so that the script can run.

Exactly whether the GPI should be in globals or locals, I'm not sure but historically we had it in both I believe.

drmarkwslater: Certainly originally we had GPI in both locals and globals and I personally think it makes sense to have it in globals as that's what I would naively expect as a user (e.g. so you can create jobs in functions). Now what in particular should be in the local_ns and (further up this function), what's passed to running the `.ganga.py`, etc. and returned is very much up for debate. From a quick look at the code it wasn't clear what the user would expect so we thought a quick fix for now followed by ore discussion later :smile:

rob-c: Shouldn't the ```local_ns``` then also be ```Ganga.GPI``` if we're worrying about polluting the interactive command line?
Historically the ```local_ns``` as it is now was mapped to both the ```local``` and ```global``` yes, I was never sure about this but copied what was being done with a much older version of IPython.

milliams: I think that probably `local_ns` should be `Ganga.GPI`+`~/.ganga.py`+`config['StartupGPI']` but this needs some thought.

rob-c: Looking like another part of the config which could do with cleaning up I suspect

PR #427, opened 2016-04-29
Search registry by identity in find()

### LINKED ISSUE #426: Why is Registry find again comparing large Ganga objects behind the scenes?
https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/Core/GangaRepository/Registry.py#L489

I'm slightly concerned that I'm seeing this line in testing. There have been bugs in the ```__eq__``` methods in the past and this is extremely expensive which is why the find operation was at some point rewritten to just compare objects based upon ```_registry_id``` which is assigned at runtime.

This is partly to blame for: http://ganga.web.cern.ch/ganga/release/6.1.19-pre/reports/latest/html/summary_Ganga.old_test.Schema.html

Can we remove this assert as I'm not expecting it to be here?

milliams:
Fixes #426

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -480,35 +480,11 @@ class Registry(object):
     def find(self, _obj):
         """Returns the id of the given object in this registry, or 
         Raise ObjectNotInRegistryError if the Object is not found"""
-
+    
         obj = stripProxy(_obj)
         try:
-            if hasattr(obj, _reg_id_str):
-                obj_reg_id = getattr(obj, _reg_id_str)
-                objects_obj = self._objects[obj_reg_id]
-                assert obj == objects_obj
-                if hasattr(obj, _id_str):
-                    if hasattr(objects_obj, _id_str):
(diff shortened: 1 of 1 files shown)


milliams:
Any problems with merging this?

rob-c:
go from me, I'm wondering if this was a source of a performance improvement or whether the drop to 15min for the tests was conincidence

milliams:
I would not be surprised if this allowed a slight performance improvement but I'd be surprised if all of the 19→15 minutes changes was due to this.

rob-c:
It is called whenever there is a ```getNodeIndexCache``` which is called more than we'd like to admit so it may, but lets wait and see.

milliams:
Looking at the [test time trend](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/buildTimeTrend), it looks like the 19→15 minute time is holding well. I guess it was this PR that was responsible for the 25% speed improvement.

rob-c:
Could this just be 25% of the tests weren't migrated, I've not checked the history.

milliams:
Ah, that could indeed be related. It happened at a similar time.

PR #429, opened 2016-04-29
Only stop monitoring if it's still running

milliams:
It seems my fix that I put in at #373 was a bit rushed and has a potential problem. If the monitoring system is already stopped and `monitoring_component.stop()` is called, it will wait on the `__cleanUpEvent` `Event` to be signaled so that it knows it's safe to shutdown the thread. However, it will wait forever since there is no other monitoring thread available to raise that signal. This PR makes it so that the monitoring system is only stopped if it's still running.


rob-c:
I keep getting confused by the Monitoring shutting down but if this catches one of the few remaining issues here then great :+1: go for merge from me :)

Issue #416, opened 2016-04-27
GangaDirac should poll DIRAC for known SE


rob-c:
GangaDirac should poll the Dirac for known SE, else we will have to keep a configuration of the known SE in the DIRAC config.

For LHCb this is nice and short and _is just_ the Tier1 sites. For GridPP Dirac this is much larger and complex and it is more correct for GangaDirac to ask for known SE unless one is specified before doing something such as a put().

Hopefully this is nice and short and there is an API call to determine this, in which case I'll change the caching method in DiracFile to only ever request this list once as it's unlikely a site will go into downtime in any given session.
If this turns into something more difficult this may get pushed back but I would be surprised if this can't be achieved correctly in a few lines of code.

rob-c:
This will likely become superseded by users knowing where they want their output to go in the mid term but this would be nice to do.

rob-c:
I don't see this getting solved for 6.1.21 but there is discussion about whether we actually want this over passing the configuration responsibility onto the user to know where they explicitly want their things to go.

Issue #401, opened 2016-04-22
Ganga DIRAC integration/communications


rob-c:
OK, @po10 met with some of the LHCbDirac developers today at CERN and there have been a few questions asked which I'd like to get people's opinions on with regard to communication between Ganga and Dirac.
### DIRAC and Ganga in the same Python environment?

As far as I know Ganga and DIRAC run in different Python environments due to the 2 projects having incompatible environments in the past.
Would we still like to continue down this route or is there any feeling that we can have Ganga work within the same environment as DIRAC?
(We have reduced the number of non pip-installable externals and we've recently improved our support for different IPython versions etc so would it be possible to test/use a version of Python/IPython which ships with a DIRAC install?)

If we chose to keep Ganga and DIRAC environments separate it may be nice to move back to a server/client model which would allow us to change the `evaluate` method to simply be something like `evaluateOnServer` which reduces the impact of the change on our current code to the 1 function as everywhere else would still work the same.
Does anyone have any good suggestions on how this may be implemented in a good way? @alexanderrichards  I know you mentioned you've looked into this a little and have some ideas.
### Should we group commonly grouped DIRAC calls into a single command which we execute within a DIRAC namespace?

It may be beneficial to simply rewrite some more complex functions such that we have a single `finalizeDIRACJob` method and equivalent for submitting subjobs to the DIRAC backend which we call once.
(there are technical reasons why submitting jobs this way may be tricky but the finalization of a DIRAC hello world job can probably be improved on compared to what we're doing now)

These methods would return a larger dictionary back to user within Ganga but runs a bit more of the risk of timeouts.
#### Option a) move back to a server/client model:

A server/client model sounds like a very nice long term solution but has a lot of potential complexity involved I suspect.
#### Option b) group DIRAC commands which can all be reasonably sent at the same time.

Grouping DIRAC calls from within Ganga to reduce the amount of overhead is potentially a nice solution (and possibly quicker to implement) but it doesn't get rid of all of our overheads it simply reduces them by a large factor.

(NB: I'm not talking about bulk submission here which would reduce the overhead of the submission of many jobs. I'm also not talking about #370 which I believe is a Ganga issue which can be solved well in 6.1.20.)

My vote would for trying **Option b)** and seeing what improvements this brings, what does everyone think?

milliams:
[...]

PR #430, opened 2016-04-29
Bugfix/cleanup vanilla DIRAC

### LINKED ISSUE #408: remove site control from DIRAC to LHCbDIRAC
This is a small issue to remind myself to clean out some LHCb(ism) in the Vanilla DIRAC. These 'defaults' should probably be moved to the LHCb plugin configuration as GridPP DIRAC doesn't attempt to submit jobs to CERN so doesn't need to be told not to.

### LINKED ISSUE #419: Dirac+Executable+inputfiles not working
For some reason the combination of using these 3 on a vanilla Dirac instance isn't currently working.

I'm playing now with what's required to get the inputfiles to be tarred up as part of the inputsandbox.

Once I have a settled on a solution I'll share a PR.

### LINKED ISSUE #428: outputSE option for DIRAC jobs
I'm planning to add an ```outputSE``` option to the ```Dirac``` backend for DIRAC jobs in 6.1.20.

This will only become a requirement when a user tries to submit a job that requires outputdata but if a user attempts to submit a job without requesting an outputSE then the job submission should fail.

There is currently no nice automated way to upload data automatically to a writable SE based upon where a user can currently write to.

This will be independent of work on the ```DiracFile``` to make it work in a more vanilla DIRAC friendly way.

Obviously I'll add/change some options for LHCb so that LHCb users are unaffected by this change.

Are there any objections to this?

### LINKED ISSUE #432: Dirac job completion bug ( Ganga + GridPPDIRAC)
Some DIRAC jobs are grabbing the output sandbox which contains an (empty) ```__postprocesslocations__``` which overwrites the one generated from talking to DIRAC...

I am opening an issue because I'm going to take a look at the order in which jobs output sandbox and data is retrieved and to check that we don't overwrite _important_ files.

Ultimately we shouldn't be creating this file on the WN for ```DiracFile``` on a Dirac WN so there is a bug there which needs to be addressed and may fix this.
This code isn't called and just creates the file which isn't wanted.

I've manually verified the file exists for the job and has a correctly assigned JDL on the GridPP DIRAC after I fixed some other outstanding bugs to do with handling files/jobs.

### LINKED ISSUE #433: Investigate removal/reduction of DiracLHCbCommands
I'm making an issue to remind me to come back and remove a lot of the ```DiracLHCbCommands``` in favor of the ```DiracCommands``` file and find some way of more closely inheriting what goes on here.

There is a large amount of code duplication which can be significantly reduced by using a common code format (i.e. ```dirac``` vs ```lhcbdirac```) . This would help minimise the small amount of bugs which have crept in between Dirac and LHCbDirac.

An obvious one I'm fixing is that getting LFN back from a GridPPDirac job doesn't work as it should when consulting the Dirac interface. I want/need this for a new app I'm writing and this should have been working for a long time.
This can currently be _fixed_ (worked around) by making Ganga do all of the heavy lifting itself by dealing with DiracFile on the WN but this is not the most optimal way of doing this and don't want to invest the time in testing this or getting it working.

### LINKED ISSUE #434: stop spamming DIRAC/bad calls
I'm seeing the fact that we're spamming DIRAC with calls such as ```status([])```. These aren't giving us any information we don't know but are taking some time on the Monitoring thread when they're there.

I'm also seeing the Monitoring thread now being _too fast_ and we're finalizing the same job twice. This comes from the fact we don't have some global register about what we've currently put into a queue to be finalized.
~~I plan to add a global register with objects popped only after they've actually completed for the Dirac backend.~~ This just needs tweaking to look at whether the job has ```.been_queued```

### LINKED ISSUE #441: Dirac/Ganga job statuses
Hi all,

From memory there is a good reason about having the ```"Done"``` and ```"Complete"``` status in Dirac marked as ```completed``` in ganga.
(I think this was something to do with data replication etc...)

However, when I run my branch #430 I'm seeing jobs consistently being marked as failed as the job is not getting the output file until some time after the job has finished.

After digging around I found my jobs were passing through the monitoring cycle too quickly and were completing whilst Dirac was still processing the output from the Job.

This is responsible for at least some of the problems of jobs completing ahead of the output sandbox or data being made available.

In interests of getting something working I'm keen to move to have only Dirac jobs which are in a ```"Done"``` state (in DIRAC) marked ```completed``` (in Ganga).

Alternatively I'm going to move to mark all jobs in the ```"Complete"``` state but expect some output data in the form of a DiracFile (aka. asking Dirac to upload some output for us) as being in the running state. This is the only way to get Ganga to wait for the output data to be made available as requesting it before it's been uploaded will _work_ and will get an empty output data set which has been a pain to debug as the data exists in DIRAC and could be manually requested in Ganga but was missing after the monitoring loop progressed a job to completed.

Does anyone have any opinions on this?

Thanks,

Rob

rob-c:
OK this PR attempts to fix #408, #428 as I'd already coded up something which loops through the DiracFile for the backend when constructing the JDL.
This would be cleaner reproduced as well at a higher level for checking ```IGangaFile``` but this can be put off until a future PR.
This also partly addresses #416 as it should be reasonable for smaller VO to know what SE they actually have write access to.

The documentation around this needs to be improved where should I look to do this?

In short on GridPP the following will work.
```
j=Job(backend=Dirac)
...
j.outputfiles=[DiracFile('*.root', defaultSE='mySEofChoice')]
j.submit()
```
but the following will now fail (unless there is a defaultSE that has been configured):
```
j=Job(backend=Dirac)
...
j.outputfiles=[DiracFile('*.root')]
j.submit()
```

Whilst LHCb users should remain unaffected by this change as DIRAC handles where to put the file for us.

A result of this is that GridPP users now have to set:

(Editted)
```
[Dirac]
userVO = gridpp
allDiracSE = [ some_list_containing_the_defaultSE ]
[defaults_DiracFile]
defaultSE = someDefaultSE
```

LHCb users may control where their outputdata goes to by default by setting: ```defaultSE```on the ```DiracFile``` or as ```DiracDefaultSE``` because ```DiracOutputDataSE``` is now removed to avoid confusion.

Edit:
This PR now also contains the changes needed to submit and collect a job with ```DiracFile``` in bot the job ```inputdata``` and ```outputdata```.

Fixes #441 
Fixes #434 
Fixes #433 
Fixes #432 
Fixes #428 
Fixes #419 
Fixes #408 
Reduces total number of DIRAC calls by a bit as per discussion in #401 
Fixing ```DiracFile.put()``` when pattern matching local namePattern.


egede:
I suggest to let `DiracDefaultSE = None` (or the empty string if that is easier in the configuration system. In `LHCb.ini` it can then be set to `DiracDefaultSE`. If a `DiracFile` object on submission time then have a None/empty value for `defaultSE` the submission will fail (with an explaining message on how to fix it).

An even better solution might be to get rid of `[Dirac]DiracDefaultSE` and just set it in the `defaults_DiracFile` section of the configuration. Then it just becomes the default value like for any other element.

rob-c:
@egede
I'm tempted to drop the ```DiracDefaultSE``` as you suggest and change the .ini to reconfigure the DiracFile.
I'll change the PR to do this as it should alleviate a few problems,

rob-c:
This PR also stops the jobs accidentally reverting their status during finalisation when using the Dirac backend.
I'm close to making a PR for LSST and this branch represents the things which I've had to fix in Core/Dirac to successfully test it.

rob-c:
OK, this PR is fixing several independent issues all around the vanilla Dirac backend. I'm keen to get this in at the start of 6.1.20 once the release is out.
As a result I am seeing my ```hello world``` jobs on a Dirac backend are only taking about 20sec to complete down from >1min.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -495,20 +497,25 @@ def _getStateTime(job, status):
             if not job.subjobs and status in b_list:
                 for childstatus in b_list:
                     if job.backend.id:
-                        logger.debug("Accessing getStateTime() in diracAPI")
-                        dirac_cmd = "getStateTime(%d,\'%s\')" % (job.backend.id, childstatus)
-                        be_statetime = execute(dirac_cmd)
+               #         logger.debug("Accessing getStateTime() in diracAPI")
+                        #be_statetime = execute("getStateTime(%d,\'%s\')" % (job.backend.id, childstatus))
```
milliams: I guess these commented-out sections should be deleted?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -495,20 +497,25 @@ def _getStateTime(job, status):
...
+                                job.time.timestamps["backend_" + childstatus] = be_statetime
+               #             logger.debug("Wrote 'backend_%s' to timestamps.", childstatus)
                     if childstatus == status:
                         break
-            logger.debug("_getStateTime(job with id: %d, '%s') called.", job.id, job.status)
+            #logger.debug("_getStateTime(job with id: %d, '%s') called.", job.id, job.status)
         else:
-            logger.debug("Status changed from '%s' to '%s'. No new timestamp was written", job.status, status)
+            pass
+            #logger.debug("Status changed from '%s' to '%s'. No new timestamp was written", job.status, status)
```
milliams:
```python
else:
    pass
```
should be deleted.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,45 +543,57 @@ def _internal_job_finalisation(job, updated_dirac_status):
         logger = getLogger()
 
         if updated_dirac_status == 'completed':
+            import time
```
milliams: Standard library imports should be at the top of the file.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,45 +543,57 @@ def _internal_job_finalisation(job, updated_dirac_status):
...
 
             # Set DiracFile metadata
             wildcards = [f.namePattern for f in job.outputfiles.get(DiracFile) if regex.search(f.namePattern) is not None]
 
-            with open(os.path.join(job.getOutputWorkspace().getPath(), getConfig('Output')['PostProcessLocationsFileName']), 'ab') as postprocesslocationsfile:
+            lfn_store = os.path.join(output_path, getConfig('Output')['PostProcessLocationsFileName'])
+
+            if not os.path.isfile(lfn_store):
+                with open(lfn_store, 'w'):
+                    pass
```
milliams: This slightly strange idiom needs a comment. I assume it's creating the file if it doesn't exist?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,45 +543,57 @@ def _internal_job_finalisation(job, updated_dirac_status):
...
 
         if updated_dirac_status == 'completed':
+            import time
+            start = time.time()
             # firstly update job to completing
             DiracBase._getStateTime(job, 'completing')
             if job.status in ['removed', 'killed']:
                 return
-            if (job.master and job.master.status in ['removed', 'killed']):
+            elif (job.master and job.master.status in ['removed', 'killed']):
```
milliams: This (and others below) doesn't benefit from being an `elif`.

python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
```diff
@@ -70,6 +75,9 @@ def dirac_outputfile_jdl(output_files):
         if outputSE != '':
             myLine = myLine.replace('###OUTPUT_SE###', str([outputSE]))
         else:
+            if empty_SE_check is True:
```
milliams: Simply `if empty_SE_check:` is sufficient.

python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py
```diff
@@ -42,31 +43,52 @@ def prepare(self, app, appsubconfig, appmasterconfig, jobmasterconfig):
...
         exe_script_name = 'exe-script.py'
 
+        logger.info("Setting Command to be: '%s'" % repr(commandline))
+
         inputsandbox.append(FileBuffer(name=exe_script_name,
                             contents=script_generator(exe_script_template(),
                                                     #remove_unreplaced = False,
                                                     # ,
-                                                    COMMAND=commandline,
+                                                    COMMAND=repr(commandline),
```
milliams: I worry about using `repr` for this. It seems potentially brittle. Does the conversion to `str` happening in `script_generator` not suffice?

python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py
```diff
@@ -42,31 +43,52 @@ def prepare(self, app, appsubconfig, appmasterconfig, jobmasterconfig):
...
-        dirac_outputfiles = dirac_outputfile_jdl(outputfiles)
+        contents=script_generator(exe_script_template(), COMMAND=repr(commandline),
+                                    OUTPUTFILESINJECTEDCODE = getWNCodeForOutputPostprocessing(job, '    ')
+                                    )
+
+        #logger.info("Script is: %s" % str(contents))
+
+        from os.path import abspath, expanduser
+
+        for _file in job.inputfiles:
```
milliams: PEP 8 suggests leading underscores only for internal names. Trailing underscores should be used to avoid conflicts with Python names. Otherwise, `f` would do.

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -11,39 +11,46 @@
...
-# def output(object):
-#    reserved_stdout.write("CHECK ME " + str(object))
-#    reserved_stdout.write(pickle.dumps(object))
-
 # Dirac commands
 #/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/
 
-def getJobGroupJobs(jg): output(dirac.selectJobs(jobGroup=jg))
+def getJobGroupJobs(jg):
+    ''' Return jobs in a group'''
```
milliams: [PEP 257](https://www.python.org/dev/peps/pep-0257/#id15) suggests triple double quotes for docstrings.

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -259,7 +307,7 @@ def getStateTime(id, status):
         checkstr = ''
 
     if checkstr == '':
-        sys.stdout.write(None)
+        print("%s" % None)
```
milliams: Why not `print(None)`?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -34,8 +34,15 @@ def getDiracEnv(force=False):
     lock = Dirac_Env_Lock
     lock.acquire()
     if DIRAC_ENV == {} or force:
-        if getConfig('DIRAC')['DiracEnvFile'] != "" and os.path.exists(getConfig('DIRAC')['DiracEnvFile']):
-            with open(getConfig('DIRAC')['DiracEnvFile'], 'r') as env_file:
+        config_file = getConfig('DIRAC')['DiracEnvFile']
+        if not os.path.exists(config_file):
+            import inspect
```
milliams: Standard library imports should be at the top of the file.

python/GangaLHCb/Lib/Server/DiracLHCbCommands.py
```diff
@@ -1,48 +1,29 @@
...
 
-#def getRootVersions(): output( diraclhcb.getRootVersions() )
+#def getRootVersions(): output( dirac.getRootVersions() )
 
-#def getSoftwareVersions(): output( diraclhcb.getSoftwareVersions() )
+#def getSoftwareVersions(): output( dirac.getSoftwareVersions() )
 
 
-def bkQueryDict(dict): output(diraclhcb.bkQuery(dict))
+def bkQueryDict(dict): output(dirac.bkQuery(dict))
```
milliams: `dirac` doesn't seem to exist in this file. Or is this one of the files that are embedded in another?

rob-c: This file is embedded in another after the GangaDirac commands. This gets rid of a couple of hundred lines of repeated code between GangaDirac and GangaLHCb when doing Dirac work.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: They could be, I'm still in the stages of testing this as some of my jobs are failing due to other reasons or are still on queues atm.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: I'm yet to confirm if the logger here was causing a performance problem. I suspect not so the logging will probably go back in before merge

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: It's not harmed by it though, I don't see any problem here?

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Yes, although this is used in enough other places in Ganga I assumed people were familiar with it.

python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
rob-c: Correct, although during development there was a more complex object here that made more sense to reduce to a bool

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: Ok, since you said "I'm keen to get this in at the start of 6.1.20 once the release is out" I thought this pull request was ready to be merged.

python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py
rob-c: I would argue this is internal to the for loop as it reminds me not to call the iterator once it's been finished with, but I suppose there is room for change here

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: No, no harm. I find it *slightly* harder to parse but not much. I was just wondering if there was a reason for the change.

python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py
rob-c: Don't know for certain. I wouldn't have said this is particularly brittle I would have said it's more belt and braces to remind me that the object is now a list being passed here rather than a string. I don't want to go checking/fixing the script_generator if I can help it as I'm sure it'll come back to bite me if I changed anything there.

python/GangaDirac/Lib/Server/DiracCommands.py
rob-c: Can't do it due to the way this file is used. The file is constructed at a much higher level making use of ```"""``` which will be broken if we use it here.
We could in future change this but it got ugly and this seemed the best compromise over using ```#``` to comment on the functions.

python/GangaDirac/Lib/Server/DiracCommands.py
rob-c: I prefer this, that probably works as well, although I don't think we test these functions properly

python/GangaDirac/Lib/Server/DiracCommands.py
milliams: I am surprised this causes problems since string escaping should only happen on the literal level, not after the strings are parsed but if you saw problems then I guess it's just being constructed in a weird way.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: The general shape is I think I'll be pushing for merge at the end of today, aside from playing with the logger here there isn't anything else I'm too worried about, (although I'd like to retest based upon some changes elsewhere in the PR)

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: No strong reason other than the fact I had an else function with some logging but this got dropped and I just stuck with what was left.

python/GangaDirac/Lib/Server/DiracCommands.py
rob-c: The file is constructed using some ```"""``` to wrap a lot of objects right at the end and then perform some string replacement/formatting I think but either way it didn't like there being multiple groups of these.
I'll keep this in mind for something to change in the future

rob-c:
I'm keen to see if I can get this in asap on Friday if possible. I've performed some test jobs by hand and everything appears to be working well on this PR and it addresses numerous papercuts and some structural things around using a Vanilla (GridPP) Dirac instance.
I'm basing my LSST work on this PR so once this is in I will focus on getting this finished.

doc/dev/index.rst
```diff
@@ -58,3 +58,56 @@ A registry is an in-memory data-store which is backed by a repository.
 
 Job monitoring
 --------------
+
+IGangaFile
+----------
+
+All file types as of Ganga 6.1 inherit from ```IGangaFile```. This main exception to this is the ```File``` object which as of 05/05/2016 is used as it still has more features than the ```IGangaFile``` inheirted classes do.
```
milliams: These should only have double backticks \`\`around them, not triple.

python/Ganga/GPIDev/Lib/File/OutputFileManager.py
```diff
@@ -8,38 +8,45 @@
...
+    """
     backendClassName = getName(job.backend)
 
     backend_output_postprocess = stripProxy(job).getBackendOutputPostprocessDict()
     if backendClassName in backend_output_postprocess:
         if outputFileClassName in backend_output_postprocess[backendClassName]:
-            if backend_output_postprocess[backendClassName][outputFileClassName] == 'client':
-                return True
+            if backend_output_postprocess[backendClassName][outputFileClassName] == when:
+                 return True
```
milliams: It looks like `return True` has too many spaces in front of it.

python/Ganga/GPIDev/Lib/File/OutputFileManager.py
```diff
@@ -8,38 +8,45 @@
...
+def outputFilePostProxessingOnSubmit(job, outputFileClassName):
+    """
+    Checks if the output files of a given job(we are interested in the backend)
+    should have been postprocessed on job submission, depending on job.backend_output_postprocess dictionary
+    """
+    return outputFilePostProcessingOnTest(job, outputFileClassName, 'submit')
 
-def outputFilePostProcessingOnClient(job, outputFileClassName):
+
+def outputFilePostProcessingOnTest(job, outputFileClassName, when):
```
milliams: What does `OnTest` mean in this function name?

python/Ganga/GPIDev/Lib/File/OutputFileManager.py
```diff
@@ -282,8 +289,17 @@ def getWNCodeForOutputPostprocessing(job, indent):
...
-                    outputfileClassName].append(outputFile)
+                outputFilesProcessedOnWN[outputfileClassName].append(outputFile)
+
+    if patternsToZip == []:
+        should_exit = True
+        for output_class in outputFilesProcessedOnWN:
+            if outputFilesProcessedOnWN[output_class] != []:
+                should_exit = False
+        if should_exit:
+            return ""
```
milliams: Why not:
```python
if not patternsToZip:
    if not any(outputFilesProcessedOnWN.values()):
        return ""
```

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -165,7 +165,8 @@ def _setParent(self, parent):
     def get(self, to_match):
         def matching_filter(item):
             if '_list_get__match__' in dir(item):
-                return item._list_get__match__(to_match)
+                ## This drops through to a base class so for sanity sake strip it !
+                return item._list_get__match__(stripProxy(to_match))
```
milliams: This should not be needed after #393

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -758,7 +758,7 @@ def outputFilesFailures(self):
 
         # check if any output was matched - be careful about master jobs
         if getConfig('Output')['FailJobIfNoOutputMatched'] and not self.subjobs:
-            for outputfile in self.outputfiles:
+            for outputfile in self.outputfiles:
```
milliams: I guess this is a typo?

python/Ganga/Lib/Localhost/LocalHostExec.py
```diff
@@ -88,14 +88,18 @@
 
 import time #datetime #disabled for python2.2 compatiblity
 
+if os.path.isfile(os.path.abspath(appscriptpath[0])):
```
milliams: This seems to be erroneously including the change from #418

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,45 +541,59 @@ def _internal_job_finalisation(job, updated_dirac_status):
...
-            import time
-            start = time.time()
             # contact dirac for information
-            job.backend.normCPUTime = execute('normCPUTime(%d)' % job.backend.id)
-            getSandboxResult = execute("getOutputSandbox(%d,'%s')" % (job.backend.id, job.getOutputWorkspace().getPath()))
-            file_info_dict = execute('getOutputDataInfo(%d)' % job.backend.id)
+            # The following single execute command is equivalent to the following few lines
+            #job.backend.normCPUTime = execute('normCPUTime(%d)' % job.backend.id)
+            #getSandboxResult = execute("getOutputSandbox(%d,'%s')" % (job.backend.id, job.getOutputWorkspace().getPath()))
+            #file_info_dict = execute('getOutputDataInfo(%d)' % job.backend.id)
```
milliams: The commented-out code here should be removed.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,45 +541,59 @@ def _internal_job_finalisation(job, updated_dirac_status):
...
+            #logger.info('Job ' + job.fqid + ' OutputSandbox: ' + str(getSandboxResult))
+            #logger.info('Job ' + job.fqid + ' normCPUTime: ' + str(job.backend.normCPUTime))
 
             # Set DiracFile metadata
             wildcards = [f.namePattern for f in job.outputfiles.get(DiracFile) if regex.search(f.namePattern) is not None]
 
-            with open(os.path.join(job.getOutputWorkspace().getPath(), getConfig('Output')['PostProcessLocationsFileName']), 'ab') as postprocesslocationsfile:
+            lfn_store = os.path.join(output_path, getConfig('Output')['PostProcessLocationsFileName'])
+
+            ## Make the file if it doesn't exist as we do in other places...
```
milliams: PEP8 suggests single `#` for line comments.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -271,7 +268,8 @@ def dirac_line_processor(self, line, dirac_file, localPath):
         guid = tokens[3]
         try:
             locations = eval(tokens[2])
-        except:
+        except Excaption as err:
```
milliams: Typo

milliams: Also, a more precise exception type would be good.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -34,8 +35,14 @@ def getDiracEnv(force=False):
     lock = Dirac_Env_Lock
     lock.acquire()
     if DIRAC_ENV == {} or force:
-        if getConfig('DIRAC')['DiracEnvFile'] != "" and os.path.exists(getConfig('DIRAC')['DiracEnvFile']):
-            with open(getConfig('DIRAC')['DiracEnvFile'], 'r') as env_file:
+        config_file = getConfig('DIRAC')['DiracEnvFile']
+        if not os.path.exists(config_file):
+            absolute_path = os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))),
+                                        '../../..', config_file)
```
milliams: At least this should probably use `os.path.dirname(os.path.realpath(__file__))` but is there not a better way to get the file rather than a relative path?

python/Ganga/GPIDev/Lib/File/OutputFileManager.py
rob-c: ```should be postprocessed 'when'```
OK I can give this a more general name but I was assuming that looking at the code this was fairly clear

python/Ganga/GPIDev/Lib/File/OutputFileManager.py
rob-c: I'll fix

python/Ganga/GPIDev/Lib/File/OutputFileManager.py
rob-c: That does look more elegant. I'll go with that

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: But it's needed now for some other code I'm wanting to get in, when can we expect #393?

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: whitespace typo

python/Ganga/Lib/Localhost/LocalHostExec.py
rob-c: No, I ended up adding this to the Dirac backend in order to get anything done sensibly during testing.
I can merge #418 first which is this short change but I was planning to mention that this supersedes it.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: This was serving as documentation as to what the ```finalize_job``` method was doing, If it's wanted to remove it i'll change it

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: This should have been ```logger.debug``` I'll get round to fixing it

python/Ganga/Lib/Localhost/LocalHostExec.py
milliams: Between your suggestion of adding the WN path to `PATH` and Ulrik's opinion that `File` is ok to use and that the PR should be closed, I'm not sure that putting this code into *this* PR is the best option.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
milliams: When anyone has reviewed it.

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: I feel it's clear that `execute("finalize_job(...` is calling `finalize_job` which has the documentation in its docstring. I don't feel there's a need to document what a function does when calling it.

doc/dev/index.rst
rob-c: Rats, wasn't planning to add this to this PR... oh well, I'll just modify it, no harm in extra docs

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -177,69 +199,71 @@ def getOutputDataLFNs(id, pipe_out=True):
...
         if 'NormCPUTime(s)' in parameters:
             ncput = parameters['NormCPUTime(s)']
-    output(ncput)
+    if pipe_out:
+        output(ncput)
+    else:
+        return ncput
+
 
+def finalize_job(id, outputDir=os.getcwd(), oversized=True):
```
milliams: We use British English in the Ganga code so this should be `finalise_job`.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Frankly I find it easier to remember what is called here for reference sake when comparing to previous implementations but I'll drop it if it's controversial

milliams: The problem is, when `def finalize_job()` is changed in the future to do something different/extra/less then this comment becomes out-of-date and confusing. Having a wrong comment is worse than having no comment at all.

python/Ganga/Lib/Localhost/LocalHostExec.py
rob-c: I'll drop it for now after reading comments in #418 but I'm under protest that using ```File``` and ```LocalFile``` is just going to cause un-needed confusion. (and frankly makes me suspect the string functionality here is something we shouldn't support)

python/GangaDirac/Lib/Server/DiracCommands.py
rob-c: Given my spelling ability in general you're lucky I stuck to a dictionary. I'll change this since it's just in 2 places, but there is a lack of conformity in a lot of other places, and I suspect this is a minor issue

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: And my argument would be that this future is when you amend your documentation, but as I've said, I'll drop it...

python/Ganga/Lib/Localhost/LocalHostExec.py
rob-c: And now this drops the ability for me to have a string and call a DiracFile... I'll just add an issue for that.

rob-c:
Right, I'm looking to merge this again after another round of comments, is there anything outstanding to change?

python/Ganga/Lib/Localhost/LocalHostExec.py
```diff
@@ -75,8 +75,7 @@
 import tarfile
 
 fullenvironment = os.environ.copy()
-for key,value in environment.iteritems():
-    fullenvironment[key] = value
+fullenvironment['PATH'] = os.getcwd():fullenvironment['PATH']
```
milliams: This looks like a typo (`:fullenvironment['PATH']`) unless I'm misunderstanding something.

rob-c: Wrong branch, removing now

rob-c:
OK to merge?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -496,13 +498,16 @@ def _getStateTime(job, status):
                 for childstatus in b_list:
                     if job.backend.id:
                         logger.debug("Accessing getStateTime() in diracAPI")
-                        dirac_cmd = "getStateTime(%d,\'%s\')" % (job.backend.id, childstatus)
-                        be_statetime = execute(dirac_cmd)
                         if childstatus in backend_final:
+                            be_statetime = execute("getStateTime(%d,\'%s\')" % (job.backend.id, childstatus))
                             job.time.timestamps["backend_final"] = be_statetime
+                            break
```
milliams: There is a line of code after this `break` statement. What is intended here?

rob-c: moved

rob-c:
One last round then I plan to merge at 5pm or so if nothing crops up

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: It looks like this was not moved in the code that you merged.

PR #431, opened 2016-05-01
Trivial bugfixes

olupton:
- Fix typo of debug function.
- Make Tasks work again.
- Re-add change that got lost somewhere allowing Tasks using GaussSplitter to seed the event number properly.

### DIFF
--- a/python/GangaLHCb/Lib/RTHandlers/RTHUtils.py
+++ b/python/GangaLHCb/Lib/RTHandlers/RTHUtils.py
@@ -57,12 +57,6 @@ j.setAncestorDepth(###ANCESTOR_DEPTH###)
 def is_gaudi_child(app):
     if isType(app, Gaudi):
         return True
-
-    if isType(app, TaskApplication):
-        from GangaLHCb.Lib.Applications import GaudiPythonTask, BenderTask
-        if not isType(app, GaudiPythonTask) and not isType(app, BenderTask):
-            return True
-
     return False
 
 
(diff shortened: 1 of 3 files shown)


milliams:
ok to test

python/GangaLHCb/Lib/RTHandlers/RTHUtils.py
```diff
@@ -57,12 +57,6 @@ def lhcbdiracAPI_script_template():
 def is_gaudi_child(app):
     if isType(app, Gaudi):
         return True
-
-    if isType(app, TaskApplication):
-        from GangaLHCb.Lib.Applications import GaudiPythonTask, BenderTask
-        if not isType(app, GaudiPythonTask) and not isType(app, BenderTask):
```
rob-c: Apologies I should get this but aren't these classes derived from the ```GangaGaudi``` codebase? Why is the check being removed?

olupton: @drmarkwslater should comment further - Tasks didn't work in the develop branch because TaskApplication doesn't exist anymore. I didn't immediately find the other two classes either so I removed the whole check.

rob-c: @milliams Can we bump this to 6.1.19 then so that this release works for LHCb people with Tasks I'd like to get as many people off 6.0.44 for LHCb asap.

milliams: Sure, we can put this in for 6.1.19. The PR is against `develop` so we'll have to merge it there and then merge that into the release branch.

milliams:
:+1:

PR #435, opened 2016-05-03
Migrate remaining tests

milliams:
As @rob-c pointed out, I missed moving some tests in #291. This covers moving the remaining ones.


python/Ganga/test/GPI/GangaList/TestCopy.py
```diff
@@ -27,18 +25,18 @@ def testCopy(self):
 
         gl = GangaList()
 
-        numberOfFiles = 5
+        numberOfFiles = 100
 
         for _ in range(numberOfFiles):
             # add something which is generally not allowed by GangaList
-            gl.append(self._makeRandomTFile())
+            gl.append([self._makeRandomTFile()])
```
drmarkwslater: doesn't this equate to:
```
[ [TFile], [TFile], [TFile], ....]
```
instead of
```
[ TFile, TFile, TFile, ...]
```

That may be what's needed (e.g. to test nested lists) but isn't that in another test?

drmarkwslater:
Out of interest, why was there a change from Unit Test -> GPI for the GangaList/TestCopy? I don't really mind, just interested 😄

milliams:
Before I made the PR #291, `GPI/GangaList/TestCopy.py` looked like [this](https://github.com/ganga-devs/ganga/blob/dd6230af6fcb9b2fcec4a7b7361114b9ab858382/python/Ganga/new_tests/GPI/GangaList/TestCopy.py). In the course of doing the PR I accidentally merged a local copy of `TestCopy.py` which I had been working on. This was an incorrect version. This PR removes my version and reinstates yours.

drmarkwslater:
Ah :) In which case, would you mind just changing the `append` call? I suspect that was my mistake :smile:

and then 👍  from me!

rob-c:
good to see this can be recovered, merge asap and as soon as everything is tested imo :+1:

milliams:
I'm going to merge this PR in first since it potentially affects the tests in other PRs.

PR #445, opened 2016-05-05
Transfer some lhcb tests

drmarkwslater:
This PR primarily allows plugins to be tested correctly. With some changes to Jenkins which I'll put in tomorrow, the tests will be made to check for a `Ganga<plugin>/testlib/setup_env.sh` script which will be sourced if found, and then the tests within `Ganga<plugin>/test` will be run. At present, the LHCb env script just sets the env variables but it can do anything that's needed to get the tests to run.

In addition:

1. Moves some LHCb tests over to the GangaLHCb/test
2. Adds an `@external` decorator to marks tests that rely on external services
3. Moves `GangaUnitTest` to `Ganga/testlib`


rob-c:
Great to see the first steps to getting this running.
I'm assuming from ```GangaLHCb/testlib/setup_env.sh``` we have access to cvmfs?
I've some ideas how to improve the environment stuff to more closely match what we do in the LHCb login but I'll wait until we have some basic tests running first before I do anything there (and for other work to get finish)

python/GangaLHCb/test/GPI/TestDataFiles.py
```diff
@@ -0,0 +1,32 @@
...
+        # LFNs
+        name = 'test.txt'
+        lfn = DiracFile(lfn=name)
+        assert lfn.lfn == name
+        lfn = DiracFile('lfn:'+name)
+        assert lfn.lfn == name
+        lfn = DiracFile('LFN:'+name)
+        assert lfn.lfn == name
+        try:
+            lfn = DiracFile('pfn:'+name)
```
rob-c: Is this even checked for I thought we dropped this entirely as of 6.1?

drmarkwslater: A warning is printed and a `LocalFile` created so it does work. However, if it's a valid test is up for debate as I just copied it wholesale from the legacy tests.

drmarkwslater:
The `setup_env.sh` script can doing anything that's needed but at present I've just got it setting environment variables as that was much less hassle. It is basically what you get if you do:
```
source /cvmfs/lhcb.cern.ch/group_login.sh
SetupProject Ganga
env > setup_env.sh
```
And then add the `export`s on.

milliams:
To reduce the noise in `setup_env.sh`, perhaps doing something like:
```bash
env > pre_env
source /cvmfs/lhcb.cern.ch/group_login.sh
SetupProject Ganga
env > post_env
diff -u pre_env post_env
```
would help?

python/GangaLHCb/test/GPI/TestDataFiles.py
```diff
@@ -0,0 +1,31 @@
...
+            raise RuntimeError('Should have got exception trying to create DiracFile from pfn.')
+        except:
+            pass
```
milliams: This could be changed to use [`pytest.raises`](https://pytest.org/latest/assert.html#assertions-about-expected-exceptions) to be slightly neater. Not a big deal, more informing for the future.

python/GangaLHCb/test/GPI/InterfaceLookFeel.py
```diff
@@ -0,0 +1,28 @@
...
+        j1 = Job(name='my',application='DaVinci')
+        j2 = Job(application = DaVinci())
+
+        j1.backend = LSF()
+        j1.backend.queue = '8nm'
+        j2.backend = j1.backend # deepcopy
+        j2.backend.queue = '16nh' # shortcut
+        bk2 = j2.backend # reference
+
+        assert(j2.backend.queue == '16nh')
```
milliams: `assert` should not use brackets around its arguments. It's error prone if a message is added in the future.

python/GangaLHCb/test/GPI/TestSubmitLocal.py
```diff
@@ -0,0 +1,26 @@
...
+        from GangaLHCb.testlib import addLocalTestSubmitter
+
+        ap = DaVinci()
+        j = Job(application=ap, backend=TestSubmitter())
+
+        # Test that submission fails before adding runtime handler
+        try:
+          j.submit()
+        except JobError:
+          pass
```
milliams: Potentially `pytest.raises` again.

python/GangaLHCb/test/GPI/InterfaceLookFeel.py
drmarkwslater: Done!

python/GangaLHCb/test/GPI/TestDataFiles.py
drmarkwslater: Actually, this doesn't work in any case as the `raise`'d exception just gets caught anyway :smile: Plus it **doesn't** raise an exception. So I've just remove the `try...except` block.

python/GangaLHCb/test/GPI/TestSubmitLocal.py
drmarkwslater: Done!

drmarkwslater:
@milliams You're right - I've cleaned the setup_env.sh script now.

Any objections to a merge?

python/GangaLHCb/test/GPI/TestDataFiles.py
rob-c: this is news (and scares me a little) how can a call to a DiracFile init return a LocalFile?

drmarkwslater: My bad - I was thinking of `LHCbDataset`. So this works (or at least doesn't raise an exception) and gives:
```
[11:13:28]
Ganga In [3]: lfn = DiracFile('pfn:test.txt')

[11:13:30]
Ganga In [4]: lfn
Ganga Out [4]: 
 DiracFile (
   defaultSE = ,
   namePattern = pfn:test.txt,
   guid = ,
   remoteDir = ,
   compressed = False,
   localDir = ,
   lfn = ,
   failureReason = ,
   locations = []  ,
   subfiles =[]
 )
```

I don't know if this is right or not but that's where we are :smile: (at least as of 6.1.18)

milliams:
It's scary how many environment variables LHCb add in their setup script! :ghost: It looks like it's mostly down to the minimum now though.

I'm wary of `PATH` being prepended with the LHCb stuff, this means that it will be finding LHCb's version of Python rather than the version that is inside the virtualenv in which the tests run. Actually, thinking some more on it, since we're running as `py.test ...`, it will find the correct `py.test` (since there's only one) which has inside a hard-coded path to the correct Python executable. I think this should be fine. It's worth remembering though.

drmarkwslater:
I agree - it could certainly be an issue but then prepending the original `PATH` may mess other things up with the LHCb envirionment. I'm personally of the opinion that if it works we leave it until it stops working 😄

PR #447, opened 2016-05-06
Change position of added PATH variable

drmarkwslater:
Needed to change where to put the original $PATH env variable as otherwise the wrong py.test is picked up.


milliams:
:+1: I'm happy with this. However, I notice that there's a `/home/mws/bin` in the `PATH`.

drmarkwslater:
Good catch 😄

egede:
Sorry, I am late to the game on this discussion. However, I am worried that by catching all the environment variables, we are not correctly following the evolution of LHCb.

Would it not be much better to source `/cvmfs/lhcb.cern.ch/lib/LbLogin.sh` which is supposed to do exactly this and is updated correctly.

drmarkwslater:
This is almost certainly true but in the (very) short term I wanted to do things this way so it was very obvious where any problems are because you know exactly what's being set. It's quite possible it would all work wonderfully out of the box but I was concerned that the (somewhat) black box nature of `LbLogin.sh` could do all sorts that we'd then need pick the particular pieces to undo to get it working. I would say that it's worth revisiting this very soon after we have all the legacy tests transferred to see what would be needed to change to setup the LHCb environment properly.

A good thing though is that it would be a very trivial PR to put in to change this - i.e. just change the contents of the shell file and the testing picks it up.

milliams:
When we do switch over to just sourcing `/cvmfs/lhcb.cern.ch/lib/LbLogin.sh` it would be a good test to make sure that simply doing:
```bash
source /cvmfs/lhcb.cern.ch/lib/LbLogin.sh
export GANGA_CONFIG_PATH=/cvmfs/ganga.cern.ch/Ganga/install/LATEST/python/GangaLHCb/LHCb.ini
export GANGA_SITE_CONFIG_AREA=/afs/cern.ch/sw/ganga/install/config/LHCb
/cvmfs/ganga.cern.ch/Ganga/install/LATEST/bin/ganga
```
would be enough to get things working. Those middle two lines are just an example but are effectively all that _should_ be needed.

milliams:
However, I am happy for this PR to be merged.

PR #450, opened 2016-05-09
Fixing prepare step bug

rob-c:
This cleans up some code in the IPrepareApp and gives a hook to allow derrived classes to quickly and easily copy a file into the prepared state folder during the prepare step.
This is useful for my LSST work where I'm already using this and it fixes a typo which was left from this function already existing 4x in the exiting code.

### DIFF
--- a/python/Ganga/GPIDev/Adapters/IPrepareApp.py
+++ b/python/Ganga/GPIDev/Adapters/IPrepareApp.py
@@ -48,7 +48,7 @@ class IPrepareApp(IApplication):
         Base class for all applications which can be placed into a prepared\
         state. 
         Args:
-            force (bool) : Fforces the prepare function to be called no matter what when True
+            force (bool) : forces the prepare function to be called no matter what when True
         """
         pass
 
--- a/python/Ganga/GPIDev/Adapters/IPrepareApp.py
+++ b/python/Ganga/GPIDev/Adapters/IPrepareApp.py
@@ -70,6 +70,7 @@ class IPrepareApp(IApplication):
     def copyIntoPrepDir(self, obj2copy):
         """
         Method for actually copying the "obj2copy" object to the prepared state dir of this application
+        obj2copy is a string address of a file to be copied as it's passed to shutil.copy2
         """
         shared_path = os.path.join(expandfilename(getConfig('Configuration')['gangadir']), 'shared', getConfig('Configuration')['user'])
 
--- a/python/Ganga/GPIDev/Adapters/IPrepareApp.py
+++ b/python/Ganga/GPIDev/Adapters/IPrepareApp.py
@@ -67,6 +67,18 @@ class IPrepareApp(IApplication):
             self.is_prepared = None
         self.hash = None
 
+    def copyIntoPrepDir(self, obj2copy):
+        """
+        Method for actually copying the "obj2copy" object to the prepared state dir of this application
+        """
+        shared_path = os.path.join(expandfilename(getConfig('Configuration')['gangadir']), 'shared', getConfig('Configuration')['user'])
+
+        shr_dir = os.path.join(shared_path, self.is_prepared.name)
+        if not os.path.isdir(shr_dir):
+            os.makedirs(shr_dir)
+        shutil.copy2(obj2copy, shr_dir)
+        logger.debug("Copying %s into: %s" % (obj2copy, shr_dir))
+
     def copyPreparables(self):
         """
         This method iterates over all attributes in an application and decides\
@@ -81,7 +93,7 @@ class IPrepareApp(IApplication):
                 logger.debug('Found preparable %s' % (name))
                 logger.debug('adding to sharedir %s' % (self.__getattribute__(name)))
                 send_to_sharedir.append(self.__getattribute__(name))
-        shared_path = os.path.join(expandfilename(getConfig('Configuration')['gangadir']), 'shared', getConfig('Configuration')['user'])
+
         for prepitem in send_to_sharedir:
             logger.debug('working on %s' % (prepitem))
             # we may have a list of files/strings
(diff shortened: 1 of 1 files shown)


milliams:
Makes sense to me. It's not obvious from the docstring what type `obj2copy` should be. Adding [argument descriptions to the docstring](http://www.sphinx-doc.org/en/stable/ext/napoleon.html#module-sphinx.ext.napoleon) in the Google style would be useful as these are also parsed by Sphinx when generating [the documentation](http://ganga.readthedocs.io/en/latest/API/Ganga.GPIDev.Adapters.IPrepareApp.html).

Also, we could potentially consider starting to use [PEP 484](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code)-style comments to define types.

rob-c:
@milliams I'll glance at the links later but I just wanted to make sure I don't end up with a huge commit addressing lots of bugs again in the future.

rob-c:
OK, frankly this is why I prefer strong typed languages. I've no idea what to do with the documentation I'm simply going to add a line stating that this is expected to be a string due to the ```shutil.copy2``` function

milliams:
For the record the Google-style docstring looks like:
```python
def copyIntoPrepDir(self, obj2copy):
    """
    Copy a file into the prepared state dir of this application

    Args:
        obj2copy (str): the path to the file to be copied
    """
```
And if you wanted to use PEP 484 comment, then it would look like:
```python
def copyIntoPrepDir(self, obj2copy):
    # type: (str) -> None
    """
    Copy a file into the prepared state dir of this application

    Args:
        obj2copy (str): the path to the file to be copied
    """
```

rob-c:
@milliams so what are we looking to standardize on?

milliams:
For now I think just the docstring format is needed, so the first of the two. The latter is only useful if we start running static analysis tools like [MyPy](http://mypy-lang.org/) over Ganga or for those of us using [PyCharm](https://www.jetbrains.com/pycharm/) but until then they are arguably just noise.

@alexanderrichards has been arguing for the former for a while now too.

alexanderrichards:
> @alexanderrichards has been arguing for the former for a while now too.

:+1:  Indeed I have :smile:.
Out of curiosity @milliams do you prefer the **type comments** from pep484 like in your second example or inline in the function prototype using the annotation style?

e.g.
```python
def copyIntoPrepDir(self, obj2copy: str) -> None:
    """
    Copy a file into the prepared state dir of this application

    Args:
        obj2copy (str): the path to the file to be copied
    """
```

rob-c:
I'm not going to start reading documentation on documentation, it's too meta and full of a lot of confusion when I'm trying to get other work done.
If I'm to write documentation like the google-style then OK, as long as it's clear what is expected. If there's a cheat sheet for those who don't care about how this works or what this does even better.

I'm sticking with the first example unless someone tells me otherwise. I want to fix and get this in so I can merge it into another branch where I want this functionality,

alexanderrichards:
> I'm sticking with the first example unless someone tells me otherwise.

I think we agree that the first example (The Google-style) one is the one we should aim for at the moment as we don't run any static analysis tools as I think that it is perfectly sufficient even when we do. The main reason I wanted us to do it this way is because it gives us the ability to auto-generate doc whilst maintaining the user friendly interactive help that users expect.

milliams:
@alexanderrichards I much prefer the inline style but that is restricted to Python 3 only. The comment style was added to the PEP in the last few months to be able to provide the feature for Python 2 and to provide a forwards compatibility path.

@rob-c Doing documentation properly is part of the job I'm afraid. I'm not saying that you have to have an opinion on what is the best way but is is important in a collaborative project to help out your colleagues by documenting your work appropriately.

alexanderrichards:
> I much prefer the inline style but that is restricted to Python 3 only.

Ah yes forgot that :wink:

rob-c:
@milliams ```Doing documentation properly is part of the job I'm afraid.``` I am _very_ aware of this and am extremely happy to document things as long as it's clear what I am doing and it is set out in a simple way what is expected of me.
This style formatting (as far as I'm aware) hasn't been discussed formally until today which is why I'm asking for a decision to be made by someone.
(I've documented the file as I prefer to do this all rather than doing just 1 function at a time as I've mentioned in the past.)

python/Ganga/GPIDev/Adapters/IPrepareApp.py
```diff
@@ -34,9 +34,17 @@ class IPrepareApp(IApplication):
     _hidden = 1
 
     def __init__(self):
+        """
+        default constructor
+        """
```
milliams: This comment does not add anything. In fact, this `__init__` method should just be removed entirely.

python/Ganga/GPIDev/Adapters/IPrepareApp.py
```diff
@@ -34,9 +34,17 @@ class IPrepareApp(IApplication):
...
         super(IPrepareApp, self).__init__()
 
     def _auto__init__(self, unprepare=None):
+        """
+        Funiction called when initializing from the Proxy layer i.e. interactive prompt or 'import ganga'
```
milliams: Typo on `Funiction`

python/Ganga/GPIDev/Adapters/IPrepareApp.py
```diff
@@ -34,9 +34,17 @@ class IPrepareApp(IApplication):
...
+        Args:
+            unprepare (bool): a bool which unprepares an app when it's created new i.e. don't copy prepared sandboxes
```
milliams: No need to repeat that it's a `bool`.

python/Ganga/GPIDev/Adapters/IPrepareApp.py
```diff
@@ -45,7 +53,8 @@ def prepare(self, force=False):
         """
         Base class for all applications which can be placed into a prepared\
         state. 
-
+        Args:
+            force (bool) : ignored by the base class. Higher up forces the prepare function to be called no matter what
```
milliams: No need to state that it's ignored by the base class. All subclasses must reimplement this function. Thedocstring should just describe what the argument is for:
```python
force (bool) : Force the prepare function to be called no matter what
```

rob-c: As I say if there is a cheat sheet according to some standard for those who care about these details I'm very happy to follow it. But this is a valid statement for this attribute in this class so why not state this?

milliams: There's no cheat sheet, just standard 'rules' for code documentation, describe the interface rather than implementation etc.

rob-c: Well we're clearly approaching this from different directions but I'll limit to interface descriptions in this case.

rob-c:
OK, once the final round of tests pass I plan to merge this unless there are any objections

python/Ganga/GPIDev/Adapters/IPrepareApp.py
```diff
@@ -45,7 +47,8 @@ def prepare(self, force=False):
         """
         Base class for all applications which can be placed into a prepared\
         state. 
-
+        Args:
+            force (bool) : Fforces the prepare function to be called no matter what when True
```
milliams: Type on 'Fforces'

rob-c: This is going to take another 20min for the sake of a typo. I'll merge the PR and edit it on develop

PR #453, opened 2016-05-09
Add runtime checks of changing registry

milliams:
The registry of an object should never change from one registry to another, only ever *to* or *from* `None`.

This is currently the case in all the Ganga code but adding this check here increases our security in the future.


rob-c:
I'm weary this could come back to bite us. Are there any tests which move objects to/from the templates/box/jobs repositories?

milliams:
There's no tests explicitly doing this as it should only ever be done in the context of creating new objects or loading them from XML. In principle, the registry of an object has two purposes:

1. Have it globally available so the user can get access to it
2. Having the code know where to flush it to or fully load it from

Having an object in more than one registry makes no sense and if it were attempted, a copy should be made (and in fact *is* for the box).

There are only four lines of code in the codebase which actually set the registry to something and one of them is setting it to `None` on deletion and another is setting the metadata registry. Of the remaining two, one of them is just unnecessary repetition. In summary, there is only one real place where `_setRegistry` is called and that is via `Registry._add()`. There are in turn only four places that `Registry._add()` is called and they are all in sensible places.

rob-c:
Any my question in this case is do we have a test that makes sure these copies are being made correctly?
I have no doubt that the correct behaviour is that the objects _shouldn't_ change repo but I would like to know there is a test against the common place(s) where a bad PR could conceptually trigger this unintended behaviour.

milliams:
I will add a test or two which check this explicitly.

milliams:
I've added a test to try to cover this. If objects stop getting copied when they're assigned to the box for example, the test will fail.

rob-c:
Go for merge from me when ready

PR #451, opened 2016-05-09
Fixing bug in setter on Proxy object construction

rob-c:
This fixes a bug in the Proxy constructor which doesn't respect the functionality which is in the ```...filter__set__``` method.

This is required for the IGangaFile objects to behave how I'm expecting them to and so should be go in as it fixes a bug which is easy to reproduce in that setting an ```.lfn``` of a ```DiracFile``` fails to modify the ```namePattern``` accordingly as would naively be expected.

Yes if we manage to get the filter system to go away this can be reverted, but for now this is a bug and needs to be fixed.

This should also fix #425 although this is just an afterthought and I've done exactly 0% testing of this particular problem.

I'm keen to get this in asap so I can carry on with LSST work.

Fixes #425

PR #457, opened 2016-05-09
Reverting to setNodeAttr as setattr doesn't work for all cases

rob-c:
This is a version of #451 which works for my original problem.

I was running a short test against #451 which seemed to indicate that it was working, however it failed to fix the problem that prompted me to write the PR.

I'll merge after the tests have patched as this isn't anything which hasn't been seen before.

### DIFF
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -788,7 +788,7 @@ def GPIProxyClassFactory(name, pluginclass):
                 val = stripProxy(getattr(self, key))
                 if isinstance(val, GangaObject):
                     val._auto__init__()
-                setattr(instance, key, val)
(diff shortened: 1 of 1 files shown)


milliams:
In my local tests with `setattr(self, key, val)`, it did indeed seem to work. Even testing with `GaudiPython` I see that doing:
```python
gp = GaudiPython(project='LHCb',version='v39r0')
assert gp._impl._data['appname'] == 'LHCb'
```
works correctly. What test or check did you run to see the failure?

milliams:
Ahh, I see what you did. In my suggestion in the other thread I said to do:
```python
setattr(self, key, val)
```
but in your change you did
```python
setattr(instance, key, val)
# and
setattr(raw_self, k, this_arg)
```
which of course is not going to call the proxy setters. You need to call it on `self` so that it goes through the proxy layer correctly.

PR #458, opened 2016-05-10
Node/GangaObject cleanup

milliams:
The primary aim of this PR is to more cleanly delineate what is `Node` and what is `GangaObject`. `Node` should be used only for keeping track of the hierarchical things like parents, `accept()` and the dirty flag. Anything to do with auto-plugins, schemas or proxy-exporting should be the domain of `GangaObject` only.

The other thing that this PR does is replace the `setNodeData`, `setNodeIndexCache`, etc. methods with Pythonic property accesses so that you can just do `obj._data['subfiles']` and it works as expected.

The separation of `Node` and `GangaObject` is needed for the `GangaList` changes I have incoming as in that branch, a `List` is not a `GangaObject` but it *is* a `Node`.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepository.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepository.py
@@ -96,7 +96,7 @@ class GangaRepository(object):
     def update_index(self, id=None):
         """update_index(id = None) --> iterable of ids
         Read the index containing the given ID (or all indices if id is None).
-        Create objects as needed , and set the _index_cache through setNodeIndexCache
+        Create objects as needed , and set the _index_cache
         for all objects that are not fully loaded.
         Returns a list of ids of jobs that changed/removed/added
         Raise RepositoryError
@@ -201,10 +201,7 @@ class GangaRepository(object):
             self._found_classes[compound_name] = cls
         cls = self._found_classes[compound_name]
         obj = cls()
-        #setattr(obj, '_parent', None)
-        #obj.__init__()
-        obj.setNodeData({})
-        obj.setNodeAttribute('id', this_id)
+        obj._data = {}
 
         obj._setFlushed()
         self._internal_setitem__(this_id, obj)
@@ -217,13 +214,12 @@ class GangaRepository(object):
         if this_id in self.incomplete_objects:
             self.incomplete_objects.remove(this_id)
         self.objects[this_id] = obj
-        setattr(obj, "_registry_id", this_id)
-        setattr(obj, "_registry_locked", False)
-        setattr(obj, "_id", this_id)
-        #if obj.getNodeData() and "id" in obj.getNodeData().keys():  # MAGIC id
-        obj.setNodeAttribute('id', this_id)
+        obj._registry_id = this_id
+        obj._registry_locked = False
+        obj._id = this_id
+        if 'id' in obj._schema.allItemNames():
+            obj.setSchemaAttribute('id', this_id)  # Don't set the object as dirty
         obj._setRegistry(self.registry)
-
 
     def _internal_del__(self, id):
         """ Internal function for repository classes to (logically) delete items to the repository."""
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -13,6 +13,7 @@
 #   and followed by
 #    obj._setDirty()
 
+import abc
 import threading
 from contextlib import contextmanager
 import functools
@@ -41,7 +42,7 @@ logger = Ganga.Utility.logging.getLogger(modulename=1)
 
 _imported_GangaList = None
 
-do_not_copy = ['_index_cache', '_parent', '_registry', '_data', '_read_lock', '_write_lock', '_proxyObject']
+do_not_copy = ['_index_cache_dict', '_parent', '_registry', '_data_dict', '_read_lock', '_write_lock', '_proxyObject']
 
 def _getGangaList():
     global _imported_GangaList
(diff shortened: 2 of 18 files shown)


python/Ganga/Core/GangaRepository/GangaRepository.py
```diff
@@ -201,10 +201,7 @@ def _make_empty_object_(self, this_id, category, classname):
             self._found_classes[compound_name] = cls
         cls = self._found_classes[compound_name]
         obj = cls()
-        #setattr(obj, '_parent', None)
-        #obj.__init__()
-        obj.setNodeData({})
-        obj.setNodeAttribute('id', this_id)
+        obj._data = {}
```
drmarkwslater: Do we not need to set the `id` as well?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -64,52 +65,43 @@ def decorated(self, *args, **kwargs):
 
 
 class Node(object):
+    """
+    The Node class is the code of the Gnaga heirachy. It allows objects to keep
```
drmarkwslater: Minor typo :smile:

python/Ganga/Core/GangaRepository/GangaRepository.py
milliams: The `id` is set by the call to `self._internal_setitem__` a few lines below.

drmarkwslater: Ignore me - I see it's done lower down!

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -696,15 +524,16 @@ def cloneVal(v):
...
 
         obj._setDirty()
 
     def __delete__(self, obj):
         """
         Delete an attribute from teh Descriptor(?) and Node
         """
-        obj.removeNodeAttribute(_getName(self))
+        raise Exception('Descriptor.__delete__')
+        del obj._data[_getName(self)]
```
drmarkwslater: Won't the `Exception` mean that this will always fail and the `del` isn't called?

drmarkwslater:
Looks good to me 👍

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -41,7 +42,7 @@ def _getName(obj):
 
 _imported_GangaList = None
 
-do_not_copy = ['_index_cache', '_parent', '_registry', '_data', '_read_lock', '_write_lock', '_proxyObject']
+do_not_copy = ['_index_cache', '_index_cache_dict', '_parent', '_registry', '_data', '_data_dict', '_read_lock', '_write_lock', '_proxyObject']
```
rob-c: What's the difference between the cache and the dict?

python/Ganga/GPIDev/Base/Objects.py
milliams: Ah, bugger. I put that exception in as a local test to see if `__delete__` is ever even called as I don't think that it should be. I didn't mean to commit it. I will remove that line but in the future I'd like to remove the method entirely.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -167,130 +162,27 @@ def _getRoot(self, cond=None):
...
 
     def printTree(self, f=None, sel=''):
         from Ganga.GPIDev.Base.VPrinter import VPrinter
         self.accept(VPrinter(f, sel))
 
-    #printPrepTree is only ever run on applications, from within IPrepareApp.py
-    #if you (manually) try to run printPrepTree on anything other than an application, it will not work as expected
-    #see the relevant code in VPrinter to understand why
-    def printPrepTree(self, f=None, sel='preparable' ):
-        ## After fixing some bugs we are left with incompatible job hashes. This should be addressd before removing
```
rob-c: ```incompatible job hashes. This should be addressd before removing```
Changing this could (and likely will) trigger all LHCb (and other I assume) prepared jobs loaded from disk to be unprepared and to throw warnings from the XML Repo

python/Ganga/GPIDev/Base/Objects.py
milliams: `_index_cache` is the descriptor and `_index_cache_dict` is the actual Python dict underlying it. Similarly for the `_data`.

The usual pattern here is for publicly accessible object, you would have an attribute called `foo` with some underlying data, perhaps stored in a class variable called `_foo` which would be accessed by:
```python
@property
def foo(self):
    return self._foo
```
Since `_data` and `_index_cache` are already semi-internal, a different name was needed.

I'm not sure that these are all needed in the `do_not_copy` though and in fact I'd very much like to get rid of it sometime soon.

python/Ganga/GPIDev/Base/Objects.py
milliams: I was careful (or at least tried to be) to keep the current behaviour. `IPrepareApp` still calls `printPrepTree` and it still has the same output. All I have done here is move that method to the `IPrepareApp` class where it should be.

python/Ganga/GPIDev/Base/Objects.py
rob-c: as far as I can tell ```do_not_copy``` will need to exist for both a) the proxy layer and b) ```_parent``` not to get into infinite copy loops. As for the object vs dictionary does deep copy actually see setters and such or just the content of ```__dict__```?

python/Ganga/GPIDev/Base/Objects.py
rob-c: But now we're still left with ```printSummaryTree``` shouldn't the IPrepareApp overload this in some way rather than have its own summary tree method and we can just replace all ```printPrepTree``` back to ```printSummaryTree```?

python/Ganga/GPIDev/Base/Objects.py
milliams: Well the point of the `memo` argument in `__deepcopy__` is exactly to prevent infinite loops. Until now we weren't using is properly but this PR adds the items correctly (I think).

I had assumed that `__dict__` would contain both the getter and the dict but now I think about it, the setter is a property of the class, not the instance. This means that I can remove `_data` and `_index_cache` from the `do_not_copy`. I will do so.

python/Ganga/GPIDev/Base/Objects.py
milliams: That's a good idea, I hadn't thought of that. I will probably leave that for another PR though, just to keep this one small?

rob-c: OK, can you add an issue if there is a chance of forgetting this :+1:

python/Ganga/GPIDev/Base/Objects.py
rob-c: Only word of caution was copying large gangalists of diracfiles (aka the book-keeping perfomance bug in LHCb) used to show that the repr of a gangalist of 5k files was ending up in the memo object which caused a bottleneck. I don't remember what memo is supposed to be off the top of my head

python/Ganga/GPIDev/Base/Objects.py
milliams: Will do.

python/Ganga/GPIDev/Base/Objects.py
milliams: Ok, I will bear that in mind, thanks.

The `memo` dict is used during the `deepcopy` operation to cache any objects which have already been deep copied. It ends up being a mapping of object id (in the Python sense) to the deep-copy of the object. The idea is that if `deepcopy` is called on an object and that object is already in the `memo` dict then it does not recurse and instead just returns the object, thereby preventing infinite loops.

rob-c: I just got put off when I printed it and it contained a string rep of the full contents of a huge list rather than just some unique name of the list I forget how this works behind the scenes but I think form of string rep of objects is taken during the copy process.
Back to the original point I think you're correct and can drop ```_index_cache``` from the list probably as well as ```_data``` but the tests will tell us if it's safe

rob-c:
Looks good to me in it's current state :+1:

PR #470, opened 2016-05-11
Patch from HC team to use a local JSON for site mapping + minor homepackage fix

drmarkwslater:
This PR contains some fixes for Hammercloud Atlas. It probably won't go in **exactly** as is but I wanted to set up the PR to allow discussion with the HC team about the changes.


python/GangaAtlas/Lib/Athena/DQ2JobSplitter.py
```diff
@@ -599,6 +601,7 @@ def split(self,job):
             allowed_sites.append('PRAGUELCG2-RUCIOTEST_DATADISK')
 
         logger.debug('allowed_sites = %s ', allowed_sites)
+        logger.warning('allowed_sites = %s ', allowed_sites)
```
drmarkwslater: A general comment on the additional logger messages: do you want these kept in? It doesn't really matter as HC are the only ones using Panda/DQ2JobSplitter but I thought it might make things a bit noisy!

python/GangaPanda/Lib/Panda/Panda.py
```diff
@@ -52,6 +53,19 @@
...
 
+##### fix by jschovan, use outputs of AGIS collector
+def load_json(filename):
+    result=None
+    f=open(filename, 'r')
+    result=json.loads(f.read())
+    f.close()
+    return result
+
+pandaresources_json = '/tmp/agis_pandaresources.json'
```
drmarkwslater: Do you want this as a config option in Ganga or hardcoded to this filename?

drmarkwslater: Also, is it worth trying to grab this info from within Ganga or stick with it being created externally?

python/GangaPanda/Lib/Panda/Panda.py
```diff
@@ -134,6 +148,16 @@ def convertDQ2NamesToQueueName(locations):
...
+    ret=[pandaresources[x]['ddm'] for x in pandaresources if pandaresources[x]['name']==queue]
+    # logger.warning('(%s) ret:%s:' % (type(ret), ret))
+    try:
+        ret=ret[0]
+    except:
+        # logger.warning('oh well')
+        import traceback
+        logger.error('queue %s error %s'%(queue, str(traceback.print_exc() )))
+    return str(ret).split(',')
+
```
drmarkwslater: Is the previous way of converting Queue name -> DDM name now completely dead? i.e. is the code following this based around:
```
    for site in Client.PandaSites[queue]['setokens'].values():
        sites.append(Client.convSrmV2ID(site))
```
broken now? I ask because if it is I'll just remove it now :smile:

python/GangaAtlas/Lib/Athena/DQ2JobSplitter.py
jschovan: Hi Mark, 
I needed that for debugging (without increasing debug verbosity elsewhere). I guess it can stay as `logger.debug(...)`. 
Thank you, Jarka

python/GangaPanda/Lib/Panda/Panda.py
jschovan: If you make it configurable, perfect. Please let me know how to update HC configs for Ganga then.

jschovan: Let's take this info externally. I hope it will save you some work ;)

python/GangaPanda/Lib/Panda/Panda.py
jschovan: Something around `convertQueueNameToDQ2Names` stopped working, that was the main reason for the intervention and patch. In HC we are very actively moving far far away from getting topology & status information from combination of sources (PanDA client vs. ToAcache vs. AGIS) and are taking this info from AGIS directly. For that we have collector that produces the json  `+pandaresources_json = '/tmp/agis_pandaresources.json' `. If Ganga can move in the similar direction, everybody wins.

drmarkwslater:
@jschovan I've made a couple of cosmetic changes to the code just to make it a bit more readable (for my eyes anyway 😄 ), make sure the JSON load doesn't happen on import (this would result in a lot of errors for normal users!) and finally added a config option to set the file:
```
[Panda]
AGISJSONFile = /tmp/agis_pandaresources.json
```

Could you possible send me an appropriate JSON file to check these are working as expected? Thanks!

milliams:
retest this please

milliams:
I'm pretty much :+1: on this. I'm not sure about the use of a function-local variable for the caching. It's better than a module-global variable but I was trying to think of a simple better way. I guess having a decorator which performed this sort of thing would wrap things nicely but I'm not convinced it's better than what you have already.

drmarkwslater:
Do we have a 'static variable' decorator available in Ganga/python? That would be quite a nice convenient as the possibly the 3-4th function variable I've added in the last few months...

milliams:
We don't but it would be as simple as something like:
```python
def static_variable(name, default_value=None):
    def wrap(f):
        setattr(f, name, default_value)
        return f
    return wrap

@static_variable('pandaresources', {})
def convertQueueNameToDQ2Names(queue):
    print(convertQueueNameToDQ2Names.pandaresources)
    convertQueueNameToDQ2Names.pandaresources[4] = 2

>>> convertQueueNameToDQ2Names(43)
{}
>>> convertQueueNameToDQ2Names(43)
{4: 2}

```

drmarkwslater:
That's what I was thinking as well - Shall I put that in this PR or save it for another one?

milliams:
I would save it for another PR to make sure we can do it properly. It's not hurting anything in its current state.

Issue #343, opened 2016-03-30
LSST Im3Shape support in Ganga


rob-c:
This issue is to hold all discussions relating to adding/improving Im3Shape support within Ganga.

Apologies this was originally a long email but I decided it's probably best put here with some formatting to have something to point back to when I push for the lsst branch to be merged in a few days.
## Intro

After meeting with @joezuntz before Easter I think I've got a clear idea of what is required for better LSST support within Ganga.
The `im3shape` executable is to be run over a large set of data (ultimately from lsst) with the amount of ways in which the data is to be processed limited by available CPU power available.

I'm trying to add a Ganga application which is capable of reproducing the behavior of https://github.com/joezuntz/grid-wl/commit/44459f3b258caf691076a398084ef4f589a1915a#diff-29c7845891b998a549bcaa7555c6adff whilst integrating more with the features offered by Ganga.
## Job requirements

A single instance of `im3shape` runs over a single input file and for a given `rank` and `size` will access a subset of the data available in this file.
### Binary distribution

In order to run the `im3shape` the application has to be made available on the worker node where it is to start running in addition to certain other pieces of information.
The best distribution method for this (at the moment) is via a DiracFile on the grid. This may change to be an installation on `cvmfs` when running over the whole dataset in the future.

The job script on the worker node has to be capable of downloading the output of `im3shape` binary (currently part of a tarball), and then run over the appropriate data file with the correct arguments.

Most of this will be checked and run by executing an `lsst_execute.py` script on the worker node which will focus on executing the following command with the following arguments.
### im3shape Configurables:

quoting https://github.com/joezuntz/grid-wl/commit/44459f3b258caf691076a398084ef4f589a1915a#diff-29c7845891b998a549bcaa7555c6adff:

im3shape is run using the following line

```
run-im3shape $DATA_LOCAL_PATH $INI $CAT $OUTPUT_LOCAL_PATH $JOB_RANK $JOB_COUNT
```

Where:
#### `$DATA_LOCAL_PATH`

Location of input file to be processed
#### `$INI` ini file

The `ini` file which configures how `im3shape` is to run. This is currently not changed very often and so can be stored as a DiracFile by default but may be changed by the user into something such as a LocalFile.
#### `$CAT`catalog

This defaults to `all` for the moment but in principle can be a file which is to be passed along with the job so can be configured to accept a LocalFile object.
#### `$OUTPUT_LOCAL_PATH`

Location of the output file
#### `$JOB_RANK`

[...]

PR #471, opened 2016-05-11
GangaLSST plugin

rob-c:
This branch represents the first semi tested working version of the GangaLSST plugin I've been developing.

This PR contains code changes in the Core and in GangaDirac. I plan to break these out into other PR(s) to discuss changes relating to these there. I only plan to try to merge this branch once the other PR are in develop and this branch only contains changes within the new GangaLSST directory.

I don't want to bog down this PR with discussions of the changes in GangaCore/GangaDIRAC.

An example configuration to make use of this branch is in ```LSST.ini``` Once you've checked out the code make sure you're at least modifying the ```RUNTIME_PATH``` in your config to make sure Ganga loads this plugin on startup.

An example configuration for using the application is given below:

```python
j=Job()
app = Im3ShapeApp(
    im3_location=DiracFile(lfn='/lsst/y1a1-v2-z/software/2016-02-24/im3shape-grid.tar.gz'),
    ini_location=LocalFile('/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/params_disc.ini'),
    blacklist=LocalFile('/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/blacklist-y1.txt')
)
j.application = app
j.backend=Dirac()
mydata=GangaDataset()
mydata.append(DiracFile(lfn='/lsst/DES0005+0043-z-meds-y1a1-gamma.fits.fz'))
j.inputdata = mydata
j.splitter=Im3ShapeSplitter(size=20)
j.outputfiles = [DiracFile('*.main.txt'), DiracFile('*.epoch.txt')]
j.submit()
```
This example will create 20 subjobs with different ranks which will iterate over the ```DES0005+0043-z-meds-y1a1-gamma.fits.fz``` file.
If you include 5 files in the dataset you will get 100 subjobs and so on.

I haven't completely tested the returning of data from these jobs yet as my test jobs are still waiting on some DIRAC test jobs to run.

Also I've been making a few light modifications today so there may be some papercuts in this branch I'm unaware of.
That being said the above code should submit a sane JDL with a sensible looking script to be executed on the WN.

This addresses #343.


rob-c:
OK, I think I'm seeing jobs running but there is still a small bug in returning the output to Ganga. I'll do some quick digging but hopefully this shouldn't take too long to fix.

rob-c:
OK, I appear to be stuck in some full queues again so I'll do my best to debug using a ```gridpp_user``` vo. I know I can access the data and execute the program, so the only thing left to do is check that the output data is collected as expected.

egede:
@rob-c Tested against the Local backend?

rob-c:
@edege not yet implemented as I need to write a dedicated RTHandler for it. (I was keen to get running on Dirac first)

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+    _schema = Schema(Version(1, 0), {
+        ## Required to configure Im3ShapeApp
+        'im3_location': GangaFileItem(defvalue=None, doc="Location of the Im3Shape program tarball"),
+        'exe_name': SimpleItem(defvalue='run-im3shape', doc="Name of the im3shape binary"),
+        'ini_location': GangaFileItem(defvalue=None, doc=".ini file used to configure Im3Shape"),
+        'blacklist': GangaFileItem(defvalue=None, doc="Blacklist file for running Im3Shape"),
+        'rank': SimpleItem(defvalue=0, doc="Rank in the split of the tile from splitting"),
+        'size': SimpleItem(defvalue=200, doc="Size of the splitting of the tile from splitting"),
+        'catalog': SimpleItem(defvalue='all', types=[IGangaFile, str], doc="Catalog which is used to describe what is processed"),
```
milliams: A `SimpleItem` shouldn't be allowed to take an `IGangaFile` It will not be safely serialised.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
```diff
@@ -0,0 +1,117 @@
...
+            force (bool): Should force the prepare step to run
+        """
+
+        if (self.is_prepared is not None) and (force is not True):
+            raise ApplicationPrepareError('%s application has already been prepared. Use prepare(force=True) to prepare again.' % getName(self))
+
+        # lets use the same criteria as the configure() method for checking file existence & sanity
+        # this will bail us out of prepare if there's somthing odd with the job config - like the executable
+        # file is unspecified, has a space or is a relative path
+        self.configure(self)
```
milliams: `self.configure(self)` seems like a strange idiom.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
```diff
@@ -0,0 +1,117 @@
...
+        logger.info('Preparing %s application.' % getName(self))
+        setattr(self, 'is_prepared', ShareDir())
```
milliams: Should be:
```python
self.is_prepared = ShareDir()
```

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
```diff
@@ -0,0 +1,117 @@
...
+            self.checkPreparedHasParent(self)
+            for file_ in [self.ini_location, self.im3_location, self.blacklist]:
+                if isinstance(file_, LocalFile):
+                    self.copyIntoPrepDir(file_.namePattern)
+                assert type(file_) in [LocalFile, DiracFile]
+            # return
+            # [os.path.join(self.is_prepared.name,os.path.basename(send_to_sharedir))]
+            self.post_prepare()
+
+        except Exception as err:
```
milliams: We shouldn't be catching all exceptions here if we can avoid it. `Exception` is dangerous as it includes `NameError` etc. which hide typos.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
```diff
@@ -0,0 +1,117 @@
...
+            logger.debug("Err: %s" % str(err))
+            self.unprepare()
+            raise err
```
milliams: Simply `raise` is better.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeRTHandler.py
```diff
@@ -0,0 +1,211 @@
...
+            appmasterconfig (tuple): This is also used to prepare the inputsandbox but contains the config of the app for the master job
+            jobmasterconfig (StandardJobConfig): This is the configuration of the master job which may or may not be the same job as owning the app
+        """
+
+        # Construct some common objects used in job submission here
+        inputsandbox, outputsandbox = sandbox_prepare(app, appsubconfig, appmasterconfig, jobmasterconfig)
+        input_data,   parametricinput_data = dirac_inputdata(app, hasOtherInputData=True)
+
+
+        job = stripProxy(app).getJobObject()
```
milliams: Shouldn't need the `stripProxy` any more.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeRTHandler.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # Construct the im3shape-script which is used by this job. i.e. the script and full command line to be used in this job
+        exe_script_name = 'im3shape-script.py'
+        output_filename = os.path.basename(job.inputdata[0].lfn) + '.' + str(app.rank) + '.' + str(app.size)
```
milliams: For simplicity, could be 
```python
output_filename = '.'.join([os.path.basename(job.inputdata[0].lfn), str(app.rank), str(app.size)])
```

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
rob-c: Copied from the Executable Dirac RT

rob-c: Which probably means I can remove it but for now I'm a little weary

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
rob-c: Copied form elsewhere as before so I'll ammend both as the latter is bad too

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
rob-c: Again, copied from another RT

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
rob-c: *copied from Executable which does bad things, frankly I'm just going to keep catching and raising I just need to have a finally unprepare here on error.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeRTHandler.py
rob-c: Yes, but it's a harmless copy

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeRTHandler.py
rob-c: Yes but it may also change in the future I suspect

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
rob-c: Then how can I have something which may be equally a file or a string?

rob-c: Tbh I'm tempted to just store a string here and if it's ever a file just check in the RTH that the file matches a namePattern in the inputfiles

milliams: There's basically no safe way in Ganga to have something which can be either a `GangaObject` or not a `GangaObject`. If it's `repr` round-tripable then it will probably work but there could be unforeseen consequences.

rob-c:
OK, the most recent test ran, uploaded the a job output and the output appeared to be the same as what I'm expecting locally. I'll finish cleaning up this PR by addressing the work in the other PR which have spawned from this, then I'll try and merge this as part of 6.1.20

rob-c:
I'm going to make the bold claim that this plugin is working so I'm keen to merge with 6.1.20 if there are no objections?
I'll work on a Local RTHandler for testing/dev in a side branch once this is out and will make a PR for that in 6.1.21.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeRTHandler.py
milliams: We still shouldn't run code that doesn't need to be.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeApp.py
```diff
@@ -0,0 +1,114 @@
...
+        'im3_location': GangaFileItem(defvalue=None, doc="Location of the Im3Shape program tarball"),
+        'exe_name': SimpleItem(defvalue='run-im3shape', doc="Name of the im3shape binary"),
+        'ini_location': GangaFileItem(defvalue=None, doc=".ini file used to configure Im3Shape"),
+        'blacklist': GangaFileItem(defvalue=None, doc="Blacklist file for running Im3Shape"),
+        'rank': SimpleItem(defvalue=0, doc="Rank in the split of the tile from splitting"),
+        'size': SimpleItem(defvalue=200, doc="Size of the splitting of the tile from splitting"),
+        'catalog': SimpleItem(defvalue='all', types=[str], doc="Catalog which is used to describe what is processed"),
+        'run_dir': SimpleItem(defvalue='im3shape-grid', types=[str], doc="Directory on the WN where the binary is"),
+        ## Below is needed for prepared state stuff
+        'is_prepared': SimpleItem(defvalue=None, strict_sequence=0, visitable=1, copyable=1, hidden=0, typelist=[None, bool, ShareDir], protected=0, comparable=1, doc='Location of shared resources. Presence of this attribute implies the application has been prepared.'),
```
milliams: Since we're in control of this app and we can decide its semantics, I feel this should be set to `typelist=[None, ShareDir]`. Especially since you do things like `self.is_prepared.name` below.

milliams:
I would have very much liked to see a set of tests being submitted along with such a large new feature as this. Due to @drmarkwslater's recent work on GangaLHCb, we can now run tests on plugins so some simple tests, even just to make sure that the modules can be imported are important.

python/GangaLSST/Lib/Im3ShapeApp/Im3ShapeRTHandler.py
rob-c: Yes but frankly this is just a nullop until it's removed... which it was in a commit which is now missing... the joys of git the history

rob-c:
OK, there is now a basic test to check that importing everything works.

Are there any objections to merge?

milliams:
Since this basically touches nothing outside of the `GangaLSST` directory, I'm happy for you to merge once the tests pass :+1:

rob-c:
@milliams Is there some trick to ensuring Ganga has loaded the correct plugins during testing?

milliams:
It should base it on the `RUNTIME_PATH` in `LSST.ini`. You need to add a `python/GangaLSST/testlib/setup_env.sh` file (like the `GangaLHCb` one) containing something like:
```bash
export GANGA_CONFIG_PATH=GangaLSST/LSST.ini
```
Does it run for you locally after that with:
```bash
$ bash -ic "source python/GangaLSST/testlib/setup_env.sh && py.test python/GangaLSST/test/"
```

rob-c:
OK, this should work now

milliams:
I looks like the LSST test passed. Great :smile:

rob-c:
and no we wait for it to re-run

PR #473, opened 2016-05-11
Bugfix/ganga dirac base changes

rob-c:
This is a small bugfix which prevents the DIRAC monitoring in Ganga from re-adding a queued job to the list of jobs being finalized.

Any objections to merging this?

Fixes #438

PR #474, opened 2016-05-11
Feature/dirac filefeature

rob-c:
This is a small addendum to DiracFile to test for LFNs being passed to the namePattern by accident and swapping them in the setter code accordingly.

This is based on #473 (was needed in testing) until it's merged then it will just touch DiracFile.

### DIFF
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -174,8 +174,7 @@ class DiracFile(IGangaFile):
         if value != "" and value is not None:
             #   Do some checking of the filenames in a subprocess
             if name == 'lfn':
-                if self.namePattern == '':
-                    self.namePattern = os.path.basename(value)
+                self.namePattern = os.path.basename(value)
                 self.remoteDir = os.path.dirname(value)
                 return value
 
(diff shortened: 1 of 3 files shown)


python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -174,8 +174,7 @@ def _attribute_filter__set__(self, name, value):
         if value != "" and value is not None:
             #   Do some checking of the filenames in a subprocess
             if name == 'lfn':
-                if self.namePattern == '':
-                    self.namePattern = os.path.basename(value)
+                self.namePattern = os.path.basename(value)
```
drmarkwslater: I'm probably forgetting how `namePattern` is used, but just overwriting it with LFN, won't this stomp over wildcards such as:
```
DiracFile.namePattern = '*.root'
DiracFile.lfn = 'hello.root'
```

I guess this may not not be a bad thing but won't it break copies of the `DiracFile` preserving the name pattern used for matching?

rob-c: yes but you have just requested that the DiracFile have the explicit LFN so you're changing the object implicitly I would argue.
The subfiles of the wildcard all work as expected but if you have ```df=DiracFile(namePattern='foo');``` ```df.lfn=bar``` you would naively assume that ```df.namePattern``` would then be ```basename('bar')```?

I think there isn't a strict matching between LFN and namePattern, but this just makes life difficult if you're uploading a local file to an LFN with a different name.

drmarkwslater: To be honest, this is probably more my lack of understanding of how `namePattern` should be used as I've always found it confusing and possibly contradictory in the code. If things are work then I'm happy 😄

rob-c: From my understanding the pattern in namePattern (if there is one) should then have multiple subfiles 1 for each match in order to do anything sensible. Then the IGangaFile acts as a sort of list which can be iterated (in some way).
If there is an IGangaFile without a wildcard (or equivalent) in the outputfiles then this is directly updated with the relevant file info after a job finishes.

If we ever support wildcards for DiracFile in the inputfiles I assume this would effectively be the same thing that we would have 1 file with the globbable string and then a set of subfiles where the actual data is.

rob-c: ...I'm working on documentation for this...

PR #477, opened 2016-05-11
Bugfix/fixing rth utils

rob-c:
This PR changes some of the code in the RTHUtils to work with a generic Dataset object which has DiracFiles within it.

I would like to try and reduce use of LHCbDataset functions everywhere apart from LHCb in future so that we can migrate most of the code/functionality out of this class and into GangaDataset so that it can  be significantly reduced and the GangaDataset object can be used fully on vanilla Dirac.

This hasn't been fully tested with the old bulk_submit but as this has been un-supported for a __long__ time I don't want to delay putting in fixes which are needed for other work.

### DIFF
--- a/python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
+++ b/python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
@@ -16,7 +16,7 @@ def mangle_job_name(app):
     Args:
         app (IApplication): This is the application belonging to the job of interest
     """
-    job = app.getJobObject()
+    job = appt.getJobObject()
 
     jobName = job.name
     jobIndex = job.getStringFQID()
@@ -118,8 +118,6 @@ def dirac_inputdata(app, hasOtherInputData=False):
         wanted_job = job.master
 
     inputLFNs = ['LFN:'+this_file.lfn for this_file in wanted_job.inputdata if isinstance(this_file, DiracFile)]
-
-    logger.info("LFNS: %s" % inputLFNs)
 
     # master job with a splitter reaching prepare, hence bulk submit
     if not job.master and job.splitter:
--- a/python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
+++ b/python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
@@ -12,7 +12,10 @@ logger = getLogger()
 
 
 def mangle_job_name(app):
-    """ Create a safe job name to send to DIRAC """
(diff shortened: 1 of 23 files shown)


python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
```diff
@@ -87,48 +96,42 @@ def dirac_outputfile_jdl(output_files, empty_SE_check):
 
 
 def dirac_inputdata(app, hasOtherInputData=False):
-    """ Construct the JDL compoenent which requests the inputdata for a job """
-    job = stripProxy(app).getJobObject()
+    """ Construct the JDL compoenent which requests the inputdata for a job
```
milliams: Typo in "compoenent"

python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
```diff
@@ -87,48 +96,42 @@ def dirac_outputfile_jdl(output_files, empty_SE_check):
...
-                raise BackendError('Dirac', 'Number of bulk submission jobs \'%s\' exceeds the maximum allowed \'%s\' if more are needed please modify your config. Note there is a hard limit in Dirac of currently 1000.' % (
-                    len(parametricinput_data), getConfig('DIRAC')['MaxDiracBulkJobs']))
+    if not job.inputdata and not job.master.inputdata:
+        return input_data, parametricinput_data
+
+    wanted_job = job
+    if not job.inputdata and job.master.inputdata:
+        wanted_job = job.master
+
+    inputLFNs = ['LFN:'+this_file.lfn for this_file in wanted_job.inputdata if isType(this_file, DiracFile)]
```
milliams: Could we use `isinstance` here?

milliams:
One or two minor comments made above but essentially :+1:

python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
rob-c: No, No, No and doing so breaks things for some awful reason.
I think this is because the Dataset is returning a proxied object from a dataset

milliams: If a proxied object is being returned then that is something that must be fixed.

rob-c: yes but not in this PR

rob-c:
This can now make use of isinstance over isType... any objections to merge?

PR #478, opened 2016-05-11
Adding Master index flush to repo flush

### LINKED ISSUE #461: master index not being updated with auto-flush
The master index was originally only intended to be updated at shutdown or once every 5min or so (I think) given the large amount of I/O.

Given we've fixed a large amount of these I/O issues adding an update of the master index cache to the auto flusher seems sensible otherwise users can potentially have a Ganga session crash and then their job will appear in the incorrect ```new``` state.

I see this when I'm doing work over a bad connection and it's becoming an irritating bug to not know the true status of my jobs.

This shouldn't require too much work, just probably cleaning away the old rate limiting code in the master index writing method and then to call it directly from some flush method (probably best done from the Registry to avoid multiple writes of the same data to disk)

rob-c:
Fixes #461.

This is a relatively minor change which auto-flushes the master index as well as all of the other data with each xml flush that is called.

This is not the most 100% optimal way to fix this but it fixes it well enough and with only a slight overhead so I'm keen to see it get merged so that the master_index stops getting out of sync with everything else on disk.


rob-c:
hmmm, either this is causing deadlocks in the Tasks tests or the tests are now taking substantially longer as a result. I'll investigate further tomorrow

rob-c:
retest this please

rob-c:
retest this please

rob-c:
Try as I might I can't get this PR to test? @milliams am I missing something?

milliams:
@rob-c I will look into it.

milliams:
retest this please

rob-c:
@milliams What exactly are the implications, are the tests running or is it just some reporting of tests running/failing that is broken?

milliams:
The tests are running but the reporting back to GitHub about the success/failure is broken. I have tried a switch to re-enable the old functionality so we'll see how that helps.

milliams:
retest this please

milliams:
Ok, seems to be working again. Poke me if you see any weirdness.

rob-c:
@milliams will do

rob-c:
OK, there seems to be a definite performance hit when writing the master index with every flush command.
I'll see if I can add some logic to determine if it's changed and only on changes write it to disk. This should only add CPU overhead and may reduce the I/O (maybe)

rob-c:
OK, this PR took 55min to pass the tests, there has to be a better solution than this one... I think I'll roll this back to 6.1.21 as I don't have the time today to fix this.

rob-c:
Superseded by #497

PR #479, opened 2016-05-11
Feature/dirac parse env and rc files

rob-c:
This PR allows for the DIRAC tools to get the environment from running the same as ```source envfile``` so we can use the bashrc from the dirac-ui directly as ```[DIRAC]DiracEnvFile```.

This should be working and makes use of the same function which existed elsewhere for getting the environment after a command has been executed.

This would be very nice to see in 6.1.20 as it simplifies setting up Ganga+dirac-ui.

### DIFF
--- a/python/Ganga/Utility/Shell.py
+++ b/python/Ganga/Utility/Shell.py
@@ -92,33 +92,29 @@ def get_env_from_arg(this_arg, def_env_on_fail = True):
     output = pipe.communicate(command)
     rc = pipe.poll()
 
+    # When good output = ( "some_env_python_dict_repr", None )
+    # When bad = ( stdout, stderr )
+
     if rc:
+
         logger.warning('Unexpected rc %d from setup command %s', rc, setup)
 
         try:
-            env2 = expand_vars(eval(eval(str(output))[0]))
+            env2 = expand_vars(eval(output[0]))
         except Exception as err:
-            logger.debug("Err: %s" % str(err))
+            logger.debug("Err: %s" % err)
             env2 = None
             logger.error("Cannot construct environ:\n%s" % str(output))
             try:
-                logger.error("eval: %s" % str(eval(str(output))[0]))
+                logger.error("eval: %s" % str(output))
             except Exception as err2:
-                logger.debug("Err2: %s" % str(err2))
-                pass
-            try:
-                logger.error("eval(eval): %s" % eval(eval(str(output))[0]))
-            except Exception as err3:
-                logger.debug("Err3: %s" % str(err3))
-                pass
+                logger.debug("Err2: %s" % err2)
 
         if env2:
             env = env2
     else:
-        try:
-            env2 = expand_vars(eval(eval(str(output))[0]))
-        except Exception as err:
-            raise
+        
+        env2 = expand_vars(eval(output[0]))
 
         if env2:
             env = env2
@@ -274,7 +270,6 @@ class Shell(object):
             if err.errno != errno.ENOENT:
                 logger.debug("Err removing shell output: %s" % str(err))
                 raise err
-            pass
 
         return rc, output, m
 
--- a/python/Ganga/Utility/Shell.py
+++ b/python/Ganga/Utility/Shell.py
@@ -64,6 +64,69 @@ def expand_vars(env):
     return tmp_dict
 
 
+def get_env_from_arg(this_arg, def_env_on_fail = True):
+    """
+    Function to return the environment after running the command "this_command"
+    Args:
+        this_arg (str): This file/command is the command which is to be run after source to alter an environment
+        def_env_on_fail (bool): True means the default environment is to be returned on failure (i.e. I expect a dict of environ), False is return None on failure
+    """
+
+    if def_env_on_fail:
+        env = dict(os.environ)
+        env = expand_vars(env)
+    else:
+        env = None
+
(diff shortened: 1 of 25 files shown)


milliams:
If I understand correctly, this is allowing us to take a DIRAC `bashrc` and extract from it an environment by sourcing it and grabbing the `env`? Do we even need to do this for DIRAC at all? Could it not be that instead of running:
```python
''';python -c ""import base64;exec(base64.b64decode('%s'))"''' % base64.b64encode(command_update)
```
we instead just effectively run
```python
'''source bashrc ;python -c "import base64;exec(base64.b64decode('%s'))"''' % base64.b64encode(command_update)
```
Is there a reason to have a copy of the DIRAC environment stored in memory as a Python `dict`?

egede:
@milliams At least in LHCb, the sourcing of the environment setup can be slow (it goes through a long dependency list of files that reads files that ...). For this reason, caching the environment is a good thing.

rob-c:
We have a generic tool for running python commands in a given env and I'm reluctant to specialize this tool or change the commands to source a file every time is possible. Besides this changes more significantly the way we use DIRAC and I'd rather avoid the change.

rob-c:
I think we're talking cross-points here. Generating the env and sourcing it each command are 2 different things. The first is extremely expensive and the second is a change in how we call python in a DIRAC context which whilst is easy involves touching things which I'd rather not touch.

python/Ganga/Utility/Shell.py
```diff
@@ -64,6 +71,69 @@ def expand_vars(env):
...
+    logger.debug('Running:   %s' % command)
+    pipe = subprocess.Popen('bash', env=env, cwd=this_cwd, shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    output = pipe.communicate(command)
+    rc = pipe.poll()
+
+    if rc:
+        logger.warning('Unexpected rc %d from setup command %s', rc, setup)
+
+        try:
+            env2 = expand_vars(eval(eval(str(output))[0]))
```
milliams: There's a lot of unexplained double `eval`ing in this function. This will need some more explanation.

python/Ganga/Utility/Shell.py
```diff
@@ -64,6 +71,69 @@ def expand_vars(env):
...
+        except Exception as err:
+            logger.debug("Err: %s" % str(err))
+            env2 = None
+            logger.error("Cannot construct environ:\n%s" % str(output))
+            try:
+                logger.error("eval: %s" % str(eval(str(output))[0]))
+            except Exception as err2:
+                logger.debug("Err2: %s" % str(err2))
+                pass
```
milliams: This `pass` is unneeded as well as the one below.

python/Ganga/Utility/Shell.py
```diff
@@ -64,6 +71,69 @@ def expand_vars(env):
...
+                logger.debug("Err3: %s" % str(err3))
+                pass
+
+        if env2:
+            env = env2
+    else:
+        try:
+            env2 = expand_vars(eval(eval(str(output))[0]))
+        except Exception as err:
+            raise
```
milliams: These 4 lines are equivalent to just ` env2 = expand_vars(eval(eval(str(output))[0]))`.

python/Ganga/Utility/Shell.py
```diff
@@ -64,6 +71,69 @@ def expand_vars(env):
...
+
+        if env2:
+            env = env2
```
milliams: If the code got to here then `env2` must be defined.

python/Ganga/Utility/Shell.py
rob-c: True, this was this way from testing

python/Ganga/Utility/Shell.py
rob-c: Yes but it can be empty or none. The expected behavior of the code as it's used is to not return this in some cases.

python/Ganga/Utility/Shell.py
rob-c: copied from some ancient code. I'll add docs if/when it's fully understood. I suspect this particular section ```if rc``` is broken but it was used in the past.

milliams: Ahh, I'd missed that it was copied.

rob-c: I'm planing to add a line saying just what this is but I don't think I'm the original author... although I did touch this a while ago...

milliams: It looks like you added the `try`/`except`s in c88324a7d2c42a83e58fd and had in fact previously added the double `eval` in 646be66753d568 so this is almost all your code.

rob-c: hmmm, there was a reason for this... either way, fix incoming

rob-c: This is almost certainly being (ab)used by one of the plugins in the past, I remember I never understood why this worked and fixed things but I could certainly demonstrate that it did... I've rewritten this few lines to something sane now with 2 lines of docs.

rob-c:
Any objections to merging this branch as part of 6.1.20?

milliams:
This looks like sensible cleanup along with added documentation :+1:

rob-c:
retest this please

PR #481, opened 2016-05-12
Stop the Datasets from adding proxies which aren't needed

### LINKED ISSUE #480: Dataset returning proxied objects
As @rob-c says in https://github.com/ganga-devs/ganga/pull/477/files/00cb2eae43ff1e5435d82b14425cdc26b90eac16#diff-cc8e6cb5a8320edcf320ba229bf84e2fR120, `Dataset` is returning proxied objects. This needs investigating as there should be nowhere that does this now.

rob-c:
I think this fixes #480, either way this is now not the correct behaviour of the code.

I'd like to merge this, then test #477 to see if this fixes it.


rob-c:
retest this please

rob-c:
retest this please

PR #485, opened 2016-05-12
Speed up tests

milliams:
This makes various changes to speed up the test execution. We'll see how much of a difference it makes when it runs on ganga-ci. Nothing here touches any runtime code except the single small change to `IBackend`.

The majority of the speed-up is due to a change from `sleep_until_completed` to `run_until_completed`.

On a few tests I have tweaked the number of 'iterations' to reduce it to a better level.

### DIFF
--- a/python/Ganga/GPIDev/Adapters/IBackend.py
+++ b/python/Ganga/GPIDev/Adapters/IBackend.py
@@ -410,8 +410,8 @@ class IBackend(GangaObject):
         updateMonitoringInformation().
         """
 
-        ## Have to import here so it's actually defined
         from Ganga.Core import monitoring_component
+        was_monitoring_running = monitoring_component and monitoring_component.isEnabled(False)
 
         logger.debug("Running Monitoring for Jobs: %s" % [j.getFQID('.') for j in jobs])
 
@@ -476,7 +476,8 @@ class IBackend(GangaObject):
 
                 for this_block in monitorable_blocks:
 
-                    if monitoring_component and not monitoring_component.isEnabled(False) or not monitoring_component:
+                    # If the monitoring function was running at the start of the function but has since stopped, break.
+                    if was_monitoring_running and monitoring_component and not monitoring_component.isEnabled(False) or not monitoring_component:
                         break
 
                     try:
--- a/python/Ganga/test/GPI/Monitoring/TestMonitoring.py
+++ b/python/Ganga/test/GPI/Monitoring/TestMonitoring.py
@@ -6,13 +6,12 @@ from Ganga.testlib.GangaUnitTest import GangaUnitTest
 
 master_timeout = 300.
 
+
 def dummySleep(someJob):
     my_timeout = 0.
     while someJob.status not in ['completed', 'failed', 'killed', 'removed'] and my_timeout < master_timeout:
         time.sleep(1.)
-        my_timeout+=1.
-
-    return
+        my_timeout += 1.
 
 class TestMonitoring(GangaUnitTest):
 
@@ -88,8 +87,6 @@ class TestMonitoring(GangaUnitTest):
         disableMonitoring()
         j=Job()
         j.submit()
-
-        time.sleep(20.)
 
         self.assertEqual(j.status, 'submitted')
 
--- a/python/Ganga/testlib/GangaUnitTest.py
+++ b/python/Ganga/testlib/GangaUnitTest.py
@@ -48,6 +48,7 @@ def start_ganga(gangadir_for_test, extra_opts=[]):
         ('Configuration', 'gangadir', gangadir_for_test),
         ('Configuration', 'user', 'testframework'),
         ('Configuration', 'repositorytype', 'LocalXML'),
+        ('Configuration', 'UsageMonitoringMSG', False),  # Turn off spyware
         ('TestingFramework', 'ReleaseTesting', True),
         ('Queues', 'NumWorkerThreads', 2),
     ]
--- a/python/Ganga/test/GPI/Bugs/TestSavannah28511.py
+++ b/python/Ganga/test/GPI/Bugs/TestSavannah28511.py
@@ -1,24 +1,23 @@
 from __future__ import absolute_import
 
 from Ganga.testlib.GangaUnitTest import GangaUnitTest
+from Ganga.testlib.monitoring import run_until_completed, run_until_state
 
 
 class TestSavannah28511(GangaUnitTest):
     def test_Savannah28511(self):
         from Ganga.GPI import Job, TestSubmitter
 
-        from GangaTest.Framework.utils import sleep_until_completed, sleep_until_state
-
         j = Job()
 
         j.submit()
-        self.assertTrue(sleep_until_completed(j, 20), 'Job is not completed')
+        assert run_until_completed(j, timeout=20), 'Job is not completed'
 
         j.resubmit()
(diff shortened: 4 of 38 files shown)


python/Ganga/GPIDev/Adapters/IBackend.py
```diff
@@ -475,10 +472,6 @@ def master_updateMonitoringInformation(jobs):
                     temp_block = []
 
                 for this_block in monitorable_blocks:
-
-                    if monitoring_component and not monitoring_component.isEnabled(False) or not monitoring_component:
```
rob-c: This is here to get out of the loop during shutdown, does it need to be removed?

python/Ganga/test/GPI/Monitoring/TestMonitoring.py
```diff
@@ -89,7 +90,7 @@ def test_f_reallyDisabled(self):
         j=Job()
         j.submit()
 
-        time.sleep(20.)
+        dummySleep(j, states=['submitted'])
```
rob-c: Surely this line shouldn't be needed if we're going straight from submit to check the status?
I think this was an issue with something not updating live in the past but we can probably drop this now (for a saving of probably 0.1s)

python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -48,6 +48,7 @@ def start_ganga(gangadir_for_test, extra_opts=[]):
         ('Configuration', 'gangadir', gangadir_for_test),
         ('Configuration', 'user', 'testframework'),
         ('Configuration', 'repositorytype', 'LocalXML'),
+        ('Configuration', 'UsageMonitoringMSG', False),  # Turn off spyware
```
rob-c: :+1:

rob-c: Actually thinking about this, is it possible to run with this on for the parallel tests at least or was this a serious bottleneck there?

milliams: We can indeed turn it on for the parallel tests if we want. It was not really a bottleneck at all but rather it just seemed silly that we were swamping the gangamon system with fake reports all the time.

python/Ganga/test/GPI/Monitoring/TestMonitoring.py
milliams: Ok, I will look at removing it.

python/Ganga/GPIDev/Adapters/IBackend.py
milliams: The issue is that we want to be able to explicitly call the monitoring functions, even if the monitoring system is not running. In the case of the tests, the monitoring is not running and we call `master_updateMonitoringInformation` in a loop. This causes any jobs with subjobs to never be monitored.

What do you think about a global var in `ShutdownManager.py` which is set to `True` at the beginning of `_ganga_run_exitfuncs()`? Any code which wants to check if ganga is in the process of shutting down could then check that flag.

python/Ganga/testlib/GangaUnitTest.py
rob-c: If you point it to a non existent server in the config the messages should silently drop. (I'll add an issue to add a test for this).
It would make sense to leave it on for the parallel stuff imo. It should be safe, but I'd like to catch any problems in testing :)

python/Ganga/GPIDev/Adapters/IBackend.py
rob-c: hmmm, This should work, but I was also hoping to allow for the monitoring to be shutdown and restarted quickly during runtime.
If this isn't here the disableMonitoring for instance doesn't take effect until after all of the jobs here have been looped over. (I think)

I think passing a special argument to run without the monitoring here may be a good idea.
Having this only bail out on shutdown risks having the disableMonitoring command becoming slow. I've not seen it auto-called in a long time but it's called in case of a repo error such as the $HOME partition becoming full and write errors occuring.

python/Ganga/testlib/GangaUnitTest.py
milliams: It looks like we might be able to set the target to be `dashb-mb-test.cern.ch` rather than `dashb-mb.cern.ch`. I'll test if this works.

rob-c: Yes, but we're not currently draining this queue. I don't think this is a problem but if anyone complains we should turn it back off because it would require setting up another collector instance for draining the tests which is a lot of effort for little to no gain

milliams: One of the reasons that I turned off the spyware for the tests is that it will make any statistics we gather about user version, backends, applications etc. worthless as it will be flooded by test run data. I think if we want to test that stomp can be submitted in parallel then we should have a test which tests that rather than relying on an unrelated test to do so.

rob-c: If you point stomp at a server which doesn't exist does the library throw an error? I was under the impression this was equivalent to udp that the user didn't notice dropped data?

milliams: 1. It turns out that `dashb-mb-test.cern.ch` doesn't exist any more
2. Setting it to an non-existent server makes a test of the submission of a simple local job go from 5 seconds to 15 seconds due to it trying to establish a connection.

```
WARNING  Could not connect to host dashb-mb-test.cern.ch, port 61113: [Errno -2] Name or service not known
```

rob-c: 1. doesn't surprise me
 2. Rats, this needs to be looked at in a PR properly, and probably shouldn't be throwing an error.

OK, lets turn it off in the tests for now as we don't gain anything really with it on

milliams:
The last test run ran in about 13 minutes which is approximately half what they were taking previously.

python/Ganga/GPIDev/Adapters/IBackend.py
milliams: For now with this line I'm thinking that I will add a line to the beginning of the function which checks if the monitoring was running when the function started. At the point of the line above it will become "was the monitoring running when the function started and has it now stopped? If so, `break`"

rob-c: This is effectively what we need a nice hook for yes

milliams:
Ok, I've made some updates based on @rob-c's comments. Any more?

alexanderrichards:
all sounds sensible to me and as you say touches almost no runtime code so I see no problem. Just quickly though (for my benefit I guess), you state:


> The majority of the speed-up is due to a change from sleep_until_completed to run_until_completed

On the face of it these functions seem to be doing similar things. So I have a couple of questions:

1) Do you know what the difference between them is?
2) Do you therefore understand the nature of the speed up from making the switch
3) Do we need both of them?

rob-c:
Is it possible to configure the monitoring system to be faster when running tests maybe? i.e. no wait times, maybe have more threads open in the monitoring code for instance? do we get anything from that approach at all? (I'm just curious if it's easy to understand where the bottleneck is or not)

milliams:
@alexanderrichards `sleep_until_completed()` will basically call `runMonitoring()` on a job which in turn sets the monitoring system's list of jobs to be just that job. It will then sleep until the various monitoring threads have finished monitoring those jobs.

`run_until_completed()` will run the monitoring code serially in the main thread without all the overhead of the multi-threaded monitoring system.

Even with the wait time between monitoring system runs set to zero seconds, it still takes longer than with `run_until_completed()`.

PR #488, opened 2016-05-13
Breaking up test and Making DiracFile work in test

rob-c:
Fixes recent minor bug in DiracFile. Better splits up the test into the components which require external resources to run.


rob-c:
Any objections to merging?

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -193,35 +190,38 @@ def locations(self):
...
-        if _lfn != "" and _namePattern != "":
-            self.lfn = _lfn
-            self.remoteDir = os.path.dirname(_lfn)
-            self.namePattern = os.path.basename(_namePattern)
-            self.localDir = os.path.dirname(expandfilename(_namePattern))
-
-        elif _lfn != "" and _namePattern == "":
-            self.lfn = _lfn
+        lfn = copy.deepcopy(_lfn)
+        namePattern = copy.deepcopy(_namePattern)
```
milliams: Is `_namePattern` a string? If so then it's immutable and there should be no need to do a deepcopy.

rob-c:
again any objections?

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -193,35 +190,38 @@ def locations(self):
...
-        if _lfn != "" and _namePattern != "":
-            self.lfn = _lfn
-            self.remoteDir = os.path.dirname(_lfn)
-            self.namePattern = os.path.basename(_namePattern)
-            self.localDir = os.path.dirname(expandfilename(_namePattern))
-
-        elif _lfn != "" and _namePattern == "":
-            self.lfn = _lfn
+        lfn = _lfn
+        namePattern = _namePattern
```
milliams: Why not just 
```python
def _setLFNnamePattern(self, lfn="", namePattern=""):
```
?

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -193,35 +190,38 @@ def locations(self):
...
+            ## TODO REPLACE THIS WITH IN LIST OF VONAMES KNOWN
+            # Check for /lhcb/some/path or /gridpp/some/path
+            if namePattern.split(os.pathsep)[0] == self.defaultSE \
+                or (len(namePattern) > 3 and namePattern[0:4].upper() == "LFN:"\
+                    or len(namePattern.split(os.pathsep)) > 1 and namePattern.split(os.pathsep)[1] == self.defaultSE):
+                # Check for LFN:/gridpp/some/path or others...
+                lfn = namePattern
+                namePattern = ""
+
+        if lfn != "" and lfn is not None:
```
milliams: Maybe just:
```python
if lfn:
```
?

rob-c:
Once more

milliams:
:+1:

PR #490, opened 2016-05-16
Move construction of DiracLFNBase into a function

milliams:
If it is in the `__init__.py` then some information set by the user will
not yet be available. Deferring it to point of use allows user settings
to be used.


rob-c:
Can you add the default value to the Config documentation do it's clear what this is expected to be upon default construction

milliams:
Ok, I have added a comment to that effect.

PR #491, opened 2016-05-16
Fixing some Plugin loading and test imports for LHCb

rob-c:
This fixes the bad runtime definitions in the GangaLHCb tests to ensure Dirac and Gaudi plugins are loaded as required.
This includes changing the schema test which should have always pointed to this in the first place.

I've also fixed a load of imports from the old GangaLHCb.test as the old functionality here wasn't ported when the tests were moved.


rob-c:
I based off the wrong branch, 1min while I fix

rob-c:
Well this should compare to #492 but with some minor changes which need to be discussed.

PR #492, opened 2016-05-16
Fix LHCb test imports

milliams:
The main important change in this PR is that to  `python/GangaLHCb/old_test/config/localxml.ini`. We now need to explicitly add `GangaDirac` to the runtime path before `GangaLHCb` so that the config has been set up by the time that the LHCb statup runs.

The other changes are just fixes to accidental leftovers from the `test`→`old_test` changes.

### DIFF
--- a/release/tools/schema_gen.sh
+++ b/release/tools/schema_gen.sh
@@ -101,7 +101,7 @@ case `whoami` in
     gangaat )
         export LOAD_PACKAGES='GangaTest:GangaAtlas' ;;
     gangalb )
-        export LOAD_PACKAGES='GangaTest:GangaLHCb' ;;
+        export LOAD_PACKAGES='GangaTest:GangaDirac:GangaGaudi:GangaLHCb' ;;
 esac
 
 ##Run the repo generation 
--- a/python/GangaDirac/old_test/GPI/Files/TestDiracFile.py
+++ b/python/GangaDirac/old_test/GPI/Files/TestDiracFile.py
@@ -4,7 +4,7 @@ from GangaTest.Framework.tests import GangaGPITestCase
 #from Ganga.GPIDev.Adapters.StandardJobConfig       import StandardJobConfig
 #from Ganga.Core.exceptions                         import ApplicationConfigurationError, GangaException
 from Ganga.GPI import *
-from Ganga.test import generateUniqueTempFile
+from Ganga.old_test import generateUniqueTempFile
 #import GangaDirac.Lib.Server.DiracServer as DiracServer
 # GangaTest.Framework.utils defines some utility methods
 from GangaTest.Framework.utils import sleep_until_completed, sleep_until_state
--- a/python/GangaLHCb/old_test/Bugs/JIRA1943.gpi
+++ b/python/GangaLHCb/old_test/Bugs/JIRA1943.gpi
@@ -1,5 +1,5 @@
 from tempfile import mkdtemp
-from GangaLHCb.test import *
+from GangaLHCb.old_test import *
 from os.path import join, basename
 from os import remove
 
--- a/python/GangaLHCb/old_test/Bugs/Savannah13943.gpi
+++ b/python/GangaLHCb/old_test/Bugs/Savannah13943.gpi
@@ -1,7 +1,7 @@
 import os,os.path
 from GangaTest.Framework.utils import sleep_until_completed,sleep_until_state
 import tempfile
-from GangaLHCb.test import *
+from GangaLHCb.old_test import *
 
 tmpdir = tempfile.mktemp()
 os.mkdir(tmpdir)
--- a/python/GangaLHCb/old_test/Bugs/Savannah20538.gpi
+++ b/python/GangaLHCb/old_test/Bugs/Savannah20538.gpi
@@ -1,5 +1,5 @@
 from tempfile import mkdtemp
-from GangaLHCb.test import *
+from GangaLHCb.old_test import *
 from os.path import join,basename
 from os import remove
 
--- a/python/GangaLHCb/old_test/Bugs/Savannah36400.gpi
+++ b/python/GangaLHCb/old_test/Bugs/Savannah36400.gpi
@@ -1,4 +1,4 @@
-from GangaLHCb.test import *
+from GangaLHCb.old_test import *
 addDiracTestSubmitter()
 
 #set a version that does't exist in the config
--- a/python/GangaLHCb/old_test/GPI/DaVinci/TestDaVinci.py
+++ b/python/GangaLHCb/old_test/GPI/DaVinci/TestDaVinci.py
@@ -4,7 +4,7 @@ from GangaTest.Framework.utils import sleep_until_completed
 
 import Ganga.Utility.Config.Config
 
-from GangaLHCb.test import getDiracAppPlatform
+from GangaLHCb.old_test import getDiracAppPlatform
 
 import Ganga.Utility.logging
 logger = Ganga.Utility.logging.getLogger()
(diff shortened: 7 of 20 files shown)


rob-c:
You're missing the required GangaGaudi plugin and the schema test will also need these changes as per #491

milliams:
Why are the `GangaGaudi` packages needed now? They weren't in the past.

rob-c:
Simply, this is what LHCb installs with on lxplus therefore this is the same env I'd expect to be in the tests. This is due to 1 or 2 bits of code historically being used through the GPI and in GangaGaudi but I'm keen to test the code as it's presented to LHCb.
This may be a good thing to change but I'm keen to try and reduce the amount of additional changes wrt LHCb in this release so that any Dirac related bugs can be spotted/fixed more easily.

milliams:
The `RUNTIME_PATH` is already being set to this by the `LHCb.ini` file as you would expect. The only reason we were seeing a problem here is that since `localxml.ini` (`GangaTest:GangaLHCb`) was loaded before `LHCb.ini` (`GangaDirac:GangaGaudi:GangaLHCb`), we ended up with a `RUNTIME_PATH` like `GangaTest:GangaLHCb:GangaDirac:GangaGaudi:GangaLHCb` which has an incorrect ordering. My PR is effectively changing this to `GangaTest:GangaDirac:GangaLHCb:GangaDirac:GangaGaudi:GangaLHCb`

We could probably set the `RUNTIME_PATH` in `localxml.ini` to an empty string and fix the problem too.

rob-c:
OK, where is the RUNTIME_PATH being defined and in what order?

I'm very confused how the runtime argument (passed in the scripts to run the tests) of ```RUNTIME_PATH=GangaTest``` isn't just shorting out this whole mess to never load the LHCb plugin as it is?

Is the LHCb plugin being loaded in order to load additional plugins after itself?
In which case how can we ever expect to load the LHCb plugin ahead of it's dependencies?

I'm bugged that we seem to be redefining the RUNTIME_PATH during the whole config/load process and the final version of which decides what we're going to load.

drmarkwslater:
This is some Ganga 'magic' for the RUNTIME_PATH (and I **think** other related variables) that will actually append things rather than overwrite unless you put in a triple colon `:::`. I'm sure there was a good reason for it at the time (probably because we don't have a good dependency system) but that's what it is at the moment, so:
```
RUNTIME_PATH = GangaTest
RUNTIME_PATH = GangaLHCb

print RUNTIME_PATH
GangaTest:GangaLHCb
```

drmarkwslater:
P.S. I **think** it's triple colon - I really can't remember 😄

milliams:
There are certain config variable types (ones that end in `_PATH`) which act in a special way (see `transform_PATH_option()`). If you set their value, the new value is appended rather than replacing the existing value. A normal LHCb user running on lxplus will have the `LHCb.ini` file loaded (via some env vars set by `SetupProject`) which sets the `RUNTIME_PATH` to `GangaDirac:GangaGaudi:GangaLHCb`. This is all fine and works.

When running the tests however we call `ganga --test -o'[Configuration]RUNTIME_PATH=GangaTest' -o'[TestingFramework]ReleaseTesting=True' -o'Config=localxml.ini' GangaLHCb/old_test/` which sets it to `GangaTest` but also explicitly tells it to load `localxml.ini`. This in addition to loading `LHCb.ini`. The order of these config files is defined in some (unknown by me) way but it seemed that it was causing problems.

The order in which plugins are loaded only affects code which runs on plugin *import*, not code which is run as part of the plugins day-to-day usage. The problem came when you edited `GangaLHCb/__init__.py` to use things which are set by `GangaDirac/__init__.py` and so it is those two plugins for which the order matters.

I'd love to be able to replace this all with a proper plugin system that does away with a lot of the nonsense that goes on at the moment but that's a bit of work that we don't have time for at the moment.

rob-c:
@drmarkwslater I don't follow what you mean, did you mean:
```
RUNTIME_PATH = GangaTest
RUNTIME_PATH = :::GangaLHCb

print RUNTIME_PATH
GangaLHCb
```

OK, so other than the tests loading things out of order with Dirac I don't know what is in Gaudi (and don't want to go checking now) so shouldn't this be loaded too before LHCb as configs within both plugins (defined in somewhere imported from ```__init__.py```)could be interfering.

Is there strong agreement here or are we looking to cleanup the config of the old tests so that this makes sense?
I'd say importing Dirac:LHCb:Gaudi should be regarded as a bug, or potentially creating one. So why not change this in the test config so that this is what we get.

drmarkwslater:
@rob-c Yep - this is what I meant. I don't have a strong feeling for loading GangaGaudi in `localxml.ini` as well other than it wasn't there before and the tests seem to work with just `GangaDirac` added. I defer to others for a ruling.

rob-c:
I would argue that this is read into an ordered dict so it only stores unique entries but we want to ensure we load the plugins in order. I don't know how ordered dict handles the same key twice but I imagine it reduces:
```Test:Dirac:LHCb:Dirac:Gaudi:LHCb``` to ```Test:Dirac:LHCb:Gaudi``` which isn't what we want. (although possibly harmless this still makes me very uncomfortable) unless we know that adding a new key changes the position of the dict entry in ```keys()```?

**Comment by milliams** [comment]
Where is this being put into an `OrderedDict`? As far as I am aware, this is just stored in a string.

**Comment by milliams** [comment]
There's also an issue here with the fact that subsequent settings of `RUNTIME_PATH` prepend rather than append. So this happens:
```python
>>> config['Configuration'].RUNTIME_PATH = 'Dirac'
>>> config['Configuration'].RUNTIME_PATH
'Dirac:'
>>> config['Configuration'].RUNTIME_PATH = 'LHCb'
>>> config['Configuration'].RUNTIME_PATH
'LHCb:Dirac:'
>>> config['Configuration'].RUNTIME_PATH = 'Dirac'
>>> config['Configuration'].RUNTIME_PATH
'Dirac:LHCb:'
```

rob-c:
so shall we just put all 3 ini config so that they get loaded in the order we expect (or adjust this elsewhere). I'm just keen that the LHCb related plugins get loaded in the correct order more than before/after the Test plugin...

milliams:
Sure, I can't see that it will hurt anything. It seems that as written, the `RUNTIME_PATH` was never really expected to be ordered and no plugins have depended on a particular load order until now.

I will update this PR to include `GangaGaudi` in the path just to be safe.

On the side, I will have a think about a better system for defining dependencies so that all we have to do is `GangaLHCb` and that will automatically know what it depends on and load them first.

release/tools/schema_gen.sh
```diff
@@ -101,7 +101,7 @@ case `whoami` in
     gangaat )
         export LOAD_PACKAGES='GangaTest:GangaAtlas' ;;
     gangalb )
-        export LOAD_PACKAGES='GangaTest:GangaLHCb' ;;
+        export LOAD_PACKAGES='GangaTest:GangaDirac:GangaLHCb' ;;
```
rob-c: Can we add GangaGaudi here too, just for completeness so we don't revisit and wonder why the change.
(the schema test should probably be one we port fairly soon so we can keep track in the integration tests I suppose)

rob-c:
I think worrying about the order of plugins should be a major task. We have much more work to do to collect the existing configs in the existing plugins into the ```__init__.py``` and after that I think we can come back to address things like the order in which the plugins are loaded. For the moment this can be changed in the config and doesn't randomly change during runtime so this is just a quirk of the designed system not something we need to develop from scratch again imo.

release/tools/schema_gen.sh
milliams: Sure, I will add it now. Again, I'm pretty sure that it's covered by the `LHCb.ini` file so there's no real change here.

milliams:
Am I ok to merge this or are there other changes to make? I'd like to rerun the legacy tests so that we can get this release out ASAP.

PR #493, opened 2016-05-16
fix broken import

mjmottram:


drmarkwslater:
Apologies - I missed that one! 👍  However, I'll wait until the release is out before merging to develop.

drmarkwslater:
Just notices this was against `master` rather than `develop!! @milliams  - is this a problem?

milliams:
It's not the end of the world. It won't be available to us until the next release but nothing should break.

rob-c:
resyncing the branch isn't difficult, this would just be another PR to merge the changes from master and then carry on (frankly it could/should even be done manually with a single command)

PR #497, opened 2016-05-17
Bugfix/fixing incomplete job object

### LINKED ISSUE #461: master index not being updated with auto-flush

### LINKED ISSUE #495: losing subjob data
this-morning I ran the following on a job which I forgot to be inaccessible to my current ganga setup:

```
jobs(326).subjobs
```

I received the appropriate Incomplete Object error and restarted into the correct Ganga environment (I was missing a plugin initially and so couldn't load the job)

Now I find that all of the subjob information associated with this job has been lost. I'm curious about this as the appropriate XML exist as well as ```subjobs.idx``` so working out why this is failing to load could be a bit tricky.

I'm a little concerned about this as it likely means the loading association of the subjobs object with a class is still a bit too fragile.

rob-c:
A potential fix for #495 . This performs a much better/safer check against whether a job has subjobs associated with it and contains some minor refactoring of the SubJobXMLList class to keep subjob logic contained within there.

I'm keen to see how this runs on Jenkins, this should work but even still...

I'm keen to add a test for this before merging but would like to get this into 6.1.21 as it saves me some headaches when switching between branches and developing plugins.

Fixes #495
Edit:
Fixes #461

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -764,10 +764,10 @@ class GangaRepositoryLocal(GangaRepository):
                     self._fully_loaded[this_id] = self.objects[this_id]
 
                 subobj_attr = getattr(self.objects[this_id], self.sub_split, None)
-                sub_attr_dirty = getattr(subjob_attr, '_dirty', False)
+                sub_attr_dirty = getattr(subobj_attr, '_dirty', False)
                 if sub_attr_dirty:
-                    if hasattr(subjob_attr, 'flush'):
-                        subjob_attr.flush()
+                    if hasattr(subobj_attr, 'flush'):
+                        subobj_attr.flush()
 
                 self.objects[this_id]._setFlushed()
 
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -354,7 +354,7 @@ class GangaRepositoryLocal(GangaRepository):
                 with open(_master_idx, 'r') as input_f:
                     this_master_cache = pickle_from_file(input_f)[0]
                 for this_cache in this_master_cache:
-                    if this_cache[4] >= 0:
+                    if this_cache[1] >= 0:
                         this_id = this_cache[0]
                         self._cache_load_timestamp[this_id] = this_cache[1]
                         self._cached_cat[this_id] = this_cache[2]
@@ -846,7 +846,7 @@ class GangaRepositoryLocal(GangaRepository):
                 raise InaccessibleObjectError(self, this_id, errs[0])
 
             has_children = (self.sub_split is not None) and hasattr(tmpobj, self.sub_split) and len(getattr(tmpobj, self.sub_split)) == 0
-            has_children = has_children or SubJobXMLList.JobHasChildrenTest(path.dirname(fn))
+            has_children = has_children or SubJobXMLList.jobHasChildrenTest(os.path.dirname(fn), self.dataFileName)
 
             if this_id in self.objects:
                 self._must_actually_load_xml(fn, this_id, load_backup, has_children, tmpobj)
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -672,7 +672,7 @@ class GangaRepositoryLocal(GangaRepository):
         if not isType(obj, EmptyGangaObject):
             split_cache = None
 
-            has_children = hasattr(obj, self.sub_split) and getattr(obj, self.sub_split)
+            has_children = getattr(obj, self.sub_split, False)
 
             if has_children:
 
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -925,11 +925,11 @@ class GangaRepositoryLocal(GangaRepository):
             logger.error("#%s Error(s) Loading File: %s" % (len(errs), fobj.name))
             raise InaccessibleObjectError(self, this_id, errs[0])
 
-        logger.debug("Testing children: %s" % str(this_id))
-	#logger.debug("Testing in: %s" % os.path.dirname(fn))
+        logger.debug("Checking children: %s" % str(this_id))
+	#logger.debug("Checking in: %s" % os.path.dirname(fn))
 	#logger.debug("found: %s" % os.listdir(os.path.dirname(fn)))
 
-        has_children = SubJobXMLList.jobHasChildrenTest(os.path.dirname(fn), self.dataFileName)
+        has_children = SubJobXMLList.checkJobHasChildren(os.path.dirname(fn), self.dataFileName)
 
         logger.debug("Found children: %s" % str(has_children))
 
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -672,8 +672,7 @@ class GangaRepositoryLocal(GangaRepository):
         if not isType(obj, EmptyGangaObject):
             split_cache = None
 
-            has_children = (not self.sub_split is None) and\
-                    (self.sub_split in obj._data) and hasattr(obj, self.sub_split) and len(getattr(obj, self.sub_split)) > 0
+            has_children = SubJobXMLList.jobHasChildrenTest(os.path.dirname(fn), self.dataFileName)
 
             if has_children:
 
@@ -765,9 +764,10 @@ class GangaRepositoryLocal(GangaRepository):
                     self._fully_loaded[this_id] = self.objects[this_id]
 
                 subobj_attr = getattr(self.objects[this_id], self.sub_split, None)
-                sub_attr_dirty = getattr(subobj_attr, '_dirty', False)
+                sub_attr_dirty = getattr(subjob_attr, '_dirty', False)
                 if sub_attr_dirty:
-                    getattr(self.objects[this_id], self.sub_split).flush()
+                    if hasattr(subjob_attr, 'flush'):
+                        subjob_attr.flush()
 
                 self.objects[this_id]._setFlushed()
 
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -1043,7 +1043,7 @@ class GangaRepositoryLocal(GangaRepository):
                 raise
 
             except Exception as err:
-                raise
+                
                 should_continue = self._handle_load_exception(err, fn, this_id, load_backup)
 
                 if should_continue is True:
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -427,7 +427,7 @@ class GangaRepositoryLocal(GangaRepository):
                 if k in self.incomplete_objects:
                     continue
                 try:
-                    if k in k in self._fully_loaded:
+                    if k in self._fully_loaded:
                         # Check and write index first
                         obj = v#self.objects[k]
                         new_index = None
@@ -645,7 +645,7 @@ class GangaRepositoryLocal(GangaRepository):
             self._internal_setitem__(ids[i], objs[i])
 
             # Set subjobs dirty - they will not be flushed if they are not.
-            if self.sub_split and self.sub_split in objs[i]._data:
+            if self.sub_split and hasattr(objs[i], self.sub_split):
                 try:
                     sj_len = len(getattr(objs[i], self.sub_split))
                     if sj_len > 0:
@@ -672,7 +672,7 @@ class GangaRepositoryLocal(GangaRepository):
         if not isType(obj, EmptyGangaObject):
             split_cache = None
 
-            has_children = SubJobXMLList.jobHasChildrenTest(os.path.dirname(fn), self.dataFileName)
+            has_children = hasattr(obj, self.sub_split) and getattr(obj, self.sub_split)
 
             if has_children:
 
@@ -704,7 +704,7 @@ class GangaRepositoryLocal(GangaRepository):
                     for sj in getattr(obj, self.sub_split):
                         job_dict[sj.id] = stripProxy(sj)
                     tempSubJList._reset_cachedJobs(job_dict)
-                    tempSubJList.flush()
+                    tempSubJList.flush(ignore_disk=True)
                     del tempSubJList
 
                 safe_save(fn, obj, self.to_file, self.sub_split)
@@ -876,16 +876,17 @@ class GangaRepositoryLocal(GangaRepository):
         # TODO investigate changing this to copyFrom
         if need_to_copy:
             for key, val in tmpobj._data.items():
-                obj.setSchemaAttribute(key, val)
+                setattr(obj, key, val)
             for attr_name, attr_val in obj._schema.allItems():
                 if attr_name not in tmpobj._data:
-                    obj.setSchemaAttribute(attr_name, obj._schema.getDefaultValue(attr_name))
+                    setattr(obj, attr_name, obj._schema.getDefaultValue(attr_name))
 
         if has_children:
             logger.debug("Adding children")
-            obj.setSchemaAttribute(self.sub_split, SubJobXMLList(os.path.dirname(fn), self.registry, self.dataFileName, load_backup, parent=obj))
+            setattr(obj, self.sub_split, SubJobXMLList(os.path.dirname(fn), self.registry, self.dataFileName, load_backup, parent=obj))
         else:
-            obj.setSchemaAttribute(self.sub_split, None)
+            if obj._schema.hasAttribute(self.sub_split):
+                setattr(obj, self.sub_split, obj._schema.getDefaultValue(self.sub_split))
 
         from Ganga.GPIDev.Base.Objects import do_not_copy
         for node_key, node_val in obj._data.items():
@@ -923,6 +924,8 @@ class GangaRepositoryLocal(GangaRepository):
             raise InaccessibleObjectError(self, this_id, errs[0])
 
         logger.debug("Testing children: %s" % str(this_id))
+	#logger.debug("Testing in: %s" % os.path.dirname(fn))
+	#logger.debug("found: %s" % os.listdir(os.path.dirname(fn)))
 
         has_children = SubJobXMLList.jobHasChildrenTest(os.path.dirname(fn), self.dataFileName)
 
@@ -1185,5 +1188,9 @@ class GangaRepositoryLocal(GangaRepository):
         Args:
             obj (GangaObject): The object we want to know if it was loaded into memory
         """
-        return id(obj) in [id(loaded_obj) for loaded_obj in self._fully_loaded.values()]
+        try:
+            _id = next(id_ for id_, o in self._fully_loaded.items() if o is obj)
+            return True
+        except StopIteration:
+            return False
 
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -874,16 +874,18 @@ class GangaRepositoryLocal(GangaRepository):
         # If the object was already in the objects (i.e. cache object, replace the schema content wilst avoiding R/O checks and such
         # The end goal is to keep the object at this_id the same object in memory but to make it closer to tmpobj.
         # TODO investigate changing this to copyFrom
+        # The temp object is from disk so all contents have correctly passed through sanitising via setattr at least once by now so this is safe
         if need_to_copy:
             for key, val in tmpobj._data.items():
-                setattr(obj, key, val)
+                obj.setSchemaAttribute(key, val)
             for attr_name, attr_val in obj._schema.allItems():
                 if attr_name not in tmpobj._data:
-                    setattr(obj, attr_name, obj._schema.getDefaultValue(attr_name))
+                    obj.setSchemaAttribute(attr_name, obj._schema.getDefaultValue(attr_name))
 
         if has_children:
             logger.debug("Adding children")
-            setattr(obj, self.sub_split, SubJobXMLList(os.path.dirname(fn), self.registry, self.dataFileName, load_backup, parent=obj))
+            # NB Keep be a SetSchemaAttribute to bypass the list manipulation which will put this into a list in some cases 
+            obj.setSchemaAttribute(self.sub_split, SubJobXMLList(os.path.dirname(fn), self.registry, self.dataFileName, load_backup, parent=obj))
         else:
             if obj._schema.hasAttribute(self.sub_split):
                 setattr(obj, self.sub_split, obj._schema.getDefaultValue(self.sub_split))
@@ -1041,7 +1043,7 @@ class GangaRepositoryLocal(GangaRepository):
                 raise
 
             except Exception as err:
-
+                raise
                 should_continue = self._handle_load_exception(err, fn, this_id, load_backup)
 
                 if should_continue is True:
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -33,7 +33,10 @@ logger = Ganga.Utility.logging.getLogger()
 save_all_history = False
 
 def check_app_hash(obj):
-    """Writes a file safely, raises IOError on error"""
+    """Writes a file safely, raises IOError on error
+    Args:
+        obj (GangaObject): This is an object which has a prepared application
+    """
 
     isVerifiableApp = False
     isVerifiableAna = False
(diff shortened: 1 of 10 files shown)


python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -225,7 +225,7 @@ def shutdown(self):
         from Ganga.Utility.logging import getLogger
         logger = getLogger()
         logger.debug("Shutting Down GangaRepositoryLocal: %s" % self.registry.name)
-        for k in self._fully_loaded.keys():
+        for k in [_k for _k in self._fully_loaded.keys() if _k not in self.incomplete_objects]:
```
milliams: So there are objects in `self._fully_loaded` which are also in `self.incomplete_objects`? That seems bad.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -352,11 +354,12 @@ def _read_master_cache(self):
                 with open(_master_idx, 'r') as input_f:
                     this_master_cache = pickle_from_file(input_f)[0]
                 for this_cache in this_master_cache:
-                    this_id = this_cache[0]
-                    self._cache_load_timestamp[this_id] = this_cache[1]
-                    self._cached_cat[this_id] = this_cache[2]
-                    self._cached_cls[this_id] = this_cache[3]
-                    self._cached_obj[this_id] = this_cache[4]
+                    if this_cache[4] >= 0:
```
milliams: What is this protecting against? Is the `id` sometimes negative?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -517,3 +495,45 @@ def _private_display(self, reg_slice, this_format, default_width, markup):
             ds += markup(this_format % tuple(vals), colour)
 
         return ds
+
+    @staticmethod
+    def JobHasChildrenTest(jobDirectory):
```
milliams: Function name should start with lower case.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: That is the time, and yes it is potentially -ve if something went wrong in a previous version of ganga were bad caches were still saved.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -517,3 +495,45 @@ def _private_display(self, reg_slice, this_format, default_width, markup):
...
+        """ Return True/False if given (job?) object has children associated with it
+        Args:
+            jobDirectory (str): name of folder to be examined
+        """
+
+        count = SubJobXMLList.countSubJobDirs(jobDirectory)
+        if count:
+            return True
+        else:
+            return False
```
milliams: This could just be:
```python
@staticmethod
def jobHasChildrenTest(jobDirectory):
    return bool(SubJobXMLList.countSubJobDirs(jobDirectory))
```
at which point the one place it's used may as well just be:
```python
has_children = has_children or SubJobXMLList.countSubJobDirs(path.dirname(fn))
```
and this function is not needed.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: I don't know it's a belt and braces check for now anyway, this may get relaxed.

this is just a short PR to fix the problems I'm seeing and see if they make worse ones (which given there's almost certainly a missing import/typo, these changes do)

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: (aside I hate casting to bools but c'est la vie)

I was planning to use/change this more in the future and to expand/change it. I'm still keen to keep a function which returns a True/False as it's much clearer what's going on.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -517,3 +495,45 @@ def _private_display(self, reg_slice, this_format, default_width, markup):
...
+                    break
+            else:
+                break
+
+            #subjob_data = self.__get_dataFile(str(i))
+            #if path.isfile(subjob_data)):
+            #    subjob_count = subjob_count + 1
+            #i += 1
+
+        return subjob_count
```
milliams: For starters, [`os.listdir` is not guaranteed to be ordered](https://docs.python.org/2.7/library/os.html?highlight=listdir#os.listdir).

Secondly, does this really need to only be sequentially numbered directories? If not, this could just be:
```python
@staticmethod
def countSubJobDirs(jobDirectory):
    return len(dir for dir in listdir(jobDirectory) if dir.isdigit())
    
```

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
milliams: A belt and braces check here will just hide potentially large problems. If we think there could be a problem with the consistency of these two lists then we need to tackle that problem, not ignore it.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: In that case I'll amend it to sort
Yes because I'm planning to check if any files are missing in another revision.

If this is so broken in the current release shall we bump a working PR to 6.1.20?

milliams: Making changes here is too late for 6.1.20. It's been working ok so far.

Linux will almost always return a sorted list (other systems may not) but the problem come in with the *way* it might sort it. I guess that subjobs 1-12 may be sorted as
```
1
10
11
12
2
3
4
5
6
7
8
9
```

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Yes, but we should never be risking user data on disk ever. I'm still paranoid that something will come along and ruin my jobs as it is still doing.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Frankly when you can just sort a list it's much simpler/quicker.

If it's not that broken I'll fix it in another revision then. (We don't support anything other than linux right now anyway :P)

rob-c: There is no need to sort this, why is this coming up? I'lll upload an amended version anyway which does some better testing

rob-c:
Added some docs this-morning whilst waiting for a few things to complete. This also contains a more updated form of what I'm wanting from the PR.

rob-c:
Hmmm, some failing tests but the thread tracer doesn't seem to like something. I'll try and re-run locally in a bit, running the tests locally for me is still fraught with difficulty.

milliams:
It's nothing to do with the thread tracer. If an early test fails very badly then the Python session is left in a weird state meaning that the thread tracer doesn't know whether to start. If you look at the [full logs](https://ganga-ci.cern.ch/jenkins/job/PR%20builder/1048/consoleFull) you'll see that you have a syntax error:
```python
self = <Ganga.Core.GangaRepository.GangaRepositoryXML.GangaRepositoryLocal object at 0x3a5a450>

    def shutdown(self):
        """Shutdown the repository. Flushing is done by the Registry
            Raise RepositoryError
            Write an index file for all new objects in memory and master index file of indexes"""
        logger = getLogger()
>       for k in k in self._fully_loaded:
E       UnboundLocalError: local variable 'k' referenced before assignment

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py:241: UnboundLocalError
```
If you can't run the tests locally then we need to sort that out as you really should be running them while developing to avoid issues exactly like this.

rob-c:
If you can suggest some way to get SetupProject to play nice with this then I'll happily use that. As it stands I'm having issues.
I'm wanting to spend another day working out why the tests won't run properly for me when there is a test box which runs them in an automated fashion. That's the point of this machine in my mind.

milliams:
If I log on to `lxplus` in a clean bash shell and do:
```bash
$ pip install --user pytest  # only need run this once
$ cd ganga  # This is the local git checkout of ganga
$ SetupProject Ganga
$ PYTHONPATH=python:$PYTHONPATH ~/.local/bin/py.test python/GangaLHCb/test/GPI/
```
then it successfully runs the LHCb tests. This also works for running the other Ganga tests if you change the path given to pytest.

rob-c:
setupproject changes the version of python does it not? also my problem is with pytest pointing to the ganga source not currently being worked on or accessed via SetupProject. I can run tests on another machine, not ideal but works for my needs

milliams:
If you make sure you *pre*pend `python` to the python path (or `.` if you're in the `python` directory) then it should always refer to your local version of Ganga. If you have in the past run `pip install -e .` then you may have a reference to that install in your `~/.local/lib/python2.6/site-packages/` which you can delete.

I believe that since `pip` is using Python 2.6, `pytest` will then be installed as Python 2.6. This means that running the code above will use Python 2.6 for running the tests. If you need to use the LHCb version of Python 2.7 to run the tests then the following will install an appropriate version of `py.test`:
```bash
$ SetupProject Ganga
$ wget https://bootstrap.pypa.io/get-pip.py
$ python2.7 get-pip.py --user
$ pip install --user pytest  # only need run this once
```
you can then run the tests using it with:
```bash
$ cd ganga  # This is the local git checkout of ganga
$ SetupProject Ganga
$ PYTHONPATH=python:$PYTHONPATH ~/.local/bin/py.test python/GangaLHCb/test/GPI/
```

rob-c:
yes and that worked but doesn't fully reflect my setup now, as I've mentioned this runs elsewhere offline so I'm testing there, I've got the local bin in my path for other tools I use. It's either yet another complexity or something which requires torching my setup and starting again, either way not something I want to spend time on right now

rob-c:
OK,

Well this unravelled to be a list of additional papercuts which all needed to be fixed before the main issue went away.

The main changes in this PR are now:

 1. I don't plan to make a test for the original condition in the PR, it's tricky and I'd like to give it some thought before putting it in
 2. I've now moved all knowledge of which objects have been fully loaded into memory into the Repository and am having the registry correctly poll for this. This fixes a chicken and egg issue that arises otherwise when the situation crops up that an object has to be loaded at startup and then made dirty to trigger a new index to be written in the future.
 3. The main index is removed after it's use on startup and only written again on shutdown. Minor performance penalty in starting a 2nd parallel session but this is minor and jobs don't get out of sync on dropped connections etc. #461
 4. The way of checking the number of subjobs is now a little more resilient as well as having the ability to check for all expected XML to be present on load. This fixes a potential bug later on of XML being missing only when working with a job which is trickier to debug/fix.
 5. I've added a setUpClass to the unittest to fix problems with previous tests making the environment dirty before we start running. This should probably be adopted by any tests not making use of this in some other way but that is for another PR.
 6. I've fixed the test for job resubmit in the tutorial to work on faster machines #503
 7. Documentation, I've added a lot of docstrings to the registry/repo layers which makes life a little easier.

As a result the determination of if a job has subjobs associated with it is now a bit more robust and there is an additional check that all of the XML are present before we load a job from disk. (This still allows for us to recover a corrupt/missing XML in favour of it's backup but we won't load a job with a missing subjob, this indicates some severe event occurred and requires some human interaction).

The loading of the XML from disk can probably be simplified even further in a future but some of the checks simply copied from old code are now fixed/removed as they got in the way of understanding this.

Whilst not just due to changes in this PR it's worth noting that:
Checking if an object is now fully loaded into memory is fully working and works in such a way that a Job object is always the same actual object all the way through the registry/repository and only has actions performed on the same object so we can perform checks using ```id(someJob)``` within this code reliably. (This works in front of the proxy as we also cache the proxy for the lifetime of the session too).

rob-c:
retest this please

rob-c:
Any objections to merging this branch if it passes the tests?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -621,8 +672,7 @@ def _safe_flush_xml(self, this_id):
         if not isType(obj, EmptyGangaObject):
             split_cache = None
 
-            has_children = (not self.sub_split is None) and\
-                    (self.sub_split in obj._data) and hasattr(obj, self.sub_split) and len(getattr(obj, self.sub_split)) > 0
+            has_children = hasattr(obj, self.sub_split) and getattr(obj, self.sub_split)
```
milliams: Could this be:
```python
has_children = getattr(obj, self.sub_split, False)
```

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -708,20 +763,32 @@ def flush(self, ids):
                 if this_id not in self._fully_loaded.keys():
                     self._fully_loaded[this_id] = self.objects[this_id]
 
+                subobj_attr = getattr(self.objects[this_id], self.sub_split, None)
+                sub_attr_dirty = getattr(subobj_attr, '_dirty', False)
```
milliams: All `GangaObject`s will have the `_dirty` flag so do we need to check for its existence here?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Hmm, it probably should be otherwise non splittable objects should throw an exception here... I wonder why this doesn't crop up in testing?

rob-c: Wait, nvm I think I understand the change you're suggesting, I misread that

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Line above can return None

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -1065,9 +1184,15 @@ def clean(self):
...
+        This will return a true false if an object has been fully loaded into memory
+        Args:
+            obj (GangaObject): The object we want to know if it was loaded into memory
         """
-        pass
+        try:
+            _id = next(id_ for id_, o in self._fully_loaded.items() if o is obj)
+            return True
+        except StopIteration:
+            return False
```
milliams: Could this be
```python
def isObjectLoaded(self, obj):
    return obj in self._fully_loaded.values()
```

rob-c: I'll give it a try, I started out with that and got into problems which may or may not have gone away now

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -189,9 +218,18 @@ def stop(self):
...
     @property
     def stopped(self):
+        """
+        Returns if the flusher has stopped as a boolean?
+        """
         return self._stop.isSet()
 
     def join(self, *args, **kwargs):
+        """
+        Not clear if this is called anywhere
```
milliams: This is an unhelpful code comment. `join` is implemented since we are subclassing `Thread` and so must make sure that if `join` is called on it, that it works in a polymorphic way and does the correct thing.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -222,7 +260,15 @@ class Registry(object):
...
     def __init__(self, name, doc, dirty_flush_counter=10, update_index_time=30, dirty_max_timeout=60, dirty_min_timeout=30):
-        """Registry constructor, giving public name and documentation"""
+        """Registry constructor, giving public name and documentation
+        Args:
+            name (str): Name of teh registry e.g. 'jobs', 'box', 'prep'
+            doc (str): This is the doc string of the registry describing it's use and contents
+            dirty_flush_counter (int): This is no longer used
+            update_index_time (int): This is no longer used
+            dirty_max_timeout (int): This is no longer used
+            dirty_min_timeout (int): This is no longer used
```
milliams: This is covered by #415

rob-c: So chicken or egg, which to resolve first?

milliams: I'll go ahead and merge my cleanup PRs since they are smaller and have already been fully reviewed.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: The tests hang at ```Ganga/test/GPI/TutorialTests.py::TutorialTests::test_c_JobManipulation``` there is at least an extreme performance penalty to doing it the simple way or something isn't behaving how we're expecting it to.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I didn't write the code I'm just trying to highlight it needs commenting, I'll copy what you've typed there unless you'd prefer to say something else?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: I gave that test ~30min on an i5 with an SSD that is probably deadlocking in a strange way, I'd prefer to keep the code as it is now because it works.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: Aside from my comments on #415 OK. Can you poke me when that finishes as we're going to have to wait for the tests to re-run several times before I can get to work on merging the changes into this PR.

python/Ganga/Core/GangaRepository/Registry.py
milliams: I'm not sure it needs much of a comment. It's a subclass of `threading.Thread` and this is the [`join` method](https://docs.python.org/2.7/library/threading.html?highlight=threading.thread.join#threading.Thread.join). The only potential addition is that is requests the loop to stop running.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
milliams: This is strange but I guess it's because the `list` `__contains__` method used equality rather than identity and out equality functions are slow and heavy.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -517,3 +561,52 @@ def _private_display(self, reg_slice, this_format, default_width, markup):
             ds += markup(this_format % tuple(vals), colour)
 
         return ds
+
+    @staticmethod
+    def jobHasChildrenTest(jobDirectory, datafileName):
```
milliams: Should this be named `jobHasChildren()`?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -517,3 +561,52 @@ def _private_display(self, reg_slice, this_format, default_width, markup):
...
+                                subjob_count+=1
+                    else:
+                        subjob_count+=1
+
+        logger.debug("count: %s len: %s" % (subjob_count, len([_folder for _folder in jobDirectoryList if _folder.isdigit()])))
+
+        if subjob_count == len([_folder for _folder in jobDirectoryList if _folder.isdigit()]):
+            return subjob_count
+        else:
+            raise GangaException("Missing subjobs data file in %s" % jobDirectory)
```
milliams: Should this be `GangaException`? I forget the rule exactly but I thought that `GangaException` was reserved for things that the user directly caused (like missing arguments to functions etc.) rather than behind-the-scenes things.

python/Ganga/test/GPI/TutorialTests.py
```diff
@@ -106,6 +106,7 @@ def test_c_JobManipulation(self):
         # -- JOBMANIPULATION JOBSLICING STOP
 
         # -- JOBMANIPULATION RESUBMIT START
+        jobs(0).kill()
```
milliams: This should be before the `# -- JOBMANIPULATION RESUBMIT START` line so that it doesn't show up in that doc section.

milliams:
I've merged the small cleanup PRs so you can go ahead and update this PR now.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: Given it's running the test for the children existing and this triggers the test to run I'd prefer to have this name, I'll update the docs to reflect this

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: In this case shall we open a PR to check all uses of this exception type and to document this?

rob-c:
Now the task tests are locking up after this PR... I'll try and investigate

rob-c:
retest this please

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: I've suspected this for a while, performing a check using this often causes significant performance drops. Which reminds me I should probably change the LHCb splitter code in time to use object ids rather than full string comparisons

rob-c:
I'm hoping to look to merge this PR today at some stage unless there are any major changes to be made?

I've changed the unittest to a) reflect the fact  the unittest code was moved without being updated and b) to place the gangadir in the ```ws``` in Jenkins to make it easier to pick up  the output of failed tests through the browser if needed.

I'm not seeing any tests fail and this addresses some XML related bugs so I'd be happier once this has been merged.

milliams:
Changing the default `gangadir` to be the workspace directory won't help much as the next run of PR Builder will immediately wipe its contents. Also, this is affecting more than just Jenkins, it will make all test runs put the gangadir into the code working directory, even when running them locally.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
milliams: That's not how function naming works. The name should reflect the API, not the implementation. If the purpose of this function is to answer the question "does this job have children" then the function should be named `jobHasChildren`. But even then, this seems like a strange place to ask this question; this class is always attached to a job object so the user should ask *that* object if it has children.

This function seems to answering the question above while also checking for the presence of the relevant directories (not raising an error if they're not found, simply printing a warning) but if there's no job directory then saying that the job has no children. That's a very confusing function to know how to use correctly.

If the API user wants to know of the job has children then they should ask *that*. if they want to know if there are any missing subjob directories then they should ask *that*. If they want to check for the presence of the job directory then they should check *that*. Combining all three into one function named `jobHasChildrenTest()` does not help with the complexity of the code.

rob-c:
Yes, this is not a bad thing in my opinion. Do we really need to have the tests here restricted to the old testing path?
I know the PR build will wipe the contents but we've got this issue already with the failed test data not persisting and accessing the file from the ws in Jenkins is the easiest way to get the data imo.
(I'm not for adding timestamps or something to the gangadir path but this requires cleaning up after ourselves which will just get messy.)

Can you suggest a better solution for getting access to the data for a failed build?

milliams:
No, I just wanted to make sure we were aware of the pros and cons. I can add an entry to my local `.gitignore` to avoid the gangadir polluting things.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
rob-c: I can change the name of the function to ```checkSubJobXML``` and then have a direct call to determine the number of  children in the XML reader.
The whole point of SubJobXMLList class is to care about how the subjobs are stored/accessed/found when on disk. You can't sensibly ask a parent job without starting to reproduce functionality which belongs here. We used to do it elsewhere but then we had no lazy loading for subjobs at all.

milliams: `checkSubJobXML` is a much better name.

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -506,3 +556,54 @@ def _private_display(self, reg_slice, this_format, default_width, markup):
...
+        for dir_entry in jobDirectoryList:
+            if dir_entry.isdigit():
+                sj_dir = path.join(jobDirectory, dir_entry)
+                if path.isdir(sj_dir):
+                    if checkDataFiles:
+                        data_file_path = path.join(sj_dir, datafileName)
+                        if path.isfile(data_file_path):
+                            subjob_count+=1
+                        else:
+                            if path.isfile(data_file_path+'~'):
```
milliams: This `else: if:` could be an `elif:`

rob-c: Yes,  development remnant

rob-c:
OK, is there anything else holding up this PR or can I push it once the tests pass?

python/Ganga/Core/GangaRepository/SubJobXMLList.py
```diff
@@ -239,24 +270,11 @@ def __len__(self):
             if this_time == last_time:
                 return self._stored_len[1]
 
-        from os import listdir, path
+        subjob_count = 0
```
milliams: There's no need for this line. it is redefined a few lines later.

PR #500, opened 2016-05-18
Use pytest features for running tests

milliams:
This PR introduces the use of pytest fixtures as a way of setting up tests. pytest fixtures are an alternative to `unittest.TestCase`-style classes that offer more flexibility and composability. If you're not familiar with them I suggest giving [this post from pythontesting.net](http://pythontesting.net/framework/pytest/pytest-fixtures-nuts-bolts/) a read and following some of the links at the beginning of that post.

The gist of it is that:
```python
def test_job_create(gpi):
    j = gpi.Job()
    assert j.id == 0
```
is a full working test module. Alternatively this can be written as:
```python
def test_job_create(gpi):
    from Ganga.GPI import Job
    j = Job()
    assert j.id == 0
```
or
```python
@pytest.mark.usefixtures('gpi')
def test_job_create():
    from Ganga.GPI import Job
    j = Job()
    assert j.id == 0
```

Behind the scenes this is calling the same `start_ganga()` and `stop_ganga()` functions that `GangaUnitTest` is currently using. Overall I have seen a definite speed improvement with these fixtures (up to 10%), especially on the smaller tests where a lot of time was being spent with the `unittest` boilerplate.

This also introduces an `@add_config()` decorator for these test function which is used like
```python
@add_config([('Configuration', 'user', 'test')])
def test_add_config(gpi):
    assert gpi.config['Configuration'].user == 'test'
```

This does not remove the current `GangaUnitTest` class so that can still be used.


python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -112,6 +112,11 @@ def start_ganga(gangadir_for_test, extra_opts=[]):
     if missing_cred:
         raise Exception("Failed due to missing credentials %s" % str(missing_cred))
 
+    # Make sure that all the config options are really set.
+    # Some from plugins may not have taken during startup
```
rob-c: I'm concerned by this, what config options fall into this category and how is this different to what is being done within normal Ganga bootstrap?

python/Ganga/test/GPI/Loading/TestLazyLoading.py
```diff
@@ -1,97 +1,80 @@
...
+            _ = jobs(0)._dirty
 
-        assert raw_j._getRegistry().has_loaded(raw_j) is False
+        assert not raw_j._getRegistry().has_loaded(raw_j)
 
         raw_j.printSummaryTree()
 
-        assert  raw_j._getRegistry().has_loaded(raw_j) is True
-
-    def test_e_JobRemoval(self):
```
rob-c: I wanted to keep this explicit as the lazy loading can (in the past) muck with the removal of a job when there are problems.
I know this is normally removed in the shutdown method but I've found most of the unittest tools a pain to show the full error when something in teardown has gone wrong.
Does this show full traces from within the teardown as it's not inside the unittest code?

rob-c:
Looks good, not something I'm familiar with but if it has clear benefits I'm all for it. A few questions:

Is there a reason not to stick to ```GPI``` over ```gpi``` as it seems more readable and that the code can be taken from a test and run directly in IPython without the need to do ```import Ganga.GPI as gpi``` (granted you probably still need ```from Ganga import GPI```, maybe it's worth maybe adding a hook into the GPI to refer to the GPI so that the tests are runnable just with a copy/paste)

Other than removing the boilerplate I'm assuming this is on a per test-function situation that the unittest boilerplate is reduced?
(I still don't understand how/if using this Ganga restarts between the different tests in the Test class and I suppose I won't unless I pull the branch and start playing).

Would it be possible to remove the Ganga startup/shutdown between tests for tests in a class that aren't concerned with XML/Registry/Repository testing?
This should gain us more than 10% in some areas I would have thought.

python/Ganga/test/GPI/Loading/TestLazyLoading.py
milliams: The traces should be shown correctly, regardless of where an exception is raised, yes. I just tested it and raising an exception in the finaliser will show the stacktrace correctly.

python/Ganga/testlib/GangaUnitTest.py
milliams: The issue is that if a test sets `[DIRAC]DiracEnvFile='something'` in its `setUp` then this function will pass it to `setConfigOption`. However, if at that point the DIRAC plugin has not been loaded, this will fail silently and when the test starts, the setting will not have been set. By line 115 the plugin will have been loaded and so it is safe to set it to the requested value.

rob-c: I'm concerned by this because to my mind we're using the correct config here from the bootstrap which doesn't exhibit this.
Is this due to restarting Ganga? In which case I'd say this should be in the else statement above (although yes, technically it doesn't cause bad things here)
Where exactly are the plugins loaded here again?

milliams:
With the example I gave above of
```python
def test_job_create(gpi):
    from Ganga.GPI import Job
    j = Job()
    assert j.id == 0
```
it is directly copy-and-pasteable into a live Ganga session (as long as you don't copy the `from .. import ...` bit). In the future we're likely to move more towards using little ganga at which point a `ganga` fixture which exactly emulates that could be used. It is possible for us to inject the GPI into the test function directly so that we don't even need the `import` line but I'm trying to avoid too much magic.

If you are in a module with test functions in it, the setup and teardown happening per-function is defined because the `gpi` fixture is defined as having a `scope='function'`. This means a separate `gpi` fixture is provided to each function. We could easily provide a `gpi_module` fixture with a different scope so that only a single fixture object is created per module and so there would be less overhead (we could still mix-and-match these in different modules). However, at the moment it's nice that we know that each test function is running in it's own (mostly) clean environment.

python/Ganga/testlib/GangaUnitTest.py
milliams: I believe (although the plugin loading is messy and multi-parted) that it happens in `bootstrap()`. The issue is likely coming from the fact that we set the config in a different way in these files to how we do in a real live Ganga session. We've also not until now been testing setting the config of plugins using this method until now. 

If this line is controversial, I'm happy to remove it as I don't believe it's needed for any of the tests in this PR.

rob-c:
frankly getting rid of that import line in the tests is something I've wanted since day 1 if it's possible, even better it makes the tests agnostic to the interface which is the way it should be imo. then behind the scenes we can change between ```ganga``` and ```GPI``` and the tests don't have to change

python/Ganga/testlib/GangaUnitTest.py
rob-c: It's not controversial, I'm just very keen to make sure this is understood as I've found in the past the config system rarely does what I'd expect it to in some cases.
If this is due to the plugins then OK, at least it's the loading of the plugins causing it to happen.
Aside from the config there's little here that isn't in the main code so we're not that different thankfully

milliams:
The problem with the magic is that it makes these test functions drift further away from being 'the Python programming language' and instead 'the Ganga programming language'. Using any IDE or static analysis tool will see `Job` and will have no idea where that name is defined.

I'd like to be moving towards a more Pythonic world where we are using little-`ganga` more so that we can do things exactly as we would in a little-`ganga` script like:
```python
import ganga
ganga.start_monitoring()
j = ganga.Job()
j.submit()
sleep(10)
assert j.status == 'completed'
```
which, with some appropriately written fixtures, could become:
```python
@pytest.mark.usefixtures('monitoring')
def test_little_ganga():
    import ganga
    j = ganga.Job()
    j.submit()
    sleep(10)
    assert j.status == 'completed'
```
The alternative is moving towards a system more like the old `foo.gpi` files and we write a runner for pytest which finds `*.gpi` files and runs then within the `gpi` fixture.

rob-c:
I don't think I agree that this is going backwards, the interface should be exposed in such a way that a test-suite should resolve it correctly. This is overhead it feels like clutter we can do away with given we know the tests are run from in-front of some sort of interface, although for 1 or 2 lines per test I'm not going to argue this.
Static analysers and such are incompatible without some pre-generated dict for the api surely, and we don't really run them on the tests do we?

milliams:
I'm open to the idea of implicitly injecting the GPI items into the function's global namespace. I'd rather leave it to another PR though so that we can make sure we're doing the right thing with it.

PR #502, opened 2016-05-20
Fix execute's update_env

milliams:
When working to port some of the DIRAC tests I came across a feature of `Ganga.Utility.execute.execute()` which allows it to update a dictionary based on the environment of the called command. This was accidentally broken in [`c637500`](https://github.com/ganga-devs/ganga/commit/c637500b68ebf9bc5e45ccbf03559868152d1949#diff-09802c04483863469997ddb0cb23bea3) but since the Dirac test was the only thing using it, it went unnoticed.

This PR fixes that command and then changes the two existing uses of `get_env_from_arg()` to use `update_env` instead.

Historically, it seems that `update_env` was added to the queues system's version of `execute()` in b46aa4ec327a494135e3092a2f02ec968949dd41 by @alexanderrichards before being migrated to the standard `execute()` function some time later.

### DIFF
--- a/python/Ganga/Utility/Shell.py
+++ b/python/Ganga/Utility/Shell.py
@@ -40,6 +40,9 @@ import tempfile
 import time
 import signal
 import subprocess
+
+from Ganga.Utility.execute import execute
+
 import Ganga.Utility.logging
 logger = Ganga.Utility.logging.getLogger()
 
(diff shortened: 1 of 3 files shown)


rob-c:
Yeah the LHCb apps have some issues with this update_env... I suspect this can be resolved but I avoided it due to the SetupProject stuff being involved. Is this best solved in LHCb or Core?

milliams:
I expect that this should be fixed in LHCb. I think the problem is a distinction between sourcing and executing a shell script. `update_env` is executing it but I think it needs to be sourced (or vice versa). I'll have a look into it.

rob-c:
If we're looking at this level of change we should probably look at the difference between the Shell and execute files, I think we can probably afford to drop one in the long term if we're intelligent

python/Ganga/Utility/Shell.py
```diff
@@ -40,6 +40,9 @@
 import time
 import signal
 import subprocess
+
+from Ganga.Utility.execute import execute
```
rob-c: If we're going to import this why not make most of the functions here wrap around to it as well? It would be nice to have just 1 place handling all of the subprocess calls so we don't have to worry about small bugs

milliams:
I agree that we should be able to merge `Shell` and `execute` if we're careful. For now their APIs are different enough that it would be a lot of work.

PR #507, opened 2016-05-23
Feature/no multiple mon loops

### LINKED ISSUE #236: Multiple Ganga sessions
Hi all,

As per #235 I think it's worth bringing up multiple concurrent Ganga sessions.

I would like to propose that if Ganga finds an active session lock that it should by default disable the monitoring thread as well as informing the user. This should prevent some level of repository clashes that we see if someone forgets and launches 2 versions of ganga in parallel.

I think a short message informing the user that they've started in a _read-only_ mode and that they can enable monitoring would be useful if we do this.

I think this also strongly merits that we check that both ```enableMonitoring()``` and ```disableMonitoring()``` do what is advertised.
I've been working on a test for this and have spotted a few minor problems which I'd like to put into a smaller PR in the near future.

Does everyone else think this sounds sane?

Thanks,

Rob

rob-c:
This PR turns off the monitoring loop when other another Ganga session is found to be running on a repo.
This should be possible to override by typing ```enableMonitoring``` from the interactive prompt if the monitoring loop is expected to run no matter what.

I've also included some code cleanup I have against ```SessionLock``` as well as a synchronize method against the GangaThread as I've seen threads change running state whilst running based upon the input from other threads so I thought it best just to keep each thread consistent even if this should rarely happen.

Fixes #236

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -959,12 +959,6 @@ class Registry(object):
                 s += ", %i other concurrent sessions:\n * %s" % (len(other_sessions), "\n * ".join(other_sessions))
         return s
 
-    def get_other_sessions(self):
-        """
-        Returns a list of information describing other sessions which are currently active
-        """
-        return self.repository.get_other_sessions()
-
     @synchronised
     def has_loaded(self, obj):
         """Returns True/False for if a given object has been fully loaded by the Registry.
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -997,9 +997,17 @@ class Registry(object):
                 s += ", %i other concurrent sessions:\n * %s" % (len(other_sessions), "\n * ".join(other_sessions))
         return s
 
+    def other_session_active(self):
+        """ Returns a True/False for if other sessions have been detected from the Repository """
+        if self.repository.get_other_sessions():
+            return True
+        else:
+            return False
+
     def print_other_sessions(self):
-        other_sessions = self.repository.get_other_sessions()
-        if len(other_sessions) > 0:
+        """ Prints a warning if other active sessions have been detected """
+        if self.other_session_active():
+            other_sessions = self.repository.get_other_sessions()
             logger.warning("%i other concurrent sessions:\n * %s" % (len(other_sessions), "\n * ".join(other_sessions)))
 
     def has_loaded(self, obj):
--- a/python/Ganga/Core/GangaRepository/SessionLock.py
+++ b/python/Ganga/Core/GangaRepository/SessionLock.py
@@ -43,7 +43,7 @@ except ConfigError, err:
 
 session_lock_refresher = None
 
-def getGlobalLockRef(session_name, sdir, gfn, _on_afs):
+def setupGlobalLockRef(session_name, sdir, gfn, _on_afs):
     global session_lock_refresher
     if session_lock_refresher is None:
         try:
@@ -404,9 +404,9 @@ class SessionLockManager(object):
                 logger.debug("Startup Session Exception: %s" % err)
                 raise RepositoryError(self.repo, "Error on session file '%s' creation: %s" % (self.fn, err))
 
-            local_session_lock_refresher = getGlobalLockRef(self.session_name, self.sdir, self.gfn, self.afs)
+            setupGlobalLockRef(self.session_name, self.sdir, self.gfn, self.afs)
 
-            local_session_lock_refresher.addRepo(self.fn, self.repo)
+            session_lock_refresher.addRepo(self.fn, self.repo)
             self.session_write()
         finally:
             self.global_lock_release()
--- a/python/Ganga/Core/GangaRepository/SessionLock.py
+++ b/python/Ganga/Core/GangaRepository/SessionLock.py
@@ -422,8 +422,7 @@ class SessionLockManager(object):
         self.locked = set()
         try:
             global session_lock_refresher
-            if session_lock_refresher and not session_lock_refresher.should_stop():
-                session_lock_refresher.stop()
+            session_lock_refresher.stop()
             if session_lock_refresher is not None:
                 # try:
                 session_lock_refresher.removeRepo(self.fn, self.repo)
--- a/python/Ganga/Core/GangaRepository/SessionLock.py
+++ b/python/Ganga/Core/GangaRepository/SessionLock.py
@@ -28,28 +28,8 @@ from Ganga.Utility.logging import getLogger
 from Ganga.Utility.Config.Config import getConfig, ConfigError
 from Ganga.GPIDev.Base.Proxy import getName
 
-try:
-    from Ganga.Core.GangaThread import GangaThread
-    from Ganga.Core.GangaRepository import RepositoryError
-except ImportError:
-    from threading import Thread
-
-    print("IMPORT ERROR SHOULD NOT OCCUR IN PRODUCTION CODE!!!!!!!!!!!!!!!!!!!!!!")
-    
-
-    class GangaThread(Thread):
-    
-        def __init__(self, name):
-            self.name = name
-            super(GangaThread, self).__init__()
-
-        def should_stop(self):
-            return False
-
-
-    class RepositoryError(Exception):
-        pass
-
+from Ganga.Core.GangaThread import GangaThread
+from Ganga.Core.GangaRepository import RepositoryError
 
 logger = getLogger()
 
--- a/python/Ganga/Runtime/Repository_runtime.py
+++ b/python/Ganga/Runtime/Repository_runtime.py
@@ -134,9 +134,11 @@ def bootstrap():
         registry.startup()
         logger.debug("started " + registry.info(full=False))
         if registry.name == "prep":
-            registry.print_other_sessions()
+            # Get a list of strings describing other running sessions
+            other_sessions = registry.get_other_sessions()
 
-            if registry.other_session_active():
+            if other_sessions:
+                logger.warning("%i other concurrent sessions:\n * %s" % (len(other_sessions), "\n * ".join(other_sessions)))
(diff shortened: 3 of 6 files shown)


python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -877,9 +877,17 @@ def info(self, full=False):
...
+            return True
+        else:
+            return False
+
     def print_other_sessions(self):
-        other_sessions = self.repository.get_other_sessions()
-        if len(other_sessions) > 0:
+        """ Prints a warning if other active sessions have been detected """
+        if self.other_session_active():
+            other_sessions = self.repository.get_other_sessions()
```
milliams: There's no need for a new function here. It could simply be:
```python
other_sessions = self.repository.get_other_sessions()
if other_sessions:
    logger.warning("%i other concurrent sessions:\n * %s" % (len(other_sessions), "\n * ".join(other_sessions)))
```

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -28,28 +28,8 @@
...
-
-        def should_stop(self):
-            return False
-
-
-    class RepositoryError(Exception):
-        pass
-
+from Ganga.Core.GangaThread import GangaThread
+from Ganga.Core.GangaRepository import RepositoryError
```
milliams: I'm happy about this removal :smile:

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -424,15 +404,14 @@ def startup(self):
                 logger.debug("Startup Session Exception: %s" % err)
                 raise RepositoryError(self.repo, "Error on session file '%s' creation: %s" % (self.fn, err))
 
-            session_lock_refresher = getGlobalLockRef(self.session_name, self.sdir, self.gfn, self.afs)
+            local_session_lock_refresher = getGlobalLockRef(self.session_name, self.sdir, self.gfn, self.afs)
```
milliams: Why the rename? This is still the same global object.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: No there isn't but this is a guaranteed test that the check was working. Functionally there is no difference and the boolean is needed elsewhere outside of the class.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: This was to track the fact that this is a local variable effectively and doesn't confuse the local variable with the global one.
We could ignore the returned value from ```getGlobalLockRef``` and simply get the global object again but it's returned by this function so we might as well use it.
This helps make it clear where the object is coming from and that we're not reassigning the global object here. (there is no need for the global)

python/Ganga/Core/GangaRepository/Registry.py
milliams: I'd rather not add new functions which are doing nothing but `return bool(self.repository.get_other_sessions())`. It's adding complexity to the API when we should we working to simplify wherever possible.

In `Repository_runtime.py` the code could just read:
```python
if registry.repository.get_other_sessions():
    logger.warning("Multiple Ganga sessions detected. The Monitoring Thread is being disabled.")
    logger.warning("Type 'enableMonitoring' to restart")
    setConfigOption('PollThread', 'autostart', False)
```

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: Frankly I don't know why it was ever there but I think it's time we can just remove it and move on

python/Ganga/Core/GangaRepository/Registry.py
rob-c: I think we have a strong disagreement about what it means to simplify the API in this case I don't think the registry code need know what external instances there are only that there are some. I don't think this information should be passed up from the SessionLock as at the moment it's very implementation specific. I can see that ```def other_session_active(self):``` can be simplified but I don't think we should be passing knowledge of other session locks outside of the sessionlock code.

python/Ganga/Core/GangaRepository/SessionLock.py
milliams: Given that `getGlobalLockRef()` itself updates the value of the global `session_lock_refresher`, it could be simpler to do:
```python
getGlobalLockRef(self.session_name, self.sdir, self.gfn, self.afs)
session_lock_refresher.addRepo(self.fn, self.repo)
```
At which point it might be worth renaming `getGlobalLockRef` to `setupGlobalLockRef`.

python/Ganga/Core/GangaRepository/Registry.py
milliams: If you rally want to hide the contents of the session list then instead how about adding a method to `GangaRepository`:
```python
def has_other_sessions(self):
    """Are there other running sessions"""
    return bool(self.get_other_sessions())
```
It makes much more sense being in the repository than in the registry. This function can then be used anywhere you want a boolean result.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: Well yes but until now we were doing something strange. I can add this change

python/Ganga/Core/GangaRepository/Registry.py
rob-c: Then why are we accessing a repository from within a registry, I know it's during bootstrap which could sensibly have knowledge of this but I dislike subverting the Registry to get details on the Repo, it causes confusion.

milliams: Because it's the repository that we care about. Registries don't have 'other sessions' since they are in in-memory construct. It is the repository that can care about other processes and we want to know if there are other processes reading the files on dick. Accessing the repository from the registry is fine; it's accessing the registry from the repository that is bad in my opinion.

rob-c: I worry that (not important here)  this encorages getting the repo which can potentially be thread unsafe as all the locking is in the Registry not the Repo

rob-c:
OK, I've moved all of the printing of any information about multiple sessions to the Runtime_regsistry code which should handle the printing and I've replaced the print method with a get method in the Registry as the Registry doesn't particularly care about what you do with a list of strings describing each of the multiple sessions found accessing the same repo.
This simplifies the logic greatly imo

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -443,7 +422,8 @@ def shutdown(self):
         self.locked = set()
         try:
             global session_lock_refresher
-            session_lock_refresher.stop()
+            if session_lock_refresher and not session_lock_refresher.should_stop():
+                session_lock_refresher.stop()
```
milliams: Is there any harm in calling `session_lock_refresher.stop()` if it's already been called?

python/Ganga/Runtime/Repository_runtime.py
```diff
@@ -134,7 +134,15 @@ def bootstrap():
         registry.startup()
         logger.debug("started " + registry.info(full=False))
         if registry.name == "prep":
-            registry.print_other_sessions()
+            # Get a list of strings describing other running sessions
+            other_sessions = registry.get_other_sessions()
```
milliams: If this was
```python
other_sessions = registry.repository.get_other_sessions()
```
then it would remove the need for the function you've had to add to the registry.

rob-c: Again, I _really_ want to avoid calling the repository from in front of the registry. We shouldn't know anything about the internals of the registry in order to do things like this. We potentially never needed the pint function in this case but it makes sense to have this distinction here I think.

milliams: And I would argue that the *registry* should not have to care about other *repository* sessions. We do care about the internals of the registry (in as much as the repository is an internal of the registry) since it is as the repository level where this is dangerous.

Maybe this warning would make a lot more sense being in `PrepRegistry.startup()`? Especially given that `if registry.name == "prep"` code-smell above.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: No idea, but from my understanding there is no need to even if the function is safe, so why do it?

milliams: I'm trying to avoid noisy code if there's no benefit. Increased code makes Ganga slower and harder to maintain. `GangaThread.stop()` already makes this check in it's first line.

rob-c: I'd argue a boolean check here doesn't really fit that criteria given there's a clearly expected behaviour but I get the point, I'll revert

python/Ganga/Runtime/Repository_runtime.py
rob-c: Frankly leaving it in the prep registry check was for a minimal change. We want to disable the monitoring at a higher level than an individual repo init imo incase we ever want to do something such as drop the prepared state repo.
If we're going to start accessing the repo and subverting the registry than I'll move for that option, but I still think this is bad practice.

milliams: Well, if it should be restricted to the prep registry then it should be in `PrepRegistry.startup()` but if it should apply to all registries then `Regsitry.startup()` would be a sensible place.

rob-c: No because then you spam the user with un-needed warnings... I'll change the PR...

rob-c:
OK, If there are no objections I'll update and merge this tomorrow once the github issues are over

rob-c:
I started hitting problems with the tests stalling again on this PR so I moved the flush_all command to the Registry shutdown code.
All of the tests are now passing. Are there any objections to the merge?

milliams:
Go for it.

PR #508, opened 2016-05-23
Simplify default object lookup

milliams:
When getting the default value of a schema item, rather than relying on the GPI being populated and then having to strip the proxy from the result, instead search in the plugin system for the requested class object.

### DIFF
--- a/python/Ganga/GPIDev/Schema/Schema.py
+++ b/python/Ganga/GPIDev/Schema/Schema.py
@@ -307,7 +307,7 @@ class Schema(object):
             from Ganga.GPIDev.Base.Proxy import isType, getRuntimeGPIObject, stripProxy, getName
             from Ganga.GPIDev.Base.Objects import Node
             if isinstance(defvalue, Node):
-                return stripProxy(getRuntimeGPIObject(getName(defvalue)))
+                return allPlugins.find(item['category'], getName(defvalue))()
             else:
                 return copy.deepcopy(defvalue)
         except ImportError:
--- a/python/Ganga/GPIDev/Schema/Schema.py
+++ b/python/Ganga/GPIDev/Schema/Schema.py
@@ -303,15 +303,7 @@ class Schema(object):
 
         # make a copy of the default value (to avoid strange effects if the
         # original modified)
-        try:
-            from Ganga.GPIDev.Base.Proxy import isType, getRuntimeGPIObject, stripProxy, getName
-            from Ganga.GPIDev.Base.Objects import Node
-            if isinstance(defvalue, Node):
-                return allPlugins.find(item['category'], getName(defvalue))()
-            else:
-                return copy.deepcopy(defvalue)
-        except ImportError:
-            return copy.deepcopy(defvalue)
+        return copy.deepcopy(defvalue)
 
 
 # Items in schema may be either Components,Simples, Files or BindingItems.


rob-c:
Do we know for certain there are no more string defaults which need to be evaluated anywhere in the codebase?

milliams:
I think so but this PR doesn't affect that anyway. All this is doing is given `'LocalFile'` instead of looking it up in `Ganga.GPI.__dict__` is instead looking in `allPlugins`.

I would have rather changed this to `return allPlugins.find(item['category'], item['defvalue'])()` but this is broken by the `defvalue=[]` we have everywhere on `sequence` items. Since sequences need looking at anyway later I went with the minimal change.

rob-c:
No, this will effect the likes of ```'str'``` and ```'int'``` won't it? which whilst bad and shouldn't be here appear to be valid

python/Ganga/GPIDev/Schema/Schema.py
```diff
@@ -307,7 +307,7 @@ def _getDefaultValueInternal(self, attr, val=None, check=False):
             from Ganga.GPIDev.Base.Proxy import isType, getRuntimeGPIObject, stripProxy, getName
             from Ganga.GPIDev.Base.Objects import Node
             if isinstance(defvalue, Node):
-                return stripProxy(getRuntimeGPIObject(getName(defvalue)))
+                return allPlugins.find(item['category'], getName(defvalue))()
```
rob-c: Can we check that it's not already an instance incase there is an instance stored here, also you would need to take a deepcopy to avoid problems with the default possibly being changed

milliams:
This is only affecting `ComponentItem`s so they will not have any defaults like that.

python/Ganga/GPIDev/Schema/Schema.py
milliams: The plugin system returns the class object, not any sort of cached object. The `()` at the end are creating a brand new object.

rob-c: ah ok

rob-c:
Aren't component Items covered by the if statement starting at: https://github.com/ganga-devs/ganga/pull/508/files#diff-ffafccc1e37d3151cc60513e63c9e5e5R277  The only component items getting here should be sequence ones shouldn't they?

milliams:
Yes, but only items with a default value of a `Node` subclass will get to the affected line (`if isinstance(defvalue, Node):`).

rob-c:
Then asking the stupid question, why are we performing a lookup if we know we could just create a new instance of this class object?
Can't we just do something closer to ```return defvalue.__class__()``` and avoid the lookup all together?

milliams:
We could but this function is going to be moving in the direction of using the plugin system more directly. `defvalue` should only ever be a name of a plugin (a `str`) or `None`. This change is one of the first steps in the right direction towards massively simpler code in this function since it is so often a hotspot for performance problems.

rob-c:
I'm sorry I don't see how performing a lookup when we already have access to the class is an improvement?

If we had a string here I'd say great stick it into the plugin system but we have effectively already resolved it elsewhere which introduces more wasted cycles getting the class instance of the object again from the plugin system.

milliams:
Because the whole function for `ComponentItem`s *should* be as simple as:
```python
if item['sequence']:
    return GangaList()
else:
    return allPlugins.find(item['category'], item['defvalue'])()
```
(modulo overwriting the default in the config).

What is currently 120 lines of code should be 10 or 20. The change in this PR is moving in the direction of that simpler future.

The fact is that this change is also needed to do plugin lookups in the case that the full bootstrap has not been run, such as during unit tests. Particularly the upcoming Dirac tests I'm porting.

If you'd rather I spend the time doing a full rewrite of the `getDefaultValue` system to show what I mean then I will but for now I thought a small change in that direction would be enough.

We already have wasted cycles (more even) due to the use of `getRuntimeGPIObject` here.

milliams:
retest this please

rob-c:
Yes we're wasting cycles and I'm trying to make sure we stop wasting as many as possible.
My entire point is that we're replacing an expensive function with a less expensive function rather than just a lookup here.

I'm not promoting a full rewrite now (yes one is clearly needed) but doing the equivalent of:
```PluginDict['cat'][getName(object)]()``` vs ```object.__class__()``` is not something I think is good idea.

If we have the object already (and it's instantiated) why are we using the name to lookup the class object?
If you want to stick to this fine, commit the PR I just think with the code in it's current form we can do better.

milliams:
I've made the simplification change. It worked on my local tests so we'll see how ganga-ci does.

milliams:
retest this please

milliams:
Ok, with that change it seems to work fine. Are there any more comments on this?

alexanderrichards:
> return copy.deepcopy(defvalue)

This is what I do in my rewrite stuff so I think it's fine and as you say a lot simpler. All seems sensible to me :+1:

rob-c:
If nothing is broken then I look forward to reducing the bottleneck here somewhat :+1:

PR #509, opened 2016-05-24
Small DIRAC tidy

milliams:
These are some change I've made in order to get the legacy DIRAC tests ported over (PR incoming soon). The changes to the imports are needed in order to correctly mock the `execute` function. 

The removal of the call to `getReplicas()` was done since the mock system was flagging that `execute` was being called multiple times for a single call to `replicate()`. It seems like the calls were added in order to be able to call `replicate()` even if `locations` was not set (by requesting the current replica SEs from DIRAC). The change was made in ecbb34b71bf14247205d9a8f81844e73939bd979 but the commit message doesn't help with working out the exact reason. In fact it doesn't matter too much since it's possible to call DIRAC's [`replicateFile()`](http://diracgrid.org/files/docs/CodeDocumentation/Interfaces/API/Dirac.html#DIRAC.Interfaces.API.Dirac.Dirac.replicateFile) without a source SE anyway (since at least 2008).

### DIFF
--- a/python/GangaDirac/Lib/Splitters/SplitterUtils.py
+++ b/python/GangaDirac/Lib/Splitters/SplitterUtils.py
@@ -3,7 +3,7 @@ from Ganga.GPIDev.Adapters.ISplitter import SplittingError
 from GangaDirac.Lib.Utilities.DiracUtilities import execute
 from GangaDirac.Lib.Backends.DiracUtils import result_ok
 from Ganga.Utility.Config import getConfig
-from GangaDirac.Lib.Files import DiracFile
+from GangaDirac.Lib.Files.DiracFile import DiracFile
 from Ganga.Utility.logging import getLogger
 logger = getLogger()
 
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -632,20 +632,23 @@ class DiracFile(IGangaFile):
     def replicate(self, destSE):
         """
         Replicate this file from self.locations[0] to destSE
+
+        If self.locations is empty, let DIRAC work it out.
         """
-        if not self.locations:
-            if self.lfn != '':
-                self.getReplicas()
-            else:
+
+        if not self.lfn:
+            if not self.locations:
                 raise GangaException('Can\'t replicate a file if it isn\'t already on a DIRAC SE, upload it first')
-        if self.lfn == '':
-            raise GangaException('Must supply an lfn to replicate')
+            else:
+                raise GangaException('Must supply an lfn to replicate')
 
         logger.info("Replicating file %s to %s" % (self.lfn, destSE))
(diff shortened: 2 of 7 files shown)


python/GangaDirac/Lib/Splitters/SplitterUtils.py
```diff
@@ -3,7 +3,7 @@
 from GangaDirac.Lib.Utilities.DiracUtilities import execute
 from GangaDirac.Lib.Backends.DiracUtils import result_ok
 from Ganga.Utility.Config import getConfig
-from GangaDirac.Lib.Files import DiracFile
+from GangaDirac.Lib.Files.DiracFile import DiracFile
```
rob-c: If this changes in this way is it corrected elsewhere in the file, if it doesn't need to be, is it even called?

rob-c: Edit, nvm I see the change in the ```__init__.py``` that's quite subtle

milliams: Indeed it is subtle. This is a common problem in Ganga where we have a class called `Foo` inside a file called `Bar/Foo.py`. This is fine until we, in `Bar/__init.py__`, do `from Foo import Foo` and suddenly `Bar.Foo` is ambiguous.

milliams:
With regards to the call to `replicateFile`, could we simply *always* omit the `sourceSE` parameter? I imagine it's main use in DIRAC is so that power users can replicate a file between two SEs and be in full control over exactly where the source is in case they know there's a fast connection or something. In Ganga, we're always blindly grabbing the zeroth element of `locations` so there's no finesse anyway. It would simplify this function a lot.

rob-c:
The fix was to populate the locations data in the DiracFile rather than risk changing the API around this which was at the time fragile for unknown reasons.
Less calls to DIRAC to do the same thing makes me happy as long as everything now works. Have the commands files been updated or how is this function being called against DIRAC?

milliams:
I've just updated the `DiracCommands.py` file to match the DIRAC API. In the incoming port of `TestDiracFile.py` I'll try to remember to add a test for this special case.

rob-c:
It would make more sense to put the ```sourceSE``` as a keyword argument imo and allow for people to use it and if it's not defined just let DIRAC decide. The use of locations was historical I fixed code that was already using this, if there is a better way then I'm happy.

milliams:
Do you mean adding `sourceSE` as an optional keyword argument to our `DiracFile.replicate()` method and then using that as the argument for the DIRAC call?

rob-c:
I would have said yes, If it's default just let DIRAC decide, if it's selected then the user knows best. We can get rid of the stuff which tries to be too clever imo

milliams:
retest this please

milliams:
Ok, I made that change to `replicate()`. Does that look better to you?

PR #510, opened 2016-05-24
Cleanup atexit handling

drmarkwslater:
This removes the messing about with atexit that was done and switches to just a single atexit handler (`_ganga_run_exitfuncs`) that shuts everything down in an order we control. This is registered in `bootstrap` as a normal atexit handler.

I believe the original code was put so plugins could at their own atexit handlers and then Ganga would shut them down but this was essentially just doing our own version of atexit. After @rob-c 's work, this also became almost entirely obsolete as the Ganga services were getting shutdown by hand anyway. One of the major problems of this original code was that it was shutting down the coverage after the first test, thus our coverage reporting was **very** wrong!

The only downside with this change is that running with coverage now adds ~50% to the time of the tests (assuming the CTracer is used) so I have also changed the PR Builder on the Jenkins box to not do the coverage and kept this for the nightlies.

### DIFF
--- a/python/Ganga/Core/InternalServices/ShutdownManager.py
+++ b/python/Ganga/Core/InternalServices/ShutdownManager.py
@@ -38,9 +38,6 @@ Extend the behaviour of the default *atexit* module to support:
   registers the function with the lowest priority (sys.maxint)
 """
 
-# System imports
-import atexit
-
 # Ganga imports
 from Ganga.Core.GangaThread import GangaThreadPool
 from Ganga.Core.GangaThread.WorkerThreads import _global_queues, shutDownQueues
@@ -66,12 +63,6 @@ def _ganga_run_exitfuncs():
     we want each shutdown function to be run (e.g. to make sure flushing is done) we put each call into a try..except
     and report to the user before continuing.
     """
-
-    # shutdown the threads in the GangaThreadPool
-    try:
-        GangaThreadPool.getInstance().shutdown(should_wait_cb=at_exit_should_wait_cb)
-    except Exception as err:
-        logger.warning("Exception raised during shutdown of GangaThreadPool: %s" % err)
 
     # Flush everything
     try:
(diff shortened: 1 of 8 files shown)


milliams:
I think that `ganga/__init__.py` needs updating too. On which note we should try to get some testing of little-`ganga` some time soon to make sure we don't miss it in future.

drmarkwslater:
Totally correct - fixed!

On that subject though, I had difficulty implementing tests with little ganga as it's difficult to do the tear down properly and you also can't run the import code twice. This has been said before but I think the only way to do this properly is to have a single bootstrap function(s) and single shutdown function(s) that little Ganga and the bootstrap in Runtime just call. We're probably not too far away from this already, certainly with shutdown.

python/Ganga/Core/InternalServices/ShutdownManager.py
```diff
@@ -38,161 +38,116 @@
...
-
-    #print("Shutting Down Ganga Repositories")
-    from Ganga.Runtime import Repository_runtime
-    Repository_runtime.flush_all()
-
-    from Ganga.Utility.logging import getLogger
-    logger = getLogger()
+    # Flush everything
+    try:
+        Repository_runtime.flush_all()
```
rob-c: I think we decided to move this to the registry shutdown after the stopTasks?

drmarkwslater: I haven't updated to develop yet - this will happen soon 😄

drmarkwslater:
So this should hopefully now be mergable. I've found an issue with coverage though that it doesn't seem to like our queued tests. With the PyTracer it is very slow but works but for CTracer it hangs on these tests. Turning off coverage everything works OK (and it also works **with** coverage on my machine, just not no ganga-CI). I've therefore decided to add another deocrator `nocoverage` that will not be run if coverage is being used. I've also added another project in Jenkins to do nightly coverage reports. There's obviously quite a bit of overlap here but I couldn't think of a better way around things. Anyone got any ideas?

Anyway, is this OK to merge?

milliams:
It's a little worrying that there's this conflict between the mutlithreading and the coverage reporting. Was there any sign in the thread trace as to the culprit? Either way, I'm happy for this PR to go in :+1:

rob-c:
I was assuming that this was more of a performance conflict with the coverage and a large number of threads?

drmarkwslater:
This has been cleaned up in #527

PR #513, opened 2016-05-24
Synchronise increase and decrease of ShareRefs

milliams:
There is currently no protection against two threads increasing or decreasing the same ShareRef at the same time. I believe this has been causing a few of the lockups that I have seen in testing PRs and locally.


drmarkwslater:
As long as this doesn't cause lockups then 👍

milliams:
It worked fine on several runs of local testing and it seems to have worked fine on ganga-ci. If there are no other comments I'd like to merge.

PR #514, opened 2016-05-24
Factor out the config file finding in bootstrap

milliams:
This makes no functional change to the bootstrap code, it simply refactors a part of the `configure()` method. It adds a new static method called `get_config_files` which gives the list of all the config files, ready to be passed to `Ganga.Utility.Config.configure()`. This allows us to load the config files in a consistent way, even when we are only running unit tests without having performed the full bootstrap procedure.

### DIFF
--- a/python/ganga/__init__.py
+++ b/python/ganga/__init__.py
@@ -1,10 +1,11 @@
+import os
+
 # Bootstrap all of ganga, setup GPI, registries, etc.
 from Ganga.Utility.Runtime import allRuntimes
-import Ganga.Utility.Config
+from Ganga.Utility.Config import getConfig, setSessionValuesFromFiles
 from Ganga.Utility.logging import getLogger
 from Ganga import _gangaPythonPath
 import Ganga.Core
-from Ganga.Core.GangaRepository import getRegistry
 
 logger = getLogger(modulename=True)
 
@@ -21,9 +22,9 @@ from Ganga.Core.InternalServices import ShutdownManager
 ShutdownManager.install()
 
 system_vars = {}
-for opt in Ganga.Utility.Config.getConfig('System'):
-    system_vars[opt] = Ganga.Utility.Config.getConfig('System')[opt]
-Ganga.Utility.Config.configure([os.path.expanduser('~/.gangarc')], system_vars)
+for opt in getConfig('System'):
+    system_vars[opt] = getConfig('System')[opt]
+setSessionValuesFromFiles([os.path.expanduser('~/.gangarc')], system_vars)
 
 import ganga
 from Ganga.Runtime import plugins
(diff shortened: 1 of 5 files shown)


rob-c:
Would it be possible to add this to ```ganga``` to make sure it gets the correct config?

milliams:
I've made a change to `ganga` to load the config files. It seems to work in my local testing.

milliams:
Any more comments on this?

python/ganga/__init__.py
```diff
@@ -20,8 +20,10 @@ def ganga_license():
 from Ganga.Core.InternalServices import ShutdownManager
 ShutdownManager.install()
 
-## TODO need to implement loading of the config system properly here.
-## loadPlugins and autoPopulateGPI will take this into account when loading objects
+system_vars = {}
+for opt in Ganga.Utility.Config.getConfig('System'):
+    system_vars[opt] = Ganga.Utility.Config.getConfig('System')[opt]
```
rob-c: Isn't this more readable as just ```for opt in getConfig('System'):```?

milliams: Perhaps, but for the call to `configure()`, I thought it best to use the explicit `Ganga.Utility.Config.configure()` rather than just `configure()` to make it clear *what* is being configured.

It becomes the difference between:
```python
from Ganga.Utility.Config import getConfig, configure

system_vars = {}
for opt in getConfig('System'):
    system_vars[opt] = getConfig('System')[opt]
configure([os.path.expanduser('~/.gangarc')], system_vars)  # This is ambiguous in my view
```
and
```python
import Ganga.Utility.Config

system_vars = {}
for opt in Ganga.Utility.Config.getConfig('System'):
    system_vars[opt] = Ganga.Utility.Config.getConfig('System')[opt]
Ganga.Utility.Config.configure([os.path.expanduser('~/.gangarc')], system_vars)
```

Perhaps a compromise would be:
```python
from Ganga.Utility import Config

system_vars = {}
for opt in Config.getConfig('System'):
    system_vars[opt] = Config.getConfig('System')[opt]
Config.configure([os.path.expanduser('~/.gangarc')], system_vars)
```
or
```python
import Ganga.Utility.Config

system_vars = {}
for opt in getConfig('System'):
    system_vars[opt] = getConfig('System')[opt]
Ganga.Utility.Config.configure([os.path.expanduser('~/.gangarc')], system_vars)
```
What are your thoughts?

rob-c: Can we just rename this ```configure``` function to reflect what it's doing in this case? If it's not obvious what it does then rename it to ```setupConfigurationSystem``` or ```applyConfigurationToSession``` or ```addConfigurationFiles``` or something reflecting more what it does (and I don't know what it does so probably all 3 of these names are bad).
Just having the full path to it in the code doesn't help understand what it's doing imo.

milliams: I guess `setSessionValuesFromFiles` would make sense as a name. I'll make that change in this PR and update it.

milliams:
@rob-c, does that look more sensible to you now?

rob-c:
looking good from me

PR #515, opened 2016-05-24
Put pytest tests' gangadir in the same place as other tests

milliams:
This is a small update to a previous change by @rob-c to make sure that the pytest-run tests' gangadir matches that of the other tests.


drmarkwslater:
Looks good to me 👍

Issue #463, opened 2016-05-11
Ganga hanging on RAL T1 cluster - any clues?


twhyntie:
Hello,

I'm trying to run Ganga on the RAL Tier-1 cluster. I've tried both running from CVMFS and installing via the install script `ganga-install` and both hang. Using the `--debug` flag it gets this far:

```
$ . /cvmfs/ganga.cern.ch/runGanga.sh
[...]
2016-05-11 10:15:25,010 MainThread __init__            ::startUpQueues       :18 DEBUG   : Starting Queues
2016-05-11 10:15:25,160 MainThread GangaPlugin         ::add                 :98 DEBUG   : adding plugin Batch (category "backends") 
2016-05-11 10:15:25,210 MainThread Schema              ::_check_type         :544 DEBUG   : valType: <type 'dict'> defValueType: <type 'dict'> name: JobTree.folders
2016-05-11 10:15:25,222 MainThread GangaPlugin         ::add                 :98 DEBUG   : adding plugin JobTree (category "jobtrees") 
2016-05-11 10:15:25,256 MainThread Repository_runtime  ::checkDiskQuota      :92 DEBUG   : Error checking disk partition: invalid literal for int() with base 10: 'Filesystem                                1024-blocks     Used  Available Capacity Mounted on\nnfs1.gridpp.rl.ac.uk:/home/tier1/twhyntie  1441830080 64705208 1362476728       5'
2016-05-11 10:15:25,276 MainThread Repository_runtime  ::checkDiskQuota      :92 DEBUG   : Error checking disk partition: invalid literal for int() with base 10: 'Filesystem                                1024-blocks     Used  Available Capacity Mounted on\nnfs1.gridpp.rl.ac.uk:/home/tier1/twhyntie  1441830080 64705208 1362476728       5'
2016-05-11 10:15:25,277 MainThread Repository_runtime  ::bootstrap           :132 DEBUG   : Registry: prep
2016-05-11 10:15:25,289 MainThread Repository_runtime  ::bootstrap           :133 DEBUG   : Loc: /home/tier1/twhyntie/gangadir/repository/twhyntie/LocalXML
2016-05-11 10:15:25,302 MainThread Registry            ::startup             :930 DEBUG   : metadata startup
2016-05-11 10:15:25,314 MainThread Registry            ::startup             :935 DEBUG   : repo startup
```

[...]

PR #516, opened 2016-05-24
Removed delay_lock_mod function which hung on no filelock FS

### LINKED ISSUE #465: Ganga should fail gracefully on an FS without File Locks
At present, if you try to start Ganga on a file system without file locks, it hangs in the Session lock. See #463. This seems like something we should allow to work.

drmarkwslater:
It was found in #463 that if you run Ganga on an FS without file locks, it hangs. I'm not sure of the exact reason for this but I'm assuming it's to do with the `delay_lock_mod` function which attempts to lock a file many times. In theory, this should fail eventually but that doesn't seem to be what's happening. Looking at the `fcntl` docs and what happened when running the `fcntl.lockf` function in plain python, if it can't acquire a lock straight away then it doesn't seem necessary to keep trying as it probably will never acquire a lock. 

The testing works on my machine and so I believe this to be safe. Let me know if not!

Fixes #465

### DIFF
--- a/python/Ganga/Core/GangaRepository/SessionLock.py
+++ b/python/Ganga/Core/GangaRepository/SessionLock.py
@@ -489,24 +489,6 @@ class SessionLockManager(object):
             except OSError as x:
                 raise RepositoryError(self.repo, "Could not open lock file '%s': %s" % (self.lockfn, x))
 
-    @staticmethod
-    def delay_lock_mod(lockfd, lock_mod):
-        i = 0
-        num_tries = 35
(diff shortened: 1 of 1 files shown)


milliams:
I'm going to guess that AFS silliness was the cause for the attempt at delayed locking.

drmarkwslater:
Possibly - but this code was specifically called when NOT on AFS 😄

drmarkwslater:
Any issues with merging?

milliams:
Go for it.

PR #518, opened 2016-05-24
Don't add the new SE to locations if it's already present

milliams:
As an addendum to #509, this makes sure that the replicated SE is not added to `locations` if it's already there.

### DIFF
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -644,7 +644,8 @@ class DiracFile(IGangaFile):
         logger.info("Replicating file %s to %s" % (self.lfn, destSE))
         stdout = execute('replicateFile("%s", "%s", "%s")' % (self.lfn, destSE, sourceSE))
         if isinstance(stdout, dict) and stdout.get('OK', False) and self.lfn in stdout.get('Value', {'Successful': {}})['Successful']:
-            self.locations.append(destSE)
+            if destSE not in self.locations:
+                self.locations.append(destSE)
             return
         logger.error("Error in replicating file '%s' : %s" % (self.lfn, stdout))
         return stdout


python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -644,7 +644,8 @@ def replicate(self, destSE, sourceSE=''):
         logger.info("Replicating file %s to %s" % (self.lfn, destSE))
         stdout = execute('replicateFile("%s", "%s", "%s")' % (self.lfn, destSE, sourceSE))
         if isinstance(stdout, dict) and stdout.get('OK', False) and self.lfn in stdout.get('Value', {'Successful': {}})['Successful']:
-            self.locations.append(destSE)
+            if destSE not in self.locations:
```
rob-c: Can we make sure we don't try and replicate to their in the first place as this will likely fail in DIRAC and produce a not nice error.
We can catch this higher up and give a nicer error to the user about why we'e not replicating a file.

milliams: As far as I am aware, if you try to replicate a file in DIRAC to an SE where it is already present, it will work just fine. It will return:
```python
{'Failed': {},
 'Successful': {'/gridpp/user/r/robot.general.gangaproject/tmp/test1.txt': {'register': 0,
                                                                            'replicate': 0}}}
```

rob-c: But this is a rather long winded way of getting here than ```if destSE in self.locations: return```, if it's harmless then ok, I suppose in some situation the locations may be out of date enough to reflect where a true lfn replica is

milliams: That was my concern I guess. I'm happy to treat `locations` as a cache of information for the user but I worry that if it gets out of date (maybe a 'remove replica' behind our backs) then we don't want to silently do nothing. On the other hand I do agree with you about avoiding making a call to DIRAC if we can avoid it, however, this particular call (a replicate to an existing replica) it is effectively a no-op at DIRAC's end.

Really, `locations` should be a `set` in which case this wouldn't be an issue any more either way.

milliams:
retest this pelase

PR #522, opened 2016-05-25
Read the env file directly

milliams:
It turns out there was an issue with how I was reading the env file for LHCb. In my Dirac config I was was using the Dirac `bashrc` file which is `source`able but the LHCb DIRAC environment that we cache is not sourceable. This PR fixes it so that our DIRAC code can read both DIRAC `envfile`-style files as well as our LHCb DIRAC cache.

It also avoids having to load a full shell and an external Python interpreter along with a few threads, just read the environment. This is what is being done in `ganga-proxy.py` anyway.

### DIFF
--- a/python/GangaDirac/Lib/Utilities/DiracUtilities.py
+++ b/python/GangaDirac/Lib/Utilities/DiracUtilities.py
@@ -49,9 +49,12 @@ def getDiracEnv(force=False):
             if getConfig('DIRAC')['DiracEnvFile'] != "" and os.path.exists(absolute_path):
 
                 env_dict = {}
-                execute('source {0}'.format(absolute_path), shell=True, env=env_dict, update_env=True)
+                with open(absolute_path) as env_file:
+                    for line in env_file:
+                        varval = line.strip().split('=')
+                        env_dict[varval[0]] = ''.join(varval[1:])
 
-                if env_dict is not None:
+                if env_dict:
                     DIRAC_ENV = env_dict
                 else:
                     logger.error("Error determining DIRAC environment")


milliams:
retest this please

rob-c:
@milliams This is sourcable it's just that the update_env is horribly broken. This was working for me in the past with a source and a print of the environ so I'm tempted to revert back to this over reading the files in this format. Besides this is prone to errors in this format as it doesn't check for bash functions being stored in the env and so causes errors and warnings for users

rob-c:
I can source this file and then access Dirac through the python interface with the resulting environ and I have tracked down that (I think) the update_env isn't passing the file descriptors around from the os.pipe. But I may be wrong. All I know is the upate_env code crashes when it tries to run. (It could be related to the bash warnings/errors when sourcing the file, but it does 'work')

rob-c:
and even worse the Exception is 'pass'-ed in the code so we never noticed. (Possibly my bad I touched this a long time ago)

milliams:
If I do `source ~.cache/LHCbDIRAC_ENV/x86_64-slc6-gcc49-opt/v8r2p27` at the command line I get a whole host of errors:
```
bash: 45: command not found
bash: 31m: command not found
bash: 31m: command not found
bash: 5: command not found
bash: 27:ln=38: command not found
bash: 5: command not found
bash: 51:mh=44: command not found
bash: 38: command not found
bash: 5: command not found
bash: 15:pi=40: command not found
bash: 38: command not found
...
```
This is due to lines like:
```bash
GREP_COLOR=37;45
LESS=-F -g -i -M -R -S -w -X -z-4
```
which are not sourceable bash commands. If they were quoted like:
```bash
GREP_COLOR="37;45"
LESS="-F -g -i -M -R -S -w -X -z-4"
```
then it works correctly. Also, in addition, each of these should have `export` in front of them so that they are exported.

We *could* avoid all this entirely by, instead of in `_store_dirac_environment()` running something like `source SetupProject.sh LHCBDIRAC && printenv > envcache` and then having to start upa  new shell to read `envcache`, we could instead pickle the environ to a file and read that.

rob-c:
We need to rewrite the LHCbDirac env store commands for cmake support anyway so any change would be welcomed.
I would rather store this in some form of source-able file so that we can just read it in the same way as the gridpp instance. Having LHCb being a special case here is just creating more work in the long run.
We have the bashrc from the dirac-ui I think we should have a similar bashrc for the LHCb version of DIRAC.

The 'command not found' errors whilst annoying don't stop the sourcing of the env file from giving an environment which is enough to get our code to run.
(As these errors should be shielded from the user then I wasn't too concerned by this)

milliams:
Ok, so here are some current problems:

1. The LHCb dirac setup can take some time so we want to cache it in some way.
2. The GangaLHCb and GangaDirac environments are currently dealt with in very different ways.
3. When installing Dirac as a user, you have to do the `env > envfile` dance manually.

Here's a proposal:

Both GangaLHCb and GangaDirac have "something we source" (`SetupProject.sh LHCBDIRAC v8r2p27 ROOT` and `dirac_ui/bashrc`). If we source that and cache the result in a Python-readable format (probably a pickle). We can then read that cache in `getDiracEnv()` in a consistent way. The pickle cache is persisted across sessions and within one session, the env is stored in memory.

What do you think? Would this fit into the cmake world?

rob-c:
1. ~Yes This is mainly an LHCb _problem_ but we don't have a gridpp solution better than our current workflow.
 2. No I think that they both point to the environment on disk and this is then sourced. If there is a bug in 1 implementation this is an isolated bug not some inconsistency in how they should be working.
 3. No, this is not required in >=6.1.20 due to the changes I put in you can source the bashrc directly. I do this myself now.

I'm _really_ against putting something into a python file it's creating another step with another file which users will get wrong and will cause more support requirements.

There is no reason why we can't just source a properly sourceable file (bashrc equivent).
The simplest solution is to correctly source the env from a file on disk.

LHCb specific code deals with updating the env file as appropriate. The LHCb plugin just needs to change the env file it generates as it currently does it in a way which is buggy. It's not the end of the world but we shouldn't be changing the workflow for an lhcb specific problem.
(Frankly this is just a case of swapping out our current command for ```lb-run``` and making sure we store the bash environment on disk in a better way.)

There is also a separate bug in the ```execute.py``` class that under some situations the update_env is crashing and returning an empty dict. This is incorrect behaviour. This should be addressed separately as there is a bug here.

milliams:
I wasn't suggesting that the user should ever have to deal with the pickle file, it would be completely automated. Anyway, I'm happy to drop this PR for the reasons you mention.

The root problem I was trying to fix is that LHCb is now not being able to load the cached environment (). I've found I can fix this by running:
```bash
$ sed --in-place 's/^/export /; s/=/="/; s/$/"/' /home/mpw/.cache/LHCbDIRAC_ENV/x86_64-slc6-gcc49-opt/v8r2p27
```
which adds `export ` to the beginning of each line, turns the first `"` into a `"=` and adds a `"` to the end of each line. I'm thinking of essentially adding this to LHCb's `_store_dirac_environment()` function after it's created the file (we can do this using `sed` or with some Python code). I'll put together another PR and we can discuss it there.

PR #524, opened 2016-05-25
Typo fixes `deamon`→`daemon`

milliams:


rob-c:
Does this change any behaviour because I'm about to commit some large changes to this class which removes amongst other things these lines

milliams:
This will only have the effect of preventing these threads from stopping the interpreter closing before they're finished. This typo has been since the code was first written so I don't think it's worth worrying about either way. If the thread code here disappears then so be it.

milliams:
Superseded by #525

PR #525, opened 2016-05-25
Bugfix/fix execute

rob-c:
This fixes the problems I'm having with the sourcing of the Dirac environment using the current LHCb scripts.
In my mind this is a failing of the execute function which didn't do what we would expect it to do.

Executing a set of commands interactively in bash worked but when they were run through here they didn't give the same result. This was due to several smaller bugs fighting with each other.

In order to debug this I have had to move away from ```os.pipe()``` in order to determine which stage in generating the env, pickling it and un-pickling it had the error.
As this is easier to debug/maintain than to go chasing memory-resident objects I am in the mind to keep the current implementation.

Whilst I was in this file I decided to add some documentation as to how it's doing what it's doing. I can now verify that I can submit jobs to LHCb and perform bookkeeping requests without any problems just by sourcing the env file which is generated by the LHCb commands.
This env file may have not been optimal but this tool should not have failed the way it was doing due to this.

### DIFF
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -217,7 +217,8 @@ def execute(command,
     logger.debug("stderr: %s" % stderr)
 
     timer.cancel()
-    timer.join()
+    if timeout is not None:
+        timer.join()
 
     # Finish up and decide what to return
     if stderr != '':
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -153,6 +153,21 @@ def start_timer(p, timeout):
     return timer, timed_out
 
 
+def update_thread(pipes, thread_output, output_key):
+    """ Functon to contrust and return background thread used to read a pickled object into the thread_output for updating
+        the environment after executing a users code
+        Args:
+            started_threads (list): List containing background threads which have been started
+            pipes (tuple): Tuple containing (read_pipe, write_pipe) which is the pipe the pickled obj is written to
+            thread_output (dict): Dictionary containing the thread outputs which are used after executing the command
+            output_key (str): Used to know where in the thread_output to store the output of this thread
+    """
+    ev = threading.Thread(target=__reader, args=(pipes, thread_output, output_key))
+    ev.daemon = True
+    ev.start()
+    return ev
+
+
 def execute(command,
             timeout=None,
             env=None,
@@ -208,6 +223,13 @@ def execute(command,
     # Start the timer thread used to kill commands which have likely stalled
     timer, timed_out = start_timer(p, timeout)
 
+    if update_env:
+        env_output_key = 'env_output'
+        update_env_thread = update_thread(env_file_pipes, thread_output, env_output_key)
+    if not shell:
+        pkl_output_key = 'pkl_output'
+        update_pkl_thread = update_thread(pkl_file_pipes, thread_output, pkl_output_key)
+
     # Execute the main command of interest
     logger.debug("Executing Command:\n'%s'" % str(command))
     stdout, stderr = p.communicate(command)
@@ -219,6 +241,10 @@ def execute(command,
     timer.cancel()
     if timeout is not None:
         timer.join()
+    if update_env:
+        update_env_thread.join()
+    if not shell:
+        update_pkl_thread.join()
 
     # Finish up and decide what to return
     if stderr != '':
@@ -231,30 +257,24 @@ def execute(command,
 
     # Decode any pickled objects from disk
     if update_env:
-        try:
-            env_output_key = 'env_output'
-            __reader(env_file_pipes, thread_output, env_output_key)
-        except Exception as err:
-            logger.error("Failed to Update Env after command: %s" % command)
-            logger.error("Error was: %s" % err)
-            logger.error("stdout was: %s" % stdout)
-            logger.error("stderr was: %s" % stderr)
-            raise
         if env_output_key in thread_output:
             env.update(thread_output[env_output_key])
+        else:
+            logger.error("Expected to find the updated env after running a command")
+            logger.error("Command: %s" % command)
+            logger.error("stdout: %s" % stdout)
+            logger.error("stderr: %s" % stderr)
+            raise Exception("Missing update env after running command")
 
     if not shell:
-        try:
-            pkl_output_key = 'pkl_output'
-            __reader(pkl_file_pipes, thread_output, pkl_output_key)
-        except Exception as err:
-            logger.error("Failed to Pickle output from command: %s" % command)
-            logger.error("Error was: %s" % err)
-            logger.error("stdout was: %s" % stdout)
-            logger.error("stderr was: %s" % stderr)
-            raise
         if pkl_output_key in thread_output:
             return thread_output[pkl_output_key]
+        else:
+            logger.error("Expected to find the pickled output after running a command")
+            logger.error("Command: %s" % command)
+            logger.error("stdout: %s" % stdout)
+            logger.error("stderr: %s" % stderr)
+            raise Exception("Missing pickled output after running command")
 
     try:
         if isinstance(stdout, str) and stdout != '':
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -2,8 +2,10 @@ import os
 import base64
 import subprocess
 import threading
-import pickle
+import cPickle as pickle
 import signal
+import shutil
+import tempfile
 from Ganga.Core.exceptions import GangaException
 from Ganga.Utility.logging import getLogger
 logger = getLogger()
@@ -19,7 +21,8 @@ def env_update_script(indent=''):
     """
     fdread, fdwrite = os.pipe()
     this_script = '''
-import os, pickle
+import os
+import cPickle as pickle
 os.close(###FD_READ###)
 with os.fdopen(###FD_WRITE###,'wb') as envpipe:
     pickle.dump(os.environ, envpipe)
@@ -48,8 +51,10 @@ def python_wrapper(command, python_setup='', update_env=False, indent=''):
     fdread, fdwrite = os.pipe()
     this_script = '''
 from __future__ import print_function
-import os, sys, pickle, traceback
+import os, sys, traceback
+import cPickle as pickle
 os.close(###PKL_FDREAD###)
+import time
 with os.fdopen(###PKL_FDWRITE###, 'wb') as PICKLE_STREAM:
     def output(data):
         print(pickle.dumps(data), file=PICKLE_STREAM)
@@ -59,9 +64,11 @@ with os.fdopen(###PKL_FDWRITE###, 'wb') as PICKLE_STREAM:
     try:
         full_command = """###SETUP### """
         full_command += """ \n###COMMAND### """
+        full_command += """\nimport sys; sys.exit(0)"""
         exec(full_command, local_ns)
     except:
         print(pickle.dumps(traceback.format_exc()), file=PICKLE_STREAM)
+
 '''
     from Ganga.GPIDev.Lib.File.FileUtils import indentScript
     script = indentScript(this_script, '###INDENT###')
@@ -75,6 +82,7 @@ with os.fdopen(###PKL_FDWRITE###, 'wb') as PICKLE_STREAM:
     if update_env:
         update_script, env_file_pipes = env_update_script()
         script += update_script
+    script += "\nimport sys; sys.exit(0)"
     return script, (fdread, fdwrite), env_file_pipes
 
 
@@ -191,6 +199,10 @@ def execute(command,
         update_env (bool): Should we update the env being passed to what the env was after the command finished running
         return_code (int): This is the returned code from the command which is executed
     """
+    if cwd is None:
+        cwd_ = tempfile.mkdtemp()
+    else:
+        cwd_ = cwd
 
     if update_env and env is None:
         raise GangaException('Cannot update the environment if None given.')
@@ -214,7 +226,7 @@ def execute(command,
         env = get_env()
 
     # Construct the class which will contain the environment we want to run the command in
-    p = subprocess.Popen(stream_command, shell=True, env=env, cwd=cwd, preexec_fn=os.setsid,
+    p = subprocess.Popen(stream_command, shell=True, env=env, cwd=cwd_, preexec_fn=os.setsid,
                          stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 
     # This is where we store the output
@@ -251,6 +263,9 @@ def execute(command,
     if timed_out.isSet():
         return 'Command timed out!'
 
+    if cwd is None:
+        shutil.rmtree(cwd_, ignore_errors=True)
+
     # Decode any pickled objects from disk
     if update_env:
         update_env_thread.join()
@@ -261,7 +276,9 @@ def execute(command,
             logger.error("Command: %s" % command)
             logger.error("stdout: %s" % stdout)
             logger.error("stderr: %s" % stderr)
-            raise Exception("Missing update env after running command")
+            # Too specific? Hopefully we'll never know
+            from Ganga.Core.exceptions import RuntimeException
+            raise RuntimeException("Missing update env after running command")
 
     if not shell:
         update_pkl_thread.join()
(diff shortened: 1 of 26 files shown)


rob-c:
any objections to merge

milliams:
I still need to review the code.

rob-c:
@milliams I was keen to get someone to review it hence the request.

milliams:
I understand but I've only had 1 hour of work time so far to do so.

python/Ganga/Utility/execute.py
```diff
@@ -4,36 +4,58 @@
...
+    """
+    this_pkl_file = NamedTemporaryFile(delete=False)
+    with open(this_pkl_file.name, 'w') as some_file:
+        pass
     this_script = '''
 from __future__ import print_function
 import os, sys, pickle, traceback
-os.close(###PKL_FDREAD###)
-with os.fdopen(###PKL_FDWRITE###, 'wb') as PICKLE_STREAM:
+with open('###TEMP_WRITE###', 'w') as PICKLE_STREAM:
```
milliams: I preferred the previous use of a pipe as it will be more efficient than fully writing a file to disk.

rob-c: But it's very difficult to debug as we lose the errors or bad pickles or bad file handlers

python/Ganga/Utility/execute.py
```diff
@@ -4,36 +4,58 @@
...
+
 def env_update_script(indent=''):
-    fdread, fdwrite = os.pipe()
+    """ This function creates an extension to a python script, or just a python script to be run at the end of the
+    piece of code we're interedted in.
+    This piece of code will dump the environment after the execution has taken place into a temporary file.
+    This returns a tuple of the script it's generated and the NamedTemporaryFile that will be used to store the env
+    Args:
+        indent (str): This is the indent to apply to the script if this script is to be appended to a python file
+    """
```
milliams: Typo "interedted"

python/Ganga/Utility/execute.py
```diff
@@ -4,36 +4,58 @@
...
+    """ This section of code wraps the given python command inside a small wrapper class to allow us to control the output.
+    Optionally we can also append to the end of this file a script to allow us to extract the environment after we've
+    finished executing our command.
+    Args:
+        command (str): This is the pthon code to be executed (can be multi-line)
+        python_setup (str): This is some python code to be executed before the python code in question (aka a script header.
+        update_env (bool): Contol whether we want to capture the env after running
+        indent (str): This allows for an indent to be applied to the script so it can be placed inside other python scripts
+    This returns the NamedTemporaryFile objects for the env_update_script, the python wrapper itself and the script which has been generated to be run
+    """
```
milliams: Typos "pthon", missing bracket after "(aka a script header", type "Contol"

python/Ganga/Utility/execute.py
rob-c: Is there a strong call to put back in support for the pipes here now that everything is debugged and working?

python/Ganga/Utility/execute.py
```diff
@@ -98,26 +137,46 @@ def execute(command,
...
+    Args:
+        command (str): This is the command that we want to execute in string format
+        timeout (int): This is the timeout which we want to assign to a funtion and it will be killed if it runs for longer than n seconds
+        env (dict): This is the environment to use for launching the new command
+        cwd (str): This is the cwd the command is to be executed within.
+        shell (bool): True for a bash command to be executed, False for a command to be executed within Python
+        python_setup (str): A python command to be executed beore the main command is
+        eval_includes (str): A string to be executed in the namespace of the output of the command
+        update_env (bool): Should we update the env being passed to what the env was after the command finished running
+        return_code (int): This is the returned code from the command which is executed
```
milliams: Typo "funtion"

python/Ganga/Utility/execute.py
```diff
@@ -98,26 +137,46 @@ def execute(command,
...
-        # note the exec gets around the problem of indent and base64 gets
-        # around the \n
-        command_update, envread, envwrite = env_update_script()
-        command += ''';python -c "from __future__ import print_function;import base64;exec(base64.b64decode('%s'))"''' % base64.b64encode(command_update)
-
+        command, temp_pkl_file, temp_env_file = python_wrapper(command, python_setup, update_env)
+    else:
+        # We want to run a shell command inside a _NEW_ shell environment.
+        # i.e. What I run here I expect to behave in the same way from the command line after I exit Ganga
+        stream_command = "bash -i "
```
milliams: I'm not sure we should be marking this as an interactive shell?

python/Ganga/Utility/execute.py
rob-c: It's faster to fix typos in the github web page than to have a round of commits where they get addressed separately, please feel free to edit it

python/Ganga/Utility/execute.py
```diff
@@ -133,6 +192,7 @@ def execute(command,
                         final_str += " ;"
                     final_str += " "
 
+    # Construct the class which will contain the environment we want to run the command in
```
milliams: This creates an object, not a class.

python/Ganga/Utility/execute.py
rob-c: To fix the problems I was having that the commands in subprocess didn't give the same result as the same commands on the command line, yes, yes we should.

python/Ganga/Utility/execute.py
```diff
@@ -142,40 +202,60 @@ def execute(command,
...
+            logger.error("Error was: %s" % err)
+            logger.error("The pickle we tried to read was in: %s" % temp_pkl_file.name)
+            logger.error("stdout was: %s" % stdout)
+            logger.error("stderr was: %s" % stderr)
+            raise
+
+    # Cleanup after ourselves
+    for file_ in [temp_env_file, temp_pkl_file]:
+        if file_:
+            os.unlink(file_.name)
```
milliams: These files don't necessarily exist. If `shell=True` and `update_env=False` are set then neither does I think.

rob-c: https://github.com/ganga-devs/ganga/pull/525/files#diff-09802c04483863469997ddb0cb23bea3R156

python/Ganga/Utility/execute.py
milliams: I'm not going to push the change as I too found it more difficult to debug using pipes. I'm just very wary about creating too many temporary files. Especially as one will be made for every `execute(shell=False)` call, just to get the output. This will be done for every single DIRAC call I believe.

python/Ganga/Utility/execute.py
```diff
@@ -4,36 +4,58 @@
...
-    fdread, fdwrite = os.pipe()
+    """ This function creates an extension to a python script, or just a python script to be run at the end of the
+    piece of code we're interedted in.
+    This piece of code will dump the environment after the execution has taken place into a temporary file.
+    This returns a tuple of the script it's generated and the NamedTemporaryFile that will be used to store the env
+    Args:
+        indent (str): This is the indent to apply to the script if this script is to be appended to a python file
+    """
+    this_env_file = NamedTemporaryFile(delete=False)
+    with open(this_env_file.name, 'w') as some_file:
```
milliams: It's semi-convention to use `_` for dummy variables which are never used.

python/Ganga/Utility/execute.py
```diff
@@ -87,6 +122,10 @@ def __timeout_func(process, timed_out):
 # /\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\
 
 
+#def execute(command, timeout=None,env=None,cwd=None,shell=True,python_setup='',eval_includes=None,update_env=False,return_code=None):
+#    with execute_lock:
+#        return _execute(command, timeout, env, cwd, shell, python_setup, eval_includes, update_env, return_code)
+
```
milliams: This commented-out block should be deleted.

python/Ganga/Utility/execute.py
milliams: Would that have been a sufficient change to fix all this?

rob-c: Not entirely clear. Tbh If I'm going back to using a pipe method I'm going to add some debugging (and the documentation on this class has helped)

rob-c:
OK, this PR has now reverted to using file handlers and pipes, although the update threads have still been dropped in favor of just calling the ```__reader``` command after everything has run (it makes it clearer what is going on and how to debug).
This also fixes a minor hidden exception which is cause when nothing is returned from the stdout.
After giving the code a bit of a kick it was also clearer that it's easier to just pass around the file handlers as tuples to track what is going on.

Given this is closer to the original but with better error handling and some improved code and documentation, are there any comments before I look to merge?
(Assuming all the tests pass)

milliams:
I've been reading about pipes etc. and my worry now is that pipes have a maximum buffer size that they can hold. This means that if the running process tries to write more data than that to the pipe (e.g. a large environment or a lot of output from a python script)it will hang indefinitely. This is why it was implemented as a concurrent thread so it could consume the pipe as it was fed.

I still very much agree with the gist of what you're doing here (cleanup, error handling and documentation etc.) but we need to be careful that we're not making things worse.

rob-c:
I strongly suspect that is you're expecting so much information to be passed by stdout/pipes (etc) I think you're doing things even more horribly wrong than whatever is going on in this class.
I will add back the threads for the sake of buffers, but I strongly suspect it's not a requirement for 99% of the use cases.

milliams:
I agree it's not needed for the majority of cases but I just want to avoid silently hanging when we hit that 1%. Pipes *can* be as small as 4096 bytes so it wouldn't be that hard to fill it completely. At the very least your round-tripping on this code will have simplified it and increased its quality.

alexanderrichards:
> I've been reading about pipes etc. and my worry now is that pipes have a maximum buffer size that they can hold. This means that if the running process tries to write more data than that to the pipe (e.g. a large environment or a lot of output from a python script)it will hang indefinitely.

yes I believe this is why I had a concurrent thread reading the pipe :wink:

rob-c:
Reverted back to threads but kept debugging code and errors when things haven't worked correctly rather than silently swallowing errors.
code is a bit more readable now if you read the final version

python/Ganga/Utility/execute.py
```diff
@@ -184,24 +251,46 @@ def execute(command,
...
+        else:
+            logger.error("Expected to find the pickled output after running a command")
+            logger.error("Command: %s" % command)
+            logger.error("stdout: %s" % stdout)
+            logger.error("stderr: %s" % stderr)
+            raise Exception("Missing pickled output after running command")
 
     try:
-        stdout = pickle.loads(stdout)
+        if isinstance(stdout, str) and stdout != '':
```
milliams: is `stdout` ever not a `str`? Also, are we actually writing any pickled output to the `stdout` ever or does this always `except`?

python/Ganga/Utility/execute.py
```diff
@@ -184,24 +251,46 @@ def execute(command,
...
+    if update_env:
+        update_env_thread.join()
+        if env_output_key in thread_output:
+            env.update(thread_output[env_output_key])
+        else:
+            logger.error("Expected to find the updated env after running a command")
+            logger.error("Command: %s" % command)
+            logger.error("stdout: %s" % stdout)
+            logger.error("stderr: %s" % stderr)
+            raise Exception("Missing update env after running command")
```
milliams: This (and the one below) should be more specific in their exception type. I would suggest at least `RuntimeError`.

python/Ganga/Utility/execute.py
```diff
@@ -98,84 +179,70 @@ def execute(command,
...
-
-    timed_out = threading.Event()
-    timer = threading.Timer(timeout, __timeout_func,
-                            args=(p, timed_out))
-    timer.deamon = True
-    started_threads = []
-    if timeout is not None:
-        timer.start()
-        started_threads.append(timer)
+    # Construct the class which will contain the environment we want to run the command in
```
milliams: "Construct the object..."

python/Ganga/Utility/execute.py
rob-c: Don't know there is a check below, didn't want to argue with it.

python/Ganga/Utility/execute.py
rob-c: This doesn't really effect anything, but I'll look into it

rob-c:
Spent too much time looking at this, I plan to merge tomorrow at about 11am unless there are any serious objections.

python/Ganga/Utility/execute.py
```diff
@@ -261,7 +276,9 @@ def execute(command,
             logger.error("Command: %s" % command)
             logger.error("stdout: %s" % stdout)
             logger.error("stderr: %s" % stderr)
-            raise Exception("Missing update env after running command")
+            # Too specific? Hopefully we'll never know
+            from Ganga.Core.exceptions import RuntimeException
```
milliams: 1. Any import like this should be at the top of the file.
2. This exception does not exist. The built-in Python exception `RuntimeError` should be used.

python/Ganga/Utility/execute.py
```diff
@@ -191,6 +199,10 @@ def execute(command,
         update_env (bool): Should we update the env being passed to what the env was after the command finished running
         return_code (int): This is the returned code from the command which is executed
     """
+    if cwd is None:
+        cwd_ = tempfile.mkdtemp()
+    else:
+        cwd_ = cwd
```
milliams: There's no need to rename this variable. Simply
```python
if cwd is None:
    cwd = tempfile.mkdtemp()
```
is sufficient.

milliams:
You see to have pushed a bunch of unrelated DIRAC changes t this PR.

rob-c:
Yes, and they've been removed as they didn't belong here

python/Ganga/Utility/execute.py
rob-c: And then I have to track I called this to clean up after the class

python/Ganga/Utility/execute.py
rob-c: I saw a circular dependence when I tried to force this to be triggered and I'll rename then

rob-c:
Well I regard that as the last round of changes unless it's broken anything, on to other work

python/Ganga/Utility/execute.py
milliams: Is there any need to create a temporary directory for the `cwd`? Doesn't `subprocess.Popen` just use the current working directory by default?

python/GangaDirac/__init__.py
```diff
@@ -80,6 +80,8 @@
                                                  'Unknown: No status for Job': 'failed'},
                                                 "Mapping of Dirac to Ganga Job statuses used to construct a queue to finalize a given job, i.e. final statues in 'statusmapping'")
 
+    configDirac.addOption('serializeBackend', False, 'Developer option to serialize Dirac code for profiling/debugging')
```
milliams: This is presumably mistakenly comitted?

python/Ganga/Utility/execute.py
```diff
@@ -2,16 +2,28 @@
...
 import subprocess
 import threading
-import pickle
+import cPickle as pickle
 import signal
+import shutil
+import tempfile
 from Ganga.Core.exceptions import GangaException
 from Ganga.Utility.logging import getLogger
+from exceptions import RuntimeError
```
milliams: There's no need to import anything. All the python standard exceptions are builtins

rob-c: Yes but if I don't import it I'm going to forget this and this does no harm

rob-c:
OK, looking to merge after the tests pass in this case I'm doing too much fighting with git with this branch and I'm working on other things

python/Ganga/Utility/execute.py
```diff
@@ -98,84 +188,74 @@ def execute(command,
...
+        cwd (str): This is the cwd the command is to be executed within.
+        shell (bool): True for a bash command to be executed, False for a command to be executed within Python
+        python_setup (str): A python command to be executed beore the main command is
+        eval_includes (str): An string used to construct an environment which, if passed, is used to eval the stdout into a python object
+        update_env (bool): Should we update the env being passed to what the env was after the command finished running
+        return_code (int): This is the returned code from the command which is executed
     """
+    shouldDeleteCWD=False
+    if cwd is None:
+        shouldDeteCWD=True
```
milliams: Typo on this variable name. Also, I'm still not sure that this CWD change is necessary. Did you see problem with it how it was before?

rob-c: Yes, I wasn't wanting to run commands which touch disk on AFS with this. They should be run in a temp directory which on many systems is faster.

milliams: Sure, but in the vast majority of cases, `cwd` is not set by the user and they are not writing any files to the working directory so we're creating a temporary directory and then deleting it for no reason.

rob-c: There's disagreement over his so I'll revert it...

milliams:
Looks good, I'll just correct a few minor typos and then you can merge :+1:

rob-c:
Good to merge?

milliams:
:+1:

Issue #520, opened 2016-05-25
Another deadlock found


drmarkwslater:
While doing the work on atexit and (by extension) the coverage reporting (see #510) I've found a definite hard lock up. It is probably made worse by the coverage now working properly and I've only seen it on the CI machines but I can certainly see the problem.

Basically, when you set a value on a `GangaObject` through `__set__` it goes through `_getWriteAccess` which in turn locks the repository, i.e. Obj Lock -> Reg Lock. Flushing however, goes the other way, i.e. Reg lock (in `_flush`) -> Obj Lock. If the timing's bad you get a lock up.

I would argue that `_getWriteAccess` isn't needed when you set a value on an object as if that object is being flushed, either the flusher or the setter has the object lock and the other one has to wait. 

This could also be the source of the other lockups I saw with the coverage but I didn't check the stack traces for those.

What do people think?

milliams:
Indeed, I really don't like the usage of `_getWriteAccess` in the code base at the moment. I understand what it was there for originally (full transaction locks for registry objects so that can be shared across concurrently running sessions) but since we never release any of these locks, I'm not sure that it's even working as it should. This is another example of just how hard it is to have thread safety *and* process safety all at once.

There's a similar potential issue in `GangaObject.__get__` but that is mostly avoided since that should only be called if the object is *not* fully loaded and `flush` will only be called if it *is*.

rob-c:
You need a lock on the file on disk once the object has been loaded I think but it makes more sense to get a writeLock when loading an object and not when modifying one already in memory imo.
We used to mark things dirty and flush them in incoherent ways and it makes more sense to get rid of the Registry locks in Object and have the registry request a file lock when an object is loaded and release one when it's unloaded.

milliams:
If I remember correctly (and I might be wrong) we grab the session lock for *all* objects in the registry when loading them anyway. We're not being that fine-grained.

rob-c:
@milliams I'm not sure I understand what you mean here. We get a lock on an object only when it's loaded else multiple sessions would never have worked.
I think we can drop the lock requests in GangaObject entirely if we're locking the object in the Registry
once they're loaded and we're currently not releasing any locks at runtime we can drop the write/read disk locks entirely for now if they're causing threading issues.

milliams:
[...]

PR #526, opened 2016-05-26
relaxing has_loaded repo locks

rob-c:
This PR relaxes the repo locks around the has_loaded check. As per discussion in #520 .

This may have problems if this is caught in any repo transaction within this class due to locks being set around too much code.

I'm the process in testing the stability of this.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -291,12 +291,12 @@ class GangaRepositoryLocal(GangaRepository):
                     obj = self._make_empty_object_(this_id, cat, cls)
                 except Exception as err:
                     raise IOError('Failed to Parse information in Index file: %s. Err: %s' % (fn, err))
-            this_cache = obj._index_cache
+            this_cache = obj._index_cache_dict
             this_data = this_cache if this_cache else {}
             for k, v in cache.iteritems():
                 this_data[k] = v
             #obj.setNodeData(this_data)
-            obj._index_cache = cache
+            obj._index_cache_dict = cache
             self._cache_load_timestamp[this_id] = fn_ctime
             self._cached_cat[this_id] = cat
             self._cached_cls[this_id] = cls
@@ -332,7 +332,7 @@ class GangaRepositoryLocal(GangaRepository):
                     logger.debug("Writing: %s" % str(new_index))
                     pickle_to_file(new_index, this_file)
                 self._cached_obj[this_id] = new_cache
-                obj._index_cache = {}
+                obj._index_cache_dict = {}
             self._cached_obj[this_id] = new_idx_cache
         except IOError as err:
             logger.error("Index saving to '%s' failed: %s %s" % (ifn, getName(err), err))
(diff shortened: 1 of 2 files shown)


rob-c:
From code inspection this doesn't appear to be a problem (aka ```has_loaded``` isn't called inside another transaction at the Registry level).
The test comes to see if it's triggered from within any transactions due to some higher level code interaction or not.

rob-c:
OK, had to fix a minor bug in that the Repo was being locked out of access to the ```_index_cache``` due to changes which pushed this to a higher level than the repo...

rob-c:
Any objections to merge?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -291,12 +291,12 @@ def index_load(self, this_id):
                     obj = self._make_empty_object_(this_id, cat, cls)
                 except Exception as err:
                     raise IOError('Failed to Parse information in Index file: %s. Err: %s' % (fn, err))
-            this_cache = obj._index_cache
+            this_cache = obj._index_cache_dict
```
milliams: I rally don't like circumventing the `_index_cache` getter here. If there are problems with it then we should look at addressing those. I'm not a big fan of the getter anyway as it's doing more blurring of the line between objects and the registry.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -968,10 +967,16 @@ def has_loaded(self, obj):
         Args:
             obj (GangaObject): Object which we want to look for in this repo
         """
+        obj_id = id(obj)
+        self.lock_transaction(obj_id, 'has_loaded')
```
drmarkwslater: I might be misunderstanding the `lock_transaction` method but doesn't this only protect against altering the registry if the same object is being operated on? e.g. it would stop you removing the object that was being checked for with `has_loaded`. Isn't the bigger problem that the `_objects` dictionary changes while the underlying `find` is going on? e.g. you call `has_loaded` with one object but are removing another at the same time. That's still allowed by this transaction lock but would be a problem overall. I would expect (though this may be wrong) if you call `remove` on an object, that requires a lock on the object itself which would stop it's changing from outside anyway.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: The getter method completely subverts the point of this code by checking that we get the same thing unless the object is not loaded. We know her the object has _not_ been loaded and in other tests the object _has just been loaded_. We know this because this class is in charge of doing this. Using the getter gets you into a strange state depending on whether the Registry/Repo is saying the object is in memory _right now_ which can potentially get out of sync during object load.
(In addition to calling potentially lockable functions higher up which don't need to be done.)

python/Ganga/Core/GangaRepository/Registry.py
rob-c: good point. Although at least this protects against this object being removed while we're checking if it's in memory... We desperately need some form of get/set lock system in the Registry.

rob-c:
closed after discussions

PR #527, opened 2016-05-26
Cleanup atexit handling (cleaned version)

drmarkwslater:
This is another version of #510 without the horrible mess while I was tracking down a couple of lockups.

### DIFF
--- a/python/Ganga/Core/InternalServices/ShutdownManager.py
+++ b/python/Ganga/Core/InternalServices/ShutdownManager.py
@@ -1,41 +1,8 @@
-##########################################################################
-# Ganga - a computational task management tool for easy access to Grid resources
-# http://cern.ch/ganga
-#
-# $Id: ShutdownManager.py,v 1.1 2008-07-17 16:40:50 moscicki Exp $
-#
-# Copyright (C) 2003-2007 The Ganga Project
-#
-# This file is part of Ganga.
-#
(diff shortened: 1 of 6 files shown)


drmarkwslater:
Any objections to merge?

python/Ganga/Core/InternalServices/ShutdownManager.py
```diff
@@ -38,157 +38,110 @@
...
-
-    from Ganga.Core.GangaThread.WorkerThreads import shutDownQueues
-    shutDownQueues()
-
-    logger.info("Shutting Down Ganga Repositories")
-    from Ganga.Runtime import Repository_runtime
-    Repository_runtime.shutdown()
-
-    from Ganga.Core.InternalServices import Coordinator
+            logger.warning("Exception raised while stopping the monitoring: %s" % err)
```
milliams: For more detailed information (e.g. the full stacktrace), perhaps these should use [`logger.exception()`](https://docs.python.org/2.6/library/logging.html#logging.Logger.exception)?

drmarkwslater: Makes sense 😄

drmarkwslater:
Anything more to change/add?

milliams:
Looks good to me :+1:

PR #530, opened 2016-05-26
Faster dirac

rob-c:
This PR tries to add a few smaller optimisations on top of the existing work that's been done with the DIRAC backend.

Specifically this tries to reduce the amount of I/O, ```os.system``` calls and other slower commands which are called within the finalization code.

I don't have any numbers for this but I'm also adding an option useful for debugging which serializes all of the backend tasks which makes profiling a _lot_ easier in the future.

This is based upon #525 which addresses a few minor points in how the execute commands were being called. There is about a 0.1s overhead in in the subprocess code which I think we can live with considering a lot of commands are taking O(10)s to complete.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -564,11 +564,7 @@ class GangaRepositoryLocal(GangaRepository):
                     if this_id in self.objects:
                         self._internal_del__(this_id)
                         changed_ids.append(this_id)
-                except Exception as x:
-                    ## WE DO NOT CARE what type of error occured here and it can be
-                    ## due to corruption so could be one of MANY exception types
-                    ## If the job is not accessible this should NOT cause the loading of ganga to fail!
-                    ## we can't reasonably write all possible exceptions here!
+                except (InaccessibleObjectError, ) as x:
                     logger.debug("update_index: Failed to load id %i: %s" % (this_id, x))
                     summary.append((this_id, x))
 
@@ -1012,7 +1008,7 @@ class GangaRepositoryLocal(GangaRepository):
         for this_id in ids:
 
             if this_id in self.incomplete_objects:
-                raise RepositoryError(self.repo, "Trying to re-load a corrupt repository id: %s" % this_id)
+                raise RepositoryError(self.registry, "Trying to re-load a corrupt repository id: %s" % this_id)
 
             fn = self.get_fn(this_id)
             if load_backup:
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -1008,7 +1008,7 @@ class GangaRepositoryLocal(GangaRepository):
         for this_id in ids:
 
             if this_id in self.incomplete_objects:
-                raise RepositoryError(self.registry, "Trying to re-load a corrupt repository id: %s" % this_id)
+                raise RepositoryError(self, "Trying to re-load a corrupt repository id: %s" % this_id)
 
             fn = self.get_fn(this_id)
             if load_backup:
--- a/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
+++ b/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
@@ -222,6 +222,14 @@ class WorkerThreadPool(object):
         """
         return [(w.gangaName, w._command, w._timeout) for w in self.__worker_threads]
 
+    def threads_matching(self, name_str):
+        """
+        Returns a list of all threads matching the given string at the start of their name
+        Args:
+            name_str (str): String to be used in the starts_with comparison of the worker_threads
+        """
+        return [w for w in self.__worker_threads if w.gangaName.startswith(name_str)]
+
     def _stop_worker_threads(self, shutdown=False):
         self._shutdown = shutdown
         for w in self.__worker_threads:
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -111,7 +111,7 @@ def execute(command,
         # note the exec gets around the problem of indent and base64 gets
         # around the \n
         command_update, envread, envwrite = env_update_script()
-        command += ''';python -c ""from __future__ import print_function;import base64;exec(base64.b64decode('%s'))"''' % base64.b64encode(command_update)
+        command += ''';python -c "from __future__ import print_function;import base64;exec(base64.b64decode('%s'))"''' % base64.b64encode(command_update)
 
     if env is None and not update_env:
         pipe = subprocess.Popen('python -c "from __future__ import print_function;import os;print(os.environ)"',
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -217,7 +217,8 @@ def execute(command,
     logger.debug("stderr: %s" % stderr)
 
     timer.cancel()
-    timer.join()
+    if timeout is not None:
+        timer.join()
 
     # Finish up and decide what to return
     if stderr != '':
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -153,6 +153,21 @@ def start_timer(p, timeout):
     return timer, timed_out
 
 
+def update_thread(pipes, thread_output, output_key):
+    """ Functon to contrust and return background thread used to read a pickled object into the thread_output for updating
+        the environment after executing a users code
+        Args:
+            started_threads (list): List containing background threads which have been started
+            pipes (tuple): Tuple containing (read_pipe, write_pipe) which is the pipe the pickled obj is written to
+            thread_output (dict): Dictionary containing the thread outputs which are used after executing the command
+            output_key (str): Used to know where in the thread_output to store the output of this thread
+    """
+    ev = threading.Thread(target=__reader, args=(pipes, thread_output, output_key))
+    ev.daemon = True
+    ev.start()
+    return ev
+
+
 def execute(command,
             timeout=None,
             env=None,
@@ -208,6 +223,13 @@ def execute(command,
     # Start the timer thread used to kill commands which have likely stalled
     timer, timed_out = start_timer(p, timeout)
 
+    if update_env:
+        env_output_key = 'env_output'
+        update_env_thread = update_thread(env_file_pipes, thread_output, env_output_key)
+    if not shell:
+        pkl_output_key = 'pkl_output'
+        update_pkl_thread = update_thread(pkl_file_pipes, thread_output, pkl_output_key)
+
     # Execute the main command of interest
     logger.debug("Executing Command:\n'%s'" % str(command))
     stdout, stderr = p.communicate(command)
@@ -219,6 +241,10 @@ def execute(command,
     timer.cancel()
     if timeout is not None:
         timer.join()
+    if update_env:
+        update_env_thread.join()
+    if not shell:
+        update_pkl_thread.join()
 
     # Finish up and decide what to return
     if stderr != '':
@@ -231,30 +257,24 @@ def execute(command,
 
     # Decode any pickled objects from disk
     if update_env:
-        try:
-            env_output_key = 'env_output'
-            __reader(env_file_pipes, thread_output, env_output_key)
-        except Exception as err:
-            logger.error("Failed to Update Env after command: %s" % command)
-            logger.error("Error was: %s" % err)
-            logger.error("stdout was: %s" % stdout)
-            logger.error("stderr was: %s" % stderr)
-            raise
         if env_output_key in thread_output:
             env.update(thread_output[env_output_key])
+        else:
+            logger.error("Expected to find the updated env after running a command")
+            logger.error("Command: %s" % command)
+            logger.error("stdout: %s" % stdout)
+            logger.error("stderr: %s" % stderr)
+            raise Exception("Missing update env after running command")
 
     if not shell:
-        try:
-            pkl_output_key = 'pkl_output'
-            __reader(pkl_file_pipes, thread_output, pkl_output_key)
-        except Exception as err:
-            logger.error("Failed to Pickle output from command: %s" % command)
-            logger.error("Error was: %s" % err)
-            logger.error("stdout was: %s" % stdout)
-            logger.error("stderr was: %s" % stderr)
-            raise
         if pkl_output_key in thread_output:
             return thread_output[pkl_output_key]
+        else:
+            logger.error("Expected to find the pickled output after running a command")
+            logger.error("Command: %s" % command)
+            logger.error("stdout: %s" % stdout)
+            logger.error("stderr: %s" % stderr)
+            raise Exception("Missing pickled output after running command")
 
     try:
         if isinstance(stdout, str) and stdout != '':
--- a/python/Ganga/Utility/execute.py
+++ b/python/Ganga/Utility/execute.py
@@ -195,7 +195,7 @@ def execute(command,
         python_setup (str): A python command to be executed beore the main command is
         eval_includes (str): An string used to construct an environment which, if passed, is used to eval the stdout into a python object
         update_env (bool): Should we update the env being passed to what the env was after the command finished running
-        return_code (int): This is the returned code from the command which is executed
+        return_code (list): This has element 0 modified to be the returned code from the command which is executed
     """
 
     if update_env and env is None:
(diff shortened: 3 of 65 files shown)


rob-c:
These changes seem to reduce the time to complete a hello world job dramatically. We're now down to 1 call to finalize a successful job with additional debug information being requested for jobs in other states. This looked to be the best compromise to getting all information on all jobs.

A revealing mock test on this branch is:
```
from Ganga.GPI import jobs
from Ganga.GPIDev.Adapters.IBackend import IBackend
from Ganga.Utility.Config import setConfigOption
setConfigOption('DIRAC', 'serializeBackend', True)
for _ in range(10):
    j = jobs(424)
    j.backend.reset()
    IBackend.master_updateMonitoringInformation([stripProxy(j)])
```
(were 424 is a completed job which was submitted to the DIRAC backend)

The main amount of time on this branch is spend in the execute commands against DIRAC (of which there are only ~20 (I think) calls to DIRAC to reset and finalize the job 10 times)

This represents a significant speed improvement over the code currently in develop which took about ~10min to execute the equivalent command but now is down to about ~3min. (I suspect this isn't due to changes in system activity)

rob-c:
This final commit was too large (by accident) but the changes not reflected in the log are that I fixed a minor outstanding bug in the XML which presented when a job of mine became corrupt.

I have also made some minor changes around how Dirac is being called to avoid that any work is being done on a shared filesystem (user-space) which is potentially extremely expensive compared to accessing the temp space on a machine. This is the case on many setups and only adds a small overhead which can be removed on a per-call basis more intelligently if/when this becomes a bigger performance hit (I don't expect us to ever reasonably limited by this).

rob-c:
Any objections to merging this branch by the end of today?

I'm seeing some monitoring bugs crop up under heavy load which I'm trying to track down but if I manage to solve this I'd be keen to get this branch in so I can rebase the other branches I have

milliams:
I should have time to review this this afternoon.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -1012,7 +1008,7 @@ def load(self, ids, load_backup=False):
         for this_id in ids:
 
             if this_id in self.incomplete_objects:
-                raise RepositoryError(self.repo, "Trying to re-load a corrupt repository id: %s" % this_id)
+                raise RepositoryError(self.registry, "Trying to re-load a corrupt repository id: %s" % this_id)
```
milliams: Why are you passing the registry to the `repo` argument?

python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
```diff
@@ -222,6 +222,14 @@ def worker_status(self):
         """
         return [(w.gangaName, w._command, w._timeout) for w in self.__worker_threads]
 
+    def threads_matching(self, name_str):
```
milliams: This function doesn't seem to be used anywhere.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -95,9 +97,20 @@ class DiracBase(IBackend):
     _hidden = True
 
     def _setup_subjob_dataset(self, dataset):
+        """
+        Unknown what this ethod does other than return None:
```
milliams: This is not a helpful docstring. Looking at its overload in `Dirac` would tell you what it does.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -95,9 +97,20 @@ class DiracBase(IBackend):
...
+        Args:
+            Dataset (Dataset): This is a GangaDataset object, todo check this isn't a list
+        """
         return None
 
     def _setup_bulk_subjobs(self, dirac_ids, dirac_script):
+        """
+        This is the old bulk submit method which is used to construct the subjobs for a paraetric job
```
milliams: What do you mean my the 'old' method? Is it not used any more? Also a typo in 'paraetric'.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -483,9 +520,30 @@ def debug(self):
             logger.error(result.get('Message', ''))
 
     @staticmethod
-    def _getStateTime(job, status):
+    def _bulk_updateStateTime( jobStateDict ):
+        """ This performs the same as the _getStateTime method but loops over a list of job ids within the DIRAC namespace (much faster)
```
milliams: If it does the same but faster, do we need both?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -483,9 +520,30 @@ def debug(self):
...
+        Args:
+            jobStateDict (dict): This is a dict of {job.backend.id : job_status, } elements
+        """
+        for this_state in jobStateDict:
+            bulk_result = execute("getBulkStateTime(\'%s\',\'%s\')" %
```
milliams: Do these `'` really need escaping?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -483,9 +520,30 @@ def debug(self):
...
+                                    (repr([job.backend.id for job in jobStateDict[this_state]]), this_state))
+            for this_job in jobStateDict[this_state]:
+                if bulk_result[this_job.backend.id]:
+                    DiracBase._getStateTime(this_job, this_state, bulk_result[this_job.backend.id])
+                else:
+                    DiracBase._getStateTime(this_job, this_state)
```
milliams: Is this `if`/`else` needed? Doesn't `bulk_result[this_job.backend.id]` default to `{}` anyway?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -795,47 +860,80 @@ def updateMonitoringInformation(_jobs):
             try:
                 job.backend.extraInfo = state[4]
             except Exception as err:
-                logger.debug("gxception: %s" % str(err))
+                logger.debug("gxecption: %s" % str(err))
```
milliams: Typo (presumably) in 'gxecption'.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Frankly something was better than the ```.repo``` argument here which was causing name errors. And there used to be 2 RepositoryError classes which added confusion.

python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
rob-c: I plan to use this in the future in a few places and was experimenting with it in this branch

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: We don't properly support the bulk submit using the parametric stuff and haven't supported it for a long time now as far as I know. It may work on the GridPP dirac without problems but it caused issues for LHCb

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Yes, we probably need independent calls for now in a few other places. I'll review this once the PR is finished, debugging a few things atm

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Don't know, copying it from elsewhere in the class

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: it should be ```if id in ...``` I think, debugging this now as it stands

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: Sorry, I assumed the PR was ready for review as you were asking to merge it. I will wait until you are ready next time. Let me know when you want this one reviewing.

rob-c: It was up until I discovered a bug about 15min ago, as it didn't change the structure of the PR I was keen to hear feedback which will be implemented before looking to merge again

rob-c:
OK, squashed several bugs again, nothing much has changes just a few 1 liners dotted around so going for merge again unless I find any more bugs.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Yes I plan to keep both functions as they address different needs.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: They didn't, as it was not a string this introduced an error. Fixed in latest code.

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -113,22 +113,23 @@ def addFile(lfn, file, diracSE, guid):
...
-        os.system('rm -rf %s' % tmpdir)
-        ganga_logs = glob.glob('%s/*_Ganga_*.log' % outputDir)
 
-        if ganga_logs:
-            os.system('ln -s %s %s/stdout' % (ganga_logs[0], outputDir))
+        if not noJobDir:
+            tmpdir = os.path.join(outputDir, str(id))
+            os.system('mv -f %s/* %s/. ; rm -rf %s' % (tmpdir, outputDir, tmpdir))
+        
+        os.system('for file in $(ls %s/*_Ganga_*.log); do ln -s ${file} %s/stdout; break; done' % (outputDir, outputDir))
```
milliams: Can we not do this and the line above using the Python API rather than falling back to shell commands?

rob-c: I don't see that we gain anything tbh given that this is run on a background python process independent to the rest of Ganga.
These have been system calls for quite a while and I've only tried to reduce amount of code here

rob-c:
OK, (hopefully) the final round of modifications hav just been pushed includes a fix in the Montoring code to allow the monitoring to run correctly when using the ```runMonitoring``` command.

I've added some extra code to reduce the amount of times we create a DIRAC instance to effectively have a ```monitorJobs``` command in the DIRAC server scripts as well as a ```finishedJobs``` command (I should check the name of the latter).
These commands mean that we have a greatly reduced number of times that we construct DIRAC instances and call DIRAC.
The monitoring in my example of a job with 100 subjobs works in a much better way, there is a slight pause at the start as the job is first read from disk. There is then a delay in the jobs updating their background status and then the job is queued for finalisation.

This addresses the strange bug that LHCb users were seeing that despite having a queues system only a few jobs at a time being actually queued. When I run
```runMonitoring(job.select(id=430))``` I find that by the time it finishes executing (~2min) some subjobs have already completed and all of the subjobs of this job are correctly queued in the queues system and taking between 15-45sec to finish depending on all the numerous other things that are done behind the scenes to finish a job.

I'm very keen to get this branch in for 6.1.21 so will spend some time looking at it after tomorrow before I get some other branches with other features in.

**Comment by rob-c** [comment]
OK, I have just updated the handling of the gridProxy slightly here. Yes this will change when the credentials stuff lands but until then this improves the experience when a proxy is destroyed either inside Ganga or externally.

**Comment by rob-c** [comment]
OK, one oddity in this PR was that you could observe Ganga moving through the list of subjobs and updating their status 1 at a time after it had finished monitoring the subjosb in DIRAC. This was a performance bug somewhere in Ganga.

File: python/GangaDirac/Lib/Utilities/DiracUtilities.py [hunk 10]
```diff
@@ -206,20 +206,42 @@ def execute(command,
...
                                   shell=shell,
                                   python_setup=python_setup,
                                   eval_includes=eval_includes,
-                                  update_env=update_env)
+                                  update_env=update_env,
+                                  return_code=return_code)
+
+    ## Command failed, lets check to see if it was due to the proxy or not
+    if return_code:
+        last_modified_valid = False
```
milliams: Have you tested this code? Because as far as I am aware, the `return_code` feature of `execute()` does not work and never has since it was first introduced in 562b5092eadee39a.

rob-c:
Does anyone have any comments on this PR? Thanks

python/GangaDirac/Lib/Utilities/DiracUtilities.py
rob-c: How can I fix it in this case then?

rob-c: This is tricky to test, especially given that I'm now checking for a valid proxy before each call.
Ah, the return_code isn't set on all returns from the function... I'll look into this

milliams: No, the problem is that in Python, setting the value of a variable passed in as a function argument will perform a name binding, not change the object that it points to. E.g.:
```python
def foo(test=None):
    test = True

var = False
foo(var)
print(var)
```
will *always* print `False`.

rob-c: OK, can you suggest a good way to return this in this case? I'm wanting to avoid changing the return type here so is there a good way to return the rc otherwise?

rob-c: Decided to go with a list of 1 entries to get this

milliams: Not in an easy way. There are a couple of problems:

1. We run the passed command in a `bash "..."` subshell and I'm not sure if error codes are propagated out correctly.
2. If `update_env` is used, we are running a number of processes chained together and only the last's rc will be caught.
3. If running a Python command, I've no idea how much we can rely on sensible return codes (this one if probably ok though).

Once those potential problems are addressed then the only way to get the information out is to pass it as a structured return value (a `dict` or a `tuple` or something) or make the `return_value` object some mutable compound type. Alternatively, raising an exception would make sense I think so that it is always noticed if a command fails.

python/Ganga/Utility/execute.py
```diff
@@ -194,7 +195,7 @@ def execute(command,
         python_setup (str): A python command to be executed beore the main command is
         eval_includes (str): An string used to construct an environment which, if passed, is used to eval the stdout into a python object
         update_env (bool): Should we update the env being passed to what the env was after the command finished running
-        return_code (int): This is the returned code from the command which is executed
+        return_code (list): This has element 0 modified to be the returned code from the command which is executed
```
milliams: This is not a nice API to use for this, it is far too likely to be used incorrectly which will cause silent failures. This makes much more sense to be done as a return value of the function (perhaps as a `{'rc':..., 'stdout':..., 'stderr':...}` dictionary or maybe better a named tuple). Using function arguments as a return mechanism is not recommended.

rob-c: If this is going down this route, I'll revert the changes here and add a fixme message to the DiracUtils as this is starting to get into the scope of some invasive changes which could cause more problems than this PR solves.
I'll revert the changes to `execute.py` and update the GangaDirac appropriately.

milliams: I think that makes sense. We'll make the fix in another PR soon. There's only a few places that call `execute()` directly so it shouldn't be too invasive.

rob-c:
OK, are there any outstanding things I should look into with this PR?

milliams:
Go for merge :+1:

milliams:
It looks like these changes have caused LHCb (and maybe vanilla DIRAC as well?) to no longer create a grid proxy by default. These `last_modified_time` and `last_modified_valid` variables are confusing to me, I'm not clear on what they're trying to do?

Regardless, it seems that since the default for `execute()` is `renew=False`, the proxy is never created in the first place. In fact, even though you introduced that parameter, you never set it in any call and so nothing is ever renewed.

Would it make sense to simple remove the `renew` parameter from `execute()` and have it always run? If so, I have a PR incoming which does exactly that.

rob-c:
We should never have been constructing these until the user explicitly calls gridProxy.renew() directly.

milliams:
But what if they *don't* call that manually? Things just break with errors like:
```
AttributeError: 'NoneType' object has no attribute 'get'
```
which is caused by our DIRAC `execute()` returning `None` if `last_modified_valid` is `False` which happens if there is no valid proxy (e.g. https://ganga-ci.cern.ch/jenkins/job/Nightly/299/console). Returning `None` on failure is seldom a safe API.

rob-c:
Well I wanted to raise a new type of exception here of no valid proxy but this potentially creates more work with handling of exceptions.
Frankly the DIRAC monitoring code has now been cleaned up enough that we can look at requiring a valid proxy and simply construct one here is no proxy has been yet created.
This would alleviate potential problems whilst still providing us with a monitoring system that demands a proxy be constructed during startup.

PR #534, opened 2016-05-27
Port DIRAC unit tests

milliams:
This ports over the DIRAC unit tests. Largely, the tests are ported over without changing much but I have tried to modernise where possible. This is in two main ways:

1. Use pytest fixtures to provide context to the tests
2. Use the `mock` library to replace the manual mocking that was happening before.

It also now work correctly on the GridPP DIRAC instance.


python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -82,8 +107,11 @@ def start_ganga(gangadir_for_test, extra_opts=[]):
         logger.info("Initializing")
         Ganga.Runtime._prog.initEnvironment()
     else:
-        from Ganga.Runtime.Repository_runtime import startUpRegistries
+        # Load the config from the appropriate files
+        load_config_files()
```
rob-c: Why are we doing something different here to loading the Config in the same way as ```ganga``` and bootstrap?
I'd much rather see us have some consistent way of loading the config between all 3 than 3 special cases of how the config is to be handled.
If there is a solid reason why this is special then OK.

python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -186,9 +214,7 @@ def stop_ganga():
     ShutdownManager._ganga_run_exitfuncs()
 
     # Undo any manual editing of the config and revert to defaults
-    from Ganga.Utility.Config import allConfigs
-    for package in allConfigs.values():
-        package.revertToDefaultOptions()
+    clear_config()
```
rob-c: Again, it would be nice to see this included somwhere in the normal Ganga shutdown rather than here, I would prefer to see this as minimal as possible. I can't think why this is a special case

python/GangaDirac/test/Unit/Backends/TestDiracBase.py
```diff
@@ -0,0 +1,422 @@
...
+from Ganga.GPIDev.Lib.File import File
+from Ganga.GPIDev.Adapters.StandardJobConfig import StandardJobConfig
+from Ganga.Lib.Splitters import ArgSplitter
+from Ganga.Lib.Executable import Executable
+from GangaDirac.Lib.Backends import Dirac
+from GangaDirac.Lib.Backends.DiracBase import DiracBase
+from GangaDirac.Lib.Files.DiracFile import DiracFile
+
+script_template = """
+# dirac job created by ganga
```
rob-c: Can we please drop the hard-coded JDL template descriptions from the tests?
We want to know if the JDL is consistent to DIRAC but keeping around the template here is just adds to the burden whenever we need to change the API.

python/GangaDirac/test/Unit/Backends/TestDiracBase.py
```diff
@@ -0,0 +1,422 @@
...
+        i = 0
+        for sj in j.subjobs:
+            sj.master = j
+            sj.id = i
+            i += 1
+
+        subjob = True
+        assert db.getOutputData() == ['a', 'b', 'c'] * 3
+        assert db.getOutputData(None, ['beta']) == ['b'] * 3
+        assert db.getOutputData(os.path.expanduser('~/gangadir_testing'), ['alpha', 'charlie']) == ['a', 'c'] * 3
```
rob-c: Isn't this safer to get the gangadir from the config?

python/Ganga/testlib/GangaUnitTest.py
milliams: This is doing the same as the normal Ganga startup. Normally, Ganga calls `GangaProgram.configure()` which, along with a bunch of other things (parsing command line, printing the welcome message, setting logging levels, setting up the spyware config, bootstrapping logging, setting command-line config options, doing some legacy test stuff, doing some `--daemon` stuff) parses the config files and sets their values.

This is pulling out just the "parse the config files and sets their values" part since the rest is rather specific to the `ganga` *application*.

python/Ganga/testlib/GangaUnitTest.py
milliams: It's a special case because in the normal use of Ganga, there's no point in clearing the config on shutdown. It takes time when the process is about to end anyway.

python/Ganga/testlib/GangaUnitTest.py
rob-c: If this can be pulled out why isn't it just pulled out as a method in the Ganga code then and called from here. I don't see any real benefit in duplicating functionality of 'read the config system' in 2 or more places. If we ever change 1 we run the risk that the other becomes out of sync for no good reason.

python/Ganga/testlib/GangaUnitTest.py
rob-c: I don't think I agree that this _should_ take time. If it is taking significant time then this is a bug, but either way I would argue it makes more sense to have in the Core shutdown as there's no reason not to clean up after these objects in the normal code as well

python/Ganga/testlib/GangaUnitTest.py
milliams: The code was factored out. That is what #514 was for.

milliams:
retest this please

python/GangaDirac/test/Unit/Backends/TestDiracBase.py
milliams: Indeed, or even better a temp dir. I will fix this.

milliams:
retest this please

milliams:
retest this please

milliams:
retest this please

rob-c:
Are we still looking for 6.1.21 would be nice to see these tests in there given the recent changes in the DIRAC code 👍

milliams:
These tests are now passing based on the latest code in `develop`. I've run the external tests as well at https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/50/ and they are all passing.

I'm happy to merge these for 6.1.21 if other people think that is a good idea? They pretty much only touch the tests.

egede:
So this is testing against the DIRAC instance at Imperial now?

milliams:
@egede Indeed, this is testing against Imperial's vanilla DIRAC instance using the DIRAC UI we have installed in our CVMFS area.

egede:
@milliams I am all for prompt inclusion then.

PR #543, opened 2016-06-01
Remove spurious '\' from newline character

### LINKED ISSUE #519: job.outputfiles "namePattern" and "locations" contain spurious r'\n' in Ganga 6.1.20
I've got some problems with a chain of transforms in Ganga 6.1.20 which runs fine in Ganga 6.0.44.

The first transform in the chain analyses 10 data files in "split" mode, one data file per PBS job.

In Ganga 6.0.44, the job is created by a call to "GPI.Job()" and then, if I add a "debug" output in form:
```
trf = self._getParent()
logger.info('trf.outputfiles = %s',trf.outputfiles) # a "debug" output
```
I get exactly what I expect:
```
trf.outputfiles = [MassStorageFile(namePattern='*_validtree_*.root')]
```

In Ganga 6.1.20, the job is created by a call to "makeRegisteredJob()" and the same "debug" output gives:
```
trf.outputfiles = [MassStorageFile(namePattern='*_validtree_*.root'), ]
```

Note the additional two characters in the end for Ganga 6.1.20 (i.e. note `"), ]"` in the end). I do not know if they may create any problems later.

All these 10 PBS jobs run fine and the first transform is marked as "completed" in both, Ganga 6.0.44 and Ganga 6.1.20.

The 10 output files from this first transform are then used by the second transform in the chain (via a "TaskChainInput"). This transform creates a single PBS job (which should get all 10 files from the first transform).

In Ganga 6.0.44, I have:
```
job = GPI.jobs(self.active_job_ids[0])
logger.info('job.outputfiles = %s',job.outputfiles) # a "debug" output
```
and, for each of the 10 files, I get exactly what I expect:
```
job.outputfiles = [MassStorageFile (
 compressed = False ,
 namePattern = 'name_of_the_validtree_file_i.root' ,
 outputfilenameformat = 'Where/These/Files/Are/Present/{fname}' ,
 localDir = '' ,
 joboutputdir = '' ,
 failureReason = '' ,
 locations = [] ,
 inputremotedirectory = None
 )]
```

Note the "namePattern" and "locations" above and compare them with what is shown below for Ganga 6.1.20.

In Ganga 6.0.44, this transform runs fine (i.e. it is "completed" without problems).

In Ganga 6.1.20, I have:
```
job = getJobByID(parent.active_job_ids[0])
logger.info('job.outputfiles = %s',job.outputfiles) # a "debug" output
```
and, for each of the 10 files, I get something like this:
```
job.outputfiles = [ MassStorageFile (
   compressed = False,
   namePattern = 'name_of_the_validtree_file_i.root\\n',
   outputfilenameformat = 'Where/These/Files/Are/Present/{fname}',
   localDir = '',
   joboutputdir = '',
   failureReason = '',
   locations = ['/...My_Mass_Storage.../Where/These/Files/Are/Present/name_of_the_validtree_file_i.root\\n'],
   inputremotedirectory = None
 )]
```

Please note "namePattern" and "locations" above and compare them to what is shown above from Ganga 6.0.44.

In Ganga 6.1.20, in the ends of both these strings, you can clearly see TWO SPURIOUS CHARACTERS: "\\" and "n" (note: this is NOT the newline character "\n", these are really two separate characters).

That is the reason why this transform fails to run in Ganga 6.1.20 with an error "No input data file given.".
It searches for files which have names which end with these two characters r'\n' (and they don not exist, of course).

Could you, please, think why / what adds these two characters ("\\" and "n") to the file name (when it is passed to the second transform as its input file, after the first transform is "completed")?

drmarkwslater:
I think in the transition from inline script to separate file, the escaped newline character was not changed and so this added extra string noise to the `__postprocesslocations__` file. This PR removes that.

Fixes #519.

### DIFF
--- a/python/Ganga/GPIDev/Lib/File/scripts/MassStorageFileWNScript.py
+++ b/python/Ganga/GPIDev/Lib/File/scripts/MassStorageFileWNScript.py
@@ -29,7 +29,7 @@ for massStorageList in ###MASSSTORAGECOMMANDS###:
 
     (exitcode, mystdout, mystderr) = execSyscmdSubprocessAndReturnOutputMAS('%s %s' % (cm_ls, pathToDirName))
     if exitcode != 0:
-        ###POSTPROCESSLOCATIONSFP###.write('massstorage %s ERROR %s\\n' % (filenameWildChar, mystderr))
+        ###POSTPROCESSLOCATIONSFP###.write('massstorage %s ERROR %s\n' % (filenameWildChar, mystderr))
         continue
 
     directoryExists = False
@@ -44,7 +44,7 @@ for massStorageList in ###MASSSTORAGECOMMANDS###:
         (exitcode, mystdout, mystderr) = execSyscmdSubprocessAndReturnOutputMAS('%s %s' % (cm_mkdir, path))
         if exitcode != 0:
             printError('Error while executing "%s %s" command, check if the ganga user has rights for creating directories in this folder' % (cm_mkdir, path) + os.linesep + mystderr)
-            ###POSTPROCESSLOCATIONSFP###.write('massstorage %s ERROR %s\\n' % (filenameWildChar, mystderr))
+            ###POSTPROCESSLOCATIONSFP###.write('massstorage %s ERROR %s\n' % (filenameWildChar, mystderr))
             continue
 
     folderStructure = ''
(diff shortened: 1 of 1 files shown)


drmarkwslater:
Any objections to merge?

rob-c:
go from me :+1:

PR #547, opened 2016-06-02
Fixing the IPython logging hook to not flush during text entry

### LINKED ISSUE #546: Logging cache flushing too early
The logging Memory cache is flushing on every single return statement.

This can be quite annoying if entering say a for loop command whilst the monitoring is busy.

The expected behaviour is that the logging only be flushed on the finishing of an input statement in IPython.

I think we can achieve this without having a custom input parser again, as I think there is a hook for post-input in their API. If this is a 1 line fix I'll look into it.

rob-c:
Fixes #546 

This PR is very short but worth documenting. I'm changing the IPython hook to be the `pre_run_code_hook` which is run after the text input in a command line has been finished.
This replaces our use of the `pre_prompt_hook` which is executed every time a user hits enter which for complex functions can be quite a few times before they've finished inputting a command.

The result is that the logging no longer flushes and interrupts the user input.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -1248,7 +1248,7 @@ under certain conditions; type license() for details.
         from Ganga.Utility.logging import enableCaching
         enableCaching()
 
-        ipshell.set_hook("pre_prompt_hook", ganga_prompt)
+        ipshell.set_hook("pre_run_code_hook", ganga_prompt)
 
         from Ganga.Runtime.IPythonMagic import magic_ganga
         ipshell.define_magic('ganga', magic_ganga)


rob-c:
Any objections to merge?

drmarkwslater:
Does this mean that a user won't see monitoring messages until they've actually inputted code rather than just pressing enter at the prompt?

alexanderrichards:
As I only ever seem to input single line commends into Ganga I didn't realise that this could be a problem but I see that if your doing multiline input you would get a prompt and hence logging output somewhere in the middle.

One question, does this hook also get called if a user hits enter without and code, i.e. just advancing to the next line?

alexanderrichards:
@drmarkwslater snap, you beat me to it :wink:

rob-c:
I'll check that this still works with no input. It's not clear but I think this is almost the equivalent of a `post_prompt_hook` (which doesn't appear to exist)

rob-c:
Yes, just hitting return means that the logger is still flushed every time I hit enter :+1: good call for checking that I wasn't concentrating enough.
I'm not experiencing any errors so I can't comment on if these are flushed as and when they happen (I expect that they are) but I can say that if I wait long enough with the monitoring having queued tasks I can now type a multi-line function without any interference from the logger.

rob-c:
There does appear to be the caveat that it doesn't completely work if you move back in your history and then edit a function in there by adding another line. But somehow I think this is likely a bug external to us.

alexanderrichards:
> There does appear to be the caveat that it doesn't completely work if you move back in your history and then edit a function in there by adding another line.

presumably though it will flush the next time you hit enter though even on a blank line right afterwards?

rob-c:
It's broken in the sense that when you're going back and editing a function by hitting return to add a new line the logger then flushes which isn't what you want. Still it doesn't do it when writing the function in the first place which I would argue is the most annoying time for this to happen.

milliams:
I think that the slight brokenness that you mention @rob-c is corner-case enough that it's worth it. Go for merge from me :+1:

alexanderrichards:
ah ok, I think though if I understand the original problem that under the old system you would also have this same caveat as you would be getting a new prompt for this new line.

I think then that this change is an improvement overall so I'm in favour unless anyone else can think of something that we have missed :+1:

PR #549, opened 2016-06-02
fixing IPython for future versions as well as IPython 1.x

rob-c:
this PR fixes some annoying warnings I get when running ganga out of the box with a different IPython, i.e. not using pip install.

This shouldn't adversely effect anything but I'm going to check again.

I just had to put this in as the flush command isn't called from the old hook in IPython 4 on my machine.


python/Ganga/Runtime/bootstrap.py
```diff
@@ -1216,10 +1216,16 @@ def launch_IPython(local_ns, args, error_handler, ganga_prompt):
 
         import IPython
 
+	ipver = IPython.__version__
```
drmarkwslater: Possible indent issue here? Looks like on github anyway 😄

rob-c: Possibly, I think we're allowed to stick in tabs here by accident (!) I'll check privately.

drmarkwslater:
Other than the indent thing, 👍

rob-c:
OK, this may warrant a further look by people.

I have now detached the flushing of the cache with the rest of the standard Ganga prompt check function.
(This function was being called on every carriage return vs being called once as the user arrives at a new prompt or just before some code has been executed)

The only consistent API I've found across quite a wide range of IPython versions is to hook into the `PromptManager` and create a dynamic prompt string which exploits putting code into a dummy classes `__str__` method.
This works as intended across multiple versions without issue.
(I'd prefer to have the cache flushed before we execute a command but doing it afterwards isn't necessarily a bad thing it just means users may have to scroll back in history to see their output)

The result of this is that I'm not sure if we want to keep the additional credential check prompt which crops up?
I suspect this to be broken atm (or at least not used by certain credentials) in addition to probably being defunct once the credentials branch is used.

I'm tempted to leave in the `ganga_prompt` function unless we all agree it's safe to remove this as it only performs a credentials check as of this PR.

drmarkwslater:
Do we want to revert #547 for the moment? I would say this is fairly low priority so may be just postpone looking at for a bit?

rob-c:
@drmarkwslater If it's simple to revert then I'm happy to delay looking properly at this, I'm probably not going to be quite done with other work until Monday. It's a __pain__ having the monitoring dump stuff to screen just as I'm finished typing several lines of code though it makes it impossible to do anything too technical while jobs are coming back.

drmarkwslater:
I certainly think we should do this but I was just saying that if this is having knock on effects it's probably not worth holding up the release on Monday for it 😄

rob-c:
I know you're not saying to delay this (or to hold the release up). The only reason I wrote this quickly was that it was really getting to me after I'd fired up ganga for the 4th or 5th time in a day, forgotten the monitoring was on, started to type some small function to test something and just before getting to the final line getting spammed by monitoring loop :P

rob-c:

drmarkwslater:
I thought we'd decided in the meeting to hold off on this because it was a bit more invasive than first thought and we didn't want to hold up 6.1.21?

rob-c:
I was under the impression we hadn't decided whether to hold off on this vs reverting the change. I was under the impression fixing this is quite good.
It's this or we add some 30/40 lines of code to deal with things properly with different IPython versions and live with the bug in 1.x which we can't then fix without dropping Python 2.6 support.
I'll push the change as part of #357 in that case if we're holding off...

rob-c:
For reference does anyone object to fixing this, this way?

drmarkwslater:
I certainly think this would good to fix in general, it was just that this seems to be a more in-depth code change than originally intended and we don't want to accidentally break something for a release which will hopefully so much else.

rob-c:
I'm not seeing any real risk associated with this? I'm inserting an object into the local_ns which is used to evaluate the prompt dynamically which is an intentional feature of IPython.
Hooking some action into the call to `__str__` is used by at least 1 or 2 other projects I've seen using IPython as an interactive prompt so it's not that unusual, even if it's a little strange at first.

drmarkwslater:
So having a look at the new implementation it seems to be fine but I'd prefer someone else to look at it just to be sure as I don't trust myself to understand it all properly. @alexanderrichards @milliams can you have a check just to confirm? I'm just worried about breaking the behaviour of the prompt!

milliams:
Every change we make *always* has risk. That's the nature of software development. I *do* think however, that injecting information like this into the prompt template is probably the best approach.

My reservation however is that `__flushCmd` is being injected to print out logging messages but expired credentials etc. are still being reported via the `post_execute`/`pre_run_code_hook` hooks (with `ganga_prompt()`). I'd prefer some standardised way for subsystems to inject information into the prompt for the user rather than simply adding another one.

rob-c:
I was wanting to avoid touching anything to do with credentials as in my mind this is all done via the credentials branch, and I was preferring to have all references to flush commands in the `Utility/logging` for readability.

alexanderrichards:
If I can just play devils advocate for just a second. Ought we to perhaps consider that since a lot of the stuff @rob-c has had to put in to make the switch to `pre_run_code_hook` work is down to us having to be backwards compatible, is there an argument that perhaps we should keep the status quo (as this is a minor bug) until #38 is merged and then we can do away with a lot of the logic here making this a much smaller merge?

At the very least we should keep in mind to revisit this and remove the extra logic once #38 is merged.

That being said I don't see anything too dangerous here and it all seems in order so if everyone else is happy then I am too :wink:  :+1:

alexanderrichards:
> I'd prefer some standardised way for subsystems to inject information into the prompt

Agreed

drmarkwslater:
@alexanderrichards I suspect #38 is still some way off in which case it would nice to have this fixed before then though you're right - we need to revisit this when that's been merged.

@rob-c  Looks like you're good to go!

PR #551, opened 2016-06-03
Cleanup DIRAC env caching

milliams:
Sourcing the cached environment for LHCb is showing to be slightly fiddly and sometimes fails to load completely. Different shells and environments can cause it to source the env file badly.

This PR allows to cache the DIRAC environment in a JSON file which is safely readable and avoids the problems of exported bash functions entirely.

This avoids the repeated use of bash for:

1. source a command in a bash session
2. `printenv` it to a file
3. source it again in a bash session
4. use it

and replaces it with:

1. source a command in a bash session
2. save it to a JSON file
3. read the JSON file
4. use it.

Other plugins can do the same but only LHCb are slow enough to need it. The environment is still cached per-session in memory. The code to do the caching is moved to the DIRAC plugin.

Deprecate the `DiracEnvFile` option and replace it with a `DiracEnvSource` option which should point to a sourceable script to provide the environment.

`DiracEnvJSON` can point to a specific cached env to override the source command.


python/Ganga/Utility/execute.py
```diff
@@ -283,7 +283,7 @@ def execute(command,
         if stdout:
             stdout = pickle.loads(stdout)
     except Exception as err:
-        logger.error("Err: %s" % str(err))
+        logger.debug("Err: %s", err)
```
rob-c: This is just hiding problems isn't it? We normally don't get here anyway as most uses don't use the stdout I think

milliams: Ahh, I think this used to be `debug` but you changed it to `error` recently and I rebased incorrectly. I will change it bask to `error`.

python/GangaLHCb/__init__.py
```diff
@@ -85,43 +85,21 @@ def _store_root_version():
 
 
 def _store_dirac_environment():
+    from GangaDirac.Lib.Utilities.DiracUtilities import write_env_cache
     diracversion = _guess_version('LHCBDIRAC')
-    import tempfile
-    import subprocess
-    import os
     platform = os.environ['CMTOPT']
     setup_script = 'SetupProject.sh'
```
rob-c: Can we change this to the `lb-run` commands and check it works in this case if we're modifying things here?

rob-c:
just an FYI, we will have to make sure we bump the LHCbDirac version on this doesn't conflict with older caches.

python/GangaLHCb/__init__.py
milliams: Sure, what would be the full `lb-run` command needed to get the environment? I'll give it a test.

milliams:
Indeed, I've made sure that old caches are invalidated by storing the new cache in a new location. They are now stored in `~/.cache/Ganga/GangaLHCb/x86_64-slc6-gcc49-opt/v8r2p27` rather than `~/.cache/LHCbDIRAC_ENV/x86_64-slc6-gcc49-opt/v8r2p27`

rob-c:
Is reading this JSON file any faster/more reliable than simply sourcing the bashrc on startup like we currently do?

python/Ganga/Utility/execute.py
```diff
@@ -207,7 +207,7 @@ def execute(command,
     else:
         # We want to run a shell command inside a _NEW_ shell environment.
         # i.e. What I run here I expect to behave in the same way from the command line after I exit Ganga
-        stream_command = "bash -i "
+        stream_command = "bash "
```
rob-c: why the change?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -31,36 +29,79 @@
...
-    return DIRAC_ENV
+    Args:
+        env_source: a command which can be sourced, providing the desired environment
+
+    Returns:
+        dict: the environment
+
+    """
+    logger.debug('Running DIRAC source command %s', env_source)
+    env = dict(copy(os.environ))
```
rob-c: why not just deepcopy?

python/GangaLHCb/__init__.py
rob-c: `lb-run LHCbDirac v8r2p45 python -c 'from __future__ import print_function; import os; print("%s" % os.environ)'``works for me

python/Ganga/Utility/execute.py
milliams: Because previously when sourcing the environment, it was giving so many errors that an interactive bash session was failing to complete. Furthermore, we are not running interactively so it's only likely so confuse scripts which check for how they are running and then try to request input from the user.

python/GangaLHCb/__init__.py
milliams: I've tested it and it fails due to the presence of `ROOT` in the command `source lb-run LHCBDIRAC v8r2p27 ROOT`. Removing that part of the command still gives problems because the model for `lb-run` is that is not sourceable. It's the difference between:

```bash
source SetupProject.sh LHCbDirac v8r2p27; python -c "import os; print([e for e in os.environ if e.startswith(\"DIRAC\")])"
```

and
```bash
lb-run LHCBDIRAC v8r2p27 python -c "import os; print([e for e in os.environ if e.startswith(\"DIRAC\")])"
```
(Note the semi-colon and the `source` command).

The problem is that `execute()` automatically adds on the `; python -c ...` part when doing an `update_env`. It looks like using the current `update_env` will not work for `lb-run`.

I'll add another commit to this branch to make LHCb do their own, simplified version of `update_env`.

milliams:
Reading the JSON file is both faster (mostly due to not having to start another process) and more reliable (due to not having to worry about bash functions and escaping etc.).

python/GangaDirac/Lib/Utilities/DiracUtilities.py
milliams: Because `copy` is simpler and faster and all we need here. `copy` is always my default and I fall back to `deepcopy` only when we need it.

python/GangaLHCb/__init__.py
rob-c: I would argue this really shouldn't rely on update_env in this case.
Why not simply get the original command you're using to capture the environment directly from within the `lb-run` environment? That way we avoid having to work with update_env which won't work here.

rob-c:
@milliams I'm not strongly convinced. If we have to run this once every time the environment changes (and there's nothing in here to update the file if the dirac-ui changes or equivalent) then we can (and _should_) trust the bashrc command to work. We're even using the same types of commands to talk to DIRAC so they best work.
I agree that I prefer to read json to a _lot_ of other formats but we're creating a 2nd file which will become out of sync with the first file just creates overhead.
Simply fixing the problems of the LHCb env storing a correct bashrc on disk, and removing the `-i` gets us 90% of the features of this branch with much less work I'm convinced. Without having to do things like checking timestamps/checksums on the env file to see if it's been updated since the last time we cached it.

python/GangaLHCb/__init__.py
milliams: That's what I've tested and indeed works locally. Do we want a switch to be able to select either based on a config or should we just switch to `lb-run` wholly?

milliams:
To clarify:

Vanilla DIRAC still does not use any sort of on-disk cache. It runs the `bashrc` at startup and stores the environment in the in-memory `DIRAC_ENV` variable. This has not changed in this PR.

LHCb DIRAC has changed to store the environment in a JSON file rather than a dump from `printenv`. Conceptually, this is the only change. As before, this file is keyed on the version of LHCbDIRAC used so will update when the version changes.

The other changes in this PR are to break out some common code into reusable functions to reduce repetition between vanilla and LHCb.

python/GangaLHCb/__init__.py
rob-c: If you mean `lb-run` vs `SetupProject` we should migrate to `lb-run` as this takes care of all situations past and present for us it reverts to `SetupProject` silently and correctly when needed.

milliams: Ok, I can make that change as a commit in this PR if you'd like? It's only about a 3-line change.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
rob-c: Surely this is teh same speed but with 2 calls vs 1? I don't know how dicts are duplicated behind the scenes so I'm just curious

python/GangaLHCb/__init__.py
rob-c: Excellent if you could that would save a shorter PR later on

rob-c:
Ah ok, so by default the `DiracEnvSource` is empty and so doesn't source from disk and the `DIRAC_ENV` is effectively pre-populated by GangaLHCb on the loading of the plugin?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
milliams: I see your point now, I think I was confusing myself. So, the `copy` was needed so that we don't change the global variable. The `dict()` is needed since `os.environ` is not a Python `dict` but rather of type `os._Environ`. You are right though that both are not needed, simply converting it to a `dict()` should be sufficient. I will update this.

milliams:
Yes, GangaLHCb does its special stuff in `GangaLHCb/__init__.py` which creates the cache if it doesn't exist and sets the config variable `DiracEnvJSON` to point to it for the duration of the session. This process is almost completely unchanged to how it worked before.

(note, I do see that the tests are failing and I am looking into it. They passed locally so there's some environment issue I guess)

milliams:
The latest commit ("Use lb-dev for the LHCbDIRAC environment" 5ab440f) makes it use `lb-run` instead of `SetupProject.sh` for getting the DIRAC environment. We'll see how it fares.

milliams:
Ok, so the tests are passing. I welcome any more comments on this.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -31,36 +29,78 @@
...
     Args:
         force (bool): This triggers a compulsory reload of the env from disk
     """
     global DIRAC_ENV
     with Dirac_Env_Lock:
-        if DIRAC_ENV == {} or force:
-            config_file = getConfig('DIRAC')['DiracEnvFile']
-            if not os.path.exists(config_file):
-                absolute_path = os.path.join(os.path.dirname(os.path.realpath(__file__)), '../../..', config_file)
+        if not DIRAC_ENV or force:
```
rob-c: I wonder if we should be overloading the 'force' method here. Or even drop it.
I've only ever used it in testing and I don't think it's used anywhere in the codebase?.
I worry that for LHCb users this could cause problems if we allow it to remain and then it gets used somewhere because it works on vanilla DIRAC.

milliams: So you're suggesting that we remove this argument? I think I agree. It won't necessarily do what is expected as it will only clear the in-memory cache, not the disk-cache. Nowhere uses in the code so it's safe to remove.

I will remove it in another commit incoming.

rob-c: OK, great to hear :+1:

rob-c:
Could/Should we remove caches older than (n days)? I'm regularly removing the cache myself for testing etc but some users could end up with a few files here over time, should we be good citizens and add a cleanup method of some form?

milliams:
I think that a timeout on the cache could make sense but it depends on what problem it is solving. All I've seen mentioned is `SetupProject.sh` failing to create a sane environment (e.g. due to AFS or CVMFS issues) and that bad env being cached and used 'forever'. A time-based cache-invalidation would not solve this issue. We should take further discussion to another issue or PR though.

rob-c:
I've never seen SetupProject fail to setup the correct env. I've seen it fail to load from disk but that's another issue that is being discussed with workarounds.
I'm more thinking about dumping every single version of LHCbDirac we support to disk in a new file could generate a lot of files if we ramp up our release process once we've got the plugins as part of the regular testing. Simply deleting all files older than 3 or 4 months or something similar in 2 or 3 lines would be us being good citizens I would have thought.

PR #552, opened 2016-06-03
Cleanup shutdown policy code

drmarkwslater:
This PR cleans up the shutdown callback code as after looking at it, I found most of it was unnecessary. The original design of this was with callback functions stored in `Ganga.Core.__init__.py` that were different depending on batch or interactive mode. Whichever had been picked was then used when `GangaThreadPool.shutdown` was called.

However, this all seemed like a lot of effort for just deciding in the shutdown method whether to ask the user if they should force the shutdown or not so I've cleaned out all this and the shutdown policy is now stored as a string in the `GangaThreadPool` class. In normal operation, this is still set in the `bootstrap` function of `Ganga.Core.__init__.py` depending on the interactive flag sent to the function. There are a few more places where the `change_atexitPolicy` is called explicitly and I've changed these appropriately as well.


python/Ganga/Core/GangaThread/GangaThreadPool.py
```diff
@@ -93,6 +95,7 @@ def _really_shutdown(self, should_wait_cb=None):
         shutdown_thread.start()
 
         t_start = time.time()
+        t_last = -t_start
```
milliams: Is this a typo? Should `t_last` be negative?

drmarkwslater: This is actually part of the logic to check for the first time the user is prompted as there are two timers involved in that. It is a bit confusing though so Il'l add a comment 😄

PR #554, opened 2016-06-03
Fixing box repo names

### LINKED ISSUE #536: Index cache not saved correctly for box objects
If I do

```
box.add(Dirac(),'Test')
```
and then quit Ganga, I see
```
ERROR    Failed to Shutdown Repository: box !!! please check for stale lock files
INFO     Stopping Job processing before shutting down Repositories
INFO     Shutting Down Ganga Repositories
INFO     Registry Shutdown
ERROR    RegistryAccessError: The object #1 in registry 'box.metadata' is not fully loaded and the registry is disconnected! Type 'reactivate()' if you want to reconnect.
ERROR    Trying to Shutdown cleanly regardless
```
When I subsequently start Ganga, the name of the box object is lost, but it works anyway with a warning.
```
Ganga In [1]: box['Test']
Ganga Out [1]: WARNING  Incorrect index cache of 'box' object #3 was corrected!

 Dirac (
   status = None,
   actualCE = None,
   settings = [CPUTime],
   normCPUTime = None,
   extraInfo = ,
   diracOpts = ,
   statusInfo = ,
   id = 3
 )

[13:50:25]
```

rob-c:
Fixes #536 
The order in which we were shutting things down was interfering with the box repo code.

This PR fixes it so that the box repo has the correct indexes.

Re-loading all effected objects from disk after this PR is in will restore any broken indexes automatically.


milliams:
This makes sense to me :+1:

PR #555, opened 2016-06-03
Create DIRAC proxy

milliams:
This causes every call to DIRAC's `execute()` function to ask to make a proxy if it has expired. Currently, if there is no proxy present when Ganga starts, no proxy will ever be created and errors like:
```
AttributeError: 'NoneType' object has no attribute 'get'
```
will be raised.

Perhaps this is unsafe because it might be called in a background thread but somehow we need to get a proxy created.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -150,7 +150,7 @@ def manualExportToGPI(my_interface=None):
     # At this point we expect to have the GridProxy already created
     # by one of the Grid plugins (LCG/NG/etc) so we search for it in creds
     # cache
-    credential = getCredential(name='GridProxy', create=False)
+    credential = getCredential(name='GridProxy')
     if credential:
         exportToInterface(my_interface, 'gridProxy', credential, 'Objects', 'Grid proxy management object.')
 
--- a/python/GangaDirac/Lib/Utilities/DiracUtilities.py
+++ b/python/GangaDirac/Lib/Utilities/DiracUtilities.py
@@ -222,7 +222,7 @@ def execute(command,
             python_setup='',
             eval_includes=None,
             update_env=False,
-            renew=False):
+            ):
     """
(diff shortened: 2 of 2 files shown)


milliams:
Alternatively, this could be set to create the proxy on Ganga startup (if it's invalid). I'm sure this is how it used to work, was it changed for a reason?

rob-c:
OK, there appear to be no more spurious calls to DIRAC by the monitoring which aren't protected by first testing that there is a valid proxy before proceeding.
I'm assuming in the new credentials branch we have some method/decorator of silently testing if there is a credential available for some code which allows the code to simply not run if the credential isn't available, sort of like a hard and soft requirement?

rob-c:
Go for merge from me as the Monitoring code appears to be sane enough to work with this.

milliams:
Ok, so I agree that the monitoring loop seems to be working sanely with this by disabling itself if the grid proxy is not valid (to avoid asking for input in a background thread and hanging forever). The issue I then saw is that since the grid proxy object is not being created at import time, it was not being exported to the GPI so I have edited bootstrap.py to make a grid proxy object always be created. This should be a silent operation and DIRAC will keep its "only ask when needed" functionality.

This should fix the nightly tests as well.

Comments?

rob-c:
I'm in favour of the merge but is there a reason the bootstrap wasn't creating this in the past?
I don't know how this compares to the behaviour in the credentials branch but is this just creating the credentials once for the whole of Ganga?

milliams:
In the past (until #556) the `gridProxy` object was being created in the DIRAC plugin at module import time (or by the LCG plugin when it starts) so by the time that line of code in bootstrap ran, there was already something in the credential store. Since #556 deferred the creation of the object, boostrap.py would find nothing and therefore not export the `gridProxy` name to the GPI. I'm not sure exactly how this should work with the new credentials system though that is my next priority since we're starting to see more credentials problems cropping up.

PR #556, opened 2016-06-03
Defer creating the proxy object until the first time it is requested

milliams:
This ensures that the LHCb plugin has finished loading before the proxy is created. Objects should not be created at import time.

This is one half (the less controversial half) of #555.


rob-c:
As long as the monitoring doesn't trigger an unwanted proxy creation at startup or during running I'm happy to move to this.
I have been toying with raising an invalid exception but that causes more work which will go away.

milliams:
Simply creating the proxy object doesn't prompt the creation of a proxy file on disk so this is safe. The change in #555 which layers on top of this is perhaps less safe and needs some thought.

PR #559, opened 2016-06-06
If default_se is not in all_storage_elements add it

milliams:
Previously, if these were not perfectly matched, an exception would be raised. This means that the user can set `defaultSE` on their `DiracFile` without also having to set the config option `allDiracSE`.

Also, add and use Mark Slater's static_vars() decorator for the global state.

### DIFF
--- a/python/Ganga/Utility/decorators.py
+++ b/python/Ganga/Utility/decorators.py
@@ -1,0 +1,16 @@
+def static_vars(**kwargs):
+    """
+    Decorate a function and provide it with some static variables.
+    >>> @static_vars(counter=0)
+    ... def foo():
+    ...     foo.counter += 1
+    ...     return foo.counter
+    >>> foo()
+    1
+    >>> foo()
+    2
+    """
+    def decorate(func):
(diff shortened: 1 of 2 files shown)


rob-c:
The whole point of this was that Ganga is supposed to check against this list of SE to see if the default one has been added.
Given we're not able to poll DIRAC to a list of SE that the user has access to (in a nice and simple way) I move to simply remove the check to see if the SE is there.

It was a nice idea to stop the user from trying to write to places they're not supposed to but this adds overhead to us that I wasn't expecting when I coded this up from an LHCb perspective a long time ago.

rob-c:
Writing to an SE which the user doesn't have access to will fail anyway in DIRAC and we can leave it to be caught there.

rob-c:
I'm not in favor of pushing this as part of 6.1.21 as this changes the logic in a way that wasn't intended and will require a lot of testing. Can we drop this to 6.1.22 at least as I think this situation can be improved by getting rid of the check here and being careful with the configuration system.

rob-c:
Superseded by #591

PR #562, opened 2016-06-07
Add quotes around lfn in hasattr call

### LINKED ISSUE #561: Problems with Dataset for LHCb users
LHCb users are reporting problems with the Dataset class (which probably shouldn't be generated anyway)

```
/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v601r20/install/ganga/python/Ganga/GPIDev/Lib/Dataset/GangaDataset.py
in getFileNames(self)
     77         names = []
     78         for i in self.files:
---> 79             if hasattr(i, lfn):
     80                 names.append(i.lfn)
     81             else:
```
I suspect this is a quick fix to put in the quotations.

drmarkwslater:
I stupidly missed this when putting this code in a month ago. This fixes the lack of quotes around 'lfn'.

Fixes #561


rob-c:
👍

PR #563, opened 2016-06-07
Making chosen_lfns use sets instead of lists

drmarkwslater:
The code in `OfflineGangaDiracSplitter` is checking for already used lfns with:
```
if iterating_LFN in chosen_lfns:
```
This is fine until the `chosen_lfns` list goes above a few hundred at which point everything grinds to a halt. This PR changes `chosen_lfns` to use a `set` instead. In a very quick & dirty splitting test over the 2015 data, this speeds this part of the function (not the whole thing) up from 2+ hours (on a ~fast desktop) to ~1 sec.

There are still other areas in this code that could be looked at to speed things up but this is a big improvement with minimal changes.

### DIFF
--- a/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
+++ b/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
@@ -384,7 +384,7 @@ def OfflineGangaDiracSplitter(_inputs, filesPerJob, maxFiles, ignoremissing):
         # NB: Can't modify this list and iterate over it directly in python
         LFN_instances = site_dict.keys()
         # Already used LFN
-        chosen_lfns = []
+        chosen_lfns = set([])
 
         for iterating_LFN in LFN_instances:
 
@@ -429,7 +429,7 @@ def OfflineGangaDiracSplitter(_inputs, filesPerJob, maxFiles, ignoremissing):
                 for lfn in _this_subset:
                     site_dict.pop(lfn)
                     allChosenSets.pop(lfn)
-                    chosen_lfns.append(lfn)
+                    chosen_lfns.add(lfn)
 
         # Lets keep track of how many times we've tried this
         iterations = iterations + 1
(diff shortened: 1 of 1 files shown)


rob-c:
@drmarkwslater Am I correct in thinking this changes the comparison from a string comparison to a pointer comparison?

drmarkwslater:
Err... Don't know 😄  But I know that sets are much better for large lists due to the implementation. I guess hashes are involved somewhere....

milliams:
It looks like this was introduced in 95c5f75e6d84e04dca8d975af38da8fbf8506dc9 last year so it's likely been a problem since then.

It's not a case of string vs pointer, it's a case of equality comparison vs binned hash comparison.

milliams:
And just to be clear, we don't care about the order of the items in this output 'list'?

drmarkwslater:
@milliams  Not for `chosen_lfns` - as far as I can tell from the code it's just a store for already processed LFNs.

rob-c:
Either way, excellent, thanks for this.
I'd planned to revisit this and make the comparisons better as I was aware that string comparisons are awful but I didn't realise this was the thing holding everything up.

This was introduced by design when I wrote this as I didn't know how that would scale. Nobody complained as the size of the subsets improved dramatically.
We don't care about the order of the elements here, only that they are in the correct sets.

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -384,7 +384,7 @@ def OfflineGangaDiracSplitter(_inputs, filesPerJob, maxFiles, ignoremissing):
         # NB: Can't modify this list and iterate over it directly in python
         LFN_instances = site_dict.keys()
         # Already used LFN
-        chosen_lfns = []
+        chosen_lfns = set([])
```
milliams: This can just be
```python
chosen_lfns = set()
```

drmarkwslater:
@milliams Of course it can 😄

milliams:
It's not the string comparisons that are slow, it's the fact that for every comparison it has to check against every item in the list. A container like a set will used binned hashes to do the search in approximately `ln(N)`-time. For 25000 items this approx 2500 times faster.

milliams:
:+1:

rob-c:
OK, well either way, learned something new as a result.

PR #564, opened 2016-06-07
Set the default value rather than the session value for the LHCb cache

milliams:
For consistency (and to prevent it being reset between test runs), set the LHCb env cache as the default value rather than the session value.

I think this should fix the nightly tests.

### DIFF
--- a/python/GangaLHCb/__init__.py
+++ b/python/GangaLHCb/__init__.py
@@ -101,8 +101,6 @@ def _store_dirac_environment():
 
 if not _after_bootstrap:
     _store_dirac_environment()
-    configDirac.setSessionValue('DiracEnvJSON', os.environ['GANGADIRACENVIRONMENT'])
-
     _store_root_version()
 
 
@@ -134,6 +132,10 @@ def loadPlugins(config=None):
     logger.debug("Importing LHCbTasks")
     import Lib.Tasks
     logger.debug("Finished Importing")
+
+
+def postBootstrapHook():
+    configDirac.setSessionValue('DiracEnvJSON', os.environ['GANGADIRACENVIRONMENT'])
 
 from Ganga.GPIDev.Lib.File.Configure import outputconfig
 


rob-c:
I take it this is an oddity from running with GangaUnitTest and the shutdown code still not fully cleaning up after itself?

milliams:
In a way, yes. It's due to the shutdown code resetting the config system back to 'defaults'. For now I think that this is the best of worlds given the work that wuld be required to separate out the config system into a independently loadable plugin.

rob-c:
I think this would be more transparent if the config system was properly cleaned up by the shutdown code between runs. It avoids anything else we thing we haven't missed that may show up in the future.

milliams:
It would indeed be better but, given the complexity of the bootstrap process, there is currently no way to do so. Too many things are done at import time rather than via explicit hooks and resetting the state of a module is not an easy game to play. It's certainly a noble goal but for now this is just as correct.

rob-c:
I would have thought this is as simple as destroying the config when ganga has shutdown and remove all global references to it. Then starting up and recreating everything from scratch every test?
Surely this just avoids problems like this which will more than likely keep showing up.
As it stands, no complaints for the PR as it works but I'll open an issue because doing this properly gives me faith that we're correctly wiping the config between tests rather than relying on some extra logic.

milliams:
I've changed how this is done, check the latest diff above. It's now using the `postBootstrapHook()` function that plugins can implement. This will run after every run of bootstrap and so will correctly set the config value. This is more correct and will be safer with future changes.

Comments?

rob-c:
I still think this shouldn't be the responsibility of the plugin to fix the tests not working with the config system if the config worked normally.
If this is a working fix for the tests then go from me.

It makes me uncomfortable that we have to change the way the config is setup so that the tests work with it. I feel this won't be the last config parameter which has problems like this.

milliams:
It probably will be one of the last. The reason this parameter was a problem is that it was only being run on import time. As I've been arguing for years now, we should not be changing the state of the program at import if we can possible avoid it.

This isn't "the plugin fixing the tests", this is "the plugin was doing something slightly silly and is now doing it in a more sensible way". We're not changing how the config is "set up", we're just changing at which point a global config value is set.

Your idea about tearing down the config at shutdown would not help this bug as the code which sets the value would still only be run once per interpreter session.

rob-c:
The config is always imported with the plugin I thought or is there a plan to change this?
It doesn't matter how a config parameter is extracted surely regardless of if it's extracted via a function or set via a string. If it's set it then it should be set this way.
If it's being reset in the tests this is a problem caused by bad behaviour in the tests, there's no escaping that imo.

Issue #566, opened 2016-06-08
LHCb proxy tool no longer is able to read disk cache


rob-c:
LHCb `ganga-proxy.py` needs to be updated to read the now JSON cache on disk in order to be able to make a dirac proxy when we change to a new version of LHCbDIRAC.

This needs to be done before 6.1.21

This should have showed up in testing. Are we making sure we're removing all of the on-disk caches on the Jenkins server between nightlies does anyone know?

milliams:
I did update the `ganga-proxy.py` tool when I made the cache change. Perhaps I missed something though. What error are you seeing?

rob-c:
My window only showed the code for reading in the old file format (which I think is no-longer needed?)

I wasn't expecting to see the new file format then read afterwards and to overwrite what is already potentially in memory.

This only effects me as I have both types of configs for the same LHCbDirac version lying around on disk. Users won't see this when we bump the version for 6.1.21.

I propose removing the old code for reading the old file format. It's no longer required.

PR #567, opened 2016-06-08
Cleanup proxy code

rob-c:
Just a minor code cleanup addressing #566 

Fixing #566

### DIFF
--- a/python/GangaLHCb/ganga-proxy.py
+++ b/python/GangaLHCb/ganga-proxy.py
@@ -28,31 +28,11 @@ except:
 
 global_env = None
 
-
-def get_env():
-    global global_env
-    if global_env is None:
-        global_env = {}
-        for k, v in os.environ.iteritems():
-            if not str(v).startswith('() {'):
-                global_env[k] = v
-            else:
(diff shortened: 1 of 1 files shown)


python/GangaLHCb/ganga-proxy.py
```diff
@@ -28,31 +28,11 @@
...
-                    final_str += str(os.path.expandvars(line)).strip()
-                    if not final_str.endswith(';'):
-                        final_str += " ;"
-                    final_str += " "
-                global_env[k] = final_str
-
-    return global_env
-
-if "GANGADIRACENVIRONMENT" not in get_env():
+temp_env = os.environ()
```
drmarkwslater: Isn't `os.environ` just a dictionary? I don't think the brackets are needed and the `temp_env` overall seems a bit excessive as you can just use `os.environ` instead.

rob-c: Fixed :)

drmarkwslater:
Assuming all the other env stuff is correct and this isn't needed anymore then 👍

milliams:
This looks good to me :+1:

PR #571, opened 2016-06-09
Allow execute pickle stream to be empty

milliams:
The recent `execute()` cleanups introduced a bug whereby a call like:
```python
execute('print("foo")', shell=False)
```
would not return `"foo"` but instead return the string:
```python
"""
  File "<stdin>", line 19, in <module>
  File "<string>", line 4, in <module>
SystemExit: 0
"""
```
This was caused by a combination of two things. Firstly the explicit `sys.exit(0)` that was added to the run command, causing an exception to be raised on exit (and therefore caught and passed to the stream) and secondly the assumption that a call with `shell=False` *must* have some data passed though the pickle pipe.

When running a Python command (i.e. `shell=False`) the two ways that data get into the pickle stream are: 1. raising an exception and 2. calling the `output()` function. If nether of those happen then the pickle stream will be empty and so the stdout of the command should be returned instead.


python/Ganga/Utility/execute.py
```diff
@@ -230,10 +228,10 @@ def execute(command,
 
     if update_env:
         env_output_key = 'env_output'
-        update_env_thread = update_thread(env_file_pipes, thread_output, env_output_key)
+        update_env_thread = update_thread(env_file_pipes, thread_output, env_output_key, require_output=True)
     if not shell:
         pkl_output_key = 'pkl_output'
-        update_pkl_thread = update_thread(pkl_file_pipes, thread_output, pkl_output_key)
+        update_pkl_thread = update_thread(pkl_file_pipes, thread_output, pkl_output_key, require_output=False)
```
rob-c: If we don't require output via pickle why are we running this collection thread in this case?
OK, we may want to run it to stop lockups if a buffer gets full but should we check once it's finished to see if it's empty as expected which is a potential bug?

python/Ganga/Utility/execute.py
```diff
@@ -80,24 +78,24 @@ def output(data):
...
     """
     os.close(pipes[1])
     with os.fdopen(pipes[0], 'rb') as read_file:
         try:
             output_ns.update({output_var: pickle.load(read_file)})
         except Exception as err:
-            logger.error("Err: %s" % str(err))
-            raise  # EOFError triggered if command killed with timeout
+            if require_output:
+                logger.error('Error getting output stream from command: %s', err)
```
rob-c: Why aren't we raising here?

python/Ganga/Utility/execute.py
milliams: Just becasue we don't *require* output into the pickle stream in this case, doesn't mean that we don't *allow* it. The point is that if `require_output` is `True` and none is found then an error is printed. The converse if not the case, `require_output` being `False` and it finding output is absolutely fine and often expected.

python/Ganga/Utility/execute.py
milliams: Because this is run in a separate thread and so the raised exception will go nowhere and be picked up be no-one.

python/Ganga/Utility/execute.py
rob-c: I'm confused now. If you expect some output you should ask for it. What is the situation where there is output through this stream where it's not been explicitly asked for? This seems like a strange mis-use of the API to me.

python/Ganga/Utility/execute.py
rob-c: Ah good point 👍

python/Ganga/Utility/execute.py
milliams: If `shell=False` then we must always *allow* for some pickled output to be sent through the stream (either from `output()` or from a caught exception). We expect that there *might* be some but also there *might not*. If there *is* then we unpickle it and end up returning it from `execute()`. We don't *require* that there is some output.

We can't, from within `execute()` know, a priori, which `command`s will be pushing though the pickle stream and which will simply be printing it through stdout.

rob-c: OK, in the case we expect output we explicitly check that we get some.

I assumed the case of `require_output=True` was the output is allowed but not needed and that `require_output=False` was that no output is expected.
I'm stuck at thinking of a better parameter name however...

Are there any situations where we are currently getting errors due to not expecting any feedback via the streams if we're executing a python command?

milliams: The example I gave in the PR description of `execute('print("foo")', shell=False)` will currently fail (as seen in the DIRAC tests PR).

rob-c: I wouldn't have thought this a particularly valid test in itself. Surely a better test is to return the string "foo" via the output command?

milliams: The test is one that has existed for three years now and has been passing all that time. It's therefore an API that is expected and `execute()` was written with this use case in mind.

rob-c: OK. Fixing a long-standing test to work again is a good motivation although I wonder if we want to decide if some like this are still valid

rob-c:
Go for merge from me when ready

PR #573, opened 2016-06-10
Reduce the difference between test run setups

milliams:
Make the code that is run to set up Ganga on each test run as similar as possible. This fixes the config being reset so aggressively between tests and moves the code to being more similar to a standard run of the `ganga` application.

### DIFF
--- a/python/Ganga/testlib/GangaUnitTest.py
+++ b/python/Ganga/testlib/GangaUnitTest.py
@@ -8,11 +8,13 @@ try:
 except ImportError:
     import unittest
 
+
 def _getGangaPath():
     file_path = os.path.dirname(os.path.realpath(__file__))
     ganga_python_dir = os.path.join(file_path, '..', '..')
     ganga_python_dir = os.path.realpath(ganga_python_dir)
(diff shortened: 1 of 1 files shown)


python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -62,49 +65,32 @@ def start_ganga(gangadir_for_test, extra_opts=[]):
...
-            # Start internal services
-            logger.info("InternalServices restarting")
-
-            from Ganga.Core.InternalServices.Coordinator import enableInternalServices
-            enableInternalServices()
-        else:
-            logger.info("InternalServices still running")
+    # Perform the configuration and bootstrap steps in ganga
+    logger.info("Parsing Configuration Options")
+    Ganga.Runtime._prog.configure()
```
rob-c: Is this now safe to call multiple times, if so fantastic 👍

milliams: It certainly seems to be. The only problem with it seemed to be needing to reset the config handlers which is now being done in the test teardown.

PR #575, opened 2016-06-10
Cleanup Config.py

milliams:
This removes a bunch of unused and unnecessary code from Config as a first step towards making it more easily loadable and un-loadable.

The biggest change in here is that `getConfig()` will no longer *create* a config section if it doesn't find one. This is possible since we are now explicit about using `makeConfig()` in our `__init__.py` files.

### DIFF
--- a/python/Ganga/Utility/Config/Config.py
+++ b/python/Ganga/Utility/Config/Config.py
@@ -115,20 +115,20 @@ class ConfigError(GangaException):
 # for example, do not do here: from Ganga.Utility.logging import logger
 # use getLogger() function defined below:
 
-logger = None
+_logger = None
 
 
 def getLogger():
     import Ganga.Utility.logging
-    global logger
-    if logger is not None:
-        return logger
+    global _logger
+    if _logger is not None:
+        return _logger
 
     # for the configuration of the logging package itself (in the initial
     # phases) the logging may be disabled
     try:
-        logger = Ganga.Utility.logging.getLogger()
-        return logger
+        _logger = Ganga.Utility.logging.getLogger()
+        return _logger
     except AttributeError as err:
         print("AttributeError: %s" % err)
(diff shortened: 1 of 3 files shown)


rob-c:
Can we bump this to 6.1.22? If we're aiming for a first release Monday then this seems like it could break a plugin potentially

python/Ganga/Utility/Config/Config.py
```diff
@@ -932,11 +877,8 @@ def setSessionValue(config_name, option_name, value):
 
     # put value in the buffer, it will be removed from the buffer when option
     # is added
-    allConfigFileValues.setdefault(config_name, {})
-    allConfigFileValues[config_name][option_name] = value
-
-
-_configured = False
+    unknownConfigFileValues.setdefault(config_name, {})
```
alexanderrichards: Worth using a defaultdict?

milliams: I'd missed that. Yes, it probably is worth it.

mesmith75:
Do we want this in the next release - it looks like there are conflicts to be resolved..

milliams:
I have one or two small tweaks to make to this so I will defer it.

rob-c:
Is this PR best against 6.2.1 or 6.3.0?

milliams:
I'm not sure whether it would count as a feature or a bugfix. It doesn't have any externally visible changes so is not a feature but at the same time it is not fixing any known bug. It should be safe and on the whole has only been removing unused code so I don't foresee any problems. I would err on the side of 6.2.1 but I'm happy to have it deferred.

rob-c:
@mesmith75 I think I'm happy to make this your call, should we include this in 6.2.1 or 6.3.0?

mesmith75:
Let's put this into 6.2.1.

mesmith75:
If there are no objections I'll merge this.

PR #576, opened 2016-06-10
Remove OutputData from LHCb

mesmith75:
I have removed the file with the OutputData class definition and the imports from the other files in the LHCb directory.


alexanderrichards:
add to whitelist

milliams:
test this please

alexanderrichards:
beat you to it this time @milliams :wink:

alexanderrichards:
mark this branch appears to be out of date. Can you merge/rebase it? Alternatively we could try pressing the update branch button here at the bottom of the PR...do we trust that yet @milliams :wink: ..scary automatic stuff

milliams:
This looks good to me. As long as there's no schema compatibility problems :+1:

@alexanderrichards The update button is relatively safe. It does a merge from `develop` into this. The merge commit it creates will usually be fast-forwarded away when it is merged into `develop` anyway. If there are any conflicts then it will not let you do it and will require manual intervention.

mesmith75:
I think that has worked.

alexanderrichards:
Only problem then I foresee is if users already have one of these objects in their repo or box.
However I believe that these objects haven't been used in LHCb workflows for some time right @rob-c? In which case we should be ok.

Go for merge from me :+1:

PR #577, opened 2016-06-10
Make global_env a dict

milliams:
#567 set `global_env` to be `None` rather than a `dict`. This makes sure that the object is item-assignable.


rob-c:
OK, feel free to merge ahead of #357 as I put it in there but hadn't factored it out yet

alexanderrichards:
:+1:

PR #578, opened 2016-06-10
CMake support

rob-c:
OK this is the first PR for CMake support.

~~I'm disabling CMake Dirac support for 6.1.20 so won't fix bugs for that here.~~ Edit (see below)
There is a short LHCb test incoming to simply create the new modules but this isn't in the PR yet.
(Also some minor docs changes/enhancements)
(and doesn't gain much as can only check the API/schema is working which is the same as everywhere else in Ganga).

Hopefully I'd like to get this merged Monday morning so the main outstanding features for 6.1.21 are in for testing.


python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
+
+from Ganga.GPIDev.Adapters.IPrepareApp import IPrepareApp
+from Ganga.GPIDev.Adapters.IRuntimeHandler import IRuntimeHandler
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+from Ganga.GPIDev.Schema import Schema, Version, SimpleItem, GangaFileItem
+
+from Ganga.Utility.Config import getConfig
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+from Ganga.GPIDev.Lib.File.File import File, ShareDir
```
milliams: Unused import `File`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+from Ganga.Core import ApplicationConfigurationError, ApplicationPrepareError, GangaException
+
+from Ganga.Utility.logging import getLogger
+
+from Ganga.GPIDev.Base.Proxy import getName, isType, stripProxy
```
milliams: Unused import `isType` and `stripProxy`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+from os import rename, path
+import shutil
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+import tempfile
+import time
+import subprocess
+
+logger = getLogger()
+cmake_sandbox_name = 'cmake-input-sandbox.tgz'
+build_target = 'ganga-input-sandbox'
+build_dest = 'input-sandbox.tgz'
+defDir = '$HOME/cmtuser'
```
milliams: There's no need for these to be module-globals. They'd be better off as class attributes to make their scope clearer.

`cmake_sandbox_name` is only used once on the whole module and needn't be a variable at all.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+def _exec_cmd(cmd, cwdir):
```
milliams: PEP 8 asks for two empty lines before module-level functions and classes.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+        time.sleep(0.5)
+    return pipe.returncode, stdout, stderr
+
+def prepare_cmake_app(myApp, myVer, myPath=defDir, myGetpack=None):
+    """
+    Short helper function for setting up minimal application environments on disk for job submission
+    """
+    full_path = expandfilename(myPath)
+    if not path.exists(full_path):
+        os.makedirs(full_path)
```
milliams: You haven't imported `os`. How did this pass your tests?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+    os.chdir(full_path)
+    _exec_cmd('lb-dev %s %s' % (myApp, myVer), myPath)
+    if myGetpack:
+        os.chdir(path.join(full_path, myApp + 'Dev_' + myVer))
+        _exec_cmd('getpack %s' % myGetpack)
```
milliams: `_exec_cmd` takes two arguments.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+            self.is_prepared = None
+        self.hash = None
+        self.uploadedInput = None
+
+    def prepare(self, force=False):
+        """
+        This method creates a set of prepared files for the application to pass to the RTHandler
+        """
+
+        if (self.is_prepared is not None) and (force is not True):
```
milliams: Could just be:
```python
if (self.is_prepared is not None) and not force:
```

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # lets use the same criteria as the configure() method for checking file existence & sanity
+        # this will bail us out of prepare if there's somthing odd with the job config - like the executable
+        # file is unspecified, has a space or is a relative path
+        self.configure(self)
+        logger.info('Preparing %s application.' % getName(self))
+        setattr(self, 'is_prepared', ShareDir())
+        logger.info('Created shared directory: %s' % (self.is_prepared.name))
+
+        build_target = self.buildGangaTarget()
```
milliams: This name shadows the global name `build_target`.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+        if (self.is_prepared is not None) and (force is not True):
+            raise ApplicationPrepareError('%s application has already been prepared. Use prepare(force=True) to prepare again.' % getName(self))
+
+        # lets use the same criteria as the configure() method for checking file existence & sanity
+        # this will bail us out of prepare if there's somthing odd with the job config - like the executable
+        # file is unspecified, has a space or is a relative path
+        self.configure(self)
+        logger.info('Preparing %s application.' % getName(self))
+        setattr(self, 'is_prepared', ShareDir())
```
milliams: Just:
```python
self.is_prepared = ShareDir()
```

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+        return 1
+
+    def configure(self, masterappconfig):
+        """
+        Required even though nothing is done in this step for this App
+        """
+        # Lets test the inputs
+        opt_file = self.getOptsFile()
+        dir_name = self.getDir()
+        return (None, None)
```
milliams: No need for these parentheses.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def getOptsFile(self):
+        """
+        This function rweturns a sanitized absolute path to the self.myOpts file from user input
+        """
+        if self.myOpts:
+            if isinstance(self.myOpts, LocalFile):
+                ## FIXME LocalFile should return the basename and folder in 2 attibutes so we can piece it together, now it doesn't
```
milliams: Single `#` to start a comment.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+
+            self.copyIntoPrepDir(build_target)
+            opts_file = self.getOptsFile()
+            if isinstance(opts_file, LocalFile):
+                self.copyIntoPrepDir(path.join( opts_file.localDir, path.basename(opts_file.namePattern) ))
+            elif isinstance(opts_file, DiracFile) and self.master and not isinstance(self.getJobObject().backend, DiracBase):
+                # NB safe to put it here as should have expressly setup a path for this job by now
+                opts_file.get(localPath=self.getSharedPath())
+            else:
+                raise ApplicationConfigurationError("Opts file type %s not yet supported please contact Ganga devs if you require this support" % getName(opts_file))
```
milliams: `ApplicationConfigurationError` takes two arguments.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,248 @@
...
+        if not path.isfile(targetFile):
+            raise GangaException("Target File: %s NOT found!" % targetFile)
+        wantedTargetFile = path.join(targetPath, cmake_sandbox_name)
+        rename(targetFile, wantedTargetFile)
+        if not path.isfile(wantedTargetFile):
+            raise GangaException("Wanted Target File: %s NOT found" % wantedTargetFile)
+
+        logger.info("Built %s" % wantedTargetFile)
+        return wantedTargetFile
+
```
milliams: Only one empty line at end of file.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
+##########################################################################
+# Ganga Project. http://cern.ch/ganga
+#
+# $Id: Executable.py,v 1.1 2008-07-17 16:40:57 moscicki Exp $
+##########################################################################
```
milliams: No need for this comment at the top.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+
+from Ganga.GPIDev.Adapters.IPrepareApp import IPrepareApp
+from Ganga.GPIDev.Adapters.IRuntimeHandler import IRuntimeHandler
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+from Ganga.GPIDev.Schema import Schema, Version, SimpleItem
+
+from Ganga.Utility.Config import getConfig
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+
+from Ganga.GPIDev.Lib.File.File import File, ShareDir
```
milliams: Unused import `File`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
+##########################################################################
+# Ganga Project. http://cern.ch/ganga
+#
+# $Id: Executable.py,v 1.1 2008-07-17 16:40:57 moscicki Exp $
+##########################################################################
+
+from Ganga.GPIDev.Adapters.IPrepareApp import IPrepareApp
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+from Ganga.GPIDev.Adapters.IRuntimeHandler import IRuntimeHandler
+from Ganga.GPIDev.Schema import Schema, Version, SimpleItem
```
milliams: Unused import

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+
+from Ganga.Utility.Config import getConfig
+
+from Ganga.GPIDev.Lib.File.File import File, ShareDir
+from Ganga.GPIDev.Lib.File.FileBuffer import FileBuffer
+from Ganga.Core import ApplicationConfigurationError, ApplicationPrepareError
```
milliams: Unused import `ApplicationPrepareError`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+import random
+import threading
+from Ganga.Utility.files import expandfilename
+
+from Ganga.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
+
+from GangaGaudi.Lib.RTHandlers.RunTimeHandlerUtils import get_share_path, master_sandbox_prepare, sandbox_prepare, script_generator
+from GangaDirac.Lib.RTHandlers.DiracRTHUtils import dirac_inputdata, dirac_ouputdata, mangle_job_name, diracAPI_script_template, diracAPI_script_settings, API_nullifier, dirac_outputfile_jdl
+from GangaDirac.Lib.Files.DiracFile import DiracFile
+from Ganga.GPIDev.Lib.File.OutputFileManager import getOutputSandboxPatterns, getWNCodeForOutputPostprocessing
```
milliams: Unused import `get_share_path`, `dirac_ouputdata`, `DiracFile`, `getOutputSandboxPatterns`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+from Ganga.Utility.Config import getConfig
+
+from Ganga.GPIDev.Lib.File.LocalFile import LocalFile
+from GangaDirac.Lib.Files.DiracFile import DiracFile
+
+logger = getLogger()
+
+data_file = 'data.py'
+
+prep_lock = threading.Lock()
```
milliams: I'd like toa void these being module-global. Do we really need a process-wide lock on preparing?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+    Generating a data.py file which contains the data we want gaudirun to use
+    Args:
+        job (Job): This is the job object which contains everything useful for generating the code
+    """
+    logger.debug("Doing XML Catalog stuff")
+
+    inputsandbox = []
+
+    data = job.inputdata
+    data_str = ''
```
milliams: Unused variable.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+    Args:
+        app (GaudiRun): This expects only the GaudiRun app
+    """
+    opts_file = app.getOptsFile()
+    if isinstance(opts_file, (LocalFile, DiracFile)):
+        # Ideally this would NOT need the basename, however LocalFile is special in this regard.
+        # TODO Fix this after fixing LocalFile
+        opts_name = os.path.basename(opts_file.namePattern)
+    else:
+        raise ApplicationConfigurationError("The filetype: %s is not yet supported for use as an opts file.\nPlease contact the Ganga devs is you wish this implemented." %
```
milliams: `ApplicationConfigurationError` takes two arguments.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+allHandlers.add('GaudiRun', 'Interactive', GaudiRunRTHandler)
+allHandlers.add('GaudiRun', 'Batch', GaudiRunRTHandler)
+
+def generateDiracInput(app):
+
+    input_files, input_folders = collectPreparedFiles(app)
+
+    job = app.getJobObject()
+
+    compressed_file = None
```
milliams: Unused variable

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+    if input_folders:
+        raise ApplicationConfigurationError('Prepared folders not supported yet, please fix this in future')
+    else:
+        prep_dir = app.getSharedPath()
+        add_timeStampFile(prep_dir)
+        prep_file = prep_dir + '.tgz'
```
milliams: Unused variable `prep_file`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+        compressed_file = os.path.join(tempfile.gettempdir(), os.path.basename(prep_dir) + ".tgz")
+
+        wnScript = generateWNScript(prepareCommand(app), job)
+        script_name = os.path.join(tempfile.gettempdir(), wnScript.name)
+        wnScript.create(script_name)
+
+        with tarfile.open( compressed_file + ".tgz", "w:gz" ) as tar_file:
```
milliams: `tar_file` is unused.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+            for name in input_files:
+                tar.add(name, arcname=os.path.basename(name))
```
milliams: `tar` is not defined.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,346 @@
...
+            tar.add(script_name, arcname=os.path.basename(script_name))
+        shutil.move(compressed_file, prep_dir)
+
+    new_df = DiracFile(namePattern = os.path.basename(compressed_file), localDir = app.getSharedPath(),
+                       remoteDir = os.path.join(DiracFile.diracLFNBase(), 'GangaInputFile/Job_%s' % job.fqid) )
+    random_SE = random.choice(getConfig('DIRAC')['allDiracSE'])
+    new_df.put(uploadSE = randomSE)
```
milliams: Typo in `randomSE`/`random_SE`.

milliams:
There are a large number of `TypeError`s, `NameError`s, typos etc in this PR. Has this been tested at all, even locally at the prompt?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Not being tested in this PR. ^^

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Not being tested/used in this PR ^^

rob-c:
`I'm disabling CMake Dirac support for 6.1.20 so won't fix bugs for that here.`

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Should also be Prepared Error but this isn't handled correctly because we're too specific higher up.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Taken from elsewhere so possibly

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Yes if this is to enable multi-threaded submit. This is currently due to a bug in Job which hasn't been fixed yet.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Unused anymore as it stands

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Should be used but can't be due to problems elsewhere

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Does this break anything?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: This comment also requires attention. I'd prefer to worry more about the FIXME than the formatting

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Copied from elsewhere. Should be fixed in Executable as well. Lots of bad code around the `is_prepared` which we've inherited

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: yes, and the global is shifting in a further commit for other reasons

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Copied from elsewhere. Needs fixing in executable if we're going to fix here

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: This also isn't being exposed yet so isn't tested

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: does this break anything?

rob-c: have we even decided that PEP8 is being impacted on the codebase, in which case we should re auto-pep the codebase. The first time it was 'auto-formatted' was not against any particular standard as far as I read, it was more of a general set of rules common across different standards.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: What is broken by this?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: https://github.com/ganga-devs/ganga/pull/578/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R197
https://github.com/ganga-devs/ganga/pull/578/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R209
This is being used isn't it?

rob-c:
Can someone take a look at how/why Jenkins failed here. I'm reluctant to just upload more code and proceed as I've seen errors like this before and I'd prefer it is checked. Thanks, Rob.

rob-c:
I have DIRAC support working in a further commit but I'll wait until I know that Jenkins is working as expected.

milliams:
Jenkins sometimes fails to connect to GitHub. I don't know whether this is a Jenkins thing or a GitHub thing but it's usually just ephemeral. I suggest just carrying on as normal.

rob-c:
@milliams Is it worth documenting these peculiarities or reporting bugs?
From my understanding it's not beyond Jenkins to try to resume from this when git fails due to a communication issue, it should retry to setup the test as the test shouldn't be marked as failed

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: https://github.com/ganga-devs/ganga/pull/578/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R63
https://github.com/ganga-devs/ganga/pull/578/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R71

rob-c:
OK, I got some free time over the weekend and have managed to get the DIRAC support for CMake into this release (this makes me a lot happier as it should largely 'just work').
I'm polishing a test now and will try testing out my helper function which will setup a vanilla-(ish) DaVinci environment on disk for the user to use to submit a test job.
Once I have the tests in I'm keen to merge this PR unless something major crops up?

egede:
@rob-c This is great news.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
milliams: Nothing is *broken* by it but it's line noise and is redundant.

rob-c:
@egede I was quite happy when the first grid job just came back with an nTuple. I'll be doing some more testing as the release takes shape but hopefully this would encourage some testers to jump on board :)

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+        ## Here if the is_prepared count reaches zero
+        self.uploadedInput = None
+
+    def prepare(self, force=False):
+        """
+        This method creates a set of prepared files for the application to pass to the RTHandler
+        """
+
+        if (self.is_prepared is not None) and (force is not True):
+            raise ApplicationPrepareError('%s application has already been prepared. Use prepare(force=True) to prepare again.' % getName(self))
```
milliams: You're missing an import for `ApplicationPrepareError`.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+        Required even though nothing is done in this step for this App
+        """
+        # Lets test the inputs
+        opt_file = self.getOptsFile()
+        dir_name = self.getDir()
+        return (None, None)
+
+    def getOptsFile(self):
+        """
+        This function rweturns a sanitized absolute path to the self.myOpts file from user input
```
milliams: Typo in "rweturns".

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+        This function returns a sanitized absolute path of the self.directory method from user input
+        """
+        if self.directory:
+            return path.abspath(expandfilename(self.directory))
+        else:
+            raise ApplicationConfigurationError(None, "No Opts File has been specified, please provide one!")
+
+    def exec_cmd(self, cmd):
+        """
+        This method exectutes a command within the namespace of the project. The cmd is placed in a bash script which is executed within the env
```
milliams: Typo in "exectutes".

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+        cmd_file = tempfile.NamedTemporaryFile(suffix='.sh', delete=False)
+
+        cmd_wrapper = './run bash -c "%s"' % cmd
+
+        cmd_file.write(cmd_wrapper)
+        cmd_file.flush()
+
+        script_run = 'bash %s' % cmd_file.name
+
+        rc, stdout, stderr = _exec_cmd(script_run, self.getDir())
```
milliams: This is going to end up running a bash shell which starts a bash shell which starts a bash shell which runs `cmd`. Surely we don't need that much nesting?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+            raise ApplicationConfigurationError(None, "No Opts File has been specified, please provide one!")
+
+    def exec_cmd(self, cmd):
+        """
+        This method exectutes a command within the namespace of the project. The cmd is placed in a bash script which is executed within the env
+        Args:
+            cmd (str): This is the command(s) which are to be executed within the project environment and directory
+        """
+
+        cmd_file = tempfile.NamedTemporaryFile(suffix='.sh', delete=False)
```
milliams: Do we need to explicitly delete this file?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+
+from GangaDirac.Lib.Backends.DiracBase import DiracBase
+
+import tempfile
+import time
+import subprocess
+
+logger = getLogger()
+
+def _exec_cmd(cmd, cwdir):
```
milliams: Do we want yet another way to run external commands? Why not `execute` or `Shell`? Also, this function will never time-out an overly-long process which can cause a hang.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+
+from Ganga.GPIDev.Lib.File.File import ShareDir
+from Ganga.Core import ApplicationConfigurationError, GangaException
+
+from Ganga.Utility.logging import getLogger
+
+from Ganga.GPIDev.Base.Proxy import getName
+
+from os import rename, path, makedirs, chdir
+import shutil
```
milliams: This import is not used.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
+
+from Ganga.GPIDev.Adapters.IPrepareApp import IPrepareApp
```
milliams: Empty line at top of file.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -0,0 +1,250 @@
...
+from Ganga.GPIDev.Schema import Schema, Version, SimpleItem, GangaFileItem
+
+from Ganga.GPIDev.Lib.File.File import ShareDir
+from Ganga.Core import ApplicationConfigurationError, GangaException
+
+from Ganga.Utility.logging import getLogger
+
+from Ganga.GPIDev.Base.Proxy import getName
+
+from os import rename, path, makedirs, chdir
```
milliams: It is convention to put all stdlib imports at the top of the file, followed by 3rd-party ones, followed by local imports.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
+
+from Ganga.GPIDev.Adapters.IRuntimeHandler import IRuntimeHandler
```
milliams: Empty line at top of file.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+from Ganga.GPIDev.Base.Proxy import getName
+
+import os
+import shutil
+from datetime import datetime
+import tempfile
+import tarfile
+import random
+import threading
+from Ganga.Utility.files import expandfilename
```
milliams: Unused import.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+    I also add 20 unique characters from the ascii and digits pool from SystemRandom which may reduce the risk of collisions between users
+    Args:
+        given_path (str): Path which we want to create the timestamp within
+    """
+    fmt = '%Y-%m-%d-%H-%M-%S'
+    time_filename = os.path.join(given_path, '__timestamp__')
+    logger.info("Constructing: %s" % time_filename)
+    with open(time_filename, 'a+') as time_file:
+        time_file.write(datetime.now().strftime(fmt))
+        time_file.write('\n'+''.join(random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(20)))
```
milliams: Don't add random characters. Use the `uuid` module.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
milliams: Those uses of `data_str` are creating a new variable, not using this one.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+    """
+    Generate the script as a file buffer and return it
+    Args:
+        commandline (str): This is the command-line argument the script is wrapping
+        job (Job): This is the job object which contains everything useful for generating the code
+    """
+    exe_script_name = 'gaudiRun-script.py'
+
+    return FileBuffer(name=exe_script_name, contents=script_generator(gaudiRun_script_template(), COMMAND=commandline,
+                                                                    OUTPUTFILESINJECTEDCODE = getWNCodeForOutputPostprocessing(job, '    ')),
```
milliams: `OUTPUTFILESINJECTEDCODE` needs two more spaces in front of it to line it up visually.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+        Args:
+            app (GaudiRun): This application is only expected to handle GaudiRun Applications here
+            appconfig (unknown): Output passed from the application configuration call
+            appmasterconfig (unknown): Output passed from the application master_configure call
+            jobmasterconfig (tuple): Output from the master job prepare step
+        """
+
+        job = app.getJobObject()
+
+        prepared_files = []
```
milliams: This variable is never set to anything and is only used as `[]`.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: I'm wanting to run all commands against the project within the `lb-run` defined environment for completeness. If I want to do something more complex, say 2-3 commands in succession then I need to construct a script (1 line or multi-line .sh) and run something so I opted to run it as a bash script and we need to make sure that the bash shell is used for a bash script as a lot of users don't have this as default.
tl;dr yes most likely.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
milliams: No, those uses are rebinding the name, not using the same variable.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: yes, I'll check this is done

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+        prep_dir = app.getSharedPath()
+        add_timeStampFile(prep_dir)
+        prep_file = prep_dir + '.tgz'
+        compressed_file = os.path.join(tempfile.gettempdir(), '__'+os.path.basename(prep_file))
+
+        wnScript = generateWNScript(prepareCommand(app), job)
+        script_name = os.path.join(tempfile.gettempdir(), wnScript.name)
+        wnScript.create(script_name)
+
+        with tarfile.open( compressed_file, "w:gz" ) as tar_file:
```
milliams: Extraneous spaces within brackets.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Yes for now, I'll review in the future changing this, I'm aware this is not ideal but for now I wanted full access to everything for testing/development unless you know a 1 line which I can replace this with?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Yes but we don't follow this across the codebase atm. It would be more efficient to try and do this automatically.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+            for name in input_files:
+                tar_file.add(name, arcname=os.path.basename(name))
+            tar_file.add(script_name, arcname=os.path.basename(script_name))
+        shutil.move(compressed_file, prep_dir)
+
+    new_df = DiracFile(namePattern = os.path.basename(compressed_file), localDir = app.getSharedPath())
```
milliams: Extraneous spaces around `=`.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,354 @@
...
+    random_SE = random.choice(getConfig('DIRAC')['allDiracSE'])
+    logger.info("new File: %s" % new_df)
+    new_lfn = os.path.join(DiracFile.diracLFNBase(), 'GangaInputFile/Job_%s' % job.fqid, os.path.basename(compressed_file))
+    new_df.put(uploadSE = random_SE, lfn=new_lfn)
```
milliams: Extraneous spaces around `=`.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
milliams: I guess I feel that when adding a new file we should do it right from the start. Saying "I'll do however I feel and fix it all later automatically" just causes work later.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: uuid is machine dependent, collisions can happen if 2 users are on the same machine, SystemRandom uses the uuid behind the scenes doesn't it?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
milliams: Well at the very least you can get rid of the `bash` on the `script_run = ...` line and just set `executable` to `/bin/bash` in the call to `subprocess.Popen()`. Starting `bash` three times for every call isn't cheap.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
milliams: `uuid` does not depend on the machine. It is for generating unique strings for exactly times like this.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Yes but there is an extremely _large_ need to do this in an automated way across a lot (everywhere else) in the codebase.
Doing this in 2 files will help the PR but I don't want to have to do this every time I open a file to fix a typo whilst I'm doing other work and I'm unaware of any large PR to fix this.

There is a lack of compliance because we haven't set a standard. Can we please add something to the dev docs to describe how we should work with the code. Preferably with some command to highlight all of these things in a useful way for the dev so that these things can be worked out without a lot of to and fro and a lot of emails? I'm happy to fix things but a back and forth on github is less than ideal in the long-term I think.
Changes incoming.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: This discussion is mixing up `uuid1` and `uuid4`. https://docs.python.org/2/library/uuid.html Although in a lot of situations `uuid1` is the norm. I'll replace this with `uuid4` although it's not much different to just having 20 random chars selected from the same pool that the `uuid`  encompasses.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Frankly calling this command isn't ever expected to be cheap. It's normal use is going to be in constructing a large project on an afs based filesystem so will take minutes to execute

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: added an unlink.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: How? This looks perfectly readable to me and is used in many places in current codebase it makes it clear that there is a tuple type being returned.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Given this is no longer used outside of `if data:` I'll just remove this for now

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
rob-c: Given the raise I'll just remove this for now. Otherwise there was an if/else calling for the same object which wasn't defined in both cases

rob-c:
OK, I'm going to fire off some test jobs to be sure nothing has broken between yesterday evening and now and then I'll look to merge at some point

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -0,0 +1,345 @@
...
+from Ganga.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
+from Ganga.GPIDev.Adapters.IRuntimeHandler import IRuntimeHandler
+from Ganga.GPIDev.Adapters.StandardJobConfig import StandardJobConfig
+from Ganga.GPIDev.Base.Proxy import getName
+from Ganga.GPIDev.Lib.File.File import File, ShareDir
+from Ganga.GPIDev.Lib.File.FileBuffer import FileBuffer
+from Ganga.GPIDev.Lib.File.LocalFile import LocalFile
+from Ganga.GPIDev.Lib.File.OutputFileManager import getWNCodeForOutputPostprocessing
+from Ganga.Utility.Config import getConfig
+from Ganga.Utility.files import unique
```
milliams: I think this should be `from Ganga.Utility.util import unique`? I'm surprised this didn't show up in testing though.

rob-c: Already fixed an hour ago, fixing another bug atm

rob-c:
OK, I'm seeing my local jobs now failing which surprises me, I didn't really touch the code there between making the PR and now.
I'm making some quick fixes but the app may go out without the ability to run with inputfiles but this isn't the end of the world for testing I hope.

rob-c:
OK, that recent commit should have fixed it. The local and DIRAC backends are tested and working with at least one LocalFile and DiracFile on each.
I'll try with a wider range of tests including a LocalFile on the DIRAC backend and a MassStorageFile on the Batch backend later this-evening and see if things 'just-work'.

I'm happy to merge now for wider testing if everything is passing but I suspect there will be bugs against this so we may want to wait until first thing tomorrow morning.

rob-c:
OK, I appear to have jobs running successfully on Localhost, LSF and Dirac for this so I think it's in a good state to merge for further testing/release.
Unless anyone has any further comments I plan to merge at about 10:30

PR #580, opened 2016-06-14
Enhancement/cleanup job

rob-c:
OK, this PR cleans up some code in the Job.py

I've added a new method which loops over the subjob statuses which avoids the bottleneck which was accidentally introduced into the code. (This explodes exponentially when you have ~100,000 jobs which is useful to find bottlenecks like this)
This now scales in a much, much flatter way.

If this passes the tests and nobody objects I'd like to see this go into 6.1.21 as it's a nice little performance fix which impacts the fact calls to `updateJobStatus` were becoming extremely slow for no good reason.

I don't object to getting this in for 6.1.22 if people would rather wait and get 6.1.21 out asap :)


python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -737,27 +738,30 @@ def outputFilesFailures(self):
 
         return postprocessFailure
 
+    def getSubJobStatuses(self):
+        """
+        This returns a set of all of the different subjob statuses whilst respecting lazy loading
+        """
+        stats = set()
```
milliams: I would define this once per branch or once in total. Probably:
```python
if isinstance(self.subjobs, SubJobXMLList):
    stats = set(self.subjobs.getAllSJStatus())
else:
    stats = set(sj.status for sj in self.subjobs)

return stats
```

rob-c:
OK, after profiling this there were a lot of calls in the monitoring to `updateJobStatus` which could potentially be reduced, but rather than track down and make sure all of these are safe I decided to be more intelligent in the way that the subjob statuses are extracted. I've added new function to the `SubJobXMLList` to get the subjob statuses in a slightly more intelligent way rather than requiring a lot of CPU be burned for no reason.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: done :+1:

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -737,27 +738,28 @@ def outputFilesFailures(self):
...
 
+    def getSubJobStatuses(self):
+        """
+        This returns a set of all of the different subjob statuses whilst respecting lazy loading
+        """
+
+        if isinstance(self.subjobs, SubJobXMLList):
+            stats = set(self.subjobs.getAllSJStatus())
+        else:
+            stats = set([sj.status for sj in self.subjobs])
```
milliams: There's no need for the square brackets here. In the case of a function which takes one argument which is a sequence you can write the generator expression directly and omit the nested brackets. It avoids creating a temporary list as well.

PR #584, opened 2016-06-15
Moving back to dynamic _data_dict in GangaObject

rob-c:
This PR seeks to make the contents of the Ganga managed dictionary behave like a defult dictionary which generates schema attributes only when needed.

This is achieved through a few changes:
1) Dropping the `._data` property as this is not intended to be external
2) Constructing this as an empty dict in `GangaObject.__init__`
3) Modifying the `Descriptor.__get__` methods to construct and populate this dictionary on the fly.
4) Renaming `_data_dict` back to `_data`
5) Adding `_inMemory` to the GangaObject (we can re-name this I'm not fixated by using this name) It's wrapped behind the GangaObject._fullyLoadedFromDisk) method.

As the `Descriptor.__get__` method is now modifying an internal object to the GangaObject we now need to acquire a full const_lock on all `__get__` and `__set__` methods. This is quite heavy although the performance gains in the extreme cases are very large so for the case of some simple tasks being much heavier this PR wins out on performance overall.

As the `_internal_lock` is no longer safe or required it is removed.

In order to make sure that an object doesn't get into a state whereby a threading deadlock can occur due to calling a registry function in 1 thread which has a registry lock before it's required to be called here I have introduced some knowledge of if an object is fully memory resident through a boolean `_inMemory` in GangaObject.

A Ganga object is defined to be in memory upon being created and hence `_inMemory=True` and only in the case of `GangaRepository._make_empty_object_` is this then set to `False` before the object is made use of.
An object not in memory will attempt to load the object from disk using a call to a registry function which does this. Currently this is `getReadAccess` although this could be any function which triggers loading the object from disk.

As some care is needed in the order of loading an object from disk the `GangaRepositoryXML` class manages whether an object has been loaded into memory before performing some important changes to ensure the object is constructed in a sane way.

The loading of an object is done within the context of the `const_lock` and hence in 1 thread and it acquires and releases a Registy lock only when it's needed.
It's not possible to trigger loading of an object twice without some direct calls to the Registry which is only accessed within the GangaObject at this time.
If we change such that the Registry is called directly elsewhere a lot then we may want to consider combining a const lock from the registry with a const lock within the GangaObject.

In addition to all of this a minor change was put in, in the wrapping/handling of the rmrf calls in the GangaRepositoryXML which was throwing errors in testing due to being on an extremely slow lxplus node using AFS.

The spirit of the changes to `._data` are to generate objects only when they're needed some additional changes have been made to the GangaObject.
1) The `__eq__` method doesn't request all objects from the Schema as this populates the object in question and makes this call extremely expensive and bad as the object is known to be the same as the default entry stored in memory in the Schema.
2) The accept method makes use of the Schema code to get the string representation of the default objects cached in the Schema code. This offers a performance boost as well as not populating the entire object being serialized with default Schema entries when they're not needed.

This doesn't effect the Generated XML filed.
This doesn't cause the loading of the XML to be changed from what it currently is.

On top of these changes several static functions of the Descriptor have now had their extra arguments placed into a common tuple. This is to allow them to be passed to `makeGangaList` and then wrapped as a partial function. This makes the generation of lists a nice 1 line call in a method rather than a nested complex for loop inside some difficult to read code.

(I can separate out the changes to the static functions in the Descriptor and the wrapping of `rmrf` commands within the XML repo but this saves about ~20 lines of code from the PR and were regarded as being fairly innocuous.)

This PR has gone through a few iterations including:
1) An attempt at constructing a defaultdict which was custom to Ganga. This was abandoned as it obfuscated what was going on quite a lot.
2) Combining the Registry and GangaObject locks to keep the internal lock. This started to mix code from several classes and made it less clear what the scope of objects was.
3) Not changing the `__eq__` and `accept` methods. This didn't really give much of a performance improvement as these functions became very heavy and caused the full schema to be populated whenever they're called.


rob-c:
As this is a somewhat invasive change I'm wanting to leave this up for review until 6.1.22, however, this PR is working against all of the integration tests, can submit large complex DIRAC jobs and offers some benefits in terms of performance improvements.

I'm keen to merge this early in the 6.1.22 release cycle unless there are any comments?

milliams:
Why are you referring to `_data_dict` throughout this PR? The whole point of Python `@property` methods is that it's a seamless interface. I didn't introduce `_data_dict` to be "the new `_data`", it's there as an internal implementation detail behind the `_data` decorator so that we can perform actions when it is set. The *only* place `_data_dict` should be referenced is inside `def _data()` and to set its initial value in `__init__()`.

rob-c:
There is no external reference to ```._data``` outside of this class and nor should there be. I can put the property back in but I fail to see what benefit this offers as it's an internal attribute to GangaObject other than allowing for any old code to still reference ```._data``` which I strongly suspect it shouldn't do?

rob-c:
Would simply renaming the dictionary in this PR back to just `._dict` make sense if it's now guaranteed to be internal?

milliams:
Yes, reverting the name back to _data would be better. The reason for the `@property` was not for data-hiding reasons but to enforce that if `_data` was set to a new value, the entries within would all have their parent set correctly.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -710,7 +713,11 @@ def _safe_flush_xml(self, this_id):
                 for idn in os.listdir(os.path.dirname(fn)):
                     split_cache = getattr(obj, self.sub_split)
                     if idn.isdigit() and int(idn) >= len(split_cache):
-                        rmrf(os.path.join(os.path.dirname(fn), idn))
+                        try:
+                            rmrf(os.path.join(os.path.dirname(fn), idn))
+                        except Exception as err:
```
milliams: This doesn't need to be such a broad exception handler.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -710,7 +713,11 @@ def _safe_flush_xml(self, this_id):
...
+                            logger.warning("Couldn't remove a file: %s" % err)
+                            pass
```
milliams: There's no need for this `pass` statement.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: What are the expected exceptions then? I'll add the ones I can see but there are likely quite a few here

milliams:
So with this change we are now keeping track of whether an object is fully loaded in two places: 1) `Regsitry._fully_loaded` and `GangaObject._inMemory`? I'd really like to keep that down to just one to avoid the chance of them getting out of sync.

It also seems this PR is making a number of changes in addition to that you mentioned in the first comment above:

1) Introducing `_inMemory`
2) Removing `_internal_lock`

These each deserve discussion in themselves and should be independent changes. A PR should aim to cover one logical change so that it can be understood in isolation.

Also, you just renamed `_data`→`_data_dict`→`_dict` instead of back to `_data`.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -335,39 +323,65 @@ def __get__(self, obj, cls):
...
             return self._bind_method(obj, self._getter_name)()
 
         # First we want to try to get the information without prompting a load from disk
 
-        # ._data takes priority ALWAYS over ._index_cache
-        # This access should not cause the object to be loaded
-        obj_data = obj._data
-        if name in obj_data:
-            return obj_data[name]
+        obj_in_schema = name in cls._schema.datadict
```
milliams: This change is unnecessary. Given how descriptors work and how they are assigned in our case, we can never have a case where we are in this function and have `name` not be matched to a schema item.

rob-c: If this works fantastic I can remove this check in this case.

rob-c:
0) Tracking the loaded status of an object should never be out of sync and I would more encourage that there is some test or runtime assertion to this fact as I think these track different things as inMemory doesn't guarantee _inRegistry_ at some level as we can create objects on the fly.
All things being equal: 
I think we can probably remove `_fully_loaded` in this PR as well if this is required. (I'll do this after reviewing the changes to GangaObject).

1) There are some changes in the `__eq__` and `__deepcopy__` methods exploiting this and avoiding creating objects within these methods which breaks the point of the PR.

2) The accept method has been changed to not generate objects on the fly within this method which is again in keeping with the spirit of the PR to not generate an object unless it's needed.

3) The movement to `input_tuple` in the static methods within the Descriptor class are to allow these methods to be passed as functions to the `makeGangaList` function and to allow them to accept additional extra arguments.

4) ooops, sed mistake I'll fix the name of the dict, this will take a min.

rob-c:
The `_internal_lock` is only removed here because it can't be used safely. It was replaced with a `const_lock` and then everything worked. I think you'd probably see a performance hit without this PR if you made this change and so it would be a bad change to make completely in isolation.
I can put the function back in but it's not safely callable and then as a result of fixing deadlocks isn't used anywhere.

rob-c:
OK, I'll now look into removing `_fully_loaded` which will expand the size and scope of this PR somewhat. (I'll put it in as a separate PR and we can merge them if we choose)

milliams:
Part of the problem is that the PR description does not match the contents of the PR. In order to review the changes I need the context.

rob-c:
I'll summarise and update the contents yet again in this case.
This has gone through a few rounds of working on my machine, breaking on Jenkins, fixing and merging the results so I'll summarise it at the top.
(I'll drop the old description entirely as it's quite big)

rob-c:
A full somewhat pedagogical description of the changes are now in the description.

rob-c:
OK, playing with the flusher mechanism breaks some XML so I think I'll revert the last commit the PR once I've understood why this happens

drmarkwslater:
Coming at this a bit late due to being on holiday and I can only really comment on the Registry/Object changes as I have looked at that code recently but I have a few issues:

1) What is the reason for the changes in the locks? These have been shown to work for some time now and I would very much prefer not to change them without a good reason. Specifically:
  * Using a raw read lock and not the internal_lock doesn't lock the root object only the object itself
  * Adding the read_lock to the const_lock - what is the motivation here?

2) This is checking if an object is loaded in the getter but due to how the descriptor works, after checking if the attribute is present in the _data dict or the index cache that means it is, by definition **not** loaded and in fact, I've done some work and the getReadAccess call can simply be replaced with a call into the Registry to load the object. In addition, as this should be the only (not including the setter) place that the Object should call into the Registry, the object itself doesn't need to know if it's loaded or not.

3) This removes `@syncronised` from a few functions which I should be kept as locked.

rob-c:
1)
The removal of the internal lock is due to the fact that getting the read lock on an object leaves you vulnerable to the object changing underneath you or worse when the getter is responsible for 'construction on the fly' of objects internal to the GangaObject.
Getting the const lock externally is the only thing which I'd argue really makes sense as it prevents the object being changed while you're doing something to it which could change it.


2)
The main thing I'm trying to tackle in the getter is:
 
1) During the construction of an object the `_data` dict isn't populated by default (in this PR).
  This means performing `__init__` on an object is now much, _much_ faster.
 2) We then attempt to get some object from `_data`.
 3a) If the object is loaded then we need to generate it based on the default Schema value
 3b) If the object is __not__ loaded then we need to load it from disk and return the value.

In solving 3) if we don't know the current loaded state we can't correctly work out whether we should return the default value from the schema or the value from disk.

The reason there is a check around the `getReadLock` is due to a problem that calling this multiple times on a loaded job causes issues (strange lockups due to some of the Registry code triggering the full `_data` dictionary to become populated probably due to some flush to disk type event).
I _only_ want to load an object into memory here so anything that the Registry does aside from this is not expected.
This is a problem lower down than in the GangaObject but can be avoided by not making multiple calls to this method.
I could fix this but it would likely involve some invasive changes I was hoping to avoid for data integrity sake.
As this method is used trigger a load from disk you in theory should never need to call it multiple times. We could and probably should be calling `self._getRegistry()._load(someID)` here anyway to load the object which is _much_ clearer.


If it wasn't for the very large performance improvements this gets I wouldn't be pushing for it, there is a very large up-front cost to constructing things like a 100k+ DiracFile objects, and whilst we can prevent some situations where this may happen we won't likely get all of them.

3) Which functions are having the synchronised statements removed? This is likely a typo. I'm adding it in several places but don't remember removing any.

rob-c:
retest this please

rob-c:
OK a massive rewrite based upon #627 is best. Moving to a defaultdict helps improve the speed of constructing very large objects which users are using

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -239,11 +239,12 @@ def shutdown(self):
         from Ganga.Utility.logging import getLogger
         logger = getLogger()
         logger.debug("Shutting Down GangaRepositoryLocal: %s" % self.registry.name)
-        for k in self._fully_loaded:
-            try:
-                self.index_write(k, shutdown=True)
-            except Exception as err:
-                logger.error("Warning: problem writing index object with id %s" % k)
+        for k in self.objects:
+            if self.objects[k]._inMemory:
```
KonstantinSchubert: This is a matter of personal taste, but I find:
```
for k, obj in self.objects.iteritems()
   if obj._inMemory():
```
easier to parse.
I understand that my nitpicking can get annoying, please let me know if I should direct my energy elsewhere.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -1140,9 +1128,9 @@ def delete(self, ids):
                 logger.debug("Delete Error: %s" % err)
             self._internal_del__(this_id)
             rmrf(os.path.dirname(fn))
-            if this_id in self._fully_loaded:
-                del self._fully_loaded[this_id]
             if this_id in self.objects:
+                if self.objects[this_id]._inMemory:
```
KonstantinSchubert: Since the delete happens anyway, do we really need the `if` here?
Would simply setting 
```
self.objects[this_id]._inMemory = False
```
without the preceding if condition do the same?

rob-c:
`OK a massive rewrite based upon #627 is best.` this PR is going to be closed and deleted I've left it open for posterity and to remind me to check that no fixes in this branch are lost otherwise

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: @KonstantinSchubert I would say the main places where we're focusing development ahead of 6.2.0 is fixing bugs/missing features in `GaudiExec` which has now superceded `GaudiRun` as well as fixing any minor issues which get in the way of a user doing something.
I imagine 6.2.0 will be about 6.1.25 or so. 
(Gauss+GaussSplitter are the main missing features that come to mind in addition to some discussion which needs to happen first around Bender/OStap support)

Shortly after 6.2.0 I think we plan to integrate the new credentials branch which will address several issues behind the scenes with regard to handling proxies and such in a better way.
Then once all of this is out of the way I was hoping to see some code rewritten around the loading/lazy-loading of objects from disk in addition to the new objects to replace the `Executable` application which exposes problems.
In addition to this the IGangaFiles need to support workerDir, localDir properly and be made to behave the same way over the next few releases which will make handling of files much easier in the long term.

(This isn't a completely exhaustive list but is a run down of some of the key development tasks that I'm aware of over the next few months).

milliams:
Before closing this I'd like to point towards the document @drmarkwslater and I put together at about the potential [`RegistryObject`](https://github.com/ganga-devs/ganga/wiki/RegistryObject) for reference since I think it solves a lot of these problems in an elegant way.

alexanderrichards:
can the branch now be safely deleted?

rob-c:
@milliams Yes the proposal you've made in your branch allows this to be made much cleaner but the default-dict like implementation of the `_data` dict does allow for performance improvements as it potentially (and in a lot of cases does) spread out the initialization of complex objects so that all of the assignments don't happen by default.
This was the whole point of the branch with teh fixes around the XML and Repo layes being secondary to getting this to actually work and demonstrating the large performance gains you can get from this.

PR #587, opened 2016-06-16
Cleaned config

rob-c:
Migrating the LHCb config to make use of `setSessionValue` rather than `overrideDefaultValue`. Will play nicer with tests etc and cleans things up a little.

Nothing major changes so I'll merge this after the tests finish


milliams:
I'm not sure that `setSessionValue` is more correct than `overrideDefaultValue`. In fact I think the latter makes more sense. It's overriding the defaults for an overlapping plugin.

rob-c:
Yes but I was seeing unrelated issues to do with working out what the value was ever supposed to be within a test I was trying to write... Given we're using 1 of the 2 possible options here it made more sense to adopt 1 strategy unless things were broken

PR #588, opened 2016-06-20
GaudiRun adding read input data

rob-c:
This PR adds the readInputData functionality into the GaudiRun app...

I was a little surprised at first but this appears to be a very quick thing to do so I think I see some appeal in addition to the LHCb users getting their datasets from bookkeeping this way.

I based this heavily on the equivalent code in the old-style LHCb apps but added some changes to make this more 'cmake'-like. This still uses gaudirun to pickle a file and such as before.

Once the test here passes I'll merge this PR and do some additional testing before looking into making a beta release to LHCb.
(I think I've spotted a bug in LHCbDataset I'd like to make 100% sure doesn't go out into testing as it would be quite annoying)

### DIFF
--- a/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
+++ b/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
@@ -174,10 +174,9 @@ def calculateSiteSEMapping(file_replicas, wanted_common_site, uniqueSE, site_to_
 
 def lookUpLFNReplicas(inputs, allLFNData):
     # Build a useful dictionary and list
-    allLFNs = []
-    LFNdict = {}
+    allLFNs = [_lfn.lfn for _lfn in inputs]
+    LFNdict = dict().fromkeys(allLFNs)
     for _lfn in inputs:
-        allLFNs.append(_lfn.lfn)
         LFNdict[_lfn.lfn] = _lfn
 
     # Request the replicas for all LFN 'LFN_parallel_limit' at a time to not overload the
--- a/python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
+++ b/python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
@@ -1,0 +1,73 @@
+import subprocess
+import time
+from Ganga.Core.exceptions import GangaException
+from GangaLHCb.Lib.Applications import PythonOptionsParser
+
+def readInputData(optsfiles, app):
+    '''Returns a LHCbDataSet object from a list of options files. The
+       optional argument extraopts will decide if the extraopts string inside
+       the application is considered or not.
+
+    Usage example:
+    # Get the data from an options file and assign it to the jobs inputdata field
+    j.inputdata = readInputData([\"~/cmtuser/DaVinci_v22r0p2/Tutorial/Analysis/options/Bs2JpsiPhi2008.py\"], j.application)
+
+    This is also called behind the scenes for j.readInputData([\"~/cmtuser/DaVinci_v22r0p2/Tutorial/Analysis/options/Bs2JpsiPhi2008.py\"])
+
+    '''
+    if not isinstance(optsfiles, []):
+        optsfiles = [optsfiles]
+
+    # use a dummy file to keep the parser happy
+    if len(optsfiles) == 0:
+        raise GangaException("Need a file to parse to call this method")
+
+    try:
+        parser = PythonOptsCmakeParser(optsfiles, app)
+    except Exception as err:
+        msg = 'Unable to parse the job options. Please check options files and extraopts.'
+        logger.error("PythonOptionsParserError:\n%s" % str(err))
+        raise ApplicationConfigurationError(None, msg)
+
+    return parser.get_input_data()
+
+exportToGPI('GetGaudiRunInputData', readInputData, 'Functions')
+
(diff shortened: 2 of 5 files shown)


python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -174,10 +174,9 @@ def calculateSiteSEMapping(file_replicas, wanted_common_site, uniqueSE, site_to_
 
 def lookUpLFNReplicas(inputs, allLFNData):
     # Build a useful dictionary and list
-    allLFNs = []
-    LFNdict = {}
+    allLFNs = [_lfn.lfn for _lfn in inputs]
+    LFNdict = dict().fromkeys(allLFNs)
```
milliams: This could have been simply `LFNdict = dict.fromkeys(allLFNs)` to avoid creating a temporary `dict` object unnecessarily.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        parser = PythonOptsCmakeParser(optsfiles, app)
+    except Exception as err:
+        msg = 'Unable to parse the job options. Please check options files and extraopts.'
+        logger.error("PythonOptsCmakeParserError:\n%s" % str(err))
+        raise ApplicationConfigurationError(None, msg)
+
+    return parser.get_input_data()
+
+exportToGPI('GetGaudiRunInputData', readInputData, 'Functions')
```
milliams: Does this need to be exported, given that it's not documented anywhere? Is it ever expected to be used by the user directly?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
```diff
@@ -0,0 +1,78 @@
...
+import time
+from Ganga.Core.exceptions import GangaException
+from Ganga.Runtime.GPIexport import exportToGPI
+from Ganga.Utility.logging import getLogger
+from Ganga.Core.exceptions import ApplicationConfigurationError
+from .PythonOptsCmakeParser import PythonOptsCmakeParser
+
+logger = getLogger()
+
+def readInputData(optsfiles, app):
```
milliams: This seems to be an almost exact copy of the function you added in `PythonOptsCmakeParser.py`. Could they not have been combined?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
rob-c: `help(GetGaudiRunInputData)` returns the docstring. If you're referring to the fact this isn't in tutorials yet I'm still busy implementing the actual application and will return to larger tutorials (including pushing users to use the box repo more) after this.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
rob-c: I'm wanting to avoid any additional changes to the latter as this is heavily co-opted from other LHCb code which I'm just adopting and using here. I want to avoid changing the way it's called in-case I replace 1 for the other at some stage (although this is looking unlikely)

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
milliams: Ok, in that case:

1) it should start with a lower-case `g`
2) Is there a reason to have the Python function name be different to the exported GPI name?

rob-c: 1) will change
2) given there is a method on the application with that name I'm wanting to highlight the fact that this doesn't take the same arguments and is somehow unrelated to whatever the application does.

python/GangaLHCb/Lib/GaudiRunApp/PythonOptsCmakeParser.py
```diff
@@ -0,0 +1,153 @@
...
+       the application is considered or not.
+
+       Usage examples:
+       # Create an LHCbDataset object with the data found in the optionsfile
+       l=DaVinci(version='v22r0p2').readInputData([\"~/cmtuser/DaVinci_v22r0p2/Tutorial/Analysis/options/Bs2JpsiPhi2008.py\"])
+       # Get the data from an options file and assign it to the jobs inputdata field
+       j.inputdata = j.application.readInputData([\"~/cmtuser/DaVinci_v22r0p2/Tutorial/Analysis/options/Bs2JpsiPhi2008.py\"])
+
+    '''
+    if not isinstance(optsfiles, []):
```
milliams: I'm not sure this does what you expect. This will just fail and raise an exception. Is this function even called anywhere?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppUtils.py
milliams: Introducing two new functions which are line-for line identical (except some small differences which make it more confusing since they imply that they are significant but they are not) is a bad idea. It increases the maintenance load and makes the code harder to understand.

PR #590, opened 2016-06-20
Feature/codecleanup

rob-c:
This PR moves some of the innocuous code changes out of #584 to make that PR more manageable.

As soon as the tests pass I'll merge this as it's only a code cleanup, fixing a rare bug in Localhost resubmit and adding some extra docs to the code


rob-c:
retest this please

python/Ganga/GPIDev/Lib/Tasks/ITransform.py
```diff
@@ -128,6 +128,9 @@ def __init__(self):
         super(ITransform, self).__init__()
         self.initialize()
 
+    def _auto__init__(self):
+        self.status = 'new'
```
milliams: Is this needed even though the schema should set the default value? I worry that something is broken if this *is* needed.

python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -27,6 +33,12 @@ def _setupGangaPath():
 
 
 def start_ganga(gangadir_for_test, extra_opts=[]):
+    """
+    Startup Ganga by calling the same set of 'safe' functions each time
+    Args:
+        gangadir_for_test (str): This is the directory which the GangaUnitTest is to be run, creates a new gangadir per test to avoid collisions
```
milliams: *This* function does not create a new gangadir per test. The calling function is responsible for that.

PR #593, opened 2016-06-20
Stopping filenames getting excessively long

rob-c:
This PR just stops the rmrf method from renaming a file which is to be deleted too many times.

Files here are moved and then deleted due to some problems with AFS lockfiles so that any folders which can't be deleted due to a stray lock file won't get in the way in the XML repo in the future.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -129,9 +129,12 @@ def rmrf(name, count=0):
     if os.path.isdir(name):
 
         try:
-            remove_name = name + "_" + str(time.time()) + '__to_be_deleted_'
-            os.rename(name, remove_name)
-            #logger.debug("Move completed")
+            if not remove_name.endswith('__to_be_deleted'):
+                remove_name = name + "_" + str(time.time()) + '__to_be_deleted_'
+                os.rename(name, remove_name)
+                #logger.debug("Move completed")
+            else:
+                remove_name = name
         except OSError as err:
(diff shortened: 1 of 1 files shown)


python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -129,9 +129,11 @@ def rmrf(name, count=0):
     if os.path.isdir(name):
 
         try:
-            remove_name = name + "_" + str(time.time()) + '__to_be_deleted_'
-            os.rename(name, remove_name)
-            #logger.debug("Move completed")
+            remove_name = name
+            if not remove_name.endswith('__to_be_deleted'):
+                remove_name = name + "_" + str(time.time()) + '__to_be_deleted_'
```
milliams:
```python
remove_name += "_" + str(time.time()) + '__to_be_deleted_'
```
would have been simpler.

rob-c: This class calls itself and once the file has been renames it's passed as the file to be deleted again.
There are files on AFS which survive multiple calls to be removed including (but not limited to) lockfiles

milliams: I was only referring to that last line, not the whole block:
```python
remove_name = name
if not remove_name.endswith('__to_be_deleted'):
    remove_name += "_" + str(time.time()) + '__to_be_deleted_'
```
or, even better,
```python
remove_name = name
if not remove_name.endswith('__to_be_deleted'):
    remove_name += "_%s__to_be_deleted_" % time.time()
```

PR #594, opened 2016-06-20
Some splitter improvements

rob-c:
This PR is an attempt to make the OfflineSplitter for DIRAC slightly more robust and correctly handle some external call failing for some reason.

I've also cleaned up the imports a little.

### DIFF
--- a/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
+++ b/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
@@ -3,8 +3,13 @@ from Ganga.GPIDev.Adapters.ISplitter import SplittingError
 from GangaDirac.Lib.Backends.DiracUtils import result_ok
 from Ganga.Utility.Config import getConfig
 from Ganga.Utility.logging import getLogger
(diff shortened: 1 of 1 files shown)


rob-c:
retest this please

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -14,10 +19,21 @@
...
 
+def wrapped_execute(command, expected_type):
+    """
+    A wrapper around execute to protect us from commands which had errors
+    """
+    try:
+        result = execute(command)
+        assert isinstance(result, expected_type)
+    except AssertionError:
+        raise SplittingError("Output from DIRAC expected to be of type: '%s', we got the following: '%s'" % (expected_type, result))
```
milliams: Why `assert` then catch `AssertionError`? Why not just raise `SplittingError` directly?
```python
result = execute(command)
if not isinstance(result, expected_type):
    raise SplittingError("Output from DIRAC expected to be of type: '%s', we got the following: '%s'" % (expected_type, result))
return result
```

PR #596, opened 2016-06-21
Proposed changes to tasks tests

rob-c:
This is a small extra test for the tasks system to verify that the tasks system (when the monitor isn't running) is obeying the lazy-loading and a few subtle changes.

### DIFF
--- a/python/Ganga/test/GPI/Tasks/TestCoreTasks.py
+++ b/python/Ganga/test/GPI/Tasks/TestCoreTasks.py
@@ -6,13 +6,15 @@ from Ganga.testlib.GangaUnitTest import GangaUnitTest
 class TestCoreTasks(GangaUnitTest):
     """A set of tests to cover both ITask and CoreTask plus related objects"""
 
+    argList = ['300', '300', '300']
+
     def setUp(self):
         """Make sure that the Tasks object isn't destroyed between tests"""
         super(TestCoreTasks, self).setUp()
         from Ganga.Utility.Config import setConfigOption
         setConfigOption('TestingFramework', 'AutoCleanup', 'False')
         setConfigOption('Tasks', 'TaskLoopFrequency', 1)
-        self._numTasks = 50
+        self._numTasks = 25
 
     def createTask(self):
         """create a task with some defaults"""
@@ -20,10 +22,10 @@ class TestCoreTasks(GangaUnitTest):
         t = CoreTask()
 
         trf = CoreTransform()
-        trf.application = Executable()
+        trf.application = Executable(exe='sleep')
         trf.unit_splitter = GenericSplitter()
         trf.unit_splitter.attribute = "application.args"
-        trf.unit_splitter.values = ['arg 1', 'arg 2', 'arg 3']
+        trf.unit_splitter.values = TestCoreTasks.argList
 
         t.appendTransform(trf)
         t.float = 20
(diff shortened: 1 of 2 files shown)


rob-c:
Relies on #584

drmarkwslater:
I don't believe Tasks actually uses lazy loading at all at present. This should be changed as Tasks is being used a lot more and loading them can be quite heavy now but at present, it's not implemented.

python/Ganga/test/GPI/Tasks/TestCoreTasks.py
```diff
@@ -6,24 +6,26 @@
...
+    argList = ['300', '300', '300']
+
     def setUp(self):
         """Make sure that the Tasks object isn't destroyed between tests"""
         super(TestCoreTasks, self).setUp()
         from Ganga.Utility.Config import setConfigOption
         setConfigOption('TestingFramework', 'AutoCleanup', 'False')
         setConfigOption('Tasks', 'TaskLoopFrequency', 1)
-        self._numTasks = 50
+        self._numTasks = 25
```
drmarkwslater: I'd keep the number of Tasks higher as this test has picked up a few deadlocks in recent weeks and I don't want to risk missing them.

python/Ganga/test/GPI/Tasks/TestCoreTasks.py
```diff
@@ -6,24 +6,26 @@
...
 
     def createTask(self):
         """create a task with some defaults"""
         from Ganga.GPI import tasks, CoreTask, CoreTransform, Executable, GenericSplitter
         t = CoreTask()
 
         trf = CoreTransform()
-        trf.application = Executable()
+        trf.application = Executable(exe='sleep')
```
drmarkwslater: Out of interest, what is the reason for changing this to a sleep?

python/Ganga/test/GPI/Tasks/TestCoreTasks.py
```diff
@@ -120,6 +123,27 @@ def test_f_CheckAccessToManyTasks(self):
...
+
+        assert len(tasks)
+
+        assert tasks(0).status != 'new'
+
+        new_t = tasks(0).copy()
+
+        assert new_t.status == 'new'
+        assert len(new_t.transforms) == 0
+        # transforms are marked (for now?) uncopyable
```
drmarkwslater: This is almost certainly a bug. I'll create an issue.

python/Ganga/test/GPI/Tasks/TestCoreTasks.py
rob-c: I was wanting to ensure that the jobs were still running and that I could more easily monitor the number of launched jobs etc when debugging. Nothing major.

rob-c:
@drmarkwslater It passes the lazy loading test fine, and it does use it (although the monitoring loop is a tad too agressive and loads completed tasks at times I think). The only difficulty is in the copying of the task which changed in #584 to fix a bug.

python/Ganga/test/GPI/Tasks/TestCoreTasks.py
drmarkwslater: Fair enough but maybe reduce the sleep time a bit? It would good to have these completing while the Task is submitting, etc.

drmarkwslater:
OK - I'll have to look into that then. We must be getting that 'for free' because I never added any specific lazy loading to Tasks myself. I suspect it's not really working as intended but I'd have to check.

python/Ganga/test/GPI/Tasks/TestCoreTasks.py
rob-c: If we're testing for this then I'll remove the sleep. Out was for debugging something solved elsewhere :)

mesmith75:
Are we putting this into a later release?

drmarkwslater:
I haven't had chance to properly look at this as I want to make sure I understand what the lazy loading is doing for Tasks. I'm concerned it's not working correctly (even if it passes the tests) and may do something odd. This may then impact the changes to the Tasks tests.

rob-c:
Happy to drop if tests are currently working

PR #598, opened 2016-06-21
removing bad option on lhcb-prepare (docs)

rob-c:
removing the -v lines on lhcb-prepare

Waiting on the test to  finish which gets in the way of editing docs on the fly

Should we consider putting the docs in a second repo so that people can contribute to them in a much more informal way because collecting things for a PR and then re-testing Ganga is a bit overkill for submitting changes at the level of typos to the docs. This is the advantages of wiki's etc...

### DIFF
--- a/doc/UserGuide/InstallAndBasicUsage.rst
+++ b/doc/UserGuide/InstallAndBasicUsage.rst
@@ -77,17 +77,17 @@ To install locally run one of the following:
 * For the official Ganga releases:
 .. code-block:: bash
 
-        ./lhcb-prepare -v v601r14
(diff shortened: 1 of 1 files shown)


rob-c:
retest this please

milliams:
As we move forward we will want to be running more automated tests over the examples in the tests and so having the tests run on doc changes is useful. For example the tutorial is run as part of the test suite to catch errors early.

rob-c:
If there is ever a nice way of testing that SetupProject is working I'd look into it, as it stands I'm wanting to point `cvmfs` users at this as part of the release testing atm.

PR #605, opened 2016-06-22
1 line fix of rmrf

rob-c:
1 line fix of rmrf

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -131,7 +131,7 @@ def rmrf(name, count=0):
         try:
             remove_name = name
             if not remove_name.endswith('__to_be_deleted'):
-                remove_name += '_%s__to_be_deleted_' % str(time.time())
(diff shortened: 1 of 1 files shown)


python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -131,7 +131,7 @@ def rmrf(name, count=0):
         try:
             remove_name = name
             if not remove_name.endswith('__to_be_deleted'):
-                remove_name = name + "_" + str(time.time()) + '__to_be_deleted_'
+                remove_name += '_%s__to_be_deleted_' % str(time.time())
```
milliams: No need for the explicit `str()`. `%s` is performing that action already.

rob-c:
@milliams Are we making use of the git plugins for Jenkins or do we do the git clone in our test scripts? This is failing quite a large amount atm and I'm just wondering if we can wrap it in a try/catch of some sort ?

rob-c:
retest this please

milliams:
We are using the built-in Jenkins system. The fault most likely lies with CERN's networking.

rob-c:
retest this please

PR #607, opened 2016-06-22
removing duplicate function

rob-c:
removing dupicate function from GaudiRun code


rob-c:
retest this please

rob-c:
If I get a chance I'm going to run a traceroute from the Jenkins server I can't work out what is causing these server<->github issues but it's happing more and more frequently atm

PR #612, opened 2016-06-23
Moving to use ._inMemory everywhere for tracking object loaded state

rob-c:
This is an extension of #584 which moves to make use of the `GangaObject._inMemory` boolean for tracking an objects loaded state (introduced in the former #PR)

This touches quite a few files so once I'm certain this works I can look at merging this into #584 ahead of merging that into develop.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepository.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepository.py
@@ -134,14 +134,6 @@ class GangaRepository(object):
         """
         pass
 
-    def isObjectLoaded(self, obj):
-        """
-        Returns if an object is loaded into memory
-        Args:
-            obj (GangaObject): object we want to know if it's in memory or not
-        """
-        raise NotImplementedError
-
 
 # Optional but suggested functions
     def get_lock_session(self, id):
@@ -301,12 +293,4 @@ class GangaRepositoryTransient(object):
             ids (list): The object keys which we want to iterate over from the objects dict
         """
         pass
-
-    def isObjectLoaded(self, obj):
-        """
-        Returns if an object is loaded into memory
-        Args:
-            obj (GangaObject): object we want to know if it's in memory or not
-        """
-        return True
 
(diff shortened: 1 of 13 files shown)


rob-c:
Getting the Registry/Repo and GangaObject to share the same way of tracking the loaded status of an object causes dead-locks...
I debug this and see if it can ever actually work. Maybe just replacing the `_fully_loaded` dict with something.

The problem is that `._inMemory` is in the GangaObject and trying to call this from within the Registry is complex when there are multiple threads all doing things. It's much cleaner if the 2 systems track if the object is loaded independently as it avoids most of these collisions in that case.
I'll see if I can get this to work else I'm not going to look to merge this

drmarkwslater:
Why does the object need to know if it's fully loaded? Isn't this just a Registry/Repo related thing that the object will only ever care about if you request something through the getter which isn't in the dictionary? This should actually be the only point where an object calls back into the Registry and the Registry should have minimal (if any - I'm not sure) calls into the Object itself. Having two separate ways of recording the same information has a good chance of coming back to bite us in the future.

rob-c:
`Why does the object need to know if it's fully loaded?` for lazy loading you can't avoid this, you need to know at some level (or in the case of develop completely ignore this and always request a re-loading of the job from disk regardless which is handled in the Registry).
I would like to get rid of `_fully_loaded` and just inspect the `inMemory` boolean I introduced in #584 but this causes deadlocks in this PR (yet not in #584 where the Registry and GangaObject don't trip over each other). I think I'll have to rethink this PR to avoid lockups in the future which involves some minor work but this has clear performance gains.

drmarkwslater:
But can't the object just call into the Registry to request to load itself and the Registry itself decide if it's fully loaded? As this call will (or at least should) only be happening in the getter/setter and after checking the dictionary and index cache for a value being present, then it should never be getting loaded multiple times. I'm not sure I understand why shifting from the Repo/Registry knowing if an object's loaded to using a flag on the object itself can get us any gains.

rob-c:
This extension to #584 makes it difficult to see what's going on. I'll tackle #584 first then deal with why this causes thread-locks in a bit. I will get back to you question.

PR #618, opened 2016-06-27
Fixing bug where test deadlocks and other nasties

### LINKED ISSUE #617: TestSJIDName::test_b_jobResubmit Keeps stalling
This is bad behaviour as this test has dramatically different behaviour on the Jenkins box to all of my other systems I've tested it on... (in develop!) I'll open a PR to fix this in a few minutes

rob-c:
Fixed #617 

I've fixed #617 by more carefully defining the SJID test and I've added an additional bit of logic to Job.py to avoid stalling (or worse!) due to very bad behaviour.

I've also added a few new short tests to test the resubmit code around subjobs as this wasn't being tested in an independent way where picking up this bug might have been slightly quicker/easier.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -227,7 +227,7 @@ class GangaRepositoryLocal(GangaRepository):
             raise RepositoryError(self.repo, "Unknown Repository type: %s" % self.registry.type)
         self.sessionlock = SessionLockManager(self, self.lockroot, self.registry.name)
         self.sessionlock.startup()
-        # Load the Ganga/Core/GangaRepository/GangaRepositoryXML.pylist of files, this time be verbose and print out a summary
+        # Load the list of files, this time be verbose and print out a summary
         # of errors
         self.update_index(verbose=True, firstRun=True)
         logger.debug("GangaRepositoryLocal Finished Startup")
@@ -243,7 +243,7 @@ class GangaRepositoryLocal(GangaRepository):
             try:
                 self.index_write(k, shutdown=True)
             except Exception as err:
-                logger.error("Warning: problem writing exception for %s" % k._id)
+                logger.error("Warning: problem writing index object with id %s" % k)
         try:
             self._write_master_cache(True)
         except Exception as err:
@@ -1190,7 +1190,10 @@ class GangaRepositoryLocal(GangaRepository):
         Clear EVERYTHING in this repository, counter, all jobs, etc.
         WARNING: This is not nice."""
         self.shutdown()
-        rmrf(self.root)
+        try:
+            rmrf(self.root)
+        except Exception as err:
+           logger.error("Failed to correctly clean repository due to: %s" % err)
         self.startup()
 
     def isObjectLoaded(self, obj):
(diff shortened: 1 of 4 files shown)


python/Ganga/test/GPI/TestSJSubmit.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def setUp(self):
+        """Make sure that the Job object isn't destroyed between tests"""
+        super(TestSJSubmit, self).setUp()
+        from Ganga.Utility.Config import setConfigOption
+        setConfigOption('Output', 'FailJobIfNoOutputMatched', 'True')
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
+        setConfigOption('Configuration', 'resubmitOnlyFailedSubjobs', 'True')
+
+    def _getSplitter(self):
```
milliams: This method doesn't use `self` so could probably be a `@staticmethod` or even a free function.

python/Ganga/test/GPI/TestSJSubmit.py
```diff
@@ -0,0 +1,102 @@
...
+class TestSJSubmit(GangaUnitTest):
+
+    n_subjobs = 10
+
+    def setUp(self):
+        """Make sure that the Job object isn't destroyed between tests"""
+        super(TestSJSubmit, self).setUp()
+        from Ganga.Utility.Config import setConfigOption
+        setConfigOption('Output', 'FailJobIfNoOutputMatched', 'True')
+        setConfigOption('TestingFramework', 'AutoCleanup', 'False')
```
milliams: This test doesn't seem to be actually testing the flush/load mechanism. As such I think it would be better (as well as faster) if we avoided doing the full Ganga unload/load between each test function. Perhaps they could all be combined into one test function or maybe use a pytest fixture with module scope.

rob-c: Yes I am relying on this happening to test different sections of code. The behaviour changes between loading a job from disk and running with it in memory. Arguably I should add a job.resubmit whilst in the same session but this is tested elsewhere.

rob-c:
Problems fixed here have been highlighted running the following in the LHCb environment.

```
py.test ./Ganga/test/GPI
```

It would be extremely nice to have a nightly which runs all of the tests in the LHCb/Atlas/Core/Other environments as well as a non-bash shell as we're still susceptible to bugs creeping in if the code paths are changed due to the different environment configs.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -240,8 +240,14 @@ def shutdown(self):
         logger = getLogger()
         logger.debug("Shutting Down GangaRepositoryLocal: %s" % self.registry.name)
         for k in self._fully_loaded:
-            self.index_write(k, shutdown=True)
-        self._write_master_cache(True)
+            try:
+                self.index_write(k, shutdown=True)
+            except Exception as err:
```
milliams: These really don't need to be such broad exception handlers.

python/Ganga/test/GPI/TestSJSubmit.py
```diff
@@ -0,0 +1,152 @@
...
+
+        sleep_until_completed(jobs(0))
+
+        jobs(0).subjobs(0).force_status('failed')
+
+        jobs(0).resubmit()
+
+        sleep_until_completed(jobs(0))
+
+    def test_e_testInMemory(self):
```
milliams: This test function really needs some comments explaining what's going on. If one of the steps fail, there'll be no way to know what the intent of the particular part of the test was.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: These were failing with KeyError, OSError and others so until the function is safer I'm leaving it as this and moving on. The index by definition doesn't (or at least shouldn't) make a job un-loadable. Yes it's an annoyance if it wasn't writted but it's not worth sacrificing the rest of the shutdown process for it.

PR #619, opened 2016-06-28
Nicely disable LCG

milliams:
As we have discussed in the past, we want to be able to disable the LCG system by default without affecting anyone who is currently using it. This PR achieves this by changing the default `[LCG]VirtualOrganisation` to `''` and adding a check to only enable the LCG system if the VO has been set to *something*.

In any case, a default VO of `dteam` made little sense anyway.

I've labelled this for 6.1.22 but it could be brought forward to 6.1.21 if people wanted.

### DIFF
--- a/python/Ganga/Lib/LCG/LCG.py
+++ b/python/Ganga/Lib/LCG/LCG.py
@@ -2120,6 +2120,10 @@ class LCGJobConfig(StandardJobConfig):
 
 logger.debug('LCG module initialization: begin')
 
+if config['GLITE_ENABLE'] and not config['VirtualOrganisation']:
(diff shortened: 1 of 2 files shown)


egede:
@milliams Thanks for finally sorting this out - and actually in a very nice way. It is fine to leave for 6.1.22 I think.

rob-c:
I personally would like to see this introduced sooner rather than later, what is the status of 6.1.21?

milliams:
@rob-c I'm happy to put this into 6.1.21. I was just waiting for your Gaudi PR to be merged but now that's in I can start the release if you're ready?

rob-c:
@milliams Go from me, everything I need to get in is there

Issue #616, opened 2016-06-27
Review getter/setter methods on Schema Items


drmarkwslater:
We seem to have at least 2 ways of applying getters and setters or callbacks to schema items, the `attribute_filter_blahblah` stuff and the `checkset` value (which is broken and hardly used either). This should be reviewed properly and a single solution settled on and documented.

rob-c:
I'd be happy to see both torched tbh, is there any real need to keep these around?
If there is a need then great, but can't most/all of this be done through different methods?

drmarkwslater:
All of these seem to be different ways of doing exactly the same thing so they certainly need to be consolidated. However, we should take a bit of care as these things are still used (to a certain degree) and there are certain cases we may want to support as well.

rob-c:
I agree there may be cases, but I can't motivate any in a particularly strong way. We can do most of this by intercepting calls to the getter I would imagine and do away with code which is fragile at best.

How many places are these 2 mechanisms used?

drmarkwslater:
Not a lot from a very quick grep but the `attribute_filter` stuff is used in `Jobs` apparently. It should be checked properly but it shouldn't be a major problem to replace with a single implementation.

rob-c:
@drmarkwslater When making changes recently I've discovered we don't have a sensible chain for inheriting the `__getattr__` method in the same way that we can overload the `__setattr__` method.

Would it be sensible to propose that we move the getter method system over to using some form of `__getattr__` call in `__get__` and have this resolve to look in `_data` in the baseclass to allow for higher up classes to better overload getter calls in the future?
(If I'm mixing up the order in which the `Descriptor.__get__` and `myClass.__getattr__` are called I apologize but it appears we can't easily overload the `__getattr__` method which I would naively expect to be able to use)

milliams:
@rob-c There is a fundamental clash between `__getattr__`/`__setattr__` and descriptors. In your recent changes you had to manually call the base class `__setattr__` in order for the descriptor to kick in, this is really a bit of a hack. Really this per-item pre-processing and potential post-precessing on the getting of an item should all be handled *within* the descriptor.

We should have a function which is called by `Descriptor.__set__` at the very beginning which can transform any incoming value (this is basically what the filter does now). Perhaps likewise, the last thing that `Descriptor.__get__` should do before returning a value is pass it though some other transformation function. These functions can have side-effects if we wish as well. The current concept of 'getters' in the schema complicates this a bit as it works very differently to everything else.

[...]

PR #627, opened 2016-06-28
Remove getwriteaccess calls where possible

### LINKED ISSUE #528: Fix/Remove _getWriteAccess and _getReadAccess in Objects.py
We have recently had issues with these functions causing deadlocks (see #527, #510, #520). We should go over these specifically and the Registry in general to make sure that Objects only call into the registry when the absolutely have to (if they ever need to) and the registry locking is done in a bit more sensible way with read/write locks as @rob-c has suggested.

drmarkwslater:
This is a rather large PR that removes the confusing `_getReadAccess` and `_getWriteAccess` calls and replaces them with more specific calls to load or lock the object. It also does a certain amount of cleanup around these and other functions in the Registry and Objects.py.

There are still some more minor changes to come in the next hour but I wanted to see it in a PR and make sure I document everything both in code and in the PR.

Summary of changes:

* Break out the single overall Registry lock into two locks: A quick 'read' lock and a specific 'flush' lock. This means functions that don't care about read actions taking place can acquire the lock they want (I have two as the flush can take significantly longer than any other call) whereas 'destructive' functions have to acquire both to continue. Practical upshot: quick registry read calls don't have to wait for the flush but write ones still need have exclusive access.

* Made the `_load` registry function decide whether to load an object or not itself and also moved this check to outside the lock acquisition.

* Removed references to `_update_index_time` and `_changed_ids` as this was only ever done semi-randomly on calls to `items` or `ids` in code that did something very strange. The call to `repository.update_index()` might need to be shifted elsewhere but certainly as was, it didn't seem to be working as intended.

* Removed the transaction locks as these were only ever called within functions that were protected in by `@syncronised` anyway.

* Did some general clean up, e.g. removed functions that were only called once, removed some `try...except` blocks that weren't doing anything with the caught exception, etc.

* Removed `_getReadAccess` and `_read_access` and replaced with explicit load calls as this just boiled down to trying to load the object from disk.

* Removed `_getWriteAccess` and `_write_access` and replaced with explicit calls as these were attempting to acquire a Session Lock and then attempting a load. Note that the only place a load is attempted is now in the setter and getter. Other calls to `_getWriteAccess` have just been replaced with the Session Lock call (which also may not be needed but at least is explicit now 😄 )

* Removed `getGangaList` function and just replaced with appropriate imports in `Objects.py`

As this touches quite a bit of low-level code I won't even consider merging this without at least 2 other devs OK 😄 

Fixes #528


drmarkwslater:
That should be the last of the changes - ready for review 😄

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -1031,55 +1017,23 @@ def _getIOTimeOut(self):
             _timeOut = 5. # 5sec hardcoded default
         return _timeOut
 
-    def _getWriteAccess(self):
-        """ tries to get write access to the object.
-        Raise LockingError (or so) on fail """
-        root = self._getRoot()
-        reg = root._getRegistry()
+    def _getSessionLock(self):
```
rob-c: Why is this used? We currently get a lock on loading the object or are there plans to change this?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -1031,55 +1017,23 @@ def _getIOTimeOut(self):
...
+            reg._release_session_lock(self._getRoot())
 
-    def _getReadAccess(self):
-        """ makes sure the objects _data is there and the object itself has a recent state.
-        Raise RepositoryError"""
-        root = self._getRoot()
-        reg = root._getRegistry()
-        if reg is not None:
-            reg._read_access(root, self)
+    def _loadObject(self):
```
rob-c: 👍

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -468,16 +451,20 @@ def __set__(self, obj, val):
...
             checkSet = self._bind_method(obj, self._checkset_name)
             if checkSet is not None:
-                checkSet(_val)
+                checkSet(val)
+
         if hasattr(obj, '_filter_name'):
             this_filter = self._bind_method(obj, self._filter_name)
             if this_filter:
-                val = this_filter(_val)
+                val = this_filter(val)
```
rob-c: How has this not turned up in testing before now?

python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py
```diff
@@ -47,22 +47,22 @@ def shutdown(self):
...
         #logger.info("ShareRef: %s" % getName(self.shareref))
-        self._lock.acquire()
-        ## THIS IS DISABLED AS IT REQUIRES ACCESS TO REPO OBJECTS THROUGH GETREADACCES...
-        ## THIS NEEDS TO BE FIXED OR IMPLEMENTED AS A SHUTDOWN SERVICE!!!
-        try:
-            stripProxy(self.shareref).closedown()  ## Commenting out a potentially EXTREMELY heavy operation from shutdown after ganga dev meeting - rcurrie
-        except Exception as err:
-            logger.error("Shutdown Error in ShareRef")
-            logger.error("Err: %s" % err)
+        with self._flush_lock:
```
rob-c: would this be better with descriptors as in the Registry?

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -259,33 +272,26 @@ class Registry(object):
...
         self._hasStarted = False
-        self.update_index_time = update_index_time
-        self._update_index_timer = 0
         self._needs_metadata = False
         self.metadata = None
-        self._lock = threading.RLock()
-        self.hard_lock = {}
-        self.changed_ids = {}
+        self._read_lock = threading.RLock()
+        self._flush_lock = threading.RLock()
```
rob-c: Have these been added to the test repo or aren't they needed there?

python/Ganga/GPIDev/Base/Objects.py
drmarkwslater: This is up for discussion as far as I'm concerned as in the original code, `_getReadAccess` would do the load but **not** acquire the Session Lock but `_getWriteAccess` would try to acquire a Session Lock but load even if that failed. I would perfectly happy to link the two (i.e. grab the Session lock when we load an object) but that leaves the question of releasing the Session lock as at the moment, we can't 'unload' an object. Now, we may **not** want to ever release a Session Lock in which case that is a moot point 😄

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -331,22 +309,25 @@ def __getitem__(self, this_id):
...
+        if this_id in self._incomplete_objects:
+            return IncompleteObject(self, this_id)
+
+        # Nope, so try to find it and raise an exception if not
         try:
             return self._objects[this_id]
         except KeyError as err:
             logger.debug("Repo KeyError: %s" % err)
             logger.debug("Keys: %s id: %s" % (self._objects.keys(), this_id))
-            if this_id in self._incomplete_objects:
```
rob-c: Is this safe? surely the object is only incomplete after failing to load in the repository?

python/Ganga/GPIDev/Base/Objects.py
drmarkwslater: This happened recently with the removal of `__atomic_set__` but these features are hardly ever used and in fact, this code doesn't work anyway as it's asking for `_filter_name` and `_checkset_name` on `obj` rather than `self`. These can basically be completely removed but I thought it better to wait until we close #616

python/Ganga/GPIDev/Base/Objects.py
rob-c: I think it's safer that we don't release it until shutdown for now as this is a bit dangerous I feel.
I didn't think we could load an object if getting the lock failed, I assumed that was the point, or whether we were loading it but unable to use it in which case I'm even more worried about the way it used to be.

In this PR are we currently locking on loading in this case then?

python/Ganga/test/GPI/Internals/TestRepo.py
```diff
@@ -16,7 +16,10 @@ def __init__(self, name):
     def _read_access(self, root, obj):
         self.repo.load([root._registry_id])
 
-    def _write_access(self, root):
+    def _acquire_session_lock(self, root):
```
rob-c: Is it better to lock the methods here in the same way as they're locked in the actual Registry?
I'm weary that this test could be hiding possible deadlocks if it doesn't reflect the locking in the actual registry

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -171,22 +161,45 @@ def __repr__(self):
...
+def synchronised_complete_lock(f):
+    """
+    Entirely lock the registry. Make sure locks are acquired in the right order!
+
+    Args:
+        f (func): Function to decorate
+    """
+    @functools.wraps(f)
+    def decorated(self, *args, **kwargs):
+        with self._flush_lock:
```
rob-c: Wouldn't it be better to lock the other way around so that it's not possible to acquire the locks in any out of order way?
(No idea how likely/possible that is I'm just wondering)

drmarkwslater: Not sure what you mean here - this only acquires one lock. If you mean the `@complete_lock' one where both locks are required, the order is determined by what functions can try to acquire what locks, i.e. when flushing, the read lock is required we need to make sure that the complete_lock gets this flush lock **first** to avoid a deadlock.

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: I don't think the locks are specifically mentioned but I've updated the tests to refer to the session locks and load functions.

python/Ganga/Core/GangaRepository/Registry.py
drmarkwslater: I'm assuming that an object won't be in both `_incomplete_objects` AND `_objects` in which case, it doesn't matter which order the check is done in and I thought it was neater to have the `_incomplete_objects` check outside of the exception handler.

python/Ganga/GPIDev/Base/Objects.py
drmarkwslater: The old code called `_load` in the `finally` part of the `try` block on the registry lock so it was going to load anyway. I agree that at the moment, if you load an object then you acquire the session lock and don't release it until shutdown, however, that would be an even bigger change and should probably wait for another PR!

This PR uses the `@complete_lock` on load where previously it was the `@syncronised' lock.

python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py
drmarkwslater: Almost certainly, but I was trying to keep the changes to parent class as I didn't want to have to debug the other registries just yet! If this PR works then looking at the child classes would be the next job.

python/Ganga/test/GPI/Internals/TestRepo.py
drmarkwslater: Almost certainly true, but again, I feel that's out of the scope of this PR at present. I really just wanted to get the basics done of removing `_getReadAccess` and `_getWriteAccess` and then after we know what's actually happening, it makes it a lot easier to do these kind of changes.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: Ah, yes I meant complete_lock.
OK, the locks are in the order I would expect them to be I may have mixed up the code I was looking at apologies.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: ... I would prefer to see that these locks are being tested in some form for deadlocks but I suppose it's unlikely not to be caught in our current tests.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: It's never in the `_incomplete_objects` until it's been removed from the `_objects` list where it sits when Ganga has loaded up. This is a bit of a mess but you will try to access an object then find it not load and be removed from the `_objects` list.

rob-c: Else you'll be returning a KeyError when an object fails to load which will get confusing unless we're taking this into account.

drmarkwslater: I'm not sure I understand - If a load fails then the object is added to `_incomplete_objects`. Then, the next time it's referenced it will be picked up in the first part of the code and an `IncompleteObject` will be returned. The only reason this code will return a `KeyError` is if the object isn't in **either** `_incomplete_objects` or `_objects` in which case, that would be expected wouldn't it?

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -646,98 +527,37 @@ def flush_all(self):
...
-            for this_d in self.changed_ids.itervalues():
-                this_d.add(this_id)
-        except (RepositoryError, RegistryAccessError, RegistryLockError, ObjectNotInRegistryError) as err:
-            raise
+            self.repository.load([obj_id])
         except Exception as err:
-            logger.debug("Unknown read access Error: %s" % err)
+            logger.error("Error Loading Job! '%s'" % obj_id)
+            ## Cleanup after ourselves if an error occured
+            ## Didn't load mark as clean so it's not flushed
```
milliams: Since leading `# ` for comments as per PEP8.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -746,137 +566,44 @@ def _write_access(self, _obj):
...
-            _obj (GangaObject): This is the object we want to get write access to and lock on disk
-        """
-        logger.debug("__write_acess")
-        obj = stripProxy(_obj)
-        this_id = id(obj)
-        if this_id in self._inprogressDict:
-            for this_d in self.changed_ids.itervalues():
-                this_d.add(self.find(obj))
-            return
+        obj = stripProxy(obj)
```
milliams: We *shouldn't* need any of these `stripProxy` calls any more. It's worth checking them but I'm pretty sure they're now redundant.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -468,16 +451,20 @@ def __set__(self, obj, val):
...
 
         self._check_getter()
 
-        # ON-DISK LOCKING
-        obj._getWriteAccess()
+        # make sure the object is loaded if it's attached to a registry
+        obj._loadObject()
+
+        # make sure we have the session lock
+        obj._getSessionLock()
```
milliams: As discussed in person, I think that these should be inverted.

milliams:
As @drmarkwslater and I discussed in person (and in addition to my few comments on the diff), the object should be flushed whenever the session lock is released. Once that is in, :+1:

(We also had some good ideas for how to progress this in the future to make the lazy-loading/locking/registry stuff work in a much nicer and simpler way. They will be forthcoming in future issues and PRs.)

rob-c:
I'm very happy if we can isolate the loading of the objects and registry access (and I'm strongly keen to see the index cache removed from the getter I think this is at very best confusing).

I'm wanting to try and move to some sort of default-dictionary like behavior for _data_dict but it ends up having to make a lot of changes around the loading of jobs and registry locking to work as there are some (possibly fundamental) problems with the current setup from a development perspective.

python/Ganga/Core/GangaRepository/Registry.py
rob-c: The flow is something like this:

 1) Ganga loads index (if it can't load the index or tries to load the job here because the index is corrupt the job is marked as incomplete and placed in `_incomplete_objects`.
  If the job is __not__ loaded here it is still in `_objects`
 2) The user (in some form) tried to load the job from disk. If this completed the object in `_objects` is removed. If it is not the object is now added to `_incomplete_objects`.
  (It may or may not still reside in the `_objects` list but I think it should be replaced by an an `IncompleteObject` class if it does).

At step 2 you cannot always know in advance if the job can be successfully loaded. You can have a job which depends on a plugin which isn't loaded at the moment and this will get into this state.

I see this a LOT due to being careless when moving from LHCb<->LSST and I'm keen not to start obliterating jobs or poisoning sessions with errors.

drmarkwslater: OK - I think I get it. So there is a point where an object can be referenced in **both** `_objects` and `_incomplete_objects` and in this case it should return the given object rather than an `IncompleteObject`. If that's the case I'll make a note and revert the code back.

rob-c: No the object is not in `_objects` here as we've arrived here through a KeyError. I suspect the object is not in the list else the incomplete jobs would show in the Jobs Registry wrapper (They currently disappear after you access them with a warning).

drmarkwslater: I'm referring to the function as a whole as in the new version, I've just moved the `incomplete_objects` check outside of the `KeyError` to before the `try..except` block so this check occurs **before** the check against `_objects`. This function will only give different behaviour if there is a point where the object is referenced in both lists as it will return an `IncompleteObject` rather than the copy in `objects`.

rob-c: @drmarkwslater OK I think I'm seeing this now. I think the object is never loaded here in this case?

Apologies I think I'm getting confused from trying to debug another issue with loading of jobs. I forgot it's elsewhere the object is actually loaded and so moves between the lists. This should be safe in that case if it's locked so my apologies again.
(I saw this happen a few times in the past but now I realize that was likely a threading issue as it was before we locked some transactions).

drmarkwslater: @rob-c I believe that's correct, yes.

drmarkwslater:
Has anyone got any more issues with this? If not, I'd like to merge ASAP as that will give us the most time to pick any problems it may have created before 6.1.22!

rob-c:
go for merge from me if everything is working :+1:

drmarkwslater:
@rob-c  Are you wanting to do a quick release to fix #651 ? If so, I'll hold off until after that, if not I'll merge after the tests complete.

rob-c:
@drmarkwslater Don't know yet, I'm not an LHCbTask expert I've barely used the codebase so I'll need to do some testing and convince myself I know what's going on.

drmarkwslater:
@rob-c given that there isn't a big push for 6.1.22 now, I'm going to merge this in the next hour if that's OK?

rob-c:
Please do :+1:

rob-c:
Actually #655 has cropped up can you give me 15-20min to make an assessment of what went wrong, thanks

rob-c:
@drmarkwslater feel free to merge asap

PR #629, opened 2016-06-28
Additional Cmake support

### LINKED ISSUE #621: GaudiRun application fails if there is not input data
If there is no input data for a GaudiRun job, the data.py file is not created. Subsequently the data.py file is expected at runtime causing a failure.

### LINKED ISSUE #622: GaudiRun application prepare step
In the prepare step of the GaudiRun application, there is a "make clean" which should *not* be executed. Also the build.<arch>/ganga directory should be completely cleaned (but the directory itself should be left in place).

### LINKED ISSUE #624: GaudiRun support for different environments
I plan to add an additional `LbLogin.sh` call to the appropriate places in GaudiRun as well as exporting the CMTCONFIG to allow for different architectures of the same code to be made and submitted to different backends.

### LINKED ISSUE #626: DiracFile on Dirac backend with GaudiRun app
The output DiracFile does not work when submitting to Dirac backend.
It is fine for the Local backend and LocalFile submits ok to both Local and Dirac backend.
The error is:
```
ERROR    Error submitting job to Dirac: Traceback (most recent call last):
  File "<stdin>", line 470, in <module>                                                                                                                     
  File "<string>", line 453, in <module>                                                                                                                    
  File "/afs/cern.ch/work/m/masmith/gangadir/workspace/masmith/LocalXML/956/input/dirac-script.py", line 17, in <module>                                    
    j.setOutputData(['test_10ev.root'], outputPath='', outputSE=[])                                                                                         
TypeError: setOutputData() got an unexpected keyword argument 'outputPath'                                                                                  

ERROR   

===                                                                                                                                                         
/afs/cern.ch/work/m/masmith/gangadir/workspace/masmith/LocalXML/956/input/dirac-script.py                                                                   
===                                                                                                                                                         

ERROR   

====                                                                                                                                                        

ERROR   
# dirac job created by ganga                                                                                                                                
parseCommandLine()                                                                                                                                          
from LHCbDIRAC.Interfaces.API.LHCbJob import LHCbJob                                                                                                        
dirac = DiracLHCb()                                                                                                                                         
j = LHCbJob()                                                                                                                                               

# default commands added by ganga                                                                                                                           
j.setName('Brunel_Nov2012_ghostFit_5_cmakeTest__{Ganga_GaudiRun_(956)}')
j.setExecutable('gaudiRun-script.py','','Ganga_GaudiRun.log')
j.setInputSandbox(['LFN:/lhcb/user/m/masmith/GangaInputFile/Job_956/__conf-3d62e484-81ea-4aeb-8d21-35fbd659f099.tgz'])
j.setOutputSandbox(['__postprocesslocations__'])
j.setInputData(['LFN:/lhcb/LHCb/Collision12/BHADRONCOMPLETEEVENT.DST/00021211/0001/00021211_00010037_1.bhadroncompleteevent.dst'])

j.setOutputData(['test_10ev.root'], outputPath='', outputSE=[])


# <-- user settings
j.setCPUTime(172800)

# user settings -->

# diracOpts added by user


# submit the job to dirac
j.setPlatform( 'ANY' )
result = dirac.submit(j)
output(result)
ERROR   
====

ERROR    BackendError: Error submitting job to Dirac: Traceback (most recent call last):
  File "<stdin>", line 470, in <module>
  File "<string>", line 453, in <module>
  File "/afs/cern.ch/work/m/masmith/gangadir/workspace/masmith/LocalXML/956/input/dirac-script.py", line 17, in <module>
    j.setOutputData(['test_10ev.root'], outputPath='', outputSE=[])
TypeError: setOutputData() got an unexpected keyword argument 'outputPath'
 (Dirac backend)
ERROR    JobManagerError: error during submit
ERROR    JobManagerError: error during submit ... reverting job 956 to the new status
```

rob-c:
This is a PR with some updates for GaudiRun.

I've changed the WN code to stream the stdout/stderr from the command directly as I've hit a few occasions where the stream has been lost from record.
(Plus the code feels more professional than an os.system call, although we may want to check return-codes)

I've tested and can switch between gcc48 and gcc49 with no issues on the Localhost and still execute jobs on the grid.

I'd like to get this into 6.1.21 if we've not tagged this for release yet as it puts the GaudiRun in a nice place where I only need to work on Gauss support and some general support for more complex workflows such as with unique data files per subjob as well as MassStorageFile support.

Fixes #621
Fixes #622
Fixes #624
Fixes #626

### DIFF
--- a/python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
+++ b/python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
@@ -2,6 +2,7 @@ from os import rename, path, makedirs, chdir, unlink
 import tempfile
 import time
 import subprocess
+import shutil
 
 from Ganga.Core import ApplicationConfigurationError, ApplicationPrepareError, GangaException
 from Ganga.GPIDev.Adapters.IPrepareApp import IPrepareApp
@@ -75,13 +76,13 @@ class GaudiRun(IPrepareApp):
         'myOpts':       GangaFileItem(defvalue=None, doc='File which contains the extra opts I want to pass to gaudirun.py'),
         'uploadedInput': GangaFileItem(defvalue=None, doc='This stores the input for the job which has been pre-uploaded so that it gets to the WN'),
         'runWithPython':SimpleItem(defvalue=False, doc='Should \'myOpts\' be run as "python myOpts.py data.py" rather than "gaudirun.py myOpts.py data.py"'),
+        'platform' :    SimpleItem(defvalue='x86_64-slc6-gcc49-opt', typelist=[str], doc='Platform the application was built for'),
 
         # Prepared job object
         'is_prepared':  SimpleItem(defvalue=None, strict_sequence=0, visitable=1, copyable=1, hidden=0, typelist=[None, bool, ShareDir], protected=0, comparable=1,
             doc='Location of shared resources. Presence of this attribute implies the application has been prepared.'),
         'hash':         SimpleItem(defvalue=None, typelist=[None, str], hidden=0,
             doc='MD5 hash of the string representation of applications preparable attributes'),
-        'arch':         SimpleItem(defvalue="x86_64-slc6-gcc49-opt", typelist=[str], doc='Arch the application was built for'),
         })
     _category = 'applications'
     _name = 'GaudiRun'
@@ -149,6 +150,8 @@ class GaudiRun(IPrepareApp):
             self.unprepare()
             raise
 
+        shutil.rmtree(path.dirname(this_build_target), ignore_errors=True)
+
         return 1
 
     def configure(self, masterappconfig):
@@ -190,6 +193,9 @@ class GaudiRun(IPrepareApp):
         else:
             raise ApplicationConfigurationError(None, "No Opts File has been specified, please provide one!")
 
+    def getEnvScript(self):
+        return 'export CMTCONFIG=%s; source LbLogin.sh --cmtconfig=%s && ' % (self.platform, self.platform)
+
     def exec_cmd(self, cmd):
         """
         This method executes a command within the namespace of the project. The cmd is placed in a bash script which is executed within the env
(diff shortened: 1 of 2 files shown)


egede:
> (Plus the code feels more professional than an os.system call, although we may want to check return-codes)

@rob-c Does this mean that the code doesn't check for return codes at the moment? Fine for the prototype, but if it is the case, we should file an issue against this straight away to make sure that we do not forget.

rob-c:
@egede No the return code is returned correctly from the command line (or should be as far as I'm aware).
I'm just not 100% certain what is expected in the case of duplicating what the _GaudiPython-like_ functionality but this feels like a secondary task as I've just copied the original implementation

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -149,6 +150,25 @@ def prepare(self, force=False):
             self.unprepare()
             raise
 
+        self.cleanGangaTargetArea(this_build_target)
```
milliams: It looks like `prepare()` is no longer returning a value. I guess you meant `return self.cleanGangaTargetArea(this_build_target)` or to put a `return 1` on the line after?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -398,6 +405,9 @@ def extractAllTarFiles(path):
                 system("tar -jxf %s" % f)
 
 def pythonScript(scriptName):
+    '''
+    This is the actual code in thr GaudiPython-like script
```
milliams: Typo in "thr"

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Just added latter in commit

milliams:
:+1:

PR #637, opened 2016-06-29
Getting rid of bad warning when it's simply not needed

rob-c:
After or if the `GLITE_ENABLED` is set to false in the config there is no need to warn the user about missing files which will/should never be used.
Goes from:
```
INFO     LCG config is not set. Please set [LCG]VirtualOrganisation to your VO (e.g. "gridpp") or disable LCG with [LCG]GLITE_ENABLE=False
WARNING  Configuration of GLITE for LCG:
WARNING  File not found: /afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh
```

```
INFO     LCG config is not set. Please set [LCG]VirtualOrganisation to your VO (e.g. "gridpp") or disable LCG with [LCG]GLITE_ENABLE=False
```

As long as we're not creating a proxy on systems when they start by default just because the tools are there I'm happy to wait until 6.1.22 for this to go in.

### DIFF
--- a/python/Ganga/Utility/GridShell.py
+++ b/python/Ganga/Utility/GridShell.py
@@ -65,6 +65,9 @@ def getShell(force=False):
         logger.warning('[LCG] configuration section not found. Cannot set up a proper grid shell.')
         return None
 
+    if not config['GLITE_ENABLE']:
+        return None
+
     s = None
 
     # 1. check if the GLITE_SETUP is changed by user -> take the user's value as session value


rob-c:
As an aside should that default point to cvmfs now more then afs? I've no idea who uses GLITE but Im under the impression cvmfs is becoming the default?

rob-c:
Any objections to squeezing this in as well before 6.1.21? It's just a small change which shouldn't effect users too much? (I only ask as #645 had to go in)

milliams:
Sure, go for it :+1:

rob-c:
ok, I'll wait for the tests to finish :+1:

PR #645, opened 2016-06-30
Fix for LHCb offline Dirac splitter

rob-c:
I'm keen to get this in (last minute I know) for 6.1.21 as this is a quick fix for splitting of jobs which has been highlighted to us as being problematic (at least for LHCbDirac users).

I'll run some submit tests now and see what I get to check this.

### DIFF
--- a/python/GangaDirac/Lib/Server/DiracCommands.py
+++ b/python/GangaDirac/Lib/Server/DiracCommands.py
@@ -55,7 +55,7 @@ def getMetadata(lfn):
 
 def getReplicas(lfns):
     ''' Return  the locations of the replicas of a given LFN in a dict format, SE: location '''
-    output(dirac.getReplicas(lfns))
+    output(dirac.getReplicas(lfns, active=True, preferDisk = True))
 
 
 def getFile(lfns, destDir=''):
--- a/python/Ganga/Utility/GridShell.py
+++ b/python/Ganga/Utility/GridShell.py
@@ -65,6 +65,9 @@ def getShell(force=False):
         logger.warning('[LCG] configuration section not found. Cannot set up a proper grid shell.')
         return None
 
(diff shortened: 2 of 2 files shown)


rob-c:
And I'll remove #637 from this now

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -55,7 +55,7 @@ def getMetadata(lfn):
 
 def getReplicas(lfns):
     ''' Return  the locations of the replicas of a given LFN in a dict format, SE: location '''
-    output(dirac.getReplicas(lfns))
+    output(dirac.getReplicas(lfns, active=True, preferDisk = True))
```
milliams: There should be no spaces around the ` = `.

milliams:
What do those new parameters do? The [DIRAC docs](http://diracgrid.org/files/docs/CodeDocumentation/Interfaces/API/Dirac.html#DIRAC.Interfaces.API.Dirac.Dirac.getReplicas) have no information about them.

rob-c:
@milliams From lhcb-ganga@cern.ch "We would like this to be really used by everyone because then we can allow the archive SEs for read access without creating problems when splitting. An attempt I made 2 days ago led to "bad" splitting as optimizers couldn't broker correctly the jobs as they ignore tape replicas then there are disk replicas, therefore sometimes the merging grouping relies only on a common tape-archive SE and then no suitable site is found :(( Therefore I went back to banning the archives for reading..."

milliams:
Ok, I'm happy to put this into 6.1.21 then.

rob-c:
OK, I'm about to submit a test job

rob-c:
OK, it appears to work for me, I'll merge now

PR #646, opened 2016-06-30
Removing __construct__ instances and fixing code

rob-c:
Here I have removed the `__construct__` functions in the classes that use them and have defined some sensible `__setattr__` functions which are now correctly called from the ProxyDescriptor.

The advantage of this is 3 fold.

 1. Classes stop calling `__construct__` when being constructed from the proxy. We'd much rather have all proxy-specific changes to an object in `_auto__init__`
 2. `__setattr__` is much nicer and cleaner than the construct list. It means that we know exactly what we're setting when and can adjust the values.
 3. The `localDir` is now always set to be `cwd` from when the class has been constructed or the dirname from a full/relative file path (if no path it's the cwd where the file is expected to be). No more guessing about what the `localDir` (although you can set it independently).

This PR has passed a large(ish) subset of the test suite locally but I'm expecting 1 or 2 corner cases I've not considered yet to come up in testing.

The Proxy `__init__` (`_init`) function is now much cleaner and the export and load of jobs still works correctly.

I'm mainly uploading this for testing (for now) but would like to see it go into 6.1.22 if possible.
There is known missing documentation on this which I'll get back to before committing it.

### DIFF
--- a/python/Ganga/GPIDev/Adapters/IPostProcessor.py
+++ b/python/Ganga/GPIDev/Adapters/IPostProcessor.py
@@ -55,10 +55,9 @@ class MultiPostProcessor(IPostProcessor):
         'process_objects': ComponentItem('postprocessor', defvalue=[], hidden=1, doc='A list of Processors to run', sequence=1)
     })
 
-    def __init__(self):
+    def __init__(self, *args):
         super(MultiPostProcessor, self).__init__()
 
-    def __construct__(self, args):
         if len(args) == 1 or len(args) > 1 and isType(args, (GangaList, list)):
             if isinstance(args, list) or isType(args, GangaList):
                 for process in args:
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -726,8 +726,6 @@ class GangaObject(Node):
         elif len(args) == 1:
             if not isinstance(args[0], type(self)):
                 logger.warning("Performing a copyFrom from: %s to: %s" % (type(args[0]), type(self)))
-                import traceback
-                traceback.print_stack()
             self.copyFrom(args[0])
         else:
             from Ganga.GPIDev.Base.Proxy import TypeMismatchError
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -726,6 +726,8 @@ class GangaObject(Node):
         elif len(args) == 1:
             if not isinstance(args[0], type(self)):
                 logger.warning("Performing a copyFrom from: %s to: %s" % (type(args[0]), type(self)))
+                import traceback
+                traceback.print_stack()
             self.copyFrom(args[0])
         else:
             from Ganga.GPIDev.Base.Proxy import TypeMismatchError
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -297,8 +297,9 @@ valueTypeAllowed = lambda val, valTypeList: _valueTypeAllowed(val, valTypeList, 
 
 class ProxyDataDescriptor(object):
 
-    def __init__(self, name):
+    def __init__(self, name, check_readOnly=True):
         self._name = name
+        self._check_readOnly = check_readOnly
 
     # apply object conversion or if it failes, make the wrapper proxy
     def disguiseComponentObject(self, v):
@@ -570,10 +571,11 @@ class ProxyDataDescriptor(object):
                 raise ReadOnlyObjectError('object %s is read-only and attribute "%s" cannot be modified now' % (repr(obj), attr_name))
             
 
-        # mechanism for locking of preparable attributes
-        if item['preparable']:
-            ## Does not modify val
-            self.__preparable_set__(obj, val, attr_name)
+        if self._check_readOnly:
+            # mechanism for locking of preparable attributes
+            if item['preparable']:
+                ## Does not modify val
+                self.__preparable_set__(obj, val, attr_name)
 
         # if we set is_prepared to None in the GPI, that should effectively
         # unprepare the application
@@ -764,20 +766,31 @@ def GPIProxyClassFactory(name, pluginclass):
             ## FIRST INITALIZE A RAW OBJECT INSTANCE CORRESPONDING TO 'pluginclass'
             ## Object was not passed by construction so need to construct new object for internal use
             if len(args) < len(getargspec(pluginclass.__init__)[0]):
-                clean_args = (stripProxy(arg) for arg in args)
-                instance = pluginclass(*clean_args)
+                if len(args) == 1 and isinstance(args[0], pluginclass):
+                    instance = deepcopy(stripProxy(args[0]))
+                elif len(args) == 1 and hasattr(args[0], '__len__') and not isinstance(args[0], str):
+                    from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
+                    instance = GangaList()
+                    instance.extend(stripProxy(i) for i in args[0])
+                else:
+                    clean_args = (stripProxy(arg) for arg in args)
+                    instance = pluginclass(*clean_args)
             else:
                 instance = pluginclass()
                 ## DOESN'T MAKE SENSE TO KEEP PROXIES HERE AS WE MAY BE PERFORMING A PSEUDO-COPY OP
                 clean_args = (stripProxy(arg) for arg in args)
                 raw_self = stripProxy(self)
-                if len(clean_args) < len(getargspec(pluginclass.__construct__)[0]):
-                    instance.__construct__(*clean_args)
+                if len(args) < len(getargspec(instance.__construct__)[0]):
+                    try:
+                        instance.__construct__(*clean_args)
+                    except TypeError:
+                        # This is bad, we should feel bad and this needs to die in flames
+                        instance.__construct__([])
                 else:
-                    instance.__construct__()
+                    instance.__construct__([])
 
         ## Avoid intercepting any of the setter method associated with the implRef as they could trigger loading from disk
-        setattr(self, implRef, instance)
+        self.__dict__[implRef] = instance
         instance.__dict__[proxyObject] = self
 
         ## Need to avoid any setter methods for GangaObjects
@@ -811,7 +824,7 @@ def GPIProxyClassFactory(name, pluginclass):
             if stripProxy(self)._schema.hasAttribute(k):
                 this_arg = stripProxy(kwds[k])
                 raw_self = stripProxy(self)
-                ProxyDataDescriptor(k).__set__(raw_self, raw_self._attribute_filter__set__(k, this_arg))
+                ProxyDataDescriptor(k, False).__set__(raw_self, raw_self._attribute_filter__set__(k, this_arg))
             else:
                 logger.warning('keyword argument in the %s constructur ignored: %s=%s (not defined in the schema)', name, k, kwds[k])
 
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -16,7 +16,7 @@ from Ganga.Core import GangaAttributeError, ProtectedAttributeError, ReadOnlyObj
 import functools
 import os
 
-from inspect import isclass
+from inspect import isclass, getargspec
 
 import types
 
@@ -623,7 +623,7 @@ class ProxyDataDescriptor(object):
         ## Does not modify val?
         self._check_type(obj, final_val)
 
-        GangaObject.__setattr__(raw_obj, attr_name, final_val)
+        raw_obj.__setattr__(attr_name, final_val)
 
 
 def proxy_wrap(f):
@@ -763,7 +763,18 @@ def GPIProxyClassFactory(name, pluginclass):
         else:
             ## FIRST INITALIZE A RAW OBJECT INSTANCE CORRESPONDING TO 'pluginclass'
             ## Object was not passed by construction so need to construct new object for internal use
-            instance = pluginclass()
+            if len(args) < len(getargspec(pluginclass.__init__)[0]):
+                clean_args = (stripProxy(arg) for arg in args)
+                instance = pluginclass(*clean_args)
+            else:
+                instance = pluginclass()
+                ## DOESN'T MAKE SENSE TO KEEP PROXIES HERE AS WE MAY BE PERFORMING A PSEUDO-COPY OP
+                clean_args = (stripProxy(arg) for arg in args)
+                raw_self = stripProxy(self)
+                if len(clean_args) < len(getargspec(pluginclass.__construct__)[0]):
+                    instance.__construct__(*clean_args)
+                else:
+                    instance.__construct__()
 
         ## Avoid intercepting any of the setter method associated with the implRef as they could trigger loading from disk
         setattr(self, implRef, instance)
@@ -790,19 +801,7 @@ def GPIProxyClassFactory(name, pluginclass):
                 instance.setSchemaAttribute(key, instance._attribute_filter__set__(key, stripProxy(val)))
 
 
-        ## THIRD(?) CONSTRUCT THE OBJECT USING THE ARGUMENTS WHICH HAVE BEEN PASSED
-        ## e.g. Job(application=exe, name='myJob', ...) or myJob2 = Job(myJob1)
-        ## THIS IS PRIMARILY FOR THE 2ND EXAMPLE ABOVE
-
-        ## DOESN'T MAKE SENSE TO KEEP PROXIES HERE AS WE MAY BE PERFORMING A PSEUDO-COPY OP
-        clean_args = [stripProxy(arg) for arg in args]
-        try:
-            stripProxy(self).__construct__(clean_args)
-        except TypeError:
-            stripProxy(self).__construct__([])
-
-
-        ## FOURTH ALLOW FOR APPLICATION AND IS_PREPARED etc TO TRIGGER RELAVENT CODE AND SET THE KEYWORDS FROM THE SCHEMA AGAIN
+        ## THIRD ALLOW FOR APPLICATION AND IS_PREPARED etc TO TRIGGER RELAVENT CODE AND SET THE KEYWORDS FROM THE SCHEMA AGAIN
         ## THIS IS MAINLY FOR THE FIRST EXAMPLE ABOVE
 
         ## THIS CORRECTLY APPLIES A PROXY TO ALL OBJECT ATTRIBUTES OF AN OBJECT CREATED WITHIN THE GPI
@@ -811,51 +810,8 @@ def GPIProxyClassFactory(name, pluginclass):
         for k in kwds:
             if stripProxy(self)._schema.hasAttribute(k):
                 this_arg = stripProxy(kwds[k])
-                if hasattr(this_arg, '_auto__init__'):
-                    this_arg._auto__init__()
-
-                ## Copying this from the __set__ method in the Proxy descriptor
-
-                if k == 'application':
-                    ProxyDataDescriptor.__app_set__(self, this_arg)
-                if k == 'is_prepared':
-                    ProxyDataDescriptor.__prep_set__(self, this_arg)
-
-
                 raw_self = stripProxy(self)
-
-                if isinstance(this_arg, str):
-                    this_arg = stripProxy(runtimeEvalString(raw_self, k, this_arg))
-                    if hasattr(this_arg, '_auto__init__'):
-                        this_arg._auto__init__()
-
-                if isinstance(this_arg, str):
-                    raw_self.setSchemaAttribute(k, raw_self._attribute_filter__set__(k, this_arg))
-                    continue
-                else:
-                    item = pluginclass._schema.getItem(k)
-
-                    # unwrap proxy
-                    if item.isA(ComponentItem):
-                        from .Filters import allComponentFilters
-                        cfilter = allComponentFilters[item['category']]
-                        stripper = lambda v: stripComponentObject(v, cfilter, item)
-                    else:
-                        stripper = None
-
-                    if item['sequence']:
-                        this_arg = ProxyDataDescriptor.__sequence_set__(stripper, raw_self, this_arg, k)
-                    else:
-                        if stripper is not None:
-                            this_arg = stripper(this_arg)
-                    # apply attribute filter to component items
-                    if item.isA(ComponentItem):
-                        this_arg = ProxyDataDescriptor._stripAttribute(raw_self, this_arg, k)
-
-                    if hasattr(this_arg, '_auto__init__'):
-                        this_arg._auto__init__()
-
-                    raw_self.setSchemaAttribute(k, raw_self._attribute_filter__set__(k, this_arg))
+                ProxyDataDescriptor(k).__set__(raw_self, raw_self._attribute_filter__set__(k, this_arg))
             else:
                 logger.warning('keyword argument in the %s constructur ignored: %s=%s (not defined in the schema)', name, k, kwds[k])
 
--- a/python/Ganga/GPIDev/Lib/File/File.py
+++ b/python/Ganga/GPIDev/Lib/File/File.py
@@ -64,21 +64,11 @@ class File(GangaObject):
         if not subdir is None:
             self.subdir = subdir
 
-    def __construct__(self, args):
-        if len(args) == 1 and isinstance(args[0], str):
-            v = args[0]
-            import os.path
-            expanded = expandfilename(v)
-            # if it is not already an absolute filename
-            if not urlprefix.match(expanded):
-                if os.path.exists(os.path.abspath(expanded)):
-                    self.name = os.path.abspath(expanded)
-                else:
-                    self.name = v
-            else:  # bugfix #20545
-                self.name = expanded
-        else:
-            super(File, self).__construct__(args)
+    def __setattr__(self, attr, value):
+        actual_value = value
+        if attr == "name":
+            actual_value = expandfilename(value)
+        super(File, self).__setattr__(attr, value)
 
     def _attribute_filter__set__(self, attribName, attribValue):
         if attribName is 'name':
@@ -126,7 +116,7 @@ def string_file_shortcut_file(v, item):
     if isinstance(v, str):
         # use proxy class to enable all user conversions on the value itself
         # but return the implementation object (not proxy)
-        return stripProxy(File._proxyClass(v))
+        return File(v)
     return None
 
 allComponentFilters['files'] = string_file_shortcut_file
@@ -255,7 +245,7 @@ def string_sharedfile_shortcut(v, item):
     if isinstance(v, str):
         # use proxy class to enable all user conversions on the value itself
         # but return the implementation object (not proxy)
-        return stripProxy(ShareDir._proxyClass(v))
+        return ShareDir(v)
     return None
 
 allComponentFilters['shareddirs'] = string_sharedfile_shortcut
--- a/python/Ganga/GPIDev/Lib/File/LocalFile.py
+++ b/python/Ganga/GPIDev/Lib/File/LocalFile.py
@@ -5,6 +5,12 @@ from __future__ import absolute_import
 # $Id: LocalFile.py,v 0.1 2011-09-29 15:40:00 idzhunov Exp $
 ##########################################################################
 
+import errno
+import re
+import os
+import os.path
+import copy
+
 from Ganga.GPIDev.Schema import Schema, Version, SimpleItem, ComponentItem
 
 from Ganga.GPIDev.Adapters.IGangaFile import IGangaFile
@@ -13,14 +19,9 @@ from Ganga.GPIDev.Base.Proxy import GPIProxyObjectFactory
 
 from Ganga.GPIDev.Lib.File.File import File
 from Ganga.GPIDev.Lib.File import FileBuffer
+from Ganga.Utility.files import expandfilename
 
 import Ganga.Utility.logging
-
-import errno
-import re
-import os
-import os.path
-import copy
 
 logger = Ganga.Utility.logging.getLogger()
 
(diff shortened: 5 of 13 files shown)


rob-c:
OK this has highlited one annoyance which apparently is already being tested for:

`LocalFile('somePath/someName')`

To me this should _always_ mean: `LocalFile(namePattern='someName', localDir='somePath')`

This is however being used such that when it's in the inputfile attribute of a job the above is true but when it's used i the outputFiles of a job its explicitly treated as: `LocalFile(namePattern='somePath/nameName')`.
This confuses things as to me if you ever create the file the user may pass: `LocalFile('someFileName.txt')` or even just `'someFileName.txt'`. These are only valid if the localDir of the file object is defined to be the cwd.
There is a hack in place to check/preserve this already but I'd rather abandon it all together and do things properly.

Is there any strong objection to this PR intentionally removing the behaviour of the following:
```
j.outputfiles=[LocalFile('somePath/someName')]
...
# This currently then works:
j.peek('somePath/someName')
# I argue _only_ the following should work
j.peek('someName')
```

I feel that there should be a dedicated parameter governing the naming of the output LocalFile for the LocalFile akin to the `outputfilenameformat` in `MassStorageFile`.
(I've looked into porting this functionality but it's beyond the scope of this PR as it touches even more I'm trying not to touch.)

Some of the work I'm doing in several places relies on the assumption that the `namePattern` is always the basename of the path passed to LocalFile.
It is extremely difficult to determine if we should/shouldn't evaluate the following:
```
LocalFile('someFile.txt') # localDir = cwd
LocalFile('./somePath/to/someFile.txt') # localDir = /absolute/of/somePath/to/someFile.txt
LocalFile('./notAPath/outputName.txt') # should localDir = ./notAPath ? should I warn the user the file doesn't exist?
```

__Every__ other IGangaFile returns the namePattern as the basename. This allows for machinery to be written which doesn't treat the LocalFile as a special case and this has been an issue against LocalFile in my mind for a while.


I'm happy to look at putting this feature back in properly but I'd rather avoid doing this in this PR unless there is a strong call for it to be preserved?

egede:
@rob-c I find it hard to understand what you write here ... Do you say that we allow the creation of filenames which inside them contain a slash? If yes, that indeed seems like a bad idea.

rob-c:
No the test I'm referring to is: https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/test/GPI/Mergers/TestStructure.py#L68

This explicitly requires that `j.outputfiles=[LocalFile('somePath/someFile')]` always generates the directory `somePath` and the file in that directory `someFile` in the output workspace.

This also means that any time I do `file.namePattern` at the moment I get `'somePath/someFile'` rather than just `'someFile'`.
The presence of this `/` within the name of a file gets in the way imo.

At the very least I feel we should correctly construct this such that the namePattern is `someFile` and the localDir is `somePath` and in the machinery in Ganga construct the localDir folder in the output workspace rather than what is currently done.

egede:
@rob-c Thanks for the explanation. The short-cut to give dir and name at the same time is quite nice, but not a requirement. I agree that we should always persist this in a uniform manner.

rob-c:
In the new mechanics the LocalFile created has a better namePattern and localDir which make sense and is consistent it's just a little tricky deciding what to do with the folders.

I may be able to add some code into the post-processing so that the file is put back into `localDir/namePattern` for the LocalFile for now and we can look to expand this in the future but if it gets too unstable or complex I may move away from it again.
(This may or may not fix the test in question)

rob-c:
OK, so moving a few things around. I've now changed the `LocalFile` to use the same machinery as the rest of the files for post-processing a file. This gets rid of one headache I've had for a while.
After I introduced this it looks like the machinery automatically adds `localDir` to the path of output files correctly :D

I'll wait and see if this adversely effects anything but this means that the LocalFile should (as a result of this) match much better the expected behavior based upon the other filetypes.
I'll have to do some testing on other backends other than Local but this is a good step forward.

rob-c:
OK I think I'm ready to ask for review for including this in 6.1.23.

This PR removes the `__construct__` framework whilst replacing all calls to `__construct__` with better function calls as well as maintaining the previous behaviour of the `IGangaFile` objects through the use of `__setattr__` which is now called in the correct way through the Proxy layer.

I've updated the top description to try and reflect the nature of the changes in the PR but mainly this is just removing a lot of dead/bad code and this is reflected in the number of lines removed.

rob-c:
OK, once I've resolved a few outstanding issues with the `DiracFile` I think I'm going to push for inclusion of this as it fixes some problems I'm having with `GaudiRun` development.

milliams:
There seems to be some unrelated documentation changes mixed in with this.

python/Ganga/GPIDev/Adapters/IPostProcessor.py
```diff
@@ -71,8 +70,8 @@ def __construct__(self, args):
 
             if hasattr(self.process_objects, 'order'):
                 self.process_objects = sorted(self.process_objects, key=lambda process: process.order)
-        else:
-            super(MultiPostProcessor, self).__construct__(args)
+        elif args and len(args) != 0:
```
milliams: if `args` evaluates to `True` then `len(args)` will *never* be `0`.

python/Ganga/GPIDev/Adapters/IPostProcessor.py
```diff
@@ -55,10 +55,9 @@ class MultiPostProcessor(IPostProcessor):
         'process_objects': ComponentItem('postprocessor', defvalue=[], hidden=1, doc='A list of Processors to run', sequence=1)
     })
 
-    def __init__(self):
+    def __init__(self, *args):
         super(MultiPostProcessor, self).__init__()
 
-    def __construct__(self, args):
         if len(args) == 1 or len(args) > 1 and isType(args, (GangaList, list)):
```
milliams: `args` will always be a `tuple` so this code will never run. Did this not show up in your local testing of this code?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -152,16 +152,17 @@ def _getRoot(self, cond=None):
...
-                if cond(root):
+                if cond(obj):
                     break
-            obj = obj._getParent()
-        return root
+            if obj._getParent() is not None:
+                obj = obj._getParent()
+            else:
+                break
+        return obj
```
milliams: This doesn't seem to be related to the removal of the `__construct__` feature?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -471,10 +494,11 @@ def __set__(self, obj, val):
         new_value = Descriptor.cleanValue(obj, val, _set_name)
 
         obj.setSchemaAttribute(_set_name, new_value)
+
         obj._setDirty()
 
-        if isinstance(val, Node):
-            val._setDirty()
+        # This is currently disabled due to bad interaction with the GangaDirac tests
+        #Descriptor.check_inheritance(obj, _set_name)
```
milliams: We shouldn't be putting in requests to merge code which is commented out like this.

milliams:
I've started reviewing this but there is a lost of seemingly unrelated changes mixed in with the removal of the `__construct__` methods. They'll need to be teased apart before I do a proper review.

python/Ganga/GPIDev/Base/Objects.py
rob-c: Minor change I can revert if needed

python/Ganga/GPIDev/Adapters/IPostProcessor.py
rob-c: This didn't expose any problems and frankly I was just looking to remove the `__construct__` method which shouldn't be called. I can fixup this code but I'm now worried why nothing has broken here in testing?

python/Ganga/GPIDev/Adapters/IPostProcessor.py
rob-c: Actually a typo, will remove

python/Ganga/GPIDev/Base/Objects.py
rob-c: I was getting errors due to non parented objects shown in this PR which exposed a horrible screwup elsewhere in the codebase. I can revert this but it's a nice realtime test to enable at some point so I'll drop it to another PR if it's felt it's not needed.

rob-c:
These were 95%+ needed to make the tests pass again. This caused/(exposed?) some horrible bugs to occur so I don't want to disentangle things which don't offer much and probably rely on the changes in this PR.
As for doc changes in the IGangaFile this is entirely an exercise in getting the IGangaFile interface to actually work, I'm hitting problems in GaudiRun and I'd rather fix it than add another set of checks to work around bad behavior.
The LocalFile before this PR is too much of a special case and it really should be using the same mechanism as the other file types which this enforces. I only had to touch this because it broke so horribly once I changed it to not have a construct method.

python/Ganga/GPIDev/Adapters/IPostProcessor.py
rob-c: Looking at this in more detail I don't see how this method can now be called as a user shouldn't be constructing a `MultiPostProcessor` from the GPI. I don't know if we do in object loading at all?
I'm going to clean it up to always expect a tuple and leave the rest of the code for now.

rob-c:
OK, only things left are the removal of the construct calls abd changes required to fix the tests that broke.
If there is anything here anyone feels shouldn't be here please let me know as afar as I'm concerned this is looking to merge as soon as I confirm I've not just broken anything

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -1121,10 +1094,15 @@ def getJobObject(self):
...
-        return r
+        if self._getParent():
+            r = self._getRoot(cond=lambda o: isinstance(o, Job))
+            if not isinstance(r, Job):
+                raise AssertionError('No Job associated with object instead root=\'%s\' for \'%s\'' % (repr(self), type(r)))
+            return r
+        else:
+            if isinstance(self, Job):
+                return self
+        raise AssertionError('No Parent associated with object instead root=\'%s\' for \'%s\'' % (repr(self), str(None)))
```
milliams: What issue is this change fixing?

Also, why on earth is this raising an `AssertionError`? There is no assertion being raised here. It should be a `RuntimeError` if anything.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -297,8 +297,9 @@ def getImpl(v):
 
 class ProxyDataDescriptor(object):
 
-    def __init__(self, name):
+    def __init__(self, name, check_readOnly=True):
```
milliams: What does this new parameter do? What has changed such that is is needed? It should also be called `check_read_only` for consistency (and PEP8).

python/Ganga/GPIDev/Base/Objects.py
rob-c: The fact that it's not clear whether you don't have a parent associated with an object or whether that parent is a Job. (I was playing with objects in a Box repo and this confused/bugged me as to why the error made no sense when I was testing something). Having no parent and having no Job object should look different to each other imo. I'll move this into another PR in this case.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -623,7 +625,7 @@ def __set__(self, obj, _val):
         ## Does not modify val?
         self._check_type(obj, final_val)
 
-        GangaObject.__setattr__(raw_obj, attr_name, final_val)
+        raw_obj.__setattr__(attr_name, final_val)
```
milliams: If you're making this change then perhaps it should go all the way to:
```python
setattr(raw_obj, attr_name, final_val)
```
?

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -763,10 +765,19 @@ def _init(self, *args, **kwds):
         else:
             ## FIRST INITALIZE A RAW OBJECT INSTANCE CORRESPONDING TO 'pluginclass'
             ## Object was not passed by construction so need to construct new object for internal use
-            instance = pluginclass()
+            # Case 1 j = Job(myExistingJob)            # We want to perform a deepcopy
+            if len(args) == 1 and isinstance(args[0], pluginclass):
+                instance = deepcopy(stripProxy(args[0]))
+            # Case 2 file_ = LocalFile('myFile.txt')   # We need to pass the (stripped) arguments to the constructor only if the 
+            elif len(args) < len(getargspec(pluginclass.__init__)[0]):
```
milliams: Should this be `<=`? Or is this to avoid the `self` clash? If so, a comment to that effect would be useful.

However, really, I think that the best approach would be to simply pass the arguments straight through and let the exception be propagated.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -763,10 +765,19 @@ def _init(self, *args, **kwds):
...
+                clean_args = (stripProxy(arg) for arg in args)
+                instance = pluginclass(*clean_args)
+            else:
+                logger.warning("Cannot use arguments: '%s' for constructing class type '%s'. Ignoring." % (args, getName(pluginclass)))
+                instance = pluginclass()
 
         ## Avoid intercepting any of the setter method associated with the implRef as they could trigger loading from disk
-        setattr(self, implRef, instance)
+        self.__dict__[implRef] = instance
```
milliams: Why circumvent the setter machinery?

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -811,51 +810,8 @@ def _init(self, *args, **kwds):
...
-                            this_arg = stripper(this_arg)
-                    # apply attribute filter to component items
-                    if item.isA(ComponentItem):
-                        this_arg = ProxyDataDescriptor._stripAttribute(raw_self, this_arg, k)
-
-                    if hasattr(this_arg, '_auto__init__'):
-                        this_arg._auto__init__()
-
-                    raw_self.setSchemaAttribute(k, raw_self._attribute_filter__set__(k, this_arg))
+                ProxyDataDescriptor(k, False).__set__(raw_self, raw_self._attribute_filter__set__(k, this_arg))
```
milliams: This is a very strange thing to do. Descriptors exist for a purpose and they are intrinsically linked with class creation. Creating one simply to call `__set__` on it is suspicious and signals that there's something wrong.

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -64,21 +64,11 @@ def __init__(self, name=None, subdir=os.curdir):
...
-                    self.name = v
-            else:  # bugfix #20545
-                self.name = expanded
-        else:
-            super(File, self).__construct__(args)
+    def __setattr__(self, attr, value):
+        actual_value = value
+        if attr == "name":
+            actual_value = expandfilename(value)
+        super(File, self).__setattr__(attr, value)
```
milliams: Is this function ever called? I'd think that the descriptor would always take precedence but I assume I'm missing something as you wouldn't have added this function if you hadn't tested it and seen it do something.

python/Ganga/GPIDev/Lib/File/LocalFile.py
```diff
@@ -5,23 +5,23 @@
 # $Id: LocalFile.py,v 0.1 2011-09-29 15:40:00 idzhunov Exp $
 ##########################################################################
 
+import errno
+import re
+import os
+from os import path
```
milliams: This import isn't needed as you can simply access them via `os.path.basename` etc.

python/Ganga/GPIDev/Lib/File/LocalFile.py
```diff
@@ -242,6 +216,25 @@ def remove(self):
 
         return
 
+    def put(self):
+	"""
+        For the LocalFile it makes no sense to do anything here
```
milliams: This comment does not match the fact that you *are* doing something in this function. Also, this sort of thing is better put in a comment that in a docstring.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: This removes some 100(ish) lines of code which duplicate the logic in the setter method of the class. The logic in parsing the command line in the constructor should be the same as the logic on assignment. If it's not then we have a mess.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Because it will throw an exception

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -362,7 +345,10 @@ def __ge__(self, obj_list):
         return self._list.__ge__(self.__getListToCompare(obj_list))
 
     def __getitem__(self, index):
-        return self._list.__getitem__(index)
+        returnable = self._list.__getitem__(index)
+        if isinstance(returnable, Node):
+            returnable._setParent(self._getParent())
+        return returnable
```
milliams: Why do we need to set the parent on *get* as well as *set*?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -424,6 +374,8 @@ def _attribute_filter__get__(self, name):
 
         if name == 'outputfiles':
 
+            # THIS IS IMMENSELY FUGLY AND STUPID WHY FOR THE LOVE OF GOD WHY!
```
milliams: Not a useful comment.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: You're probably correct about letting any error be propagated but I'm wanting to protect against older broken classes for now and just warn the user which is less strict but relatively safe.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Probably, I'm not sure how the function maps but I suspect that should work I'll try the change

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -671,13 +625,19 @@ def postprocessoutput(self, outputfiles, outputdir):
                     for currentFile in glob.glob(os.path.join(outputdir, outputfile.namePattern)):
                         os.system("gzip %s" % currentFile)
 
+
             backend_output_postprocess = self.getBackendOutputPostprocessDict()
             if backendClass in backend_output_postprocess:
                 if outputfileClass in backend_output_postprocess[backendClass]:
-                    if backend_output_postprocess[backendClass][outputfileClass] == 'client':
+                    if backend_output_postprocess[backendClass][outputfileClass].lower() == 'client':
```
milliams: Are there cases where this is not already lowercase?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -671,13 +625,19 @@ def postprocessoutput(self, outputfiles, outputdir):
...
                         logger.info("Job %s Putting File %s: %s" % (self.getFQID('.'), getName(outputfile), outputfile.namePattern))
                         outputfile.put()
                         for f in glob.glob(os.path.join(self.outputdir, outputfile.namePattern)):
+                            # FIXME should this be controlled via some per filetype config?
```
milliams: No, it should most likely be solved using polymorphism rather than an explicit type check.

python/Ganga/GPIDev/Lib/File/File.py
rob-c: There's a bug that I think line 71 should read actual_value rather than value (for automatic filename expansion). I think this just passes through in it's current form without doing anything which seems surprising but behind the scenes in Ganga we try to use `File` with absolute filenames and avoid exposing it to the GPI due to now having LocalFile.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: I'm assuming users may type something crazy here in the config

python/Ganga/Utility/files.py
```diff
@@ -12,6 +12,8 @@
 
 def expandfilename(filename, force=False):
     """expand a path or filename in a standard way so that it may contain ~ and ${VAR} strings"""
+    if not filename:
+        return filename
```
milliams: It probably makes more sense to raise an exception that to silently return an empty string.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Yes but where, where do we make sure all other files remove the file apart from the LocalFile, some new function or do we now place the onus on all other filetypes to clean up after themselves. The latter seems like a policy change which needs discussion, whether it's in this PR or later I don't mind.

python/Ganga/test/GPI/Parser/TestSelect.py
```diff
@@ -50,7 +50,7 @@ def test_c_SelectSJTests(self):
         """ Is is a bird is it a plane... no it's a test for selecting subjobs now"""
         from Ganga.GPI import jobs, Job, ArgSplitter
         
-        j=Job(splitter=ArgSplitter(args=job_names))
+        j=Job(splitter=ArgSplitter(args=[[_] for _ in job_names]))
```
milliams: Is this essentially
```python
j = Job(splitter=ArgSplitter(args=list(job_names)))
```
?

python/Ganga/Utility/files.py
rob-c: Then we need to add a __lot__ of protection against this everywhere else and there are mehods passing None vs '' which also causes problems.

python/Ganga/GPIDev/Lib/GangaList/GangaList.py
rob-c: I can remove this, typo. was left in when tracking down some madness due to missing parents

python/Ganga/GPIDev/Lib/File/LocalFile.py
rob-c: Will update docstr

python/Ganga/GPIDev/Lib/File/LocalFile.py
rob-c: Yes but it reduces the amount of something.somethingelse.something which gets annoying imo

python/Ganga/test/GPI/Parser/TestSelect.py
rob-c: I'm not sure what you mean, comment or change the line?

rob-c:
OK, I decided to add a ```cleanUpClient``` method to the IGangaFile which handles the cleaning up of files from the client workspace after they've been put.

Aside from adding/changing any more docs are there any problems with looking to merge this tomorrow?

rob-c:
OK, I've no idea why Travis is having trouble but I think I'll look back at this tomorrow

rob-c:
I'm keen to get this branch in asap so I can start working again on GaudiRun this was fixing some things which was really broken from the perspective of writing a new app.
Are there any additional changes I need to make to this PR?

milliams:
Yes, it still needs review. I'm particularly cautious about the use of `__setattr__` and how it conflicts with descriptors. I'm going to have to check out the branch locally and run through some of by hand.

rob-c:
OK, well I'm going to continue working assuming that this in some form goes in no matter how/where/why we hook some `__setattr__` like method I think there is an strong need for one on a per class level, even if the method is named something else or called from elsewhere.

rob-c:
OK I can't even run some of my example jobs anymore due to the bug in the prompt parser that this fixes... I'll make some modifications in my code and work around this for now but this PR is high on my priority list.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -811,51 +820,10 @@ def _init(self, *args, **kwds):
...
-                    if item.isA(ComponentItem):
-                        this_arg = ProxyDataDescriptor._stripAttribute(raw_self, this_arg, k)
-
-                    if hasattr(this_arg, '_auto__init__'):
-                        this_arg._auto__init__()
-
-                    raw_self.setSchemaAttribute(k, raw_self._attribute_filter__set__(k, this_arg))
+                # This calls the same logic when assigning a named attribute as when we're assigning it to the object
+                # There is logic here which we 'could' duplicate but it is over 100 lines of code which then is duplicating funtionality written elsewhere
+                ProxyDataDescriptor(k, False).__set__(raw_self, raw_self._attribute_filter__set__(k, this_arg))
```
milliams: In what cases are we creating a new proxy object which has keyword arguments which is also wrapping an object which is set to read-only? I can't think of a case where we need to circumvent the read-only flag on an object when creating it. Pretty much the only read-only objects in Ganga are Jobs and Tasks which have started.

milliams:
What 'bug in the prompt parser' does this fix? I haven't reports of anything like that before.

rob-c:
@milliams The init parser code doesn't fully follow the path of the code in the proxy descriptor set method and in addition the base `__setattr__` called in the setter isn't the overloaded one, it's explicitly the baseclass.
This means auto-expanding a string from say `job.application.some_val = '$HOME/$SOMEDIR/file.py'` doesn't auto-expand the string fully when the code to do this is in the application.
I'm working around this by fully expanding all of my vars in my test but my 20 char script is now over 100 chars long and painful to read

python/Ganga/GPIDev/Base/Proxy.py
rob-c: Loading an object from disk pushes the object through the parser in this way. Also I would argue it's a passable (albeit strange) thing to do something like:
`j=Job(read_only_attr='someConstructedValue')`
or `app = myApplication(someReadOnlyVal='valOnConstructionAsUserKnowsBest')`.

This means we can construct a prepared application which shares the same prepared state as another one only at construction and then the application is in a read-only state after that and read-only attributes can't be modified. (Of course if a user modifies the object to be insane via construction then there's little we can do as we should probably respect that decision).

milliams: There's no such thing as read-only schema attributes, only read-only objects.

It looks like the real problem is in `export`ing and `load`ing jobs in the GPI. For example, a job which has been creaded and submitted and is then `export(j)`ed will make a file like:
```python
Job (
   comment = '',
   parallel_submit = False,
   do_auto_resubmit = False,
   name = '',
   inputfiles =[],
   backend =   Local (
     nice = 0
   ),
   application =   Executable (
     exe = 'echo',
     hash = 'cfa689697ba25ae38f7a47b232ea3e7c',
     env = [],
     args = ['Hello World'],
     is_prepared = ShareDir(name='conf-47de7c78-e402-4c59-af7f-29138635fcfa',subdir='.')
   ),
   inputdata =None,
   outputfiles =[],
   postprocessors =   MultiPostProcessor (
   ),
   info =   JobInfo (
     monitor =None
   ),
   splitter =None,
   inputsandbox =[]
 )
```

The problem come when Python is parsing the
```python
Executable (
  exe = 'echo',
  hash = 'cfa689697ba25ae38f7a47b232ea3e7c',
  env = [],
  args = ['Hello World'],
  is_prepared = ShareDir(name='conf-47de7c78-e402-4c59-af7f-29138635fcfa',subdir='.')
)
```
part. Since the order in which the arguments are processed is not fixed, the `is_prepared` attribute is being assigned before e.g. the `exe` attribute. This means that Ganga raises an error saying "This object is in the prepared state so you cannot change it".

The underlying problem here is that we want to ignore the read-only state of an object while it is being constructed and only start enforcing it once its state is fully defined. I'm working on a patch to this PR which will fix this is a slightly different way.

rob-c: *replace schema with attribute object, I was going for clarity over being exactingly true

If there is a better way great, but we __still__ need to call the setter method here in order for string expansion to work in a sane way.

rob-c: Is this PR going to be a fix for this in the construction of objects or a large PR which touches a lot of the proxy?

milliams: It will be a small PR which factors out the common code from the `ProxyDataDescriptor` so the relevant code can be used by `_init()`.

rob-c: This is everything aside from the read only statement isn't it? As far as I can tell this just adds another function which is called by both the `__set__` and `_init` methods or is there other machinery not applicable to the initialization of an object?

milliams: I've ended up factoring out almost the entirety of `__set__` into a new `@staticmethod` called `_process_set_value()` which is called by both `__set__` and `_init`. The difference compared to your code is that instead of the `checkReadOnly` flag being on the `ProxyDataDescriptor` object at initialisation and having to create a temporary object, the `_init` can call this `_process_set_value` method with `check_read_only=False`. I'll make this a PR against this PR for your review once the tests have completed locally.

python/Ganga/GPIDev/Lib/File/LocalFile.py
```diff
@@ -39,7 +39,7 @@ class LocalFile(IGangaFile):
                                      })
     _category = 'gangafiles'
     _name = "LocalFile"
-    _exportmethods = ["location", "remove", "accessURL", "processOutputWildcardMatches", "hasMatchedFiles"]
+    _exportmethods = ["location", "remove", "accessURL"]
```
milliams: I think I agree with this removal but is it safe? Are we sure that people aren't using these methods?

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I would have argued the cost of making this object is surprisingly cheap but if we're saved from that then fantastic. OK, feel free to merge it against this PR when ready if there aren't any problems

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -671,34 +622,26 @@ def postprocessoutput(self, outputfiles, outputdir):
...
-                            outputfile.setLocation()
-                    else:
                         try:
                             outputfile.setLocation()
                         except Exception as err:
                             logger.error("Error: %s" % err)
 
-            if outputfileClass == 'LocalFile':
-                outputfile.processOutputWildcardMatches()
+                    if backend_output_postprocess[backendClass][outputfileClass].lower() == 'client':
```
milliams: I still don't think we want to be doing this `lower()` here.

python/Ganga/test/GPI/Parser/TestSelect.py
milliams: I guess I'm asking why:
```python
j = Job(splitter=ArgSplitter(args=job_names))
```
isn't sufficient?

python/Ganga/GPIDev/Lib/File/LocalFile.py
rob-c: `processOutputWildcardMatches` was exclusively only used in the `Job.py` in favor of `setLocation` which wasn't defined for this class (and probably should have been).
I can put the `processOutputWildcardMatches` method back and add a warning that this shouldn't be used and will likely be removed soon. As for `hasMatchedFiles` I can add it back in, I just assumed this was only an internal thing but if people want access to it then there's no particular problem that I can see.

python/Ganga/GPIDev/Base/Proxy.py
milliams: Indeed it is cheap to construct, what's expensive is the time needed to understand why descriptor object are being created on-the-fly, outside of a class, and the maintenance effort of a more complicated `ProxyDataDescriptor` class.

python/Ganga/test/GPI/Parser/TestSelect.py
rob-c: The args are expected to be a list of lists not a list of strings from what I can tell. I think a list of strings would meet most of the requirements but I hit a problem at least once that a string (whilst looking like a list) isn't a list.

milliams: Ahh, yes, you're quite right.

python/Ganga/GPIDev/Lib/File/LocalFile.py
milliams: No, I think it's ok overall. They *should* be private implementation details but for some reason (the commit message for 61666d2dd6acd4edf5374bb9985cc161f5c971d3 doesn't give any justification) they were exposed as public API. The problem with removing functions from the public API is anyone's scripts using them will now break. In reality I doubt that anyone will be using them so they should be fine to remove.

rob-c:
OK, there's been some minor cleanup work in the `Proxy.py`. Whilst not strongly related to this PR these are mostly removing multiple calls to `stripProxy` on the same object. Hopefully the tests will pass without problem soon. Are there any outstanding parts of this PR I should look at or is it feasible to merge this by the end of today?

milliams:
I think I'm happy with this PR in its current state. Thanks for the hard work on this :+1:

PR #647, opened 2016-07-01
Fixing shutdown badness

rob-c:
This PR stops ganga from registering the shutdown method for every single GPI unittest.

I noticed as the same message appeared a few hundred times upon exiting which got me slightly concerned about a few bits.

Running the tests with `--capture=no` revealed this so I'm wondering if it might be good to use this on some nightlies maybe?


python/Ganga/Core/InternalServices/ShutdownManager.py
```diff
@@ -21,7 +23,17 @@
 
 # Globals
 logger = getLogger()
+_has_registered = False
```
drmarkwslater: This could be a static function variable rather than a global. Otherwise 👍

rob-c:
Any objections to merging?

PR #650, opened 2016-07-04
Fix broken import in WN script

drmarkwslater:
It seems that the merge of #493 didn't make it into develop from master after the release so I'm just re-applying this fix directly.

### DIFF
--- a/python/Ganga/GPIDev/Lib/File/OutputFileManager.py
+++ b/python/Ganga/GPIDev/Lib/File/OutputFileManager.py
@@ -162,7 +162,7 @@ def getWNCodeForOutputSandbox(job, files, jobid):
                 patternsToSandbox.append(outputFile.namePattern)
 
     insertScript = """\n
-from Ganga.Utility.files import recursive_copy
+from files import recursive_copy
 import glob
 import os
 


rob-c:
@drmarkwslater I'm assuming this is just because this PR was against master? I've checked 1 or 2 random PR from during the release and they appear to have been merged correctly so I'm assuming we don't need to be too concerned by this?

Is the master branch supposed to remain read-only or are we not correctly merging it into develop as part of the release? I've always been confused about it's purpose on our project anyway if I'm honest.

drmarkwslater:
Indeed - it's just because it was against `master` (my bad - I didn't check this when I merged the PR from @mjmottram). The other PRs will all be fine 😄

rob-c:
Should we avoid future PR against the master branch then to avoid this happening again or should we integrate it with part of the release process to sync it with develop?

drmarkwslater:
I'd go for the former but I don't know if this can be done. I'll let @milliams answer that one :)

rob-c:
In the case if the former is there any automated way of avoiding PRs ending up against the master branch. e.g. can we restrict write access to 1 branch on github to another user/group at all?

PR #651, opened 2016-07-04
Fixing bug in old_tests

rob-c:
This bug was highlighted by looking through the old_tests that were run against 6.1.21.

At what point in the release process are these tests run? Was there an email saying where the test results are located? I think I may have missed that email as I can't find one in my inbox.

As it stands this looks like a bug against LHCb Tasks so I'll try and see if I can run some simple task examples for LHCb and if not I'll push out the release with a caveat for tasks users. (I'm hoping to avoid this)

### DIFF
--- a/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
+++ b/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
@@ -71,7 +71,7 @@ class LHCbDataset(GangaDataset):
                 return isinstance(_file, IGangaFile)
             areFiles = all([isFileTest(f) for f in files._list])
             if areFiles:
-                self.file._list.extend(files._list)
+                self.files._list.extend(files._list)
                 process_files = False
         elif isinstance(files, LHCbDataset):
             self.files._list.extend(files.files._list)


rob-c:
Are these tests supposed to be passing? Either way I imagine this particular bug to be an annoyance. I'm going to try and find an archived LHCbTask .py and try to run it as the tests are struggling.

rob-c:
After some exploring this doesn't appear to be a 'major' bug but I'm not sure if the code in the Tasks example is representative of the way that some people will use the Tasks system.
I'll ask for the release 6.1.21 to LHCb later today and I think we should be ok with this issue in the release given the many other things it's addressing.

rob-c:
Any objections to merging this and pushing for 6.1.22? @milliams Are you able to make the release?

drmarkwslater:
@rob-c Matt is on holiday this week so won't be able to do the release until next week. Are you able to yourself?

rob-c:
@drmarkwslater I've been looking into this, (emailed the dev list) This should be possible for me to do looking at things. I don't think I'll re-run the old tests as historically we haven't for bugfix style releases.

PR #663, opened 2016-07-07
GaudiRun app Input dir fix

mesmith75:
I have added in a check in the local host run time handler for the GaudiRun app when taking the objects from the shared dir and putting them in the tar for the input sandbox tar. If there is a tar file it is ignored to save having the cmake tar in both the shared dir and the input.


rob-c:
@mesmith75 Can we not just remove https://github.com/ganga-devs/ganga/pull/663/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R191 and the following checks? I don't think anything is required in the RTHandler which isn't done in the machinery elsewhere.
(We may be able to support folders in the shared dir too! I'm just not aware if we do, it may crash elsewhere but I'm not using them in the GaudiRun app intentionally then.)

It used to be that we couldn't handle non-tarred files in the sandbox but I fixed this a while ago withe mimetype check before we try to extract files on the WN.

mesmith75:
@rob-c I don't think so. This part also takes the options file from the shared directory.

rob-c:
Yes but is this auto copied because it's in the shared folder or are we having to explicitly request this be sent to the worker node via some mechanism?

mesmith75:
Hi Rob, I'm not sure I follow.
As far as I can make out the options are copied here because they are in the shared directory. There doesn't seem to be anything else going on to send it to the WN.

rob-c:
@mesmith75 https://github.com/ganga-devs/ganga/pull/663/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R99 doesn't modify the contents of the shared directory so I think that what is happening is this method is (mistakenly) also adding the same file to the inputsandbox of the job which is creating duplicate files in the input sandbox (largest of which is the cmake target).
If the file is not added to the input sandbox and is only placed in the shared directory I am expecting that it is passed to the WN by some other mechanism.
(I strongly suspect this is the case due to an old bug with afs lockfiles being passed to worker-nodes).

I think that we somewhere else in Ganga copy all files in the shared directory (regardless of the filetype) to the WN correctly and therefore GaudiRun is misbehaving.

There is a bug that the contents of the whole shared directory have to be tarred up and uploaded to DIRAC. (I think I'm missing an unlink call after the file has been uploaded). Ideally the `generateDiracInput` method would operate in a temp dir which would reduce problems with duplicating data in the users shared area (https://github.com/ganga-devs/ganga/pull/663/files#diff-8fb2d96aad0323b4634f00d2b0a026b9R214).

rob-c:
@mesmith75 I'm doing some testing right now. It seems useful to copy all non-tarred files from the shared directory to the WN (I'm not sure but I don't think this is happening). I'll investigate making sure this is done in a more central way than 'just in this app'

rob-c:
If #679 goes in then all files from the shared directory are automatically handled by Ganga and I can remove these calls to check/process what is in the input sandbox. Else I need to make sure all non-tarred files are incorporated in the input sandbox of each job.
There is a separate issue to do with the temporary DiracFile which is created as part of uploading the data to the Grid. I'll address this in a future PR as I'm working to have a 2nd archive uploaded containing the extra opts which the users will want to run on the WN.

mesmith75:
Ok, I'll wait for that update and see if there remains anything to be done.

rob-c:
@mesmith75 I think the changes highlighted here have been implemented in the `GaudiExec` class. If that is working correctly am I alright to close this PR and delete the branch?

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunAppHandlers.py
```diff
@@ -192,6 +193,9 @@ def prepare(self, app, appconfig, appmasterconfig, jobmasterconfig):
             raise ApplicationConfigurationError(None, 'Prepared folders not supported yet, please fix this in future')
         else:
             for f in input_files:
```
KonstantinSchubert: Would it be possible to simplify this to:
```

rob-c: `I think the changes highlighted here have been implemented in the GaudiExec class.`
The GaudiExec class now in develop is the new place where the LHCb Cmake support is being developed.

mesmith75:
@rob-c I ran a quick test with GaudiExec and it seems to be working fine for both Local and Dirac backends so by all mean close this and delete the branch

rob-c:
@mesmith75 great! thanks :+1:  Closing as issue has been found and resolved 'upstream' in `GaudiExec`.

PR #664, opened 2016-07-07
Change PYTHONPATH to syspath for rucio import

drmarkwslater:
In the recent changes to the plugin setup and removal of REEXEC I missed a change to the rucio client import that was needed for (very) old Athena releases. This fixes that.


drmarkwslater:
Any objections to merge?

PR #669, opened 2016-07-14
Adding realtime check for inheritance

rob-c:
This was originally written as part of the remove `__construct__` PR which exposed some nasty behaviour with getters and such.

As a result of some of the tests doing bad things with this enabled it was disabled. It is now in this PR.

I'm keen to work out why/how stuff is failing and fix it. This may be a problem in the tests that fail or it may be a bug in the code I don't know yet I've not had the time to look.

I'm not sure I want to leave this enabled 100% of the time but it would be nice to enable it in testing for instance.


rob-c:
This is now highlighting the problems with the tests. The subjobs are defined and stored using a getter in the `Job.py` file subverting this means that things will likely go wrong you should instead alter the unproxied object directly! (which is now stored in the Job correctly as of a fix a few months ago now)

rob-c:
I'd like to look to merge this into `6.3-features`.

This PR checks for all objects after the setter has finished to check that the correct parent is set at runtime.
This has a negligible performance hit and means that we can be sure that the inheritance of all objects (not from getter methods) are set correctly.
Should an error occur which breaks the inheritance there will be a `RuntimeError` raised which will inform us of a problem. However nothing breaks in the current test-suite so I think the chances of this happening are remote.

Does anyone have a strong opinion about this?
I'm happy to drop this check if we trust that the code is now working correctly, but I think it adds an extra layer of security in case of any future developments/changes.

rob-c:
Closing to remove stale branch

PR #670, opened 2016-07-14
doc changes

rob-c:
This is moving do changes out of a PR. The barrier to entry for editing these docs is still far too high imo. This takes time that's probably better spent populating a wiki out of other tasks.

I'm merging this once the tests pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ doc/_build
 .idea
 doc/API/Ganga*
 doc/API/ganga.rst
+doc/API/python.*
 doc/API/modules.rst
 doc/GPI/plugins.rst
 doc/GPI/classes.rst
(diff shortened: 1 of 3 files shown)


milliams:
Why isn't the information in the table put into the docstrings of `IGangaFile`? The information in the dev guide should be higher-level conceptual design documentation.

rob-c:
This went in a while ago without comment. Where should this go I'm confused as there is little documentation on how to use the documentation which frankly just confuses me even more.

rob-c:
I'm not going to document the IGangaFile interface until it's clear what belongs there. I'd rather have some place where we can edit/look/agree about what should or shouldn't be the standard and then make the changes else there would be a PR with lots of discussions about docs for the IGangaFile I fear.

milliams:
If you just want a scratch space for discussion about interfaces then the wiki is the best place for it. The dev guide should be an authoritative source of design information.

mesmith75:
Is this for the next release or shall we defer it?

PR #671, opened 2016-07-14
Attempt to speed up logic of getParent

rob-c:
Small attempt to alter this function which is now called an extremely high number of times to be a little faster.

### DIFF
--- a/python/Ganga/GPIDev/Base/Objects.py
+++ b/python/Ganga/GPIDev/Base/Objects.py
@@ -152,16 +152,18 @@ class Node(object):
         """
         if self._getParent() is None:
             return self
-        root = None
         obj = self
         cond_test = cond is not None
-        while obj is not None:
-            root = obj
+        while True:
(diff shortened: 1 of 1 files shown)


drmarkwslater:
I don't see how this can speed up the `getRoot` call as it just seems to move the code around a bit. Do you have a test to show the performance benefit?

KonstantinSchubert:
I do not know if this speeds it up, but here is the (hopefully) same function in fewer lines:
```
def _getRoot(self, cond=None):
    obj = self
    while obj is not None:
        if cond is not None:
            if cond(obj):
                break
        obj = obj._getParent()
    return obj
```

KonstantinSchubert:
Also @rob-c , the doc string of the method says: 
> cond is an optional function which may cut the search path: when it returns True, then the parent is returned as root

In your modified implementation, the obj itself is returned if cond(obj) is true, not the parent.

PR #672, opened 2016-07-14
Raise more detailed exception in getJobObject

rob-c:
This makes a distinction between an object with no Parent and an object with no Job associated with it. This is useful when debugging code using objects in a Box repo which (correctly) don't necessarily have a Job associated with them.


python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -1121,10 +1121,15 @@ def getJobObject(self):
...
-        return r
+        if self._getParent():
+            r = self._getRoot(cond=lambda o: isinstance(o, Job))
+            if not isinstance(r, Job):
+                raise AssertionError('No Job associated with object instead root=\'%s\' for \'%s\'' % (repr(self), type(r)))
+            return r
+        else:
+            if isinstance(self, Job):
+                return self
+        raise AssertionError('No Parent associated with object instead root=\'%s\' for \'%s\'' % (repr(self), str(None)))
```
drmarkwslater: I would say that either we don't need an `else` here or the `raise` needs to be indented to make it clearer what's happening.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -1121,10 +1121,15 @@ def getJobObject(self):
...
         """
         from Ganga.GPIDev.Lib.Job import Job
-        r = self._getRoot(cond=lambda o: isinstance(o, Job))
-        if not isinstance(r, Job):
-            raise AssertionError('no job associated with object ' + repr(self))
-        return r
+        if self._getParent():
+            r = self._getRoot(cond=lambda o: isinstance(o, Job))
+            if not isinstance(r, Job):
+                raise AssertionError('No Job associated with object instead root=\'%s\' for \'%s\'' % (repr(self), type(r)))
```
drmarkwslater: Are the arguments the wrong way round here? `root` is set to `repr(self)` rather than `type(r)` (`r` being the return of `_getRoot`).

milliams:
This should not be raising an `AssertionError`. There is no `assert` here. Instead probably something like a `RuntimeError` is more appropriate.

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -1121,10 +1121,14 @@ def getJobObject(self):
...
-            raise AssertionError('no job associated with object ' + repr(self))
-        return r
+        if self._getParent():
+            r = self._getRoot(cond=lambda o: isinstance(o, Job))
+            if not isinstance(r, Job):
+                raise AssertionError('No Job associated with object instead root=\'%s\' for \'%s\'' % (repr(r), type(r)))
+            return r
+        elif isinstance(self, Job):
+            return self
+        raise AssertionError('No Parent associated with object instead root=\'%s\' for \'%s\'' % (repr(self), str(None)))
```
drmarkwslater: This error message should probably just be `No parent associated with object '%s' % repr(self)`

rob-c:
@milliams The AssertError and RuntimeError are handled differently I'd rather not get into the situation that some code that used to run and caught exceptions now gets really messed up because I'm effectively just changing the string of the exception.

rob-c:
retest this please

mesmith75:
Is this ready to go into the release?

rob-c:
I'm happy to merge if there are no complaints?

PR #675, opened 2016-07-15
Allowing us to track latest stable LHCb code

rob-c:
This changes the source-d file for the LHCb tests to be a bit more minimal.
This ran the external/integration LHCb tests for me locally without any issue.

I didn't appear to have a problem source-ing a bash env script which contains a source command on the 2 machines I tested it on (SLC-something).


rob-c:
Any objections to merging this? it means that the tests are tracking the dependencies better than had-coding Ganga 6.1.18 into the paths.

milliams:
I'm happy with this. I'd preferred this way of doing it from the beginning but I seem to remember there was a reason it was not originally done this way. @drmarkwslater, do you remember?

drmarkwslater:
My initial issues were simply that I wasn't sure what `SetupProject` would do and whether it would actually setup everything correctly on the ganga-ci machines. If this works, then definitely 👍

PR #679, opened 2016-07-18
Adding copy non-tarred files to WN

rob-c:
This PR adds a step to the WN script for some backends to copy over files from the ganga shared directory.

The WN already auto-extract from the shared directory all gzip and bzip2 archives but I'm adding a step to copy onto the WN any files which are in the shared directory but which aren't archived files.

If anyone thinks this is a bad thing then can they let me know?

I'd like to use this if possible with my work on GaudiRun unless anyone thinks this is particularly bad?


python/Ganga/Lib/Batch/Batch.py
```diff
@@ -509,7 +509,7 @@ def get_status(f):
 
         from Ganga.Utility.Config import getConfig
         for j in jobs:
-            stripProxy(j)._getSessionLock()
+            stripProxy(j)._getWriteAccess()
```
drmarkwslater: This doesn't exist anymore

rob-c: @drmarkwslater Yes I appear to be fixing the conflicts now

rob-c:
OK, this is now just a few calls to `shutil.copy` does anyone have any objections to merging this?

rob-c:
retest this please

rob-c:
Any objections to merging this? It appears that the files are directly exposed to the running jobs via the inputsandbox but files are often expected to be in the cwd of the running job by the users payload which is what this achieves.

I was under the assumption that there was some mechanism to auto-compress the contents of the shared directory. There isn't. There is some machinery to auto-extract all archives on the WN once the ganga script starts executing but there is no machinery to auto-compress all files in the shared folder into a single `.tgz`

PR #681, opened 2016-07-19
Fix error message in FileSplitter

saschastahl:
The message says that property ignoreMissing = False but the property
of FileSplitter is called ignoremissing. I adjusted the error message.


milliams:
@ganga-ci test this please

milliams:
Looks good to me. Once the tests pass I will merge it.

milliams:
@saschastahl Thanks for the patch, it will be there in the next release of Ganga, 6.1.23.

PR #682, opened 2016-07-19
Stop task monitoring with --no-mon

### LINKED ISSUE #678: Ganga --no-mon should stop tasks
Just a quick issue to remind me to tie the `--no-mon` option to disable the task monitoring in addition to the job monitoring. This should alleviate the problem that tasks users have no way of launching ganga without getting many (many) jobs loaded from disk during startup.

rob-c:
Fixes #678


milliams:
Makes sense to me. Feel free to merge :+1:

Issue #654, opened 2016-07-04
accessURL method not doing the right thing for DiracFile


egede:
The accessURL method for DiracFile seems just to return an empty string. It should rather return a URL that can be openeddirectly in ROOT or similar. See https://lhcbqa.web.cern.ch/lhcbqa/317/how-to-automatically-replicate-to-cern-user for a discussion of what a user would have to do now to get it to work. The example given there, just returns the URL for the first place the file is replicated. One might want to do something slightly cleverer.

rob-c:
Does it make sense for the method to accept an argument which stears the filetype to know where to look for the file on multiple locations?
I know it doesn't for other filetypes but I don't think it's too bad to have DiracFile overload the method and allow for a an optional argument to be passed.
I think DiracFile is the unique one (for now) which only has 1 lfn to multiple PFN but how are we to know which SE we should access? I think it's sensible to use the defaultSE for the DiracFile to first order and then allow for it to be overloaded so people can use this for testing file access and such if required.

egede:
@rob-c I think using `defaultSE` from the configuration as the default, and to allow an extra argument is fine. The important thing is that a URL should be returned even if it is not matching those (ie. the argument gives a preference but not a requirement). We have to teach users that it is perfectly fine to read files across the WAN.

rob-c:
@egede OK, I hadn't appreciated that we are promoting/enabling accessing files remotely over the steaming protocols. This could ultimately allow Ganga to be used to construct some test jobs which can monitor/debug file transfer across DIRAC or to/from a given site.

egede:
@rob-c It works really well and can through promotion avoid the often needless transfers from `Dirac` storage to EOS or to local disk. As can be seen from the original discussion (linked into the first post of this issue), it is a bit cumbersome at the moment to get right.

mesmith75:
I'm not sure this function in ganga ever worked. The requisite DiracCommand was missing. I have it approximately working now. I thought I would just get it to return a list of the SEs and the URL for each rather than picking one.

Before I commit anything can you explain the reason for the matching between the SE and the site? Is it really necessary? If not I would like to get rid of that part of the code. I am referring to about lines 517 to 531.
https://github.com/ganga-devs/ganga/blob/develop/python/GangaDirac/Lib/Files/DiracFile.py#L517

As an aside the defaultSE element is returning nothing - is that to be expected?

rob-c:
[...]

PR #684, opened 2016-07-19
A fix to the accessURL function in DiracFile

mesmith75:
#654 Fixed the function to get an accessURL for a DiracFile. It just fills a dict with the accessURLs for all the SEs with replicas of the LFN. If no SE is specified by the user a random one is chosen.


egede:
@mesmith75 Looks good.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -475,51 +475,33 @@ def accessURL(self, thisSE=''):
...
-
-            # Currently only written to cope with 1 replica per DIRAC file
-            # I think adding multiple accessURL for the same file at 1 site
-            # adds confusion
-            return [this_accessURL]
+            # If the SE isn't specified return a random choice.
+            if thisSE == '':
+              return random.choice(list(this_accessURL.items()))[1]
+            # If the SE is specified and we got a URL for a replica there, return it.
+            elif thisSE in this_accessURL.keys():
```
milliams: This can be simply:
```python
elif thisSE in this_accessURL:
```

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -475,51 +475,33 @@ def accessURL(self, thisSE=''):
...
+                logger.info( 'Cannot find a replica for the LFN %s' % self.lfn)
                 return []
-
-            # Currently only written to cope with 1 replica per DIRAC file
-            # I think adding multiple accessURL for the same file at 1 site
-            # adds confusion
-            return [this_accessURL]
+            # If the SE isn't specified return a random choice.
+            if thisSE == '':
+              return random.choice(list(this_accessURL.items()))[1]
```
milliams: This could probably at least be:
```python
return random.choice(list(this_accessURL.values()))
```
but since this is basically getting 'any' entry from the list, we could just grab the first with:
```python
return next(iter(this_accessURL.values()))
```

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -475,51 +475,33 @@ def accessURL(self, thisSE=''):
...
-                        elif type(default_site) == type(''):
-                            if this_site == default_site:
-                                this_accessURL = this_URL
-                                break
-                if this_accessURL != '':
-                    break
-
-            # Cannot find an accessURL
+                this_accessURL[this_SE] = []
+                myurl = execute('getAccessURL("%s" , "%s")' % (self.lfn , str(this_SE)))
```
milliams: There should be no space between `self.lfn` and the `,`. Also, there is no need to pass `this_SE` via `str()`

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -475,51 +475,33 @@ def accessURL(self, thisSE=''):
...
+                this_accessURL[this_SE] = myurl['Value']['Successful'][self.lfn]
+            # Cannot find an accessURL. Doesn't really do anything as Dirac throws an error before this point.
             if this_accessURL == '':
+                logger.info( 'Cannot find a replica for the LFN %s' % self.lfn)
```
milliams: Should be no space between `info(` and `'Cannot`.

python/GangaDirac/Lib/Files/DiracFile.py
egede: @milliams We do not know what the ordering is, so I think it is better to 
return a random entry then. The documentation should explain that it is random.

rob-c:
@mesmith75 Looks like this is heading in the right direction. Perhaps showing my inability to parse python but does this always return a list of possible URLs? The MassStorageFile always just returns a list so I think we want to be compatible with this. I see we're returning a list in some places but some returned values look like strings unless I'm mistaken.

Also, if the requested URL for a given site can't be found should we make the message a warning as this is potentially not what the user wants (e.g. file should be replicated at CERN but according to these functions the replica is unknown). I think returning some other valid URL is a good fall-back as determining if a remote file can/can't be streamed is more of a user task I would argue.

mesmith75:
@milliams I've implemented your comments except that I have persisted with returning a URL from a random SE.

mesmith75:
@rob-c If there are no subfiles then it returns just the single URL. If there are subfiles it currently returns a list of the accuess URLs for all the subfiles only.

I think it would be better to get rid of the part of the code about the subfiles and just return the accessURL of the given DiracFile - If the user wants the accessURL of a subfile it is trivial to get it just by running the command on the subfile.

rob-c:
@mesmith75 I think there is code already to expect that the list of URL from a file (regardless of if it has any subfiles) in Ganga? If not this needs to be checked.
I think the only working implementation is the `MassStorageFile` class which is used to make available the list of appropriate URLs for all known subfiles to the parent file.

The intention behind this is that we can call the 1 file with a wildcard in the namePattern and get all of the accessURL which correspond to the matching files on storage. i.e. `MassStorageFile("/some/path/*.root").accessURL()` returns a list of all files rather than nothing.
Otherwise when the *.root matches just 1 file you need to do special things that you don't have to do when just asking for a list of all URLs as we have at the moment.
(Plus if we're always returning a list we can reduce the amount of constant checks in our code about what a given method returned).

I'm more in favor of keeping the list API and making it clear in DiracFile that the returned list is composed of URLs which are each randomly selected from the available replicas for each sub-file in the DFC.

If there is no code explicitly expecting a list here then we can change the API but if there is code expecting a list somewhere then we should stick to the standard already used elsewhere.

egede:
@mesmith75 @rob-c Always returning a list is fine for me. This allows for wildcards and there is never doubt about the return type.

mesmith75:
Ok I agree returning a list is better.
For those files with subfiles I will return a list with the top file, with an SE can be chosen, followed by the subfiles which will be chosen at random.
It looks as if the getReplicas() function only applies to the subfile - is that deliberate?

rob-c:
@mesmith75 I'm not sure what you mean when you say that it 'only applies to the subfile' can you elaborate?

mesmith75:
If I have a DiracFile with some subfiles and use getReplicas() it will only return the replicas of the subfiles but not the DiracFile itself. I can see this is implemented in the code:
https://github.com/ganga-devs/ganga/blob/develop/python/GangaDirac/Lib/Files/DiracFile.py
```
def getReplicas(self, forceRefresh=False):
#        Get the list of all SE where this file has a replica
#       This relies on an internally stored list of replicas, (SE and  unless forceRefresh = True
        if self.lfn == '':
            self._optionallyUploadLocalFile()
        if self.lfn == '':
            raise GangaException("Can't find replicas for file which has no LFN!")
        these_replicas = None
        if len(self.subfiles) != 0:
            allReplicas = []
            for i in self.subfiles:
                allReplicas.append(i.getReplicas())
            these_replicas = allReplicas
```
in

rob-c:
@mesmith75 In the case that a file has subfiles the parent file itself is always expected to contain some wildcard or pattern matching so it makes no sense to ask for the `accessURL` or something similar on the parent file but it makes sense to ask about the subfiles which each correspond to an actual file on the remote storage.
(I hope this helps)

mesmith75:
@rob-c Ok I see, I had missed the point of subfiles. In which case I shall just return a list of the subfiles from random SEs.
If there are no subfiles then the SE can be specified.

rob-c:
@mesmith75 I would say it's sane to pass on any user requested SE to the subfiles calls so each subfile can pass the correct replica if there is one for the chosen SE.

mesmith75:
@rob-c Yes that seems reasonable. 
I have one more question. I have noticed that getReplicas updates the _remoteURLs value for every file. For example if I call getReplicas on a single subfile then the other subfiles all have their _remoteURLs updated with that list of replicas.
Is that deliberate? It is a bit of a pain if the subfiles are spread over different locations.

rob-c:
@mesmith75 I think the _remoteURLs should be updated by this method but each subfile should only contain the URLs corresponding to the actual file it describes.
I think the point of this is to cache the result of this lookup which is potentially very expensive.

mesmith75:
The last returns a list of URLs. If there are no subfiles it just returns one URL, if there are subfiles it returns a URL for each one.
In case of a SE being specified that is passed to the subfile which will use that one, or a random one if there is no replica there.

rob-c:
@mesmith75 Looks good to merge from me :+1:

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -58,6 +58,11 @@ def getReplicas(lfns):
     output(dirac.getReplicas(lfns, active=True, preferDisk = True))
 
 
+def getAccessURL(lfn,SE):
+    ''' Return the access URL for the given LFN and storage element '''
+    output(dirac.getAccessURL(lfn,SE))
```
KonstantinSchubert: Stupid question: The docstring says "Return", but this function does not return anything?

milliams: These functions are run in a special DIRAC Python process which defines the `output` function so that it pushes its argument to the stdout (or maybe some other pipe, I can't remember). The process which originally called the function consumes this pipe and so get the string representation of the `output` argument which it then `eval`s.

Confusing I agree but the best solution given the constraints.

KonstantinSchubert: Okay. Thanks for the explanation :)

mesmith75:
If there are no objections I'll merge this into the develop bramch and close the issue.

mesmith75:
This fixes issue #654

PR #690, opened 2016-07-25
New credentials system

milliams:
This adds the new credentials system to Ganga. The best summary is at the [dev manual page](http://ganga.readthedocs.io/en/feature-credentials/dev/credentials.html).

I believe that this is all working for core LCG as well as DIRAC so please test that thoroughly. The ATLAS support is provisional but those changes will be more involved and I have no way of testing it at present. @drmarkwslater could you run some ATLAS and give me some feedback about what doesn't work.


python/Ganga/Core/InternalServices/Coordinator.py
```diff
@@ -234,10 +180,10 @@ def enableInternalServices():
...
 
     # make sure all required credentials are valid
-    missing_cred = getMissingCredentials()
-    if missing_cred:
-        log.error("The following credentials are still required: %s."
-                  "Make sure you renew them before reactivating this session" % ','.join(missing_cred))
+    invalid_afs = [afsToken for afsToken in credential_store.get_all_matching_type(AfsToken) if not afsToken.is_valid()]
+
+    if invalid_afs:
+        log.error('No valid AFS token was found. Please re-authorise before reactivating this session.')
```
rob-c: Given the registry flushing and lock files don't respect this stop shouldn't this cause an emergency shutdown?

python/Ganga/GPIDev/Credentials2/AfsToken.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class AfsTokenInfo(ICredentialInfo):
+    """
+    A wrapper around an AFS token
+
+    For now it is very CERN-specific (or at least only follows the CERN use-case)
+    """
+
+    def __init__(self, requirements, check_file=False, create=False):
```
rob-c: What do these arguments correspond to?

python/Ganga/GPIDev/Credentials2/ICredentialRequirement.py
```diff
@@ -0,0 +1,57 @@
...
+
+    # This attribute refers to the ICredentialInfo subclass which can satisfy the requirements
+    info_class = None  # type: Type[ICredentialInfo]
+
+    def __init__(self, **kwargs):
+        super(ICredentialRequirement, self).__init__()
+        for key, value in kwargs.items():
+            if key not in self._schema.allItemNames():
+                raise Exception()
+            setattr(self, key, value)
```
rob-c: Wouldn't this be better in GangaObject as this is overloading the values which are constructed there?

python/Ganga/GPIDev/Credentials2/VomsProxy.py
```diff
@@ -0,0 +1,148 @@
...
+        logger.debug(command)
+        return_code = self.shell.system(command)
+        if return_code == 0:
+            logger.debug('Grid proxy %s created. Valid for %s', self.location, self.time_left())
+        else:
+            raise CredentialRenewalError('Failed to create VOMS proxy')
+
+    @property
+    def shell(self):
+        if self._shell is None:
```
rob-c: Is there a reason this can't be `if not self._shell`?

python/Ganga/GPIDev/Credentials2/VomsProxy.py
```diff
@@ -0,0 +1,148 @@
...
+        # TODO Make this support multiple groups and subgroups
+        group_role_list = vo_list[1].split('/')
+        if len(group_role_list) <= 2:
+            return None  # No group specified in command
+        return group_role_list[-1].strip()
+
+    @cache
+    def expiry_time(self):
+        status, output, message = self.shell.cmd1('voms-proxy-info -file %s -timeleft' % self.location)
+        if status != 0:
```
rob-c: Is there a reason this isn't `not status`?

python/Ganga/Lib/LCG/Grid.py
```diff
@@ -1,9 +1,9 @@
 import os
 import re
 import tempfile
-import time
+import datetime
 
-from Ganga.GPIDev.Credentials import getCredential
+from Ganga.GPIDev.Credentials2 import credential_store
```
rob-c: Wouldn't it be better to just replace the Credentials module with a new one rather than one with a new name? Why should external modules have to know about this?

python/Ganga/Lib/LCG/Grid.py
```diff
@@ -188,23 +149,14 @@ def __resolve_no_matching_jobs__(cmd_output):
     return glite_ids
 
 
-def list_match(jdlpath, ce=None):
+def list_match(jdlpath, cred_req, ce=None):
```
rob-c: what is this attribute?

python/Ganga/__init__.py
```diff
@@ -299,27 +299,13 @@ def getLCGRootPath():
...
-# set default values for the configuration parameters
-lcg_config.addOption(
-    'EDG_ENABLE', False, 'DEPRECATED enables/disables the support of the EDG middleware')
-
-lcg_config.addOption('EDG_SETUP', '/afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh',
-                 'DEPRECATED sets the LCG-UI environment setup script for the EDG middleware',
-                 filter=Ganga.Utility.Config.expandvars)
-
-lcg_config.addOption(
-    'GLITE_ENABLE', True, 'Enables/disables the support of the GLITE middleware')
```
rob-c: Rather than remove the old config options (I'm not sure Ganga always picks up that options are invalid in a useful way)
Could we not add a method which prevents Ganga starting if the user has configured everything for the old setup to save some confusion about what they have to change.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -145,7 +149,7 @@ def _common_submit(self, dirac_script):
         self.statusInfo = ''
         j.been_queued = False
         dirac_cmd = """execfile(\'%s\')""" % dirac_script
-        result = execute(dirac_cmd)
+        result = execute(dirac_cmd, cred_req=self.credential_requirements)
```
rob-c: Why not just wrap execute save repeating this change all over the DIRAC code?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -760,13 +768,7 @@ def job_finalisation(job, updated_dirac_status):
...
                 if job.status in ['completed', 'killed', 'removed']:
                     break
-                # make sure proxy is valid
-                if DiracBase.checkDiracProxy():
-                    # perform finalisation
-                    DiracBase._internal_job_finalisation(job, updated_dirac_status)
-                else:
-                    # exit gracefully
-                    logger.warning("Cannot process job: %s. DIRAC monitoring has been disabled. To activate your grid proxy type: \'gridProxy.renew()\'" % job.fqid)
+                DiracBase._internal_job_finalisation(job, updated_dirac_status)
```
rob-c: Does this still exit gracefully or does it demand a user give proxy details because the monitoring system has found a grid job?
This can be annoying when wanting to fire up ganga for a non-grid related task when the repo may have large numbers of jobs known to the user not to need updating.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -974,8 +957,26 @@ def updateMonitoringInformation(jobs_):
         #logger.debug('Monitor jobs    : ' + repr([j.fqid for j in monitor_jobs]))
         #logger.debug('Requeue jobs    : ' + repr([j.fqid for j in requeue_jobs]))
 
+        # Split all the monitorable jobs into groups based on the
+        # credential used to communicate with DIRAC
```
rob-c: Isn't each proxy for each group an individual credential object? In which case I agree with the `TODO` this should be more generic as can't any tools using the VOMS tools potentially have multiple groups for 1 user.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -974,8 +957,26 @@ def updateMonitoringInformation(jobs_):
...
+        for j in monitor_jobs:
+            try:
+                cred_req = j.backend.credential_requirements
+                cred = credential_store[cred_req]
+                if not cred.is_valid():
+                    logger.debug('Required credential %s is not valid', cred)
+                    needed_credentials.add(cred_req)
+                    continue
+                monitor_jobs_by_credential[cred].append(j)
+            except KeyError:
```
rob-c: Can't this just be avoided by using `if element in dict`? The latter is much faster once you start throwing many KeyErrors and I've a dislike for writing code to intentionally throw an exception in a large number of cases it just feels like the code is doing something not quite right.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -47,6 +49,7 @@ class DiracFile(IGangaFile):
                                                                typelist=['GangaDirac.Lib.Files.DiracFile'], doc="collected files from the wildcard namePattern"),
                                      'defaultSE': SimpleItem(defvalue='', copyable=1, doc="defaultSE where the file is to be accessed from or uploaded to"),
                                      'failureReason': SimpleItem(defvalue="", protected=1, copyable=0, doc='reason for the upload failure'),
+                                     'credential_requirements': ComponentItem('CredentialRequirement', defvalue=DiracProxy()),
```
rob-c: Should this be hidden?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -87,7 +89,8 @@ class DiracBase(IBackend):
                                 doc='DIRAC API commands to add the job definition script. Only edit '
                                 'if you *really* know what you are doing'),
         'settings': SimpleItem(defvalue={'CPUTime': 2 * 86400},
-                               doc='Settings for DIRAC job (e.g. CPUTime, BannedSites, etc.)')
+                               doc='Settings for DIRAC job (e.g. CPUTime, BannedSites, etc.)'),
+        'credential_requirements': ComponentItem('CredentialRequirement', defvalue=DiracProxy()),
```
rob-c: Should this be hidden or is a user expected to know how to make/edit a DiracProxy?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -237,19 +175,17 @@ def execute(command,
...
 
-    # We're about to perform an expensive operation so being safe before we run it shouldn't cost too much
-    _checkProxy(force = True)
-
-    #logger.debug("Executing command:\n'%s'" % str(command))
-    #logger.debug("python_setup:\n'%s'" % str(python_setup))
-    #logger.debug("eval_includes:\n'%s'" % str(eval_includes))
+    if cred_req is not None:
+        env = env or {}
+        env['X509_USER_PROXY'] = credential_store[cred_req].location
```
rob-c: I'm concerned by the env potentially being just the X509 path which I would expect to cause bad problems

rob-c:
Is this branch working for LHCb?

milliams:
@rob-c It should be, yes. Give it a try and let me know if anything doesn't work. Also, I appreciate your comments on the code, I just haven't had time to go through them yet as I'm working on local grid stuff at the moment.

rob-c:
@milliams I've just been cautious about spending too much time on this given the failing LHCb test. Once I've finished the work I'm doing on CMake I'll try and give this a spin with some real world jobs.

milliams:
The tests all pass on my local system so I'll look into why there's a difference on Jenkins.

python/Ganga/Core/InternalServices/Coordinator.py
milliams: I've tried to avoid changing the logic behind the runtime systems too much. For now I just wanted to maintain the current behaviour. I agree in general though that having a startup check for an invalid AFS token may be a good idea. How does it react at present?

python/Ganga/GPIDev/Credentials2/AfsToken.py
milliams: These arguments are the same as in the base class (`ICredentialInfo`) and are documented there.

python/Ganga/GPIDev/Credentials2/ICredentialRequirement.py
milliams: Probably, yes. This is a general bit of code which could apply to every `GangaObject`. Perhaps I should propose this in a separate PR?

python/Ganga/GPIDev/Credentials2/VomsProxy.py
milliams: Because I want to be explicit about a missing value. I want to explicitly check for an unassigned value rather than checking for truthiness.

python/Ganga/Core/InternalServices/Coordinator.py
rob-c: If the AFS token expires we get horrible errors and nothing makes sense and eventually everything bails out because we can't even log to disk.
I would just be worried about submitting jobs to a remote backend and then losing all track of them here because Ganga isn't flushing until after the jobs have been submitted. It's a bad situation to be in.

python/Ganga/GPIDev/Credentials2/VomsProxy.py
milliams: In the case of dealing with Unix commands which have the usual `True==1`, `False==0` definition inverted it is confusing reading:
```python
status = <some unix command>
if status:
    raise RuntimeError('command failed')
```
since normally an `if something` suggests truth and therefore success. Explicitly comparing against the `0` literal shows that it really is that value we care about, not some other definition of truth.

python/Ganga/Lib/LCG/Grid.py
milliams: Indeed. Unfortunately while this PR is being finalised, there is still some ATLAS code which depends on the old credentials system. Once that is ported, `Credentials` can be removed and `Credentials2` can be renamed.

python/Ganga/GPIDev/Credentials2/ICredentialRequirement.py
rob-c: If we can offer support for all kwds for all classes by default that would be nice :+1:

python/Ganga/Lib/LCG/Grid.py
milliams: Across the codebase, `cred_req` is used to designate a `ICredentialRequirement` object being passed around. I can go through `Grid.py` and add in docstrings explaining this.

python/Ganga/GPIDev/Credentials2/AfsToken.py
rob-c: Would a summary help here rather than having to read several files to know how this works?

python/Ganga/__init__.py
milliams: In the new system there are no config settings needed to enable the LCG system. It is always enabled by default. That said, it will give an
```
ERROR    unknown options [LCG]GLITE_ENABLE
```
message on startup which may be annoying. However, we need to get rid of these at some point and there's no better way of doing so.

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: In the new system, Dirac's `execute()` will *always* require an `ICredentialRequirement` object which must be provided by the calling scope so I can't see an easy way of wrapping this. Unless I'm missing something?

python/Ganga/Lib/LCG/Grid.py
rob-c: Shouldn't we be documenting the functions which rely on this so that it's easy to see when just looking at one function in isolation?

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: The idea is that the check will be preformed at a higher level (in `updateMonitoringInformation()`) so that if the proxy is not present or invalid then it never gets this far.

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: I was unsure where to put it at first but I think I will move it to `IBackend`. *TODO*

python/Ganga/__init__.py
rob-c: Yes but I would imagine it's easier when we first introduce this to give some error about the option being deprecated and construct it to be a read-only object so that if a user tires to set it an exception is thrown?
That way Ganga won't start with an old config which will probably lead to confusion about what parameter has been set and what hasn't.
I imagine the users should read the errors and should read the release notes but I'm of the opinion that if we want to enforce this we need to ensure that Ganga won't start with a bad configuration setup.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Atm I'm unable to see where the credential requirement is defined, probably because of something fancy in the decorator.
I'm reading the code passing an undeclared parameter rather than some new execute function where the credentials are worked out.

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: I'm not sure it will be any faster as the standard definition of `__contains__` is:
```python
def __contains__(self, key):
    try:
        self[key]
    except KeyError:
        return False
    else:
        return True
```
The point is that in the `try` block you should be able to write code as if there was no problem without checking and then use the `except` block for situations where that assumption failed (see 'EAFP' in https://docs.python.org/2/glossary.html).

The throwing and catching of exceptions as logic flow is generally considered idiomatic in Python, as long as it's not overdone.

That said, I will look at rewriting this part in that style and see if it's more readable.

python/GangaDirac/Lib/Files/DiracFile.py
milliams: No, this is the mechanism by which the user can specify special requirements needed to access that file.

python/GangaDirac/Lib/Backends/DiracBase.py
milliams: If they need to change anything then they are indeed required to know how to create one, yes:
```python
j = Job()
j.backend = Dirac(credential_requirements=DiracProxy(group='my_group'))
```

python/GangaDirac/Lib/Utilities/DiracUtilities.py
milliams: Ahh, this is a change from when this code was first written. Now we have
```python
if env is None:
    env = getDiracEnv()
```
I can get rid of this so that it's:
```python
if cred_req is not None:
    env['X509_USER_PROXY'] = credential_store[cred_req].location
```

python/Ganga/Core/InternalServices/Coordinator.py
drmarkwslater: I think I agree with @rob-c here - maybe this is the time to introduce Ganga bailing out as soon as the AFS token expires? In theory you could keep running and allow the user to get a new AFS token but I suspect that the only real time this happens is with sessions left running overnight in which case they wouldn't be able to recreate the token,

egede: So can we catch that the afs token is about to expire and then shut down nicely. Otherwise we will just abort which is not great.

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -664,10 +665,7 @@ def runMonitoring(self, jobs=None, steps=1, timeout=300, _loadCredentials=False)
         if not self.enabled:
             # and there are some required cred which are missing
             # (the monitoring loop does not monitor the credentials so we need to check 'by hand' here)
-            if _loadCredentials is True:
-                _missingCreds = Coordinator.getMissingCredentials()
-            else:
-                _missingCreds = False
```
drmarkwslater: Because the default for `runMonitoring` was `False` and never called with something else (as far as I can tell), it seems the missing credentials was never checked. This is changed in the new code which I think I agree with but just wanted to check this was expected.

python/Ganga/Core/InternalServices/Coordinator.py
rob-c: @drmarkwslater Would it be sensible to suggest that Ganga attempt (via some mechanism) to renew it's token as I think 5 automated renewals are typically allowed which can extend a session to 5 days without a password being needed?
In this case are we going to mark the Registries as requiring an AFS token if they're on an AFS filesystem?

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: @milliams Does the defvalue work correctly for the ComponentItem? I think I've seen some oddities in this part of the Schema? If it does 'just work' great :D

python/Ganga/GPIDev/Credentials2/AfsToken.py
drmarkwslater: I think it's fair enough to expect people to read up on the base class otherwise it'll just be repeating what's there. And I would guess there would be links in the generated docs anyway.

python/Ganga/GPIDev/Credentials2/AfsToken.py
```diff
@@ -0,0 +1,122 @@
...
+import subprocess
+from getpass import getpass
+from glob import glob
+
+import Ganga.Utility.logging
+from Ganga.Utility.Shell import Shell
+
+from .ICredentialInfo import ICredentialInfo, cache
+from .ICredentialRequirement import ICredentialRequirement
+from .exceptions import CredentialRenewalError
```
drmarkwslater: Entirely personal comment but I'm not a great fan of relative imports 😄

python/Ganga/GPIDev/Credentials2/CredentialStore.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Returns:
+            ICredentialInfo: A single credential object. If more than one is found, the first is returned
+        """
+
+        matches = self.matches(query)
+        if len(matches) == 1:
+            return matches[0]
+        if len(matches) > 1:
+            logger.debug('More than one match...')
```
drmarkwslater: Very minor - add the query to this debug message?

python/Ganga/GPIDev/Credentials2/ICredentialInfo.py
```diff
@@ -0,0 +1,192 @@
...
+            self.cache[method.func_name] = method(self, *args, **kwargs)
+
+        return self.cache[method.func_name]
+    return wrapped_function
+
+
+class ICredentialInfo(object):
+    """
+    The interface for all credential types.
+    Each object covers one credential file exactly.
```
drmarkwslater: We don't have any credentials that need this, but is there any way at all to get this to work with a Credential that needs multiple files? Just interested to see if this cannot work at all or could be made to work somehow - just in case 😄

python/Ganga/GPIDev/Credentials2/AfsToken.py
rob-c: I'll be honest I'm confused at this point I've had the similar opinion for a while but other advice I've had is to document every occurrence of a method which overloads the baseclass, are the `__init__` methods different? I'll admit I don't think many people would be calling `help` on them.

python/Ganga/GPIDev/Credentials2/VomsProxy.py
```diff
@@ -0,0 +1,148 @@
...
+        return_code = self.shell.system(command)
+        if return_code == 0:
+            logger.debug('Grid proxy %s created. Valid for %s', self.location, self.time_left())
+        else:
+            raise CredentialRenewalError('Failed to create VOMS proxy')
+
+    @property
+    def shell(self):
+        if self._shell is None:
+            self._shell = Shell()
```
drmarkwslater: Am I right in saying this assumes the default shell contains a working UI? In the past, this was not required if you specified the setup script in the config...

python/Ganga/Runtime/bootstrap.py
```diff
@@ -915,6 +903,15 @@ def bootstrap(interactive):
 
         manualExportToGPI()
 
+        from Ganga.Runtime import Workspace_runtime, Repository_runtime
+        from Ganga.GPIDev.Credentials2 import credential_store
+        if Workspace_runtime.requiresAfsToken() or Repository_runtime.requiresAfsToken():
```
rob-c: Are these methods safe from soft linking?

python/Ganga/Lib/LCG/Grid.py
```diff
@@ -1332,39 +1121,26 @@ def arc_cancel(jobid):
...
-    if not credential().isValid('01:00'):
-        logger.warning('GRID proxy lifetime shorter than 1 hour')
-        return False
 
     idsfile = tempfile.mktemp('.jids')
     with open(idsfile, 'w') as ids_file:
         ids_file.write('\n'.join(jobids) + '\n')
 
-    # compose the cancel command
+    # compose the cancel comman
```
drmarkwslater: Lost a 'd'

python/Ganga/__init__.py
drmarkwslater: Just discussed this with @milliams and I'm OK with people seeing some `ConfigError` messages as they're harmless and won't represent a 'bad config' as such (these are just not applicable anymore) but I think we should have a semi-permanent message telling people not to worry about that and pointing them at docs for the new system. If it's just shown on the release notes then people probably won't read it and will only get one chance at seeing it anyway.

python/Ganga/__init__.py
```diff
@@ -299,27 +299,13 @@ def getLCGRootPath():
...
-
 lcg_config.addOption('GLITE_SETUP', '/afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh',
                  'sets the LCG-UI environment setup script for the GLITE middleware',
                  filter=Ganga.Utility.Config.expandvars)
 
 lcg_config.addOption('VirtualOrganisation', '',
                  'sets the name of the grid virtual organisation')
 
-lcg_config.addOption('ConfigVO', '', 'DEPRECATED sets the VO-specific LCG-UI configuration script for the EDG resource broker',
-                 filter=Ganga.Utility.Config.expandvars)
```
drmarkwslater: So is there now a single place to put the default VO? is it `[LCG][VirtualOrganisation]`? If so what about `[GridProxy_defaults]voname` - is the GridProxy object still a thing?

python/Ganga/Core/InternalServices/Coordinator.py
drmarkwslater: @rob-c That sounds good - in fact the `renew` function would probably do that as it uses `kinit -R`.

python/Ganga/__init__.py
rob-c: @drmarkwslater I'm not sure what you mean here, are you saying should we add a 2nd message to the launch of Ganga after integrating the credentials branch?

I still think failing to load until people have read a message and taken action is more likely to get people's attention and not to bug those who don't use these features. (new users and people on local batch systems). These errors should only be presented to people who are directly effected by the changes in this PR from what I can tell.

drmarkwslater: On 08/08/2016 14:01, Robert Currie wrote:
>
> In python/Ganga/__init__.py 
> <https://github.com/ganga-devs/ganga/pull/690#discussion_r73870496>:
>
> > @@ -299,27 +299,13 @@ def getLCGRootPath():
> >  lcg_config = makeConfig('LCG', 'LCG/gLite/EGEE configuration parameters')
> >  #gproxy_config = getConfig('GridProxy_Properties')
> >
> > -# set default values for the configuration parameters
> > -lcg_config.addOption(
> > -    'EDG_ENABLE', False, 'DEPRECATED enables/disables the support of the EDG middleware')
> > -
> > -lcg_config.addOption('EDG_SETUP', '/afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh',
> > -                 'DEPRECATED sets the LCG-UI environment setup script for the EDG middleware',
> > -                 filter=Ganga.Utility.Config.expandvars)
> > -
> > -lcg_config.addOption(
> > -    'GLITE_ENABLE', True, 'Enables/disables the support of the GLITE middleware')
>
> @drmarkwslater <https://github.com/drmarkwslater> I'm not sure what 
> you mean here, are you saying should we add a 2nd message to the 
> launch of Ganga after integrating the credentials branch?
>
Yes - partly to highlight the improvements we've put in (being able to 
submit jobs from different VOs in the same Ganga session/repo is quite a 
big thing) but also to tell people why they are seeing errors.

> I still think failing to load until people have read a message and 
> taken action is more likely to get people's attention and not to bug 
> those who don't use these features. (new users and people on local 
> batch systems). These errors should only be presented to people who 
> are directly effected by the changes in this PR from what I can tell.
>
Failing to load seems quite a major thing to do when Ganga can continue 
without issue. And the users don't *have* to take any action as these 
errors can be ignored. Plus, at present the release notes are only shown 
once so if you miss that or aren't reading the notes properly, you can't 
start Ganga and have just got an 'Unkown Config Option' error to go on, 
to which people may not think to just delete from their .gangarc. And 
actually, if this option is set on a site-wide/install level, the user 
may not be able to remove the option anyway if it's set in a .ini file 
somewhere.

@egede, @alexanderrichards  any thoughts?

drmarkwslater:
I've found some issues with Atlas and `VomsProxy`. Basically, when running in the Atlas environment, starting Java pollutes the output somewhat:
```
bash-4.1$ voms-proxy-info  -timeleft
Picked up _JAVA_OPTIONS: -Xms64m -Xmx128m
343744
bash-4.1$ 
```

which in turn then breaks at least the `expiry_time` function in `VomsProxy` and probably others. Not sure of the best way around this - maybe a regex to find the correct line?

drmarkwslater:
Also, there's a more general issue with the Atlas version of this code in that to properly go over it I need to get at least some of the DQ2 removal and cleanup stuff done (see #712 and more to come). I'll  spend a week or so pushing that a bit more and then I can hopefully come back to this more properly.

rob-c:
Can't seem to load ganga using this branch due to:

```
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py", line 314, in index_load
    self.objects[this_id] = self._make_empty_object_(this_id, self._cached_cat[this_id], self._cached_cls[this_id])
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/Ganga/Core/GangaRepository/GangaRepository.py", line 184, in _make_empty_object_
    obj = cls()
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/GangaDirac/Lib/Files/DiracFile.py", line 73, in __init__
    super(DiracFile, self).__init__()
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/Ganga/GPIDev/Adapters/IGangaFile.py", line 22, in __init__
    super(IGangaFile, self).__init__()
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/Ganga/GPIDev/Base/Objects.py", line 690, in __init__
    setattr(self, attr, self._schema.getDefaultValue(attr))
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/Ganga/GPIDev/Schema/Schema.py", line 221, in getDefaultValue
    returnable = self._getDefaultValueInternal(attr)
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/Ganga/GPIDev/Schema/Schema.py", line 299, in _getDefaultValueInternal
    return _found_components[category]()
  File "/afs/cern.ch/user/r/rcurrie/cmtuser/GANGA/GANGA_LSST/install/ganga/python/GangaDirac/Lib/Credentials/DiracProxy.py", line 103, in __init__
    raise ValueError('DIRAC Proxy `group` is not set')
ValueError: DIRAC Proxy `group` is not set
```

Are there any objections to me pushing bugfix PR(s) against this branch?

This is from my LSST dev environment using the latest code.
I'm also not a big fan of seeing all of the old configs to do with proxy creation ignored as it would be helpful to tell the user something has changed here as they will probably ignore the startup release notes (esp if the ssh connection dies right after they try and login for instance)

rob-c:
Also, minor(ish) comment. Why is the `DiracProxy` not in `GangaDirac`?

milliams:
Regarding "DIRAC Proxy `group` is not set": there's currently no setting in the config that maps to a DIRAC group so we have to request a new one. In the past the VO was abused to do this but it's not the same thing at all.

rob-c:
Can we add some startup checks for things like this not being set. I get a horrible rolling screen of red when I try to load ganga from this branch otherwise.
It should only have to tell a user once to change their config and then fail to load if this hasn't been set correctly.

milliams:
Is the user group in DIRAC for LSST `lsst_user`? If so then I will add:
```ini
[defaults_DiracProxy]
group = lsst_user
```
to LSST.ini and the problem will go away.

rob-c:
@milliams yes it goes away but I'm keen to see that users shouldn't be greeted to 6.3.0 with a rolling screen of red just because they've not set this yet, especially if we can just determine at startup that this hasn't been set.

rob-c:
I also see a lot of `ERROR    Problem in the monitoring loop: AfsTokenInfo checking failed!` when trying to run the monitoring loop on a DIRAC job. I'll see if I can get something more sensible from a machine on lxplus.

rob-c:
I'm seeing:
```Warning, some credentials needed by the monitoring are missing or invalid:
  DiracProxy(group='lsst_user')
Call `credential_store.renew()` to update them.

Generating proxy...
Enter Certificate password:
Requested group lsst_user is not valid for DN /DC=ch/DC=cern/OU=Organic Units/OU=Users/CN=rcurrie/CN=675917/CN=Robert Andrew Currie
ERROR    Error: CredentialRenewalError:  ('Failed to create DIRAC proxy',)```

I can successfully create a proxy with:
`source /cvmfs/ganga.cern.ch/dirac_ui/bashrc && /cvmfs/ganga.cern.ch/dirac_ui/scripts/dirac-proxy-init -g lsst_user -M -t`
so I'm guessing I need to point the branch to the correct DIRAC setup in some new way?

**Comment by milliams** [comment]
The config should be automatically loaded via the usual DIRAC environment stuff. Have you set `[DIRAC]/DiracEnvSource` in your config file? If so then what does:
```python
from GangaDirac.Lib.Utilities.DiracUtilities import getDiracEnv
getDiracEnv()['DIRAC']
```
give you?

rob-c:
Yes, without this branch I can submit jobs. Either something is interfering with the order in which my `.ini` files are loaded or something else a bit funny is going on. I'm investigating atm.
Still no luck on why the AFS credentials are having issues. Is there a test for this?

rob-c:
for some reason my default env and not the expected bashrc is loaded when I try to run with `GangaLSST` loaded using the appropriate `.ini`.
Tracking down this AFS problem is looking like it's going to take a while so I'll target this first as until that's fixed I have to make horrible hacks in the monitoring to do anything else.

rob-c:
Is there a plan to put `Ganga/GPIDev/Credentials2/ICredentialInfo.py` into the `Ganga/GPIDev/Adapters/`?

rob-c:
OK the monitoring loop expects a True/False return on a credential renewal.
From what I can tell the new credentials system always returns `None` but throws on any and all exceptions and failures am I correct?

In which case I can make a small PR or 2 to fix this and another fairly serious bug I spotted in the Monitoring code.

rob-c:
OK, I can create a credential now, there was something screwing with my environment from when I was running the `LHCbDIrac` externals earlier.

rob-c:
OK, aside from needing a sane startup check for the group and fixing the credentials problem with the AFS checking I think we're OK, I can complete a job which was already in my repo from earlier. Now just to test submitting a new one from scratch and playing around with enabling/disabling the proxy.

I also think I'll give the disk space checker a quick look as this should also be running I think.

rob-c:
If I've followed this correctly I think we need to make a smaller PR to add a `DiracProxy` and `VomsProxy` object into the schema in 6.2.x in order to enable some level of backwards compatibility?

python/Ganga/test/GPI/Credentials/TestCredentials.py
```diff
@@ -32,12 +32,9 @@ def test_voms_proxy_life_cycle(gpi):
         _ = credential_store[VomsProxy()]
 
     default_cred = credential_store.create(VomsProxy())
-    lhcb_cred = credential_store.create(VomsProxy(vo='lhcb'))
-    assert len(credential_store) == 2
     explicit_default_cred = credential_store.create(VomsProxy(vo=getConfig('LCG')['VirtualOrganisation']))
     assert explicit_default_cred == default_cred
-    assert default_cred != lhcb_cred
-    assert len(credential_store) == 2
+    assert len(credential_store) == 1
```
rob-c: Is it possible to just add a `FakeProxy` here to check that the length is 2 rather than just dropping the `'lhcb'` proxy?

milliams: I would be, yes. I'll see about importing it across.

python/Ganga/GPIDev/Adapters/IBackend.py
```diff
@@ -511,3 +513,32 @@ def updateMonitoringInformation(jobs):
...
+            cred = credential_store[cred_req]
+            if not cred.is_valid():
+                logger.debug('Required credential %s is not valid', cred)
+                needed_credentials.add(cred_req)
+                continue
+            jobs_by_credential[cred].append(j)
+        except KeyError:
+            logger.debug('Required credential %s is missing', cred_req)
+            needed_credentials.add(cred_req)
+    return list(jobs_by_credential.values())
```
rob-c: This is a list already isn't it?

python/Ganga/GPIDev/Adapters/IBackend.py
```diff
@@ -511,3 +513,32 @@ def updateMonitoringInformation(jobs):
...
+    for j in jobs:
+        try:
+            cred_req = j.backend.credential_requirements
+            cred = credential_store[cred_req]
+            if not cred.is_valid():
+                logger.debug('Required credential %s is not valid', cred)
+                needed_credentials.add(cred_req)
+                continue
+            jobs_by_credential[cred].append(j)
+        except KeyError:
```
rob-c: Doing this catches problems from `credential_store` as well as `jobs_by_credential` from reading the code.
Which of these are expected to be checked here?

python/Ganga/GPIDev/Adapters/IBackend.py
milliams: It is in Python 2 but in Python 3 it becomes a dictionary view object so I'm covering this feature in a forwards-compatible way.

python/Ganga/GPIDev/Adapters/IBackend.py
milliams: `jobs_by_credential[something]` will not raise a `KeyError` as it is a `defaultdict` so it will always return a `list` of some kind.

milliams:
There shouldn't be any need to include `VomsProxy` and `DiracProxy` in old versions since running a 6.3 job in 6.2 will silently ignore the attribute and running a 6.2 job in 6.3 will just use the default values. This is no different to any other new `GangaObject` or schema change.

python/Ganga/GPIDev/Credentials2/CredentialStore.py
```diff
@@ -0,0 +1,259 @@
...
+        Args:
+            credential_object (ICredentialInfo):
+        """
+
+        self.credentials.remove(credential_object)
+
+    def __str__(self, interactive=False):
+        # Get the string information with which to fill the table
+        headers = ['Type', 'Location', 'Valid', 'Time left']
+        cred_info = [[str(f) for f in (type(cred).__name__, cred.location, cred.is_valid(), cred.time_left())] for cred in self.credentials]
```
rob-c: Can this be made more readable. I struggle to follow the full set of logic here due to the 2 for loops

milliams: I can certainly try to format it in a nicer way. The code in this function could perhaps be pulled out into a function in utils since most of it is just doing table formatting?

rob-c:
@milliams Ganga still fully evaluates the __entire__ XML file even if the attribute is lost when an object to disk again in the future. The thing that happens is that ganga doesn't care about assigning the attribute (I still think this is not great but c'est la vie). But if Ganga cannot construct the object it bails out the same as if I try to load a LHCb job when I haven't imported the LHCb Plugin.

I get errors which are captured from within the index generation and file loading which causes the file to revert to an earlier state and job corruption to occur in some cases and jobs to fail to load entirely in others if the backup also contains the proxy object in the XML. This is bad and we should try to either warn users or `fix` 6.2.x so that this doesn't occur. This can be fixed by adding 2 lightweight objects which can't be made at the proxy layer within Ganga (possibly entirely in the plugin system and hidden from the user).

python/Ganga/GPIDev/Credentials2/CredentialStore.py
```diff
@@ -0,0 +1,259 @@
...
+
+        if not isinstance(query, ICredentialRequirement):
+            raise TypeError('Credential store query should be of type ICredentialRequirement')
+
+        match = self.match(query)
+        if match:
+            return match
+
+        try:
+            cred = self.create(query, create=False, check_file=True)
```
rob-c: Is there a reason this tries to create a credential when non exists if it's not creating a new file on disk?
I've probably completely missed something obvious here so apologies if this is the case I'm just wondering what the logic is?

python/Ganga/GPIDev/Credentials2/CredentialStore.py
```diff
@@ -0,0 +1,259 @@
...
+            if not cred.is_valid() or cred.time_left() < timedelta(hours=1):
+                cred.renew()
+        for cred_req in needed_credentials - self.credentials:
+            try:
+                self[cred_req].renew()
+            except KeyError:
+                self.create(cred_req)
+
+    @export
+    def clear(self):
```
rob-c: Should this call some destroy method on credentials within the system rather than just forgetting about them?
I don't know the pros/cons of this but it just seems like a sensible question that I don't know the answer to,

python/Ganga/GPIDev/Credentials2/ICredentialInfo.py
```diff
@@ -0,0 +1,192 @@
...
+
+def cache(method):
+    """
+    The cache decorator can be applied to any method in an ``ICredentialInfo` subclass.
+    It stores the return value of the function in the ``self.cache`` dictionary
+    with the key being the name of the function.
+    The cache is invalidated if ``os.path.getmtime(self.location)`` returns
+    greater than ``self.cache['mtime']``, i.e. if the file has changed on disk
+
+    Not having to call the external commands comes at the cost of calling ``stat()``
```
rob-c: This is probably nothing but is this safe on AFS systems? I'm probably being overly paranoid here as I think proxies are normally stored in `/tmp/` on disk which shouldn't be a distributed fs.

python/Ganga/GPIDev/Credentials2/CredentialStore.py
milliams: This line is attempting to create an `ICredentialInfo` object which would try to wrap an existing credential file on disk if one exists. When Ganga starts, the credential store is empty even if there are proxy files in `/tmp`. This will try to wrap one of those files and add an entry in the credential store for it.

python/Ganga/GPIDev/Credentials2/CredentialStore.py
milliams: I can't think of a time when the user will want to destroy *all* credentials that are recorded in Ganga. Particularly since this would also destroy the AFS token for example. If they really want to do so anyway then:
```python
for cred in credential_store:
    cred.destroy()
```
would do the trick.

python/Ganga/GPIDev/Credentials2/ICredentialInfo.py
milliams: All the credentials that we care about so far (grid proxies and AFS tokens) are stored in `/tmp` so those will be safe. If someone implemented a credential to store, for example, some other username/password in a local file on AFS then I think it should still be fine since AFS supports `mtime`.

python/GangaDirac/Lib/Credentials/DiracProxy.py
```diff
@@ -0,0 +1,108 @@
...
+        Creates the grid proxy.
+
+        Raises:
+            CredentialRenewalError: If the renewal process returns a non-zero value
+        """
+        group_command = ''
+        logger.debug('require ' + self.initial_requirements.group)
+        if self.initial_requirements.group:
+            group_command = '--group %s --VOMS' % self.initial_requirements.group
+        command = 'dirac-proxy-init --out %s %s' % (self.location, group_command)
```
rob-c: I think we may need to add the `-t` option here to be strict about our requirement to generate a correct proxy and to avoid problems when DIRAC accepts a bad one.
(unless this comes in through the group attribute above in which case feel free to ignore me :)

milliams: That's a good idea, I'd forgotten about the `--strict` option. I'll put that in now.

python/GangaTest/Framework/utils.py
```diff
@@ -46,7 +46,7 @@ def sleep_until_state(j, timeout=None, state='completed', break_states=None, sle
     current_status = None
     while j.status != state and timeout > 0:
         if not monitoring_component.isEnabled():
-            monitoring_component.runMonitoring( jobs=jobs.select(j.id,j.id), _loadCredentials=False )
+            monitoring_component.runMonitoring( jobs=jobs.select(j.id,j.id))
```
rob-c: There's inconsistent whitespace here

milliams: I was simply removing an argument from this function so didn't edit the whitespace at all. I'll make a fix to it now though.

python/GangaPanda/Lib/Panda/Panda.py
```diff
@@ -1053,7 +1065,8 @@ def master_resubmit(self,jobs):
         for job in jobs: 
             jobIDs[job.backend.id] = job
 
-        rc,jspecs = Client.getFullJobStatus(jobIDs.keys(),False)
+        with inject_proxy(self.credential_requirements):
+            rc,jspecs = Client.getFullJobStatus(jobIDs.keys(),False)
```
rob-c: whitespace after `,` I think

rob-c:
What is the timeline for removing the old credentials code as part of this PR? I think functionally we're mostly there we just need to move some files and remove the unused old code.

milliams:
This PR removes as much of the old code as it can. As long as ATLAS is not fully ported, we can't remove all of it. I've left some code to aid @drmarkwslater's porting process but once that's done the whole module can be removed.

rob-c:
I was under the impression that as we are _not_ shipping with Atlas support we can just remove this and have a nicer transition in 6.3.0. with Atlas support coming in something like 6.4.0.
@drmarkwslater What are your thoughts on this?

This aside I would still rather see the work to integrate this properly with the core done, I still think exceptions belong with other core exceptions as do interface classes.
This is the design we have adopted and used elsewhere to stick to the conventions already in the project at the expense of the modularity of new features.

drmarkwslater:
Apologies for not being very involved with this - the start of term has proved rather hectic! Basically, there are two options from my point of view:

1) Leave things in but prepare the PR for removal later on

2) Take things out now and I'll refer back to previous code/this PR to do the Atlas stuff

I would have a preference for 1 as it makes things easier for me but probably 2 is better/cleaner. Maybe keep a branch around without the removal that I can refer to but add the actual removal to this PR?

rob-c:
I think I'm leaning toward 2 as there is no chance the old code is loaded if it's not needed to and avoid any possibility of confusion if we're left trying to debug some problems if any crop up. I know it's unlikely but I'm aware that user-support is going to reduce quite quickly over the next few weeks.

Would it be worth branching the code when it's finished (I think it is now?) as a reference and they do the work to integrate this fully with Core?
(I imagine aside from moving a few files and changing a few import statements nothing much changes).

milliams:
At minumum I will push a commit which moves `Credentials2` to `Credentials`. For now I have moved the legacy `Credentials` to `Credentials_old` and have updated all imports to match. If we want to delete `Credentials_old` before releasing 6.3, we can but I don't see that we need to rush doing so.

Next I will move the exceptions to the central place and the interfaces too.

rob-c:
OK, if we want to keep around the old code but we're sure it's not loaded then I'm very happy, I just want to remove the possibility of something somewhere loading from the old Credentials folder (although I'm sure it doesn't happen after a quick grep of the codebase).
Sounds like a good plan 👍

milliams:
Ok, I think this is now in a mergeable state. I'd like to get this in so that we can move towards a 6.3 release as soon as possible. Does anyone have any outstanding problems or other comments?

python/Ganga/Core/exceptions.py
```diff
@@ -270,6 +270,34 @@ def listFailedJobs(self):
     def getOriginalJobError(self, id):
         return self.details.get(id)
 
+
+class CredentialsError(GangaException):
+    """
+    Base class for credential-related errors
+    """
+    pass
```
rob-c: I don't think the pass is strictly needed here and I have seen several tools autostrip this from codebases.

rob-c:
I'm checking something for LHCb can I get 20min or so

rob-c:
also, @milliams I've not looked at the DiracProxy but does the auto-detection of an existing proxy check the group so that I wouldn't accidentally use a lhcb proxy with an lsst system or vice versa?

rob-c:
OK, this must just be me being stupid but what is the new equivalent to `gridProxy.destroy()`. This is useful for developers but not 100% essential for users to have.
Also, is it possible to list and access individual credentials in the store without coming up with a complex requirement for which credential I want to know about. i.e. accessing the DiracProxy from the command line quickly.

python/GangaDirac/Lib/Credentials/DiracProxy.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @cache
+    def expiry_time(self):
+        status, output, message = self.shell.cmd1('voms-proxy-info -file "%s" -timeleft' % self.location)
+        if status != 0:
+            return datetime.now()
+        return datetime.now() + timedelta(seconds=int(output))
+
+    def default_location(self):
+        return os.getenv('X509_USER_PROXY') or '/tmp/x509up_u' + str(os.getuid())
```
rob-c: Can we put this in the proper tmp directory which is defined differently on different systems and allows for `/tmp/<username>` on systems like lxplus. This means standard tools don't pick up a proxy by default and there are problems with users knowing which file is theirs given the random name. Also `uuid.uuid4` means there is a smaller chance of collision between users from what I've read but this becomes mitigated if the file is placed in the correct user's temp folder.

rob-c:
OK, the standard `voms-proxy-destroy` doesn't work. If we can't destroy the Dirac proxy easily I can see us having this requested of us.

So before we merge can we do:
```
 1. Fix the temp folder used to get the temp folder from the system config.
 2. If it's not there (I'm about to read the docs to check. Add some way of listing and accessing known credentials which would allow a user to do something like `credential_store.list()` and access one by name or some id to that it's easy to renew/create/destroy a proxy on command such as when an LHCb user creates an invalid proxy externally and then gets it picked up via Ganga.
```

python/GangaDirac/Lib/Credentials/DiracProxy.py
milliams: This *is* using that standard location for proxies as closely as we can. The reason that normal tools don't pick up our proxies is because we encode some information about them in the file name so for example an LHCb proxy will be called `/tmp/x509up_u34874:lhcb` while a GridPP proxy will be `/tmp/x509up_u34874:gridpp`. I have deliberately named these as closely as possible as the usual names to make them easy to find. There is no risk of collision between users as every proxy is named after the user who created it. For example, log onto any `lxplus` node and do `ls -hlt /tmp/x509up_u*`:
```
-rw-------. 1 anmehta  zh     7.3K Sep 29 14:04 /tmp/x509up_u44856
-rw-------. 1 rwatari  zp     9.0K Sep 29 09:50 /tmp/x509up_u77518
-rw-------. 1 tkunigo  zp     9.0K Sep 29 05:54 /tmp/x509up_u49252
-rw-------. 1 spigazzi zh     7.7K Sep 28 17:40 /tmp/x509up_u68758
-rw-------. 1 kmistry  zp     9.1K Sep 28 17:32 /tmp/x509up_u55942
...
```

milliams:
The existing proxy check will indeed ensure that there is a match against the correct VO or DIRAC group.

You can list all credentials in the store by simply typing `credential_store`. You can iterate through all credentials in the normal way so you can do whatever you want to them.

You can destroy any credential by calling `destroy()` on it. `voms-proxy-destroy` by itself will of course not work since if there are multiple proxies then it does not know which to delete, also they are stored is slightly non-standard places.

An externally created proxy will only be picked up by Ganga if it matches the requirements. If it really is dodgy in some unspecified way then it can destroyed as mentioned above.

It's worth noting that `voms-proxy-destroy` is really just a call to `rm`, it's nothing special.

rob-c:
@milliams There have been issues on some lxplus systems where the sheer number of files in `/tmp` gets unmanageable (I've been on systems where the inode limit has been reached!). Given we're already giving the files custom names I think it's not unreasonable to ask for the tmp folder as defined by the environment.
For users who can't cope with setting the X509 variables in the env themselves asking them to `ls -l` and grep the results will just add an overhead which is detrimental to LHCb from a support perspective.
Yes we're doing good clever things to set the environment we pass on from Ganga to use the correct files but we should really place these in a much more sensible place imo.
Can we at least have the folder name configurable for lhcb so I can look to add some support for other lhcb workflows this interrupts by having custom filenames.

How on earth do you access a given credential without iterating through them and checking the object type?
I would have expected to be able to do something like: `credential_store[1].destroy()` or `credential_store[1].renew()` not being able to do this is annoying and will annoy some uses.

milliams:
You cannot access a credential by an integer index since they are not ordered in any way. If you want to destroy your DIRAC proxy then doing
```python
credential_store[DiracProxy()].destroy()
```
will work.

As for putting the credentials in `env['TMPDIR']` I can do that easily. I'll add that now.,

milliams:
I've addressed the issues above and made the handling of credentials in the store which have been destroyed a bit more sensible. I'm happy to merge unless there are any more comments?

rob-c:
Looks OK from me. No access to a computer atm but looks good to go in, great work from submitting test jobs earlier in the week.

rob-c:
In the case of `DiracProxy` is there a reason why we aren't making use of proxy inspection (parsing `dirac-proxy-info`) to determine the group of existing proxies?
If we just place the proxy on disk with the same format as the normal tools we reduce a lot of hassle with interacting with other tools such as `lhcb-proxy-init` which users have been trained to use.
I'm trying to come up with some solution(s) involving symlinks for LHCb but each solution has it's downsides.

PR #691, opened 2016-07-26
GaudiRun improvements

rob-c:
Firstly this PR is slightly larger than it needs to be. That being said there are bugs that this PR fixes so I'll break these out and merge these before the PR goes in, but I fully intend to merge all of the changes here in some form unless there are any strong objections.

The GaudiRun app has been heavily modified since the 6.1.22 release. The main changes are:

 1. The Schema has now been almost entirely rewritten
 2. A unique WN script is generated ahead of each job as is a unique `extraOpts.py` python file
 3. The extraOpts are imported before GaudiPython wrapper script executes the main payload. This is the best behaviour I can come up with to support extraOpts here.
 4. Splitting jobs has been tested and appears working
 5. ~~Should I rename the app to `GaudiExec` ?~~
 6. Only 2 files per master job are uploaded to DIRAC which speeds up job submission with many almost identical subjobs. The subjob is quided how to run the job on the WN.
 7. Still same code run on Localhost and Dirac WN 
 8. I've moved the code to the correct locations within GangaLHCb
 9. I've renamed `GauiRun` to `GaudiExec`

The main job format I'm using to test this looks like:
```
j=Job(application=GaudiExec(options=LocalFile('~/ntuple_options.py')))
j.application.directory='$HOME/DaVinciDev_v40r2'
j.outputfiles=[LocalFile('*.root')]
j.backend=Dirac()
j.application.extraOpts='print(\'hello\')';
j.splitter=GenericSplitter();j.splitter.attribute='inputdata'
j.splitter.values=[[DiracFile(lfn='/lhcb/MC/2012/ALLSTREAMS.DST/00035742/0000/00035742_00000002_1.allstreams.dst')] for _ in range(2)]
j.application.useGaudiRun=True
j.parallel_submit=True
j.submit()
```
I've tested this with Dirac/Local and with `LocalFile` and `DiracFile` outputs

TODO:
 0. Clean up non GaudiRun changes in PR to other smaller PR
 1. ~~Move App into the _normal_ app folders within LHCb (and rename)~~
 2. Fix problem where `DiracFile`s generated on submit remain locally and aren't cleaned up.
 3. Add more docs for how newer CMake bender apps etc can inherit from this

This is close to being merged but I imagine there may be some breakage and/or code cleanup to do but this is the majority of the work I'd like to see done for 6.1.23 on GaudiRun.


egede:
@rob-c This looks good overall.

The name `GaudiExec` is not great but I have failed to come up with a better idea. And it is better than the now slightly misleading `GaudiRun`.

> The subjob is quided how to run the job on the WN.

I do not understand this.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -57,40 +65,62 @@ class GaudiRun(IPrepareApp):
...
         # Options created for constructing/submitting this app
-        'directory':    SimpleItem(preparable=1, defvalue=None, typelist=[None, str], comparable=1,
-            doc='A path to the project that you\'re wanting to run.'),
-        'build_opts':   SimpleItem(defvalue=[""], typelist=[str], sequence=1, strict_sequence=0,
-            doc="Options to be passed to 'make ganga-input-sandbox'"),
-        'myOpts':       GangaFileItem(defvalue=None, doc='File which contains the extra opts I want to pass to gaudirun.py'),
-        'uploadedInput': GangaFileItem(defvalue=None, doc='This stores the input for the job which has been pre-uploaded so that it gets to the WN'),
-        'runWithPython':SimpleItem(defvalue=False, doc='Should \'myOpts\' be run as "python myOpts.py data.py" rather than "gaudirun.py myOpts.py data.py"'),
+        'directory':    SimpleItem(defvalue=None, typelist=[None, str], comparable=1, doc='A path to the project that you\'re wanting to run.'),
+        'options':       GangaFileItem(defvalue=None, doc='File which contains the extra opts I want to pass to gaudirun.py'),
```
egede: Remove the word "extra" from the documentation as it is confusing with the `extraopts` below.

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
```diff
@@ -179,70 +279,68 @@ def configure(self, masterappconfig):
...
-
-    def getEnvScript(self):
         return 'export CMTCONFIG=%s; source LbLogin.sh --cmtconfig=%s && ' % (self.platform, self.platform)
 
-    def exec_cmd(self, cmd):
+    def execCmd(self, cmd):
         """
         This method executes a command within the namespace of the project. The cmd is placed in a bash script which is executed within the env
+        This will adopt the platform associated with this application and all commands requiring to be run within the project env
+        will be required to be executed with the './run' explcitly called
```
egede: Give an explicit example or two here. Presumably `j.application.execCmd('make')` is the most obvious one.

rob-c: Will do :+1:

python/GangaLHCb/Lib/GaudiRunApp/GaudiRunApp.py
rob-c: Thanks for catching this I'm starting to miss the wood for the trees when looking at the Schema now I've been working on it for too long.

rob-c:
@egede Regarding `The subjob is guided how to run the job on the WN.` 
Apologies I wasn't exactly clear about this.

Each subjob has a unique job script and extra options file created with a unique name. In the case of the Local backends these are generated as appropriate and reach the WN via the standard machinery.

In the case of DIRAC these scripts are now all generated at the same time before submitting any jobs and stored in a single archive.
This archive is then sent to the WN where all of the scripts are auto-extracted (into a folder) and DIRAC is simply told which script to run for this job.

This means that the JDL on a per job basis is just a set of instructions for DIRAC to know what files to pick up from storage before running the job and there is no need to upload an inputsandbox which just contains the script for the job to run.

egede:
@rob-c That is great.

rob-c:
OK, I spotted a bug where I was accidentally mis-mapping the scripts between subjobs. I'll push the fix for this and a few other things then I'll look to extract the changes from this PR which aren't GaudiRun and make some additional PR to merge them.

rob-c:
OK I've migrated the code to GaudiExec from GaudiRun, and made 2 other PR to factor out the changes in Core and GaudiDirac.
The rest of the changes are best left in this PR although I need to migrate the old GaudiRun test as well as  fixing the problem of not unlinking the files uploaded to DIRAC storage during job submission.

rob-c:
OK, once I've ported the old GaudiRun test I plan to put this up for review for merge, but I don't expect any of the other code to change so people can feel good to review this PR.

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -38,73 +51,106 @@ class GaudiRun(IPrepareApp):
...
+        actual_value = value
+        if attr == 'directory':
+            if value:
+                actual_value = path.abspath(fullpath(expandfilename(value)))
+        elif attr == 'options':
+            if isinstance(value, str):
+                new_file = allComponentFilters['gangafiles'](value, None)
+                actual_value = [ new_file ]
+            elif isinstance(value, IGangaFile):
+                actual_valye = [ value ]
```
KonstantinSchubert: Typo: valye

rob-c: Thanks, will fix

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -179,88 +290,93 @@ def configure(self, masterappconfig):
...
-
-    def getEnvScript(self):
         return 'export CMTCONFIG=%s; source LbLogin.sh --cmtconfig=%s && ' % (self.platform, self.platform)
 
-    def exec_cmd(self, cmd):
+    def execCmd(self, cmd):
         """
         This method executes a command within the namespace of the project. The cmd is placed in a bash script which is executed within the env
+        This will adopt the platform associated with this application and all commands requiring to be run within the project env
+        will be required to be executed with the './run' explcitly called
```
rob-c: I'll update this as this is no longer correct...

rob-c:
OK, looking to merge whenever ready now, I've fixed a few minor bugs and have updated the integration test to point to `GaudiExec` rather than `GaudiRun` and have hanged the API approriately

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -1,31 +1,42 @@
...
 
-
-class GaudiRun(IPrepareApp):
+class GaudiExec(IPrepareApp):
     """
+
     Welcome to the new GaudiApp for LHCb apps written/constructed making use of the new CMake framework
 
+    Requirements:
+    ===============
```
milliams: In [RestructuredText](http://www.sphinx-doc.org/en/stable/rest.html#sections), the header underlines should match the length of the title.

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -38,73 +49,108 @@ class GaudiRun(IPrepareApp):
     The output from this command can be quite large and Ganga will save it to disk and store it at least once per (master) job
     If your build target is large I would advise that you consider placing your gangadir in your AFS workspace where there is more storage available
 
+
+    Usage:
+    ===============
```
milliams: As above.

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -38,73 +49,108 @@ class GaudiRun(IPrepareApp):
...
 
-    For GaudiPython style of running:
+
+    How it works:
+    ===============
 
     The actual command run on the WN is:
 
-        ./run gaudirun.py myOptsFile.py data.py
+        ./run gaudirun.py optionsFile.py data.py
```
milliams: Make code like:
```rst
.. code-block:: bash

    ./run gaudirun.py optionsFile.py data.py
```

rob-c: What exactly does this do? Does this impact what the user sees from `help(myApp)`?

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -154,9 +201,76 @@ def prepare(self, force=False):
...
+
+        df = master_job.application.sharedOptsInput
+
+        folder_dir = master_job.getInputWorkspace(create=True).getPath()
+
+        if not df or df.namePattern == '':
+            master_job.application.sharedOptsInput = LocalFile(namePattern=GaudiExec.sharedOptsFile_name, localDir=folder_dir)
+            tar_filename = path.join(folder_dir, GaudiExec.sharedOptsFile_name)
+            if not path.isfile(tar_filename):
+                with tarfile.open(tar_filename, "w") as tar_file:
```
milliams: There's no need for an ` as tar_file` if it's not being referenced.

python/GangaLHCb/Lib/Applications/GaudiExecUtils.py
```diff
@@ -8,14 +11,56 @@
...
+importOptions('%s')
+
+# Run the command
+execfile('%s')
+""" % (sys_args, options, data_file, script_file)
+    return wrapper_script
+
+def getTimestampContent():
+    """
+    Returns a string containing the current time in a gven format and a unique random uuid
```
milliams: 'gven'→'given'

python/GangaLHCb/Lib/LHCbDataset/LHCbDatasetUtils.py
```diff
@@ -38,13 +38,13 @@ def strToDataFile(name, allowNone=True):
...
             raise GangaException( "Cannot construct file object: %s" % str(name) )
 
-def getDataFile(file):
-    if isType(file, DiracFile):
-        return file
-    if isType(file, LocalFile):
-        return file
-    if type(file) == type(''):
-        return strToDataFile(file)
+def getDataFile(_file):
```
milliams: PEP 8 suggests `single_trailing_underscore_` is "used by convention to avoid conflicts with Python keyword".

python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -223,29 +231,64 @@ def generateDiracInput(app):
...
+    """
+
+    new_df = DiracFile(namePattern, localDir=localDir)
+    random_SE = random.choice(getConfig('DIRAC')['allDiracSE'])
+    new_lfn = os.path.join(getInputFileDir(job), namePattern)
+    returnable = new_df.put(force=True, uploadSE=random_SE, lfn=new_lfn)[0]
+
+    return returnable
+
+def getInputFileDir(job):
```
milliams: There should be two blank lines between module-level constructs.

python/GangaLHCb/Lib/Applications/GaudiExec.py
milliams: It makes it be formatted correctly when generating the Sphinx docs (syntax highlighting in a code block with a proportional font). You're right though that the in-built Ganga docs don't understand it. In that case, the compromise of:

```rst
The actual command run on the WN is::

    ./run gaudirun.py optionsFile.py data.py
```
will work on both (note the double colon). It won't syntax highlight in Sphinx but it will put it in a proportional font.

rob-c:
At this point I think I'm good to merge unless anyone has any objections?

milliams:
:+1:

Issue #692, opened 2016-07-26
Error in recursive proxy strip for unicode Type


alexanderrichards:
While writing code for the LZ collaboration I get the following error while running Tasks:

```python
ERROR    Exception occurred in task monitoring loop: <type 'exceptions.AttributeError'> 'unicode' object has no attribute 'append'
The offending task was paused.                                                                                                                                                                                                                
ERROR    Full traceback:
 Traceback (most recent call last):                                                                                                                                                                                                           
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Lib/Tasks/TaskRegistry.py", line 92, in _thread_main                                                                                                                                
    p.update()                                                                                                                                                                                                                                
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Lib/Tasks/ITask.py", line 99, in update                                                                                                                                             
    if trf.update() and not self.check_all_trfs:                                                                                                                                                                                              
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Lib/Tasks/ITransform.py", line 235, in update                                                                                                                                       
    self.createUnits()                                                                                                                                                                                                                        
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Lib/Tasks/CoreTransform.py", line 58, in createUnits                                                                                                                                
    subjobs = self.unit_splitter.split(j)                                                                                                                                                                                                     
   File "/home/hep/arichard/git/ganga/python/Ganga/Lib/Splitters/GenericSplitter.py", line 117, in split                                                                                                                                      
    setattr(obj, attr, vallist[i])                                                                                                                                                                                                            
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Base/Proxy.py", line 1030, in _setattr                                                                                                                                              
    GPIProxyObject.__setattr__(self, x, new_v)                                                                                                                                                                                                
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Base/Proxy.py", line 651, in __set__                                                                                                                                                
    final_val = ProxyDataDescriptor._process_set_value(raw_obj, _val, attr_name)                                                                                                                                                              
   File "/home/hep/arichard/git/ganga/python/Ganga/GPIDev/Base/Proxy.py", line 569, in _process_set_value                                                                                                                                     
    val = ProxyDataDescriptor.__recursive_strip(_val)                                                                                                                                                                                         
[...]

PR #693, opened 2016-07-26
Unicode proxystrip

alexanderrichards:
This PR resolves https://github.com/ganga-devs/ganga/issues/692. Essentially checking for base type `basestring` rather than `str` such that unicode objects are also caught. It also includes some code clean up as mentioned in the above issue.


rob-c:
retest this please

PR #694, opened 2016-07-26
Code cleanup and LocalFile construction fix

rob-c:
This is some minor code cleanup and fixing a minor bug in LocalFile.

I'm planning to merge this tomorrow unless there are any objections

### DIFF
--- a/python/Ganga/GPIDev/Lib/File/LocalFile.py
+++ b/python/Ganga/GPIDev/Lib/File/LocalFile.py
@@ -62,6 +62,9 @@ class LocalFile(IGangaFile):
         else:
             logger.error("Unkown type: %s . Cannot Create LocalFile from this!" % type(namePattern))
 
+        if localDir:
+            self.localDir = localDir
+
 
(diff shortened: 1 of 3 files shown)


python/Ganga/GPIDev/Lib/File/LocalFile.py
```diff
@@ -62,6 +62,9 @@ def __init__(self, namePattern='', localDir='', **kwds):
         else:
             logger.error("Unkown type: %s . Cannot Create LocalFile from this!" % type(namePattern))
 
+        if localDir:
+            self.localDir = localDir
```
milliams: Isn't `self.localDir` already set by the `if`/`elif`/`else` branches above? The only one that doesn't cover it is the first so perhaps it should be moved there.

rob-c: I'l test if this fixes my problems then make the change if it does

milliams:
As in #695, this needs a PR title describing the change. Ignoring that, :+1:

PR #695, opened 2016-07-26
Fixes for DIRACserver job kill problem and DiracFile used as Batch/Local outputfile

### LINKED ISSUE #685: GangaDirac not handling externally killed jobs well
Some users are reporting that jobs killed in the (LHCb?)DIRAC system are causing problems in the monitoring loop and triggering:
`ERROR    Unexpected dirac status 'killed' encountered`

I think this shouldn't take too long to track down and add a try/except handle around but the difficulty could be if this is being thrown from within a bulk request against DIRAC which would then need to be repeated.

rob-c:
These are some changes broken out from #691

I've confirmed the changes in `GangaDirac/__init__.py` Fix #685 

I've also made several small changes to the DiracFile and it's scripts which are inserted into jobs. This fixes the DiracFile so that it works correctly as an outputfile for Local jobs.

In addition I've fixed some minor bugs where the DiracFile wasn't uploaded to the correct LFN.

I think this needs to be merged given the DiracFile is currently broken (I suspect a lot of changes due to the `__construct__` changes).

### DIFF
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -667,14 +667,11 @@ class DiracFile(IGangaFile):
                 logger.warning("LFN will be generated automatically")
                 self.lfn = ""
 
-        if self.remoteDir == '':
+        if not self.remoteDir:
             try:
                 job = self.getJobObject()
-            except AssertionError:
-                job = None
-            if job:
                 lfn_folder = os.path.join("GangaUploadedFiles", "GangaJob_%s" % job.getFQID('.'))
-            else:
+            except AssertionError:
                 t = datetime.datetime.now()
                 this_date = t.strftime("%H.%M_%A_%d_%B_%Y")
                 lfn_folder = os.path.join("GangaUploadedFiles", 'GangaFiles_%s' % this_date)
@@ -838,14 +835,11 @@ for f in glob.glob('###NAME_PATTERN###'):
         WNscript_location = os.path.join( script_path, 'WNInjectTemplate.py' )
         script = FileUtils.loadScript(WNscript_location, '')
 
-        if self.remoteDir == '':
+        if not self.remoteDir:
             try:
                 job = self.getJobObject()
-            except AssertionError:
-                job = None
-            if job:
                 lfn_folder = os.path.join("GangaUploadedFiles", "GangaJob_%s" % job.getFQID('.'))
-            else:
+            except AssertionError:
                 t = datetime.datetime.now()
                 this_date = t.strftime("%H.%M_%A_%d_%B_%Y")
                 lfn_folder = os.path.join("GangaUploadedFiles", 'GangaFiles_%s' % this_date)
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -1,10 +1,12 @@
 import copy
 import os
 import datetime
+import inspect
 import hashlib
 import re
 import os.path
 import random
+import glob
 from Ganga.GPIDev.Base.Proxy import stripProxy, GPIProxyObjectFactory, isType, getName
 from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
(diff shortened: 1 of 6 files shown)


milliams:
Can this be given a title which matches the contents please?

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -661,15 +667,18 @@ def put(self, lfn='', force=False, uploadSE="", replicate=False):
...
+                job = self.getJobObject()
+            except AssertionError:
+                job = None
+            if job:
+                lfn_folder = os.path.join("GangaUploadedFiles", "GangaJob_%s" % job.getFQID('.'))
+            else:
+                t = datetime.datetime.now()
+                this_date = t.strftime("%H.%M_%A_%d_%B_%Y")
+                lfn_folder = os.path.join("GangaUploadedFiles", 'GangaFiles_%s' % this_date)
+            self.lfn = os.path.join(DiracFile.diracLFNBase(), lfn_folder, self.namePattern)
```
milliams: This could simply be:
```python
if not self.remoteDir:
    try:
        job = self.getJobObject()
        lfn_folder = os.path.join("GangaUploadedFiles", "GangaJob_%s" % job.getFQID('.'))
    except AssertionError:
        t = datetime.datetime.now()
        this_date = t.strftime("%H.%M_%A_%d_%B_%Y")
        lfn_folder = os.path.join("GangaUploadedFiles", 'GangaFiles_%s' % this_date)
    self.lfn = os.path.join(DiracFile.diracLFNBase(), lfn_folder, self.namePattern)
```

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -838,24 +838,27 @@ def getWNInjectedScript(self, outputFiles, indent, patternsToZip, postProcessLoc
...
```
milliams: As above

python/GangaDirac/Lib/Files/uploadScript.py
```diff
@@ -1,39 +1,40 @@
...
 
-lfn= os.path.join('###LFN_BASE###', file_name)
+lfn=os.path.join('###LFN_BASE###', file_name)
 wildcard='###WILDCARD###'
 storage_elements=###SEs###
 
-from contextlib import closing
 with closing(open('###LOCATIONSFILE_NAME###','ab')) as locationsfile:
     for se in storage_elements:
+        sys.stdout.write('\nUploading file: \"%s\" as \"%s\" at \"%s\"\n' % (file_name, lfn, se))
```
milliams: Why not:
```python
sys.stdout.write('\nUploading file: "%s" as "%s" at "%s"\n' % (file_name, lfn, se))
```

rob-c:
I felt the title did reflect the changes, I'm happy for someone else to change it if they have a better suggestion.

python/GangaDirac/Lib/Files/uploadScript.py
rob-c: I was wanting to be certain these were escaped, also I was originally wrapping the whole string in `"` rather than `'`.

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: Must admit I'm not a fan of this method throwing exceptions I still feel it's a bit messy. I'll look into making the changes

milliams:
"Some Dirac fixes from #691" doesn't describe anything. All it tells me is that it's fixing *something* in Dirac. *What* does it fix? *What* does it change?

rob-c:
To me that implies the title is short to describe well the changes to encourage you to read the other commit info if you need to know what changed.
As I've said, please feel free to change it to something expressing the changes in the PR I'm not strongly attached to anything I just find that it works for me.

milliams:
To me it suggests that there are too many unrelated changes going into one PR. It was useful to be able to read through the PR titles when composing the release notes but with titles like this it becomes difficult. Since I'm no longer the release manager I guess it's not my problem any more.

rob-c:
(Completely aside from this PR) Are the release notes reviewed before they go into the release I've not seen a PR for them?

PR #696, opened 2016-07-27
Attempt to stop ShareDir making excess folders

### LINKED ISSUE #642: delete `gangadir/shared/USERNAME` directories when job object is `remove()`'d
It appears that the in ganga v601r21-BETA the directories holding the `cmake-input-sandbox.tgz` files do not get deleted when the corresponding job is deleted with `jobs(2637).remove()` in ganga.

What I did is, try to submit a `GaudiRun` job:
```
Created job for the jobs:  GaudiRun_test
Calls splitter
Finally submit the job
INFO     submitting job 2637
INFO     job 2637 status changed to "submitting"
INFO     Preparing GaudiRun application.
INFO     Created shared directory: conf-e153ca7d-178f-4897-a528-1cb2d33d7195
```
(the submission fails, as expected / discussed on the mailinglist)
then call `jobs(2637).remove()` to clean up my afs quota and do
```
find ~/gangadir/shared/pseyfert -name cmake-input-sandbox.tgz
```
to observe that `/afs/cern.ch/user/p/pseyfert/gangadir/shared/pseyfert/conf-e153ca7d-178f-4897-a528-1cb2d33d7195/cmake-input-sandbox.tgz` still exists.

### LINKED ISSUE #662: Excess directories created in `gangadir/shared/`
AFS has a limit of 64K file names per directory. This limit is reduced depending on the length of the file names; it is reduced to less than 22K if the file names are longer than 33 letters.

https://lists.openafs.org/pipermail/openafs-info/2002-September/005812.html

My `gangadir/shared/koschube/` directory contains `31 707` files currently, and I appear to have hit the limit.

It is no longer possible for me to create new jobs in ganga, I am getting errors like:

```
OSError: [Errno 27] File too large: '/afs/cern.ch/user/k/koschube/workspace/gangadir2/shared/koschube/conf-1e3fe559-6b5e-4db3-a9cb-25912de073ed'`
```

Please note that I had started with a completely new gangadir just a few weeks ago and that I have submitted less than 100 jobs in total (!). So this definitely isn't caused by some outrageous use case.

rob-c:
This PR reduces the number of excess folders created which are unused by the `ShareDir` class.

This passes some basic tests locally but this may cause issues over the full test suite.

Fixes #662 
Fixes #642

### DIFF
--- a/python/Ganga/GPIDev/Lib/File/File.py
+++ b/python/Ganga/GPIDev/Lib/File/File.py
@@ -290,9 +290,6 @@ def cleanUpShareDirs():
                     os.rmdir(this_dir)
                 except OSError:
                     logger.debug("Failed to remove: %s" % this_dir)
-                    pass
-
-    return
 
 exportToGPI('cleanUpShareDirs', cleanUpShareDirs, 'Functions')
 
--- a/python/Ganga/GPIDev/Lib/File/File.py
+++ b/python/Ganga/GPIDev/Lib/File/File.py
@@ -137,7 +137,7 @@ class ShareDir(GangaObject):
     the Executable() application. A single ("prepared") application can be associated to multiple jobs.
 
     """
-    _schema = Schema(Version(1, 0), {'name': SimpleItem(defvalue='', doc='path to the file source'),
+    _schema = Schema(Version(1, 0), {'name': SimpleItem(defvalue='', getter="_getName", doc='path to the file source'),
                                      'subdir': SimpleItem(defvalue=os.curdir, doc='destination subdirectory (a relative path)')})
 
     _category = 'shareddirs'
@@ -151,22 +151,10 @@ class ShareDir(GangaObject):
         self._setRegistry(None)
 
         if not name is None:
-            self.name = name
+            self._name = name
         else:
-            # continue generating directory names until we create a unique one
-            # (which will likely be on the first attempt).
-            while True:
-                name = 'conf-{0}'.format(uuid.uuid4())
-                if not os.path.isdir(os.path.join(getSharedPath(), name)):
-                    os.makedirs(os.path.join(getSharedPath(), name))
-
-                if not os.path.isdir(os.path.join(getSharedPath(), name)):
-                    logger.error("ERROR creating path: %s" %
-                                 os.path.join(getSharedPath(), name))
-                    raise GangaException("ShareDir ERROR")
-                else:
-                    break
-            self.name = str(name)
+            name = 'conf-{0}'.format(uuid.uuid4())
+            self._name = name
 
             # incrementing then decrementing the shareref counter has the effect of putting the newly
             # created ShareDir into the shareref table. This is desirable if a ShareDir is created in isolation,
@@ -175,6 +163,27 @@ class ShareDir(GangaObject):
         #shareref = GPIProxyObjectFactory(getRegistry("prep").getShareRef())
         # shareref.increase(self.name)
         # shareref.decrease(self.name)
+
+    def __setattr__(self, name, value):
+
+        if name == 'name':
+            self._name = value
+        else:
+            super(ShareDir, self).__setattr__(name, value)
+
+
+    def _getName(self):
+
+        share_dir = os.path.join(getSharedPath(), self._name)
+        if not os.path.isdir(share_dir):
+            logger.info("Actually creating: %s" % share_dir)
+            os.makedirs(share_dir)
+        if not os.path.isdir(share_dir):
+            logger.error("ERROR creating path: %s" % share_dir)
+            raise GangaException("ShareDir ERROR")
+
+        return self._name
+
 
     def add(self, input):
         from Ganga.Core.GangaRepository import getRegistry
--- a/python/Ganga/GPIDev/Lib/File/File.py
+++ b/python/Ganga/GPIDev/Lib/File/File.py
@@ -25,6 +25,8 @@ import Ganga.Utility.Config
 
 from Ganga.GPIDev.Lib.File import getSharedPath
 
+from Ganga.Runtime.GPIexport import exportToGPI
+
 # regex [[PROTOCOL:][SETYPE:]..[<alfanumeric>:][/]]/filename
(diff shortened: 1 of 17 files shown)


egede:
@rob-c Great you got onto this. Can a small function exposed to the GPI be made which will remove old empy directories from user areas.

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -151,22 +151,10 @@ def __init__(self, name=None, subdir=os.curdir):
...
-
-                if not os.path.isdir(os.path.join(getSharedPath(), name)):
-                    logger.error("ERROR creating path: %s" %
-                                 os.path.join(getSharedPath(), name))
-                    raise GangaException("ShareDir ERROR")
-                else:
-                    break
-            self.name = str(name)
+            name = 'conf-{0}'.format(uuid.uuid4())
+            self._name = name
```
milliams: Why not:
```python
if name is None:
    name = 'conf-{0}'.format(uuid.uuid4())
self._name = name
```

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -176,6 +164,27 @@ def __init__(self, name=None, subdir=os.curdir):
...
 
+    def __setattr__(self, name, value):
+
+        if name == 'name':
+            self._name = value
+        else:
+            super(ShareDir, self).__setattr__(name, value)
+
+
+    def _getName(self):
```
milliams: Should be a single blank line.

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -176,6 +164,27 @@ def __init__(self, name=None, subdir=os.curdir):
...
+        if not os.path.isdir(share_dir):
+            logger.info("Actually creating: %s" % share_dir)
+            os.makedirs(share_dir)
+        if not os.path.isdir(share_dir):
+            logger.error("ERROR creating path: %s" % share_dir)
+            raise GangaException("ShareDir ERROR")
+
+        return self._name
+
+
```
milliams: Should be a single blank line.

rob-c:
@egede I can add a small function into this PR to clean the remote dirs which are already empty.
I'm just surprised that the changes here didn't lead to the whole project falling down around me, I'll have to run some more realistic LHCb jobs to verify that this doesn't break anything in a way we don't catch in the tests.

python/Ganga/GPIDev/Lib/File/File.py
rob-c: Given what this used to be probably not but the old code was just broken

python/Ganga/GPIDev/Lib/File/File.py
rob-c: I also plan to add docs in a further amendment but it's not in yet.

python/Ganga/GPIDev/Lib/File/File.py
rob-c: As above

rob-c:
OK, I've added `cleanUpShareDirs` into the GPI which removes all empty folders from the shared directory.

I'll try and remember to add this to the release notes when 6.1.23 is a bit closer to going out.

Assuming that nothing has broken as a result of the changes I plan to merge this-evening as this should fix the problem that was being seen (I _think_ it does for me at least).

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -257,6 +271,31 @@ def string_sharedfile_shortcut(v, item):
...
+        this_dir = os.path.join(share_path, item)
+        if os.path.isdir(this_dir):
+            # NB we do need to explicitly test the length of the returned value here
+            # Checking is __MUCH__ faster than trying and failing to remove folders with contents on AFS
+            if len(os.listdir(this_dir)) == 0:
+                try:
+                    os.rmdir(this_dir)
+                except OSError:
+                    logger.debug("Failed to remove: %s" % this_dir)
+                    pass
```
milliams: This `pass` is not needed.

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -257,6 +271,31 @@ def string_sharedfile_shortcut(v, item):
...
+
+    return
```
milliams: This `return` is not needed.

KonstantinSchubert:
As an outsider, may I comment that I think it is a bad decision to introduce a new method to the API which only serves to resolve a hopefully temporary issue?

As soon as everybody has updated to the new ganga version, `cleanUpShareDirs` should no longer be needed, right? But if it is in the API, people will grow to depend on it and then it cannot be removed any more. 

If this was my project, I would recommend to remove the `cleanUpShareDirs` method again before the release and instead recommend users to use @egede s one-liner:

    find . -depth -type d -empty -delete

drmarkwslater:
[Coming at this late thanks to my daughter having chicken pox!]

I agree with @KonstantinSchubert - if we're going to have to tell people to run a GPI command anyway, why not just tell them to run the cmd line option and save the additional code (that won't be used again) from being added to the codebase? @egede Is there a reason to prefer a GPI function?

rob-c:
@drmarkwslater My feeling is we then need to educate users to go find the shared directory which they may have moved due to their `.gangarc` it's a trivial thing to find from within Ganga as the path to it is easy to resolve.
@KonstantinSchubert From my perspective the plan is to retire this method within a few versions (possibly even before 6.2.0) and let users who don't update deal with the problem themselves. This is entirely there as a quick call that can be made from the IPython prompt which will be explained in the release notes.

From experience people tend to expect a fix from within Ganga for a Ganga problem, even if the fix requires manual intervention.

(I could even hide this as part of the bootstrap but then we're encouraging extremely bad dev practices imo)

egede:
To have it as a manual command or as a function is a choice. In my mind the function is safer and it doesn't encourage users to fiddle around with deleting files inside the Ganga managed area themselves. We know that migration to new releases is often slow so it would have to stay around for a while though. I am happy to follow the majority here, but I am voting for it as a function.

PR #698, opened 2016-07-28
Cleaning up local files after they've been uploaded to Dirac

rob-c:
This PR deletes the local files generated during the submission of a CMake job to the grid to reduce the files left on disk.

I'm exploring removing the shared directory when it's no longer needed locally but this will be in another PR as I need to download missing files again if we copy a prepared job and run it on a Local/Batch machine vs Dirac.


rob-c:
Unless there are objections I'll merge this after the meeting

PR #700, opened 2016-07-28
Fix GaudiExec prep

rob-c:
I've refactored some of the code to fix a few minor bugs in how the jobs work when a new job is copied.

Now every new job regenerates all of the script files from scratch. However the prepared state of the application remains the same.

I have tried copying a Local job, modifying it's backend to Dirac and resubmitting it and everything now works as expected.

This also makes the code in the master_prepare methods a bit easier to read.

I've also cleaned up the schema slightly so that all of the generated scripts are stored in 1 LFN and all of the scripts for the application prepared state are stored in another.

Now when unpreparing an application the reference to the LFN is set to None and there is no way to automatically remove these files which may be on external storage (this is potentially annoying but not the end of the world).

(aside and looking at 6.1.24/6.2.0)
I've a small additional PR incoming which I'd like to propose which fixes this which I'd like to see in 6.2.0 so we don't have lost files and LFN everywhere. I think I can do this in a fully backwards compatible way so that only future jobs store the information this way. I propose we keep the current system for removing the shared dir, and just add a step to remove additional files not in the sharedDir on disk but are registered with the sharedDir.
(The problem comes in fixing a long standing bug in how we handle these sharedDir objects in XML which will hold the PR up a bit).

### DIFF
--- a/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
+++ b/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
@@ -191,7 +191,7 @@ class GaudiExecRTHandler(IRuntimeHandler):
 
         job_command = prepareCommand(app)
 
-         # Generate the script which is to be executed for us on the WN
+        # Generate the script which is to be executed for us on the WN
         scriptToRun = generateWNScript(job_command, app)
         input_sand.append(scriptToRun)
 
@@ -259,9 +259,7 @@ def generateJobScripts(app, appendJobScripts):
     job = app.getJobObject()
 
     if not job.master:
-        rjobs = job.subjobs
-        if not rjobs:
-            rjobs = [job]
+        rjobs = job.subjobs or [job]
     else:
         rjobs = [job]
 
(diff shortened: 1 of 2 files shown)


python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -205,7 +191,7 @@ def prepare(self, app, appconfig, appmasterconfig, jobmasterconfig):
 
         job_command = prepareCommand(app)
 
-        # Generate the script which is to be executed for us on the WN
+         # Generate the script which is to be executed for us on the WN
```
milliams: Accidental extraneous space I guess?

python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -250,30 +236,80 @@ def generateDiracInput(app):
...
+    """
+
+    job = app.getJobObject()
+
+    if not job.master:
+        rjobs = job.subjobs
+        if not rjobs:
+            rjobs = [job]
+    else:
+        rjobs = [job]
```
milliams: How about:
```python
if job.master:
    rjobs = [job]
else:
    rjobs = job.subjobs or [job]
```
?

rob-c:
Any objections to merging?

PR #701, opened 2016-07-28
Fixing parsing of XML when GangaObject is in SimpleItem

rob-c:
This fixes the bug in the XML that the ShareDir is stored as a string representation of the object rather than the full object in XML serialized form.

This allows us to trivially extend the object and add additional schema attribute to this class (and any other classes stored this way in the future).

As far as I know this breaks nothing as we've made a lot of effort to improve/fix the XML parsing recently so this should be a transparent fix.


rob-c:
If there are no objections I plan to merge this today

milliams:
I don't think that this is the right fix. The correct answer is that these things should be `ComponentItem`s, that's their whole point.

rob-c:
@milliams Unless we plan to change/port all of the existing apps this bug exists. What risk is there to porting an existing app and then loading the existing strings (already in XML files) into the place of the ComponentItem?
Are strings from here correctly parsed into real objects or are they treated as strings? If they are never parsed then this PR could potentially be the only solution I'm thinking.
Are there any dangers with changing the schema attribute type from one Ganga version to the next?

Frankly this solves the problem and deals with it with far less code changes so I'm strongly leaning on this solution for practicality.

rob-c:
Frankly looking at the code generated this generates the same code as the ComponentItem in this case which makes me wonder why we even have the distinction between the 2 schema attribute types at this level if with some logic they can be passed in the same way?
The code to read in the XML has no distinction between Component/Simple items at this level so why should we be enforcing this behavior in the writing logic?

I'm aware there are runtime and config differences between the Component/Simple attributes but in the XML I'm not sure there needs to be?

milliams:
If we take there being no difference between the simple and component items to the logical extreme then we end up not writing XML at all but simply the `VPrinter` output. Writing and parsing the XML is useful as it allows us to put in more information than you would get through the `SimpleItem`'s `repr`/`eval` serialisation cycle. It makes forward and backward compatibility easier since we can construct the schema attributes manually and manipulate them before giving them to the object, we can store the plugin class and schema version too.

All we need to do to fix the `ShareDir` item is to change (in e.g. `Executable`) the `is_prepared` attribute to be defined as:
```python
'is_prepared': ComponentItem('shareddirs', defvalue=None, optional=True,  load_default=False, doc='Location of shared resources...'),
```

I've just tested this and it works as far as I can tell. It works because when reading the XML, the parser bases its decision on how to read the object on the XML text, not the Ganga type so it simple `eval`s it and assigns it to `is_prepared`. Next time it writes it, it will write it as full XML as it should be.

As an aside though, why do we store the `ShareRef` on an application object at all? It stores the location of the share dir but can that not be calculated from the hash of the application? I feel like all `ShareRef` objects should only exist within the prep registry anyway and not also sometimes be attached to job objects.

rob-c:
@milliams I'm not promoting taking it any further than looking at what XML we expect to be generated. This is a corner case and a unique one. In my mind there is no requirement at the XML layer for this distinction to exist, hence the PR.
This is a philosophical/practical difference I fear. I'm preparing a 2nd PR which highlights just why this is much smaller and less intrusive.
If we're not allowing GangaObjects to be stored this way I'm adding an exception as this is a bug which needs to be fixed regardless of the fix.

I'm not sure I follow the only way you can know where an application has shared information stored is to store in some format the location of where the information is? I don't see where the ShareRef is being stored other than the repo. You can prepare an application without a Job if you're wondering why the information is stored here.

rob-c:
After doing some exploring of the code I'm keen to push this as part of 6.1.24 quite soon into the release cycle.
I'm not interested in fixing the numerous problems in the Schema system required by not having something like this in the XML layer and I am keen to start exploring changing the `SharedDir` object which requires these changes.
I see no harm in this PR and it removes an annoying bug which has been around for quite a while now. The results from this PR or fixing it at a higher level are identical from an XML perspective so I see no real motivation for fixing this by fixing half the problems in the Schema.

This may be _related_ to a Schema problems but from my perspective the issue here is that the XML file doesn't contain the correct XML and this should be addressed asap.

egede:
@rob-c @milliams After reading throug the thread here, I am quite happy to merge this. 

@rob-c Are we sure that this does not create problems when switching forwards and backwards between Ganga releases where this fix was implemented?

rob-c:
@egede This doesn't remove the ability to process old XML files it simply means that new XML files moving forward can be processed in a sane way with all of the information recorded properly. The XML parser will still fall back to checking string values in the XML in-case they're actually repr of a GangaObject class.

egede:
OK, just go ahead and merge then,

PR #702, opened 2016-07-29
Force status transition in TestSJSubmit to prevent race condition failures

rob-c:
This test randomly failed on Jenkins due to having to kill a job to mark it as failed. This is bad so I'm adding some minor changes which should fix it.


rob-c:
I'm not sure I agree with the race condition based on the error that was thrown but either way I'm merging this when everything passes

milliams:
Well it's a race condition between processes where the test checks the status and then tries to do something based on that state but in the meantime the running process has carried on and changed its status to something else.

PR #703, opened 2016-07-29
Migrating is_prepared to ComponentItem and cleaning up Schema

rob-c:
OK this PR cleans up some strings which still don't need to be in the Schema as they have to be parsed into Python objects which can be stored directly in lists.

I've also migrated the `is_prepared` attribute for several classes where it's clear that a `GangaObject` (`SharedDir`) is stored in the attribute.

I've also added an exception to throw when trying to store the `repr` of a `GangaObject` which is something we should **not** be doing within the XML.

At the very least this is just a PR with some more spring cleaning in the Schema of some classes and should help things run smoothly as there used to be an overhead associated with parsing all of these strings back into real python every time they were parsed.

### DIFF
--- a/python/Ganga/Core/GangaRepository/VStreamer.py
+++ b/python/Ganga/Core/GangaRepository/VStreamer.py
@@ -1,7 +1,7 @@
 from __future__ import print_function, absolute_import
 from Ganga.Core.exceptions import GangaException
 from Ganga.Utility.logging import getLogger
-from Ganga.GPIDev.Base.Proxy import stripProxy, isType, getName
+from Ganga.GPIDev.Base.Objects import _getName
 
 from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList, makeGangaListByRef
 
@@ -50,7 +50,7 @@ class XMLFileError(GangaException):
 def _raw_to_file(j, fobj=None, ignore_subs=[]):
     vstreamer = VStreamer(out=fobj, selection=ignore_subs)
     vstreamer.begin_root()
-    stripProxy(j).accept(vstreamer)
+    j.accept(vstreamer)
     vstreamer.end_root()
 
 def to_file(j, fobj=None, ignore_subs=[]):
@@ -122,7 +122,7 @@ def fastXML(obj, indent='', ignore_subs=''):
         return sl
     elif hasattr(obj, '_data'):
         v = obj._schema.version
-        sl = ['\n', indent, '<class name="%s" version="%i.%i" category="%s">\n' % (getName(obj), v.major, v.minor, obj._category)]
+        sl = ['\n', indent, '<class name="%s" version="%i.%i" category="%s">\n' % (_getName(obj), v.major, v.minor, obj._category)]
         for attr_name in obj._schema.allItemNames():
             k = attr_name
             o = getattr(obj, k)
@@ -140,7 +140,7 @@ def fastXML(obj, indent='', ignore_subs=''):
         sl.append('</class>')
         return sl
     else:
-        return ["<value>", escape(repr(stripProxy(obj))), "</value>"]
+        return ["<value>", escape(repr(obj)), "</value>"]
 
 ##########################################################################
 # A visitor to print the object tree into XML.
@@ -182,7 +182,7 @@ class VStreamer(object):
         return
 
     def print_value(self, x):
-        print('\n', self.indent(), '<value>%s</value>' % escape(repr(stripProxy(x))), file=self.out)
+        print('\n', self.indent(), '<value>%s</value>' % escape(repr(x)), file=self.out)
 
     def showAttribute(self, node, name):
         return not node._schema.getItem(name)['transient'] and (self.level > 1 or name not in self.selection)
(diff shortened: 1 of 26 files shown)


python/Ganga/Core/GangaRepository/VStreamer.py
```diff
@@ -203,6 +203,8 @@ def simpleAttribute(self, node, name, value, sequence):
                 print(self.indent(), '</attribute>', file=self.out)
             else:
                 self.level += 1
+                if hasattr(value, 'accept'):
```
milliams: This should probably be an actual type check rather than an attribute check. Furthermore, it should probably be done at the schema/descriptor level to avoid them ever being assigned rather than simply preventing them being written.

python/Ganga/Lib/Executable/Executable.py
```diff
@@ -55,8 +55,8 @@ class Executable(IPrepareApp):
...
         'args': SimpleItem(defvalue=["Hello World"], typelist=[str, File, int], sequence=1, strict_sequence=0, doc="List of arguments for the executable. Arguments may be strings, numerics or File objects."),
         'env': SimpleItem(defvalue={}, typelist=[str], doc='Dictionary of environment variables that will be replaced in the running environment.'),
-       'is_prepared': SimpleItem(defvalue=None, strict_sequence=0, visitable=1, copyable=1, hidden=0, typelist=[None, ShareDir], protected=0, comparable=1, doc='Location of shared resources. Presence of this attribute implies the application has been prepared.'),
-        'hash': SimpleItem(defvalue=None, typelist=[None, str], hidden=0, doc='MD5 hash of the string representation of applications preparable attributes')
+       'is_prepared': ComponentItem('shareddirs', strict_sequence=0, visitable=1, copyable=1, hidden=0, typelist=[None, ShareDir], protected=0, comparable=1, doc='Location of shared resources. Presence of this attribute implies the application has been prepared.'),
```
milliams: This should be more like:
```python
'is_prepared': ComponentItem('shareddirs', defvalue=None, optional=True, load_default=False, doc='Location of shared resources. Presence of this attribute implies the application has been prepared.'),
```
as many of these arguments are either defaulted or irrelevant.

python/Ganga/Core/GangaRepository/VStreamer.py
rob-c: That should catch them yes but I'm wanting to test for this __ever__ showing up first.

python/Ganga/Lib/Executable/Executable.py
rob-c: I'm leaving things like this well enough alone in this PR it is going to touch a __lot__ more than I want it to. If we want to start removing default like this then I vote for a 2nd PR for house cleaning as this PR is going to have to fix things in the Proxy/Application layer I suspect seeing the number of tests failing.

python/GangaGaudi/Lib/Applications/GaudiBase.py
```diff
@@ -30,39 +30,40 @@ class GaudiBase(IPrepareApp):
...
     docstr = 'Location of shared resources. Presence of this attribute implies'\
              'the application has been prepared.'
-    schema['is_prepared'] = SimpleItem(defvalue=None,
+    schema['is_prepared'] = ComponentItem('shareddirs',
+                                       defvalue=None,
                                        strict_sequence=0,
                                        visitable=1,
                                        copyable=1,
-                                       typelist=['type(None)', 'str', ShareDir],
+                                       typelist=[None, str, ShareDir],
```
milliams: This highlights an underlying problem (ignoring the fact that `typelist` is ignored on `ComponentItem`s). This is a problem in the type system if this attribute can be a `str` or `ShareDir`. Why should it be a `str`?

rob-c:
Some work needs to be done here to work out why the Proxy is stopping the user from setting `job.application.exe` I suspect this is due to the poor defaults when using the `ComponentItem`.

I'll come back to this for 6.1.24 because I was opting for a quick fix #701 rather than rewriting the Proxy layer due to changing the Schema.

python/Ganga/Lib/Executable/Executable.py
milliams: `typelist` is not a valid argument for a `ComponentItem`. If you're changing the type then you have to change the arguments passed.

Also, this PR is already doing a load of `'str'`→`str`-type changes anyway, perhaps they'd be better in a separate PR?

python/GangaGaudi/Lib/Applications/GaudiBase.py
rob-c: I don't know but it was allowed to be so I'm not going to go chasing down whether it is ever set to be this or not. (I suspect not). and I'm not sure I agree that there is an underlying problem here?

python/Ganga/Lib/Executable/Executable.py
rob-c: If it's not a valid argument shouldn't it be causing some exception on construction? else the `ComponentItem` has a bug.

milliams: Indeed the schema system has a bug (plenty in fact). This is something we should try to be strict about in our usage so that we can fix and simplify the schema code. This will allow us to make things more predictable and avoid us getting in situations like we are with `ShareRef` in the future.

rob-c: This small fix is quickly turning into several PR and a lot of changes I was not wanting to make.
I'll try and look into this for 6.1.24 but for 6.1.23 I've some other smaller tasks to try and finish so I'll focus on them for now.

rob-c:
Closing as this is unlikely to be picked up and looks like an undesirable way to go for schema changes

PR #706, opened 2016-08-02
Fixing slice access

### LINKED ISSUE #705: Cannot support negative indexes
I've had a stab at re-enabling the negative index notation for the RegistrySlices and have hit a problem:

I have a registry with 90 jobs with the jobIDs ranging from 1 to 353. i.e. lots of deleted jobs:

`jobs[-1:]` actually passes the slice `slice(90:sys.maxsize:None)`
(i.e. the `-1` string is pre-processed away at some level in IPython/ast as far as I can tell and `len(jobs)` returns 90.)

This causes a problem because doing the following works:
`jobs[90:]` gives `slice(90:sys.maxsize:None)`

This can be fixed by getting the `len(jobs)` to return `353` but this would cause problems due to then trying to do: `assert len(jobs[5:6]) == 2` as `len(jobs[5:6])` has to be 6 for `jobs[5:6][-1]` to return `jobs(6)`.

At this stage I'm thinking we can't sensibly enable negative indexes due to the problem of mapping these indices against the position in the registry dictionary.

Does anyone have any experience with this section of the code? I'm suspecting this could be an IPython feature(bug?) but I'm yet to test against IPython4.

(I'll likely leave this on the back burner for now as I was only interested in pursuing this if I could get it working in a few hours in my spare time.)

rob-c:
Fixes #705 

This PR fixes the problems with negative list indexes and slicing not working.

`__getslice__` is deprecated so has been removed from the RegistrySlice over the slice object being supported correctly by `__getitem__`.

Whilst being a nicety this also fixes a bug where users would get random jobs/objects returned from valid looking requests against slices.

~~This also fixes a bug in the construction of slices handling subjobs.~~ Edit, this was a mistake...

### DIFF
--- a/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
+++ b/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
@@ -80,8 +80,10 @@ class RegistrySlice(object):
 
     def ids(self, minid=None, maxid=None):
         "Get the list of job ids. 'minid' and 'maxid' specify optional (inclusive) slice range."
-        maxid = maxid or sys.maxsize
-        minid = minid or 0
+        if maxid is None:
+            maxid = sys.maxsize
+        if minid is None:
+            minid = 0
         return [k for k in self.objects.keys() if minid <= k <= maxid]
 
     def clean(self, confirm=False, force=False):
--- a/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
+++ b/python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
@@ -1,9 +1,12 @@
+import collections
+import fnmatch
+import re
+import repr
 import sys
+from inspect import isclass
 import Ganga.Utility.logging
 from Ganga.Core import GangaException
-from Ganga.Core.GangaRepository.Registry import RegistryKeyError, RegistryIndexError, RegistryAccessError
-import fnmatch
-import collections
+from Ganga.Core.GangaRepository.Registry import RegistryKeyError, RegistryIndexError, RegistryAccessError, IncompleteObject
 
 from Ganga.GPIDev.Schema import ComponentItem
 from Ganga.Utility.external.OrderedDict import OrderedDict as oDict
@@ -19,6 +22,12 @@ config = Ganga.Utility.Config.getConfig('Display')
 class RegistrySlice(object):
 
     def __init__(self, name, display_prefix):
+        """
+        Constructor for a Registry Slice object
+        Args:
+            name (str): Slightly descriptive name of the slice
+            display_prefix (str): Prefix for slice to help modify display and other string reps of the object
+        """
         super(RegistrySlice, self).__init__()
         self.objects = oDict()
         self.name = name
@@ -46,8 +55,7 @@ class RegistrySlice(object):
         """
         """
         if not isinstance(keep_going, bool):
-            raise GangaException("The variable 'keep_going' must be a boolean. Probably you wanted to do %s(%s).%s()" % (
-                self.name, keep_going, method))
+            raise GangaException("The variable 'keep_going' must be a boolean. Probably you wanted to do %s(%s).%s()" % (self.name, keep_going, method))
         result = []
         for _id in self.objects.keys():
             obj = self.objects[_id]
@@ -72,10 +80,8 @@ class RegistrySlice(object):
 
     def ids(self, minid=None, maxid=None):
         "Get the list of job ids. 'minid' and 'maxid' specify optional (inclusive) slice range."
-        if maxid is None:
-            maxid = sys.maxsize
-        if minid is None:
-            minid = 0
+        maxid = maxid or sys.maxsize
+        minid = minid or 0
         return [k for k in self.objects.keys() if minid <= k <= maxid]
 
     def clean(self, confirm=False, force=False):
@@ -95,7 +101,6 @@ class RegistrySlice(object):
         return self.objects.clean(force)
 
     def select(self, minid=None, maxid=None, **attrs):
-        import repr
         from Ganga.GPIDev.Lib.Job.Job import Job
 
         if isType(minid, Job):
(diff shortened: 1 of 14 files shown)


python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -352,13 +352,16 @@ def __getitem__(self, x):
         """
         if isinstance(x, int):
             try:
-                return addProxy(self.objects[x])
+                if x < 0:
+                    return addProxy(self.objects[self.ids()[x]])
```
milliams: `self.ids()` needs to be sorted I believe.

rob-c: Shouldn't it be sorted as it's derived from the keys in the registry?

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -72,10 +80,8 @@ def do_collective_operation(self, keep_going, method, *args, **kwds):
 
     def ids(self, minid=None, maxid=None):
         "Get the list of job ids. 'minid' and 'maxid' specify optional (inclusive) slice range."
-        if maxid is None:
-            maxid = sys.maxsize
-        if minid is None:
-            minid = 0
+        maxid = maxid if maxid not None else sys.maxsize
+        minid = minid if minid not None else 0
```
milliams: I think if the expression gets this complicated (ignoring the syntax error) then it's not worth using a ternary operator.

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
milliams: Ahh, yes. I was confusing it with `values()`

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
alexanderrichards: I'll admit that this PR looks more complicated than my simple example and I haven't touched the JobRegistry in ages but why do we use `sys.maxsize` here rather than just a `max` as in my example?

rob-c: I'm not sure why we use sys.maxsize, I'll revert touching this function to keep this minimal

rob-c:
Any objections to merging this? It fixes a genuine set of bugs from the user perspective and fixes a nicety from the perspective of a _power-user_.

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
alexanderrichards: @milliams 
> I think if the expression gets this complicated (ignoring the syntax error) then it's not worth using a ternary operator.

Personally, I agree in general about the length of the expression however this expression is pretty small really, the only way it could be smaller is if the conditional check was a single word i.e. `minid if minid else 0` instead of three words `minid not None`

**Note** I realise the logic is different for the above two, it was illustrative purposes only

milliams: @alexanderrichards firstly, the code above is invalid Python and should instead be:
```python
maxid = maxid if maxid is not None else sys.maxsize
```
and secondly, in a syntax-light language like Python, by the time that a statement is 50% keywords it gets very hard to read and keep the correct associations in your mind.

Ternary operators work well when they are branching on a simple boolean variable but combining them with equality checks means that too many things are happening on one line.

rob-c:
Hmm, this appears to be having some problems with the subjobs, I'll check if I can revert the fix I put into the Job.py as that may be no longer needed.

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
alexanderrichards: > firstly, the code above is invalid Python

good spot

> means that too many things are happening on one line.

As I said generally I agree but in the above case I have no strong preference as I think to expand such a short ternary makes the code more bulky than perhaps needed and so I feel that the intention or flow of the code is not as easily conveyed or slightly hindered. It comes down to preference as always :wink:

alexanderrichards: having looked again at the code I think that I now agree with you because of the fact that maxid does not have to be set before the conditional check owing to the fact it's input as an argument. In this case the ternary does seem long.

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -368,21 +373,17 @@ def __getitem__(self, x):
...
-        import sys
+            returnable = self.__class__("%s[%s:%s]" % (self.name, start, stop))
+            for id_ in self.ids()[x]:
+                returnable.objects[id_] = self.objects[id_]
 
-        if i2 == sys.maxsize:
-            endrange = ''
-        else:
-            endrange = str(i2)
+            return addProxy(returnable)
```
milliams: Do we need this `addProxy()` here (or elsewhere in this PR)? Is this not covered by the proxy fixes I made a few months ago?

rob-c: I'm avoiding changing things as much as possible in this. We can fix this in another PR as I'm wanting to focus on as fewer things as possible in this one.
These used to be a proxy applied here and this class does not make use of a lot of the standard proxy machinery so I'm not sure and don't want to rock the boat by changing the workflow in this PR.

(Arguably I should revert the changes I made to the imports but as that's PEP8 and relatively obvious the consequence I'll leave those changes in.)

rob-c:
OK, removing the 'fix' I put into the Job.py fixes the problems in this PR. I'm confident that this should pass the integration tests now.

python/Ganga/GPIDev/Lib/Registry/RegistrySlice.py
milliams: No problem. I'm happy to play it safe here and look at removing these `addProxy` in the future.

rob-c:
OK, If there aren't any significant problems with this PR i plan to merge

PR #707, opened 2016-08-02
Removing all un-needed proxy generation

rob-c:
This PR removes all un-needed references of `addProxy`, `GPIProxyObjectFactory` and `GPIProxyClassFactory`.

Some of these are just plain bad code i.e. adding subfiles with a proxy into a class from behind the proxy is very bad!

Sometimes this should be safe but I'm not sure (addProxy around the Job may effect the getter).

If this PR just works then great I'll push for merging but otherwise this will highlight some broken areas in need of attention that we need to address and I'll make a smaller PR to address the clear problems this exercise has highlighted.


drmarkwslater:
Out of interest, do we have any more unneeded `stripProxy` calls that could also be removed at this point?

rob-c:
@drmarkwslater I don't think we do, I'm being cautious and removing 1 set of add or strip first and then doing the rest to check that everything makes sense. There is already 1 strange error from the queues system, but it's only 1 error so I suspect is an isolated bug or possibly a pseudo-race condition in the tests.

Once this is working I think I'll make a 2nd PR to try and remove the `stripProxy` references and then I'll look for a tool to highlight the unused imports we have lying around.

drmarkwslater:
That makes sense - just thought I'd ask 😄

rob-c:
OK, I had to fix the tutorials to respect the Schema. This isn't a bad thing and I think we should have been doing this before now. Once the tests pass I'm tempted to merge this once someone has had the chance to look through this as it does touch some 38 files.

drmarkwslater:
This last commit just put in a **massive** number of changes from 2e356ec. Though I'm not necessarily averse to this, it should definitely be in a separate PR.

rob-c:
@drmarkwslater This was in a separate PR against another PR I thought... I may have my branches messed up again...

rob-c:
OK, I'm confused how this has now worked. I'll remove the PR and start again with cherry-pick, or abandon this based upon upcoming priorities. It was just a code cleanup

egede:
This seems abandoned. I am closing.

PR #708, opened 2016-08-03
Enabling GaussSplitter support for GaudiExec

### LINKED ISSUE #631: Gauss Support is needed for GaudiRun
Support for Gauss and the GausSplitter is needed in GaudiRun.

The original implementation pickles the resulting python file and passes it to `gaudirun.py`. This means an additional level of work is needed to find out what exactly (and in what order) various options are passed to the running Job.

I'm aware of at least 1 minor bug in the original Gauss+GaussSplitter implementation which has been found by a PhD student trying to get their analysis working where there is some minor disagreement between different parts of the code to the number of jobs being generated.

In the original implementation there is 1 pickle file per subjob which means a unique inputsandbox.
I'm keen to explore passing the unique options here as runtime script arguments which would mean that submitting many subjobs should be much quicker.
(This may require reassembling a pickle object on the WN, I don't yet know)

rob-c:
Fixes #631 

After running some quick tests this should be all that is required for GaussSplitter to be correctly supported by the `GaudiExec` application.

I've added the appropriate splitter options into the `extraOpts` attribute for each subjob and have verified that each subjob is correctly unique and that the job scripts appear to be picking up the correct extraOpts script which is created and sent to the WN.

I've managed to get as far as getting back a file, but I'm not an MC expert I was cobbling together a solution from an old tutorial so this may be rough around the edges.

If the 6.1.23 release hasn't been finalised I'd like to merge this as it should be working.

### DIFF
--- a/python/GangaLHCb/Lib/Splitters/GaussSplitter.py
+++ b/python/GangaLHCb/Lib/Splitters/GaussSplitter.py
@@ -1,13 +1,16 @@
+import pickle
+import os
+import copy
 from Ganga.GPIDev.Adapters.ISplitter import ISplitter
+from Ganga.GPIDev.Base.Proxy import stripProxy
 from Ganga.GPIDev.Schema import Schema, Version, SimpleItem
 from Ganga.Utility.Config import getConfig
 from Ganga.Utility.files import expandfilename
 from Ganga.GPIDev.Lib.Job import Job
-import pickle
-import os
-import copy
-import Ganga.Utility.logging
-logger = Ganga.Utility.logging.getLogger()
+from Ganga.Utility.logging import getLogger
+from GangaLHCb.Lib.Applications.GaudiExec import GaudiExec
+
+logger = getLogger()
 
 class GaussSplitter(ISplitter):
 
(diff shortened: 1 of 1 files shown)


python/GangaLHCb/Lib/Splitters/GaussSplitter.py
```diff
@@ -39,39 +42,59 @@ def _create_subjob(self, job, inputdata):
...
 
     def split(self, job):
+        """
+            Method to do the splitting work
+            Args:
+                job (Job): master job to be used as a template to split subjobs
+        """
+
+        # NB Not sure if we can ever _not_ import the Gauss app from the GPI
+        from Ganga.GPI import Gauss
```
milliams: You should be able to do:
```python
from GangaLHCb.Lib.Applications import Gauss
```
Though looking at the code that generates that name (`GangaLHCb/Lib/Application/__init__.py`) is not recommended :unamused:

milliams:
The release has already started so it's probably too late for 6.1.23 though that's up to @mesmith75 to decide.

python/GangaLHCb/Lib/Splitters/GaussSplitter.py
rob-c: So what should be the solution here? I'm not a fan of either solution but if I can avoid the GPI and it works then that sits a little better with me.

egede:
@mesmith75 I think it is fine to delay this by one release.

rob-c:
I'll merge this right after 6.1.24 as it appears to work without problem

python/GangaLHCb/Lib/Splitters/GaussSplitter.py
milliams: Your solution in a44814872b0b2ab77964a8c36c3f81c4c653f074 is correct. The problem I have is not with this PR but rather with the magical way we are generating these applications in the LHCb plugin.

egede:
@rob-c Given that @drmarkwslater has asked for a delay of 1 day, I think you can just merge this now.

rob-c:
@egede Has the release branch been abandoned or do we have some strategy as how we cope with releases that get halted yet?
basically, where is best for this to be merged?

egede:
@rob-c You ask a git question to somebody with no knowledge about git. In my mind the best would be just to delete the release branch and then recreate it by starting the release process again. But I do not know if that is possible.

rob-c:
@mesmith75 @milliams  If a release is aborted is the release branch deleted or does the branch simply pull from develop again if required to continue the release?

If problems are found with a release during the process where should we submit PR against?
I'm assuming this would all have to be done manually to sync the release and develop branches if they're not abandoned?

milliams:
There is no "abort release" button we can press. All that is needed is to delete the release branch and start the release again.

If the release is underway and a fix needs to be made then the PR should be made against the release branch.

rob-c:
@milliams Great but in the case of situations like this where the PR is against the develop will the release branch pull from develop or would that have to be done manually?

rob-c:
OK, I propose that I'll merge this into develop and as @drmarkwslater has PR against develop we finish merging in develop then someone manually pulls develop into the release branch for the release moving forward to 6.1.24.
I'm assuming nothing has changed in this branch yet?

PR #710, opened 2016-08-03
Lots of small changes for PEP8 compliance

rob-c:
These are a lot of small changes to move the codebase closer to PEP8.

This should be all whitespace and formatting changes to make the code more readable and maintainable.

I'm not expecting this to effect any tests so if it does I'll not press for this to be merged but the code works just fine for me in a quick subset of the tests.

(This was done using autopep8 with only the safe default options enabled. This PR should only effect the formatting of code).


rob-c:
Scratch that this has modified the WN scripts. I'll make a PR to rename them to something safer first.

PR #713, opened 2016-08-03
Remove unwanted traces and fix registryslice summary

### LINKED ISSUE #709: should we introduce a ProxyError
Would people be objecting to introducing a ProxyError exception or even one called UserError?

This would allow us to raise exceptions which are defined as being due to a typo or mistake or bad code entered at the IPython prompt, rather than a bug in Ganga.

I'd like to make the prompt a little more friendly as some people associate the trace with something going horribly wrong rather than the trace being there to help them.
(I know a lot of people simply didn't care for the old logging information for instance)

Or, even should we disable the trace from a normal production release?

rob-c:
This addresses 2 areas where stack traces are generated within the IPython prompt that aren't particularly wanted/required.
The 2 stack traces I'm looking to avoid here are, 1) mis-typed commands from the IPython prompt, and 2) attributes not being found from in front of the Proxy.
Fixes #709 

I've also added a small fix which removes empty objects from the _display method for objects which have been marked as incomplete during runtime. (I think this effects very few users).


rob-c:
retest this please

rob-c:

rob-c:
retest this please

rob-c:
Rats forgot to mark this for inclusion before 6.2.0, 6.2.1 will have to be good enough although it improves the experience of users on the IPython prompt imo.

rob-c:
I'd strongly regard it as a bugfix as the only thing being done here is to improve the handling of some exceptions thrown at the IPython prompt.

PR #711, opened 2016-08-03
Moving template files to a .template extension

rob-c:
I've moved a bunch of the templates of what is run on the WN to `.template` files from having a `.py` extension as this causes problems with code review and such.

I aim to merge this after 6.1.24 to do some static code analysis.

PR #714, opened 2016-08-04
Remove un-needed proxy methods and unused imports

rob-c:
This PR builds on #707 and removes some dead code as well as removing un-needed proxy methods from code which doesn't need to know about the proxy.

Specifically I've removed a lot of _safe_ un-needed calls to `stripProxy` and `getName` which pull in the Proxy.
In addition I've removed a lot of unused imports.

This will merge atop #707 if everything goes well and between the 2 PR a lot of un-needed proxy manipulation behind the scenes will have been removed.

The 2 areas where I haven't removed `stripProxy/addProxy` calls are the Proxy Registry classes and Tasks. Both of these require a bit of care and in the latter I think I can reduce the number of calls to these methods but these require being careful.

Edit: I've integrated #711 to reduce problems when automatically examining the code


mesmith75:
Do we want to try and merge this? I imagine a lot of the code has changed since August but a cleanup would be nice?

rob-c:
@mesmith75 I was wanting to move down this line but won't have time to look at this over the week. Feel free to adopt the branch and continue as the next time I'll likely get to play with this will be next weekend.

mesmith75:
ok, can i rebase this to develop?

rob-c:
This might as well be regenerated from scratch, but it's just an example of code cleanup so can be done in the future.

PR #731, opened 2016-08-12
Fixing Job Templates to work again

### LINKED ISSUE #730: Can't make jobs from jobtemplates
The functionality of:

```
j = Job(JobTemplate(...))
```

has been broken since the removal of the `__construct__` method which means that the Job `__init__` method needs to be designed to accept an argument and to check that the argument is a Job/JobTemplate object in order to fully construct a job.
I'll put together a quick PR for this but I think this warrants fixing pre 6.2.0 but that I think we can make a 6.1.26/6.2.0 release with both of these fixes if needed. What are people's opinions wrt to this?

rob-c:
This should fix the problem spotted by LHCb users that construction of Job objects using JobTemplates doesn't work since 6.1.23.

Fixes #730


rob-c:
I'd prefer to see a 6.1.25 release next week for this but it may be worth holding off on 6.1.25 until some users have had the chance to test the past release or 2.
What are people's opinions?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -250,6 +254,32 @@ def __init__(self):
 
         self._stored_subjobs_proxy = None
 
+
+        if old_job:
+            self._unsetSubmitTransients()
```
drmarkwslater: As this is at creation, is this call needed?

drmarkwslater:
As this is a bug fix, I'd be happy for this to go in for a 6.1.25. I'm just a bit concerned that we keep pushing 6.2 back and never get to it...

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -232,7 +232,11 @@ class Job(GangaObject):
 
     # TODO: usage of **kwds may be envisaged at this level to optimize the
     # overriding of values, this must be reviewed
-    def __init__(self):
+    def __init__(self, old_job=None):
```
drmarkwslater: I think I'd prefer something like `job_tpl` or maybe`prev_job` as `old_job` could be misleading. Also, some docstrings + comments just to document what's happening?

rob-c: I'll look into switching to `prev_job`

rob-c:
@drmarkwslater I was keen to get this resolved on Friday as I'm technically away until Wednesday now.
If this is the only bug which is effecting LHCb users I would be keen to see it go in and 6.2.0 to out immediately afterwards.
What are people's opinions? not being able to use the Templates is a big thing when looking to go to 6.2.0 hence why the PR.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: This is safe to drop as it's during construction.

drmarkwslater:
I agree that this is a significant bug to start the 6.2 branch with so should probably be fixed before or with. I'm not sure what the consensus is though - anyone any thoughts?

alexanderrichards:
> this is a significant bug to start the 6.2 branch with so should probably be fixed before

reluctantly I think I agree. We should get these small niggles sorted and then have a chance for the 6.1.2X/6.2-pre to be tested by the community before making the switch to 6.2. I don't think that we want to introduce anything at the same time as the switch as this leaves up open to breaking everything just as we move.

We should however aim to turn this next bugfix release around relatively quickly such that we can define a timescale for moving to 6.2 at this weeks meeting.

rob-c:
I've updated the PR and would like to look into merging it asap

drmarkwslater:
👍

PR #735, opened 2016-08-18
Adding test for MassStorageFile

### LINKED ISSUE #733: MassStorage file problem
Problem found by LHCb user is:

```
I am using a fresh: v601r23 version and I ma getting the following error:

ERROR    Error Putting or cleaning up file: *.root, err::global name 'glob' is not defined

In the ganga submission script I did define the following:

j.outputfiles=[MassStorageFile(namePattern="*.root", outputfilenameformat=project+'/'+unique_production_name+'/{jid}/{sjid}/{fname}’)]

I think all looks OK and the files are at least stored on the desired eos location.

Can you please advise what to do?
```

The script used for this is:

```
import os

project = 'DiMuon'
configfile = "DiMuon4.py"

unique_production_name = '2011_21r1p1a_EW_up'

# for testing
#datalist = 'test.py'

# 2011
#datalist = 'LHCb_Collision11_90000000_Beam3500GeVVeloClosedMagUp_Real_Data_Reco14_Stripping21r1p1a_EW_DST_26_4_2016.py'
datalist = BKQuery('/LHCb/Collision11/Beam3500GeV-VeloClosed-MagUp/Real Data/Reco14/Stripping21r1p1a/90000000/EW.DST', dqflag=['OK']).getDataset()

projectpath = os.environ['HOME']+'/workspace/LHCb/'+project+'/'

j=Job(application=DaVinci( version = "v39r1p1" ), backend=Dirac() )

j.name = unique_production_name
print "\n["+j.name+"] Working on: " + j.name

print "\n["+j.name+"] Initalizing ganga running of project: " + project + " with unique_production_name version: " + unique_production_name
print "["+j.name+"] Using project path: "+ projectpath +"\n"

print "["+j.name+"] Config file: " + configfile
print "["+j.name+"] Data list in file/from query:\n"
print datalist
print "\n"

j.application.optsfile = [ File ( projectpath+configfile ) ]

j.application.extraopts="""
DaVinci().PrintFreq = 1000
DaVinci().EvtMax = -1
DaVinci().InputType = 'DST'
DaVinci().DataType = '2011'
DaVinci().Simulation = False
"""
print "\n["+j.name+"] Here are the changed job options: "
print j.application.extraopts

#j.inputdata = j.application.readInputData(projectpath+datalist )
j.inputdata = datalist

j.splitter = SplitByFiles( filesPerJob = 10 )
j.parallel_submit = True
j.outputfiles=[MassStorageFile(namePattern="*.root", outputfilenameformat=project+'/'+unique_production_name+'/{jid}/{sjid}/{fname}')]
#j.outputfiles=[DiracFile(DiracFile(namePattern='*.root',locations=["CERN-USER"])]

print "\n["+j.name+"] Writing the following files to EOS:"
print j.outputfiles

print "\n["+j.name+"] Now submitting the job ..."
j.submit()
#queues.add(j.submit)
print "\n["+j.name+"] ... submitting job done.\n"
```

This concerns me slightly as there is an error somewhere around the handling of complex file names which could do with tracking down.
I'm worried that files could be being copied multiple times or not at all or multiple times but once is broken.
Given this effects the integrity of the output data for the user I think we should be sure there are no problems around this in 6.2.0.
(This functionality should also exist across all IGangaFile object types eventually!)

rob-c:
This PR addresses the issue that some MassStorageFile objects weren't being cleaned up from disk.

This doesn't mean that data is lost but it could lead to situations where the disk becomes excessively full rather quickly as copies of potentially very large files are left on 'local' storage.

I've added a new integration test which tests the behavior of the `MassStorageFile` using the local file system and the standard `'mkdir -p'`, `'cp'`, and `'ls'` tools.

This works as intended for me on my test machine and should also work on the Jenkins boxes.

I cannot enable the tests which make use of the 'client' code with a 'Local' backend which is due to the limitations of the way that this combination is hard-coded out of the file management code.
(This appears to be a bad idea imo and is being removed in a future large PR which will arrive after 6.2.0 as it impacts _everyone_ so will not arrive before 6.2.0 for stability/safety reasons)

Edit:
Fixes #733

### DIFF
--- a/python/Ganga/GPIDev/Adapters/IGangaFile.py
+++ b/python/Ganga/GPIDev/Adapters/IGangaFile.py
@@ -1,6 +1,7 @@
 from Ganga.GPIDev.Base import GangaObject
 from Ganga.GPIDev.Schema import Schema, Version, SimpleItem
 from fnmatch import fnmatch
+import os
 import glob
 import re
 
--- a/python/Ganga/GPIDev/Adapters/IGangaFile.py
+++ b/python/Ganga/GPIDev/Adapters/IGangaFile.py
@@ -1,6 +1,7 @@
 from Ganga.GPIDev.Base import GangaObject
 from Ganga.GPIDev.Schema import Schema, Version, SimpleItem
 from fnmatch import fnmatch
+import glob
 import re
 
 regex = re.compile('[*?\[\]]')
@@ -164,7 +165,7 @@ class IGangaFile(GangaObject):
         # For all other file types (not LocalFile) The file in the outputdir is temporary waiting for Ganga to pass it to the storage solution
         job = self.getJobObject()
 
-        for f in glob.glob(os.path.join(job.outputdir, outputfile.namePattern)):
+        for f in glob.glob(os.path.join(job.outputdir, self.namePattern)):
             try:
                 os.remove(f)
             except OSError as err:
--- a/python/Ganga/GPIDev/Adapters/IGangaFile.py
+++ b/python/Ganga/GPIDev/Adapters/IGangaFile.py
@@ -111,8 +111,7 @@ class IGangaFile(GangaObject):
         mystderr = ''
 
         try:
-            child = subprocess.Popen(
-                cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            child = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
             (mystdout, mystderr) = child.communicate()
             exitcode = child.returncode
         finally:
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -250,7 +250,7 @@ class MassStorageFile(IGangaFile):
 
         (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, pathToDirName))
         if exitcode != 0:
-            self.handleUploadFailure(mystderr)
+            self.handleUploadFailure(mystderr, '%s %s' % (ls_cmd, pathToDirName))
             return
 
         directoryExists = False
@@ -262,7 +262,7 @@ class MassStorageFile(IGangaFile):
         if not directoryExists:
             (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (mkdir_cmd, massStoragePath))
             if exitcode != 0:
-                self.handleUploadFailure(mystderr)
+                self.handleUploadFailure(mystderr, '%s %s' % (mkdir_cmd, massStoragePath))
                 return
 
         # the folder part of self.outputfilenameformat
@@ -312,7 +312,7 @@ class MassStorageFile(IGangaFile):
             command = '%s -p %s' % (mkdir_cmd, massStoragePath)
             (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess(command)
             if exitcode != 0:
-                self.handleUploadFailure(mystderr)
+                self.handleUploadFailure(mystderr, command)
                 return
 
         # here filenameStructure has replaced jid and sjid if any, and only not
@@ -330,7 +330,7 @@ class MassStorageFile(IGangaFile):
                 d.compressed = self.compressed
 
                 if exitcode != 0:
-                    self.handleUploadFailure(mystderr)
+                    self.handleUploadFailure(mystderr, '%s %s %s' % (cp_cmd, currentFile, os.path.join(massStoragePath, finalFilename)))
                 else:
                     logger.info('%s successfully uploaded to mass storage as %s' % (currentFile, os.path.join(massStoragePath, finalFilename)))
                     d.locations = os.path.join(massStoragePath, os.path.basename(finalFilename))
@@ -346,7 +346,7 @@ class MassStorageFile(IGangaFile):
             finalFilename = filenameStructure.replace('{fname}', os.path.basename(currentFile))
             (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s %s' % (cp_cmd, currentFile, os.path.join(massStoragePath, finalFilename)))
             if exitcode != 0:
-                self.handleUploadFailure(mystderr)
+                self.handleUploadFailure(mystderr, '%s %s %s' % (cp_cmd, currentFile, os.path.join(massStoragePath, finalFilename)))
             else:
                 logger.info('%s successfully uploaded to mass storage as %s' % (currentFile, os.path.join(massStoragePath, finalFilename)))
                 location = os.path.join(massStoragePath, os.path.basename(finalFilename))
@@ -405,14 +405,15 @@ class MassStorageFile(IGangaFile):
 
         return (True, '')
 
-    def handleUploadFailure(self, error):
+    def handleUploadFailure(self, error, cmd=''):
 
         self.failureReason = error
         if self._getParent() != None:
-            logger.error("Job %s failed. One of the job.outputfiles couldn't be uploaded because of %s" % (
-                str(self._getParent().fqid), self.failureReason))
+            logger.error("Job %s failed. One of the job.outputfiles couldn't be uploaded because of %s" % (str(self._getParent().fqid), self.failureReason))
         else:
             logger.error("The file can't be uploaded because of %s" % (self.failureReason))
+        if cmd:
+            logger.error("Attempted to run: '%s'" % (cmd))
 
     def getWNInjectedScript(self, outputFiles, indent, patternsToZip, postProcessLocationsFP):
         """
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -8,6 +8,7 @@ import errno
 import re
 import os
 import copy
+import glob
 
 from Ganga.GPIDev.Schema import Schema, Version, SimpleItem, ComponentItem
 
@@ -203,7 +204,6 @@ class MassStorageFile(IGangaFile):
         Creates and executes commands for file upload to mass storage (Castor), this method will
         be called on the client
         """
-        import glob
         import re
 
         sourceDir = ''
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -223,11 +223,12 @@ class MassStorageFile(IGangaFile):
             self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
             raise GangaException(mystderr)
         
-        logger.info("--looking for: %s" % pathToDirName)
+        logger.info("--looking for: '%s'" % massStoragePath)
+        logger.info("--looking in: '%s'" % pathToDirName)
 
         directoryExists = False
         for directory in mystdout.split('\n'):
-            logger.info("--found: %s" % directory.strip())
+            logger.info("--found: '%s'" % directory.strip())
             if directory.strip() == dirName:
                 directoryExists = True
                 break
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -220,11 +220,9 @@ class MassStorageFile(IGangaFile):
         directoryExists = False
 
         (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, quote(pathToDirName)))
-        if exitcode != 0:
-            if exitIfNotExist:
-                self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
-                raise GangaException(mystderr)
-            directoryExists = False
+        if exitcode != 0 and exitIfNotExist:
+            self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
+            raise GangaException(mystderr)
 
         for directory in mystdout.split('\n'):
             if directory.strip() == dirName:
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -334,12 +334,10 @@ class MassStorageFile(IGangaFile):
 
         # create the folder structure
         if folderStructure:
-            folderStructure = folderStructure.strip('/')
             massStoragePath = os.path.join(massStoragePath, folderStructure)
-            command = '%s -p %s' % (mkdir_cmd, massStoragePath)
-            (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess(command)
-            if exitcode != 0:
-                self.handleUploadFailure(mystderr, "3) " + command)
+            try:
+                self._mkdir(massStoragePath)
+            except GanagException:
                 return
 
         # here filenameStructure has replaced jid and sjid if any, and only not
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -211,22 +211,23 @@ class MassStorageFile(IGangaFile):
 
         massStorageConfig = getConfig('Output')['MassStorageFile']['uploadOptions']
         mkdir_cmd = massStorageConfig['mkdir_cmd']
-        cp_cmd = massStorageConfig['cp_cmd']
         ls_cmd = massStorageConfig['ls_cmd']
 
         # create the last directory (if not exist) from the config path
         pathToDirName = os.path.dirname(massStoragePath)
         dirName = os.path.basename(massStoragePath)
 
+        directoryExists = False
+
         (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, pathToDirName))
         if exitcode != 0:
-            self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
-            raise GangaException(mystderr)
-        
+            #self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
+            #raise GangaException(mystderr)
+            directoryExists = True
+
         logger.info("--looking for: '%s'" % massStoragePath)
         logger.info("--looking in: '%s'" % pathToDirName)
 
-        directoryExists = False
         logger.info("Stdout: %s" % mystdout)
         for directory in mystdout.split('\n'):
             #logger.info("--found: '%s'" % directory.strip())
@@ -236,7 +237,7 @@ class MassStorageFile(IGangaFile):
 
         if not directoryExists:
             logger.info("---not-exists")
-            (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (mkdir_cmd, massStoragePath))
+            (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s -p %s' % (mkdir_cmd, massStoragePath))
             if exitcode != 0:
                 self.handleUploadFailure(mystderr, '2) %s %s' % (mkdir_cmd, massStoragePath))
                 raise GangaException(mystderr)
@@ -280,14 +281,13 @@ class MassStorageFile(IGangaFile):
 
         massStorageConfig = getConfig('Output')['MassStorageFile']['uploadOptions']
 
-        mkdir_cmd = massStorageConfig['mkdir_cmd']
         cp_cmd = massStorageConfig['cp_cmd']
         ls_cmd = massStorageConfig['ls_cmd']
         massStoragePath = massStorageConfig['path']
 
         try:
             self._mkdir(massStoragePath)
-        except GanagException:
+        except GangaException:
             return
 
         # the folder part of self.outputfilenameformat
@@ -295,7 +295,7 @@ class MassStorageFile(IGangaFile):
         # the file name part of self.outputfilenameformat
         filenameStructure = ''
 
-        if self._getParent() != None:
+        if self._getParent() is not None:
             jobfqid = self.getJobObject().fqid
 
             jobid = jobfqid
@@ -337,7 +337,7 @@ class MassStorageFile(IGangaFile):
             massStoragePath = os.path.join(massStoragePath, folderStructure)
             try:
                 self._mkdir(massStoragePath)
-            except GanagException:
+            except GangaException:
                 return
 
         # here filenameStructure has replaced jid and sjid if any, and only not
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -125,13 +125,14 @@ class MassStorageFile(IGangaFile):
                     d = MassStorageFile(namePattern=pattern)
                     d.compressed = mass_file.compressed
                     d.failureReason = line[line.find('ERROR') + 5:]
-                    mass_file.subfiles.append(GPIProxyObjectFactory(d))
+                    mass_file.subfiles.append(d)
                 else:
-                    d = MassStorageFile(namePattern=name)
-                    d.compressed = mass_file.compressed
-                    d.outputfilenameformat = mass_file.outputfilenameformat
-                    mass_file.subfiles.append(GPIProxyObjectFactory(d))
-                    mass_line_processor(line, d)
+                    if pattern == self.namePattern:
+                        d = MassStorageFile(namePattern=name)
+                        d.compressed = mass_file.compressed
+                        d.outputfilenameformat = mass_file.outputfilenameformat
+                        mass_file.subfiles.append(d)
+                        mass_line_processor(line, d)
             elif name == mass_file.namePattern:
                 if outputPath == 'ERROR':
                     logger.error("Failed to upload file to mass storage")
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -346,11 +346,6 @@ class MassStorageFile(IGangaFile):
         if regex.search(fileName) is not None:
             for currentFile in glob.glob(os.path.join(sourceDir, fileName)):
                 finalFilename = filenameStructure.replace('{fname}', os.path.basename(currentFile))
-                try:
-                    folder_ = os.path.dirname(os.path.join(massStoragePath, finalFilename))
-                    self._mkdir(folder_)
-                except GanagException:
-                    return
                 (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s %s' %\
                                                 (cp_cmd, escapeWhiteSpace(currentFile), escapeWhiteSpace(os.path.join(massStoragePath, finalFilename))))
 
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -222,18 +222,24 @@ class MassStorageFile(IGangaFile):
         if exitcode != 0:
             self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
             raise GangaException(mystderr)
+        
+        logger.info("--looking for: %s" % pathToDirName)
 
         directoryExists = False
         for directory in mystdout.split('\n'):
+            logger.info("--found: %s" % directory.strip())
             if directory.strip() == dirName:
                 directoryExists = True
                 break
 
         if not directoryExists:
+            logger.info("---not-exists")
             (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (mkdir_cmd, massStoragePath))
             if exitcode != 0:
                 self.handleUploadFailure(mystderr, '2) %s %s' % (mkdir_cmd, massStoragePath))
                 raise GangaException(mystderr)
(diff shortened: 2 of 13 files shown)


rob-c:
I'll try and run a test on the DIRAC backend and see if I can test the client side code, but no promises as this relies on the jobs finishing in a timely manner.

rob-c:
Can someone check the code in this PR.

I don't understand 100% how the `subfiles` are supposed to work hence the multiple 'FIXME' in this test.
I think the fact that there is a custom getter function in the Job.py for the outputfiles is getting in the way of doing something sensible.
(I'm not a fan of this getter method as it confuses things. The outputfiles from in front of the proxy is a different object to the outputfiles behind the proxy an this just causes problems imo!)

mesmith75:
@rob-c I think this needs an extra ```import os``` as it gives me
```ERROR    Error Putting or cleaning up file: *.root, err::global name 'os' is not defined```
on my quick test job.

**Comment by rob-c** [comment]
@mesmith75 Do you know where I'm tracking this down but so far no luck :/

File: python/Ganga/test/GPI/FileTests/TestMassStorage.py [hunk 1]
```diff
@@ -0,0 +1,304 @@
...
+
+        sleep_until_completed(j)
+
+        assert j.status == 'completed'
+
+        assert len(j.subjobs) == TestMassStorage.sj_len
+
+        for i in range(0, TestMassStorage.sj_len):
+            # FIXME?
+            #assert (stripProxy(j.subjobs[i].outputfiles[0]).subfiles) == 2
```
drmarkwslater: Why isn't this working? I feel it should looking at the code...

python/Ganga/test/GPI/FileTests/TestMassStorage.py
```diff
@@ -0,0 +1,304 @@
...
+        assert j.status == 'completed'
+
+        assert len(j.inputfiles) == 1
+        assert len(j.outputfiles) == 1
+        # FIXME?
+        #assert len(stripProxy(j.outputfiles[0]).subfiles) == 1
+
+        #print("namePattern: %s" % (j.outputfiles[0].namePattern, )) 
+        #print("Locations: %s" % (j.outputfiles[0].locations, ))
+        #print("accessURL: %s" % (j.outputfiles[0].accessURL(), ))
```
drmarkwslater: Probably best to just remove these rather than comment them out

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: The getter in Job.py returns effectively an auto-expanded form of the subjobs so I'd have to add a stripProxy to the `j.subjobs[i]` object additionally I think but I've not tested.
(I'm really not a fan of returning completely different objects based upon if we're infront of the Proxy or not it confuses me more than helps me imo but we shouldn't change this now as user scripts rely on it etc...)

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: Ah, yes, sorry forgot to add a commit doing that

rob-c:
OK there is still a bug relating to the `MassStorageFile` client side code which I see when I submit a job which has a `MassStorageFile` in the `outputfiles` as well as being on the `Dirac` backend.
I'm still trying to track down how to fix this as the only way to reproduce this is currently difficult.

python/Ganga/test/GPI/FileTests/TestMassStorage.py
```diff
@@ -0,0 +1,304 @@
...
+import shutil
+import copy
+
+def generateUniqueTempFile( ext = '.txt' ):
+    """ Generate a unique file with a given filename with some random contents and return the name of the file on disk
+    Args:
+        ext (str): This is the extension (including '.') to give to the file of interest
+    """
+
+    myFile = tempfile.NamedTemporaryFile(mode='w',delete=False)
```
drmarkwslater: `NamedTemporaryFile` takes a suffix argument so you don't have to rename the file afterwards

python/Ganga/test/GPI/FileTests/TestMassStorage.py
```diff
@@ -0,0 +1,304 @@
...
+
+    file_string = file_string + str( rand ) + "\n"
+
+    urand = os.urandom(20)
+
+    file_string = file_string + str( urand ) + "\n"
+
+    myFile.write( file_string )
+
+    myFile.close()
```
drmarkwslater: This could be in a context manager:
```
with tempfile.NamedTemporaryFile as myFile:
```

python/Ganga/test/GPI/FileTests/TestMassStorage.py
```diff
@@ -0,0 +1,304 @@
...
+    random.seed( unix_t )
+    rand = random.random() * 1E10
+
+    file_string = file_string + str( rand ) + "\n"
+
+    urand = os.urandom(20)
+
+    file_string = file_string + str( urand ) + "\n"
+
+    myFile.write( file_string )
```
drmarkwslater: If you just want a random string, can't you just write some random text, e.g.
```
myFile.write( ''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(20)) )
```

python/Ganga/test/GPI/FileTests/TestMassStorage.py
drmarkwslater: Apologies - just read the other comments. I'll see if I can have a look.

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: I have been using this script elsewhere I just copied/pasted it here... I'll clean it up

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: OK, adding in that stripProxy doesn't work.

This probably needs looking into but not right now as long as everything else is working as this is technically an implementation detail I suppose.

python/Ganga/GPIDev/Adapters/IGangaFile.py
```diff
@@ -164,7 +165,7 @@ def cleanUpClient(self):
         # For all other file types (not LocalFile) The file in the outputdir is temporary waiting for Ganga to pass it to the storage solution
         job = self.getJobObject()
 
-        for f in glob.glob(os.path.join(job.outputdir, outputfile.namePattern)):
+        for f in glob.glob(os.path.join(job.outputdir, self.namePattern)):
             try:
                 os.remove(f)
```
drmarkwslater: As has been pointed out - this needs an `import os` at the top.

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: I think it's unlikely this code would leave a file object open but yes I'll make the change

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: I'll make the change, the rename was an evolution to change an existing piece of code a while ago rather than to do this properly

mesmith75:
@rob-c I think in the IGangaFile alongside ```import glob```

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: The piece of code is used elsewhere in a section of code where I tested something for being able to use non ascii stuff (not ganga I'm afraid) I'll simplify it for this example.

python/Ganga/test/GPI/FileTests/TestMassStorage.py
drmarkwslater: So the following should work and tests that the appropriate number of files were matched:
```
assert j.subjobs[i].outputfiles == 2
```
The subfiles is hidden from the GPI and is how `IGangaFile`'s expand do wildcard expansion but keep the knowledge that they originated from wildcards, e.g if you copied the above job you'd get back to a single output file with `*.root`.

python/Ganga/GPIDev/Lib/File/scripts/MassStorageFileWNScript.py
```diff
@@ -38,7 +38,7 @@ def execSyscmdSubprocessAndReturnOutputMAS(cmd):
             directoryExists = True
             break
 
-    print("Directory, %s found: %s" % (str(path), str(directoryExists)))
+    #print("Directory, %s found: %s" % (str(path), str(directoryExists)))
```
drmarkwslater: Just remove?

rob-c:
OK, I've given the code a cleanup and assuming it all passes can people give this a second pass. The grid gods aren't favoring me so @mesmith75 Can you try testing the client side code for the MassStorageFile in this PR?
I'm still looking into how to make the test work with Local+MassStorageFile+client as this combination would be very helpful to have a test for but this won't arrive until 6.2.0.

python/Ganga/GPIDev/Lib/File/scripts/MassStorageFileWNScript.py
rob-c: Ah, missed that, thanks

python/Ganga/test/GPI/FileTests/TestMassStorage.py
rob-c: @drmarkwslater I'm aware this is how it's intended to be used and works but it does it by returning a lot of objects which until recently didn't correctly inherit the Job as a parent and still don't really do it in a full proof way because fixing this at the object level causes issues for some reason.
I can't see how to ever access the raw outputfiles to test this which is potentially an issue but one I'll leave for another day to worry about.

rob-c:
OK, all of the comments should be implemented now to let some tests run. I'll look to merge this at about 5pm or the end of today and I think aim for the release to start tomorrow morning.

python/Ganga/test/GPI/FileTests/TestMassStorage.py
drmarkwslater: But aren't you just wanting to test what that the returned files according to the GPI is what you'd expect? If they're not then that would show a bug anyway. Certainly for this test, it doesn't matter if the returned files have the correct parent set.

rob-c: @drmarkwslater I'll add a len check in that case once I've finished checking a few other things I'm wanting to test

mesmith75:
@rob-c I've run a quick test with Dirac backend that works. The output files are uploaded to eos and cleaned from the repository successfully.

rob-c:
OK, problem `assert len(j.subjobs[i].outputfiles) == 3` highlights a problem when there are multiple files.

I'm tempted to disable this test and move forward if this is difficult to diagnose what are people's thoughts.

This is __yet__ another problem from the custom getter and outputfiles so fixing this may take 24hr or more.

rob-c:
Fixed the bug. The fact that this has been there for so long uncaught I think summarises the situation with the outputfiles. It's difficult to write tests for these and even more difficult when the actual file objects aren't accessible for debugging.

rob-c:
Right I'm go for merge unless there any comments?

drmarkwslater:
I'd still like to check some things about this - give me a bit...

drmarkwslater:
Sorry for the delay - I agree, it's very strange this bug wasn't caught before but there we go. I think the tests you've got here go a long way to picking up on these errors in the future though and I don't think there's a major problem with outputfiles as they stand (at least GPI-wise), it's just we didn't have tests for them until now (which is bad in itself of course!).

Anyway, 👍  and good catch!

mesmith75:
is this the last thing for 6.1.25?

drmarkwslater:
I may have an Atlas issue to fix - let you know in the next hour!

rob-c:
I'm still investigating something with the client side tests, that will take an hour or 2

drmarkwslater:
I'm done as the Atlas issue was an external problem, not a Ganga code problem!

mesmith75:
@drmarkwslater ok great. Can you generate an ATLAS proxy for the release tests. I'll got started once this is finished,

rob-c:
Well I've just hit a bug in the config system. This is really kind of serious so I'm doing some digging to make sure what is effected if anything.

rob-c:
OK. Think I solved this. It's really quite an annoying subtlety and I'll make a 2nd PR which is __exclusive to the tests__ (touches zero lines of code elsewhere) which I'd like to see put in for 6.1.25 as well.
There is now a difference in behaviour of different parts of the code based upon the state of the configuration in the test. This is obviously not ideal and so I'm going to have a go at correcting the tests to make sure that the correct code paths are being tested and not just ones which happen to be safe.

rob-c:
Interesting. The new test passes fine for me and not for Jenkins. That's annoying.
I'll do some digging and try to reproduce this.

rob-c:
OK, in summary the methods for this were matching all wildcards, were incompatible with directories containing whitespaces, and there is an annoying bug in the way that the setUp functions are now run. This took a while to debug/fix but we now have a working test for the MassStorageFile on a local fs.
I'm going to clean this up and will try for merge Monday.
Also I'll migrate this default config to a new file type which inherits from MassStorageFile in another PR.
This touches a lot of things but there were very broken and needed attention.

rob-c:
Can someone else review this so I can merge it today so we can go for 6.1.25?

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -25,6 +30,22 @@
...
+    4) This does NOT perform any level of security/safety check and I'm sure a user could inject 'rm -fr' commands into their job description\
+       but other security practices should isolate their choices from other users.
+    Args:
+        somePath (str): somePath which may contain whitespaces
+    """
+    returnable = str(somePath)
+    for i in (' ', ';', '(', ')',):
+        returnable = returnable.replace(i, "\\"+i)
+    return returnable
+
```
drmarkwslater: I would personally prefer using `pipes.quote` rather than running our own function. I know it's deprecated from 2.7 but it is documented in 2.7 and simply becomes `shlex.quote` in Python 3.3. I don't if others think differently though.

rob-c: @drmarkwslater I was concerned about strange differences in behavior between the instance we run here and the code run on the WN in this case.
I'm happy to change to this (if I can find an example) I just generally avoid things with big deprecation warnings but I'm not tied to this implementation by any means.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -198,20 +217,51 @@ def getWNScriptDownloadCommand(self, indent):
...
+        pathToDirName = os.path.dirname(massStoragePath)
+        dirName = os.path.basename(massStoragePath)
+
+        directoryExists = False
+
+        (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, escapeWhiteSpace(pathToDirName)))
+        if exitcode != 0:
+            #self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
+            #raise GangaException(mystderr)
+            directoryExists = True
```
drmarkwslater: If the dir **doesn't** exist, exitcode will be non-zero. Surely at that point it should try to create it? As it stands this will never actually create the directory.

drmarkwslater: Apologies - got confused with the names. I see how this works now!

rob-c: Tbh, I think you're correct I'll have a check of this I just tried factor out the original code without too many changes.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -324,39 +353,37 @@ def put(self):
         if regex.search(fileName) is not None:
             for currentFile in glob.glob(os.path.join(sourceDir, fileName)):
                 finalFilename = filenameStructure.replace('{fname}', os.path.basename(currentFile))
-                (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s %s' % (cp_cmd, currentFile, os.path.join(massStoragePath, finalFilename)))
+                try:
+                    folder_ = os.path.dirname(os.path.join(massStoragePath, finalFilename))
+                    self._mkdir(folder_)
```
drmarkwslater: This isn't needed as `finalFilename` is derived from an `os.basename` call so can't have a subdir prepended to it. Plus the `massStoragePath` has already been created further up.

rob-c: Ah, I think you're correct. I was getting errors that directories didn't exist here during development so decided to try this. I'll remove the extra call to `_mkdir` as it  should not needed :+1:

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -324,39 +353,37 @@ def put(self):
...
+                self.subfiles.append(d)
         else:
             currentFile = os.path.join(sourceDir, fileName)
             finalFilename = filenameStructure.replace('{fname}', os.path.basename(currentFile))
-            (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s %s' % (cp_cmd, currentFile, os.path.join(massStoragePath, finalFilename)))
+            (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s %s' %\
+                                                        (cp_cmd, escapeWhiteSpace(currentFile), escapeWhiteSpace(os.path.join(massStoragePath, finalFilename))))
             if exitcode != 0:
-                self.handleUploadFailure(mystderr)
+                self.handleUploadFailure(mystderr, '%) %s %s %s' % (cp_cmd, currentFile, os.path.join(massStoragePath, finalFilename)))
```
drmarkwslater: Should there be a number here? It's just got `'%)...'`.

rob-c: Yep, that should be 5, thanks for catching that typo

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
```diff
@@ -0,0 +1,129 @@
...
+                                                    'PBS': 'client', 'Interactive': 'client', 'Local': 'client', 'CREAM': 'client'}}
+
+    def setUp(self):
+        """
+        Configure the MassStorageFile for the test
+        """
+        extra_opts=[('PollThread', 'autostart', 'False'),
+                    ('Local', 'remove_workdir', 'False'),
+                    ('TestingFramework', 'AutoCleanup', 'False'),
+                    ('Output', 'MassStorageFile', copy.deepcopy(TestMassStorageClient.MassStorageTestConfig)),
```
drmarkwslater: It doesn't really need a `deepcopy` here does it? In fact, can't the config be a local variable to the `setUp` function?

python/Ganga/test/GPI/FileTests/TestMassStorageWN.py
```diff
@@ -0,0 +1,249 @@
...
+import tempfile
+import datetime
+import time
+import random
+import os
+import shutil
+import copy
+import string
+
+def generateUniqueTempFile( ext = '.txt' ):
```
drmarkwslater: This should be shifted to `testlib` and imported to avoid repeated definitions

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
rob-c: Ah, I was debugging why the options weren't being applied here. I can remove the deepcopy.

And no, unfortunately this can't be a parameter in the `setUp` function as the monitoring loop then starts with the default config and proceeds to start processing the completed jobs until it gets past the super call below.
(aka the monitoring loop is too fast!)
Tracking that down cost a lot of time but raises some concerns that some tests are passing incorrectly.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -25,6 +31,14 @@
...
 logger = getLogger()
 
+def escapeWhiteSpace(somePath):
+    """
+    This makes the aruments safe to pass to the command line
+    Args:
+        somePath (str): somePath which may contain whitespaces or special chars
+    """
+    return quote(somePath)
+
```
drmarkwslater: Why not just remove the function completely and replace with `quote` directly in the code?

python/Ganga/test/GPI/FileTests/TestMassStorageWN.py
rob-c: If people are happy for me to move stuff there. I was planning on breaking this out into a common file in this dir but it felt overkill for what is about ~10 lines of code.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: Will do, it just takes longer and I wanted to know if it works

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -198,20 +210,51 @@ def getWNScriptDownloadCommand(self, indent):
...
+        for directory in mystdout.split('\n'):
+            if directory.strip() == dirName:
+                directoryExists = True
+                break
+
+        if not directoryExists:
+            (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s -p %s' % (mkdir_cmd, escapeWhiteSpace(massStoragePath)))
+            if exitcode != 0:
+                self.handleUploadFailure(mystderr, '2) %s %s' % (mkdir_cmd, massStoragePath))
+                raise GangaException(mystderr)
```
milliams: If this is a `GangaException` intended for end-users then it would be helpful to give a user-facing message in addition to the raw output from the mkdir command.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -238,39 +281,21 @@ def put(self):
...
-        # create the last directory (if not exist) from the config path
-        import os.path
-        pathToDirName = os.path.dirname(massStoragePath)
-        dirName = os.path.basename(massStoragePath)
-
-        (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, pathToDirName))
-        if exitcode != 0:
-            self.handleUploadFailure(mystderr)
+        try:
+            self._mkdir(massStoragePath)
```
drmarkwslater: This is now explicitly changing the behaviour from 'Check the parent dir exists, if not bail' to 'Create the whole dir structure as provided using `mkdir -p`'. I'm not saying this is necessarily a bad idea but it is a change. I'm not sure if anyone else has thoughts about this?

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: this is what the code used to do so, I'm extremely reluctant to start adding yet more functionality. I just added a well defined exception here to allow escaping from `_mkdir` where there used to be `return` statements in the logic

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -198,20 +210,51 @@ def getWNScriptDownloadCommand(self, indent):
...
+        pathToDirName = os.path.dirname(massStoragePath)
+        dirName = os.path.basename(massStoragePath)
+
+        directoryExists = False
+
+        (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, escapeWhiteSpace(pathToDirName)))
+        if exitcode != 0:
+            #self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
+            #raise GangaException(mystderr)
+            directoryExists = False
```
drmarkwslater: You may as well just do a `mkdir -p` and not bother with the `ls` check as `mkdir -p` will return no error if the dir exists already.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
milliams: I think that `mkdir -p`-style creation makes more sense to the user. It makes typos more 'dangerous' as it will create more unneeded directories but I think it will help more with the user's expectations.

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
```diff
@@ -0,0 +1,129 @@
...
+import tempfile
+import datetime
+import time
+import random
+import os
+import shutil
+import copy
+import string
+
+def generateUniqueTempFile( ext = '.txt' ):
```
milliams: Should be no spaces inside parentheses by PEP8.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: I'll add an abort on not existing option of some kind to restore the original functionality here.

rob-c:
If all of the tests are still passing I'm of a mind to merge this.

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
drmarkwslater: I think you might be misunderstanding: All I meant was:
```
def setUp:
    MassStorageTestConfig = ...
    extra_opts = [ ('Output', 'MassStorageFile',MassStorageTestConfig) ...]
```

`extra_opts` would still be used but you just don't have a class level variable as it's not referred to elsewhere.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -198,20 +202,52 @@ def getWNScriptDownloadCommand(self, indent):
...
+        dirName = os.path.basename(massStoragePath)
+
+        directoryExists = False
+
+        (exitcode, mystdout, mystderr) = self.execSyscmdSubprocess('%s %s' % (ls_cmd, quote(pathToDirName)))
+        if exitcode != 0:
+            if exitIfNotExist:
+                self.handleUploadFailure(mystderr, '1) %s %s' % (ls_cmd, pathToDirName))
+                raise GangaException(mystderr)
+            directoryExists = False
```
drmarkwslater: This should be just:
```
if exitcode != 0 and exitIfNotExist:
    self.handleUploadFailure...
```
The additional `directoryExists = False` isn't needed anymore.

python/Ganga/testlib/functions.py
```diff
@@ -0,0 +1,18 @@
...
+import tempfile
+import string
+
+def generateUniqueTempFile(ext = '.txt'):
+    """ Generate a unique file with a given filename with some random contents and return the name of the file on disk
+    Args:
+        ext (str): This is the extension (including '.') to give to the file of interest
+    """
+
+    with tempfile.NamedTemporaryFile(mode='w',suffix=ext,delete=False) as myFile:
```
milliams: Spaces after commas for PEP8.

python/Ganga/testlib/functions.py
```diff
@@ -0,0 +1,18 @@
+
```
milliams: I don't think that `functions.py` is a useful name for this file. `utils.py` or `file_utils.py` or similar describes the purpose much better.

python/Ganga/testlib/functions.py
```diff
@@ -0,0 +1,18 @@
...
+import random
+import tempfile
+import string
+
+def generateUniqueTempFile(ext = '.txt'):
```
milliams: The rest of `testlib` is following PEP8-style naming for functions so `generate_unique_temp_file()` would fit better.

python/Ganga/GPIDev/Lib/File/scripts/MassStorageFileWNScript.py
```diff
@@ -15,6 +15,8 @@ def execSyscmdSubprocessAndReturnOutputMAS(cmd):
 
     return (exitcode, mystdout, mystderr)
 
+from pipes import quote
```
milliams: This import should be at the top of the file.

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
rob-c: I explicitly changed it at one point in the past and plan to again in a future test (if I ever get around to writing it) so I'd prefer to keep these class level

python/Ganga/test/GPI/FileTests/TestMassStorageWN.py
milliams: I would have preferred a common file in this dir but `testlib` could be used if it's going to be used in other tests too.

python/Ganga/testlib/functions.py
rob-c: This is yet another reason why I didn't want to put these common 4/5 lines in a global function. I'll move them back to the tests if this starts to become a costly exercise.

rob-c:
I'm going to merge this at 2:30 unless there are any functional problems left against this PR. Then I'll look at making a test for #744 once I've finished some outstanding work.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -405,14 +411,15 @@ def validate(self):
 
         return (True, '')
 
-    def handleUploadFailure(self, error):
+    def handleUploadFailure(self, error, cmd=''):
```
milliams: You've named this argument `cmd` (implying it's the command that was run) but most of the time you;re passing it something like `'1) %s %s' % (ls_cmd, pathToDirName)` which is not a command.

rob-c: Yes but we now know what command caused the failure. This was frankly impossible before now.

rob-c:
OK, so after some minor name changes I'm going to merge this PR once the tests finish again.

PR #736, opened 2016-08-18
Removing LHCbApp references in GaussSplitter

rob-c:
Removing LHCbApp references in GaussSplitter

Having this in the splitter causes problems. I'm removing it from 6.1.25 onwards as any bugs from not having this I would argue are user/application bugs.


rob-c:
This needs to go into 6.1.25 or whatever the next version is as this effects LHCb users and makes MC generation impossible.

drmarkwslater:
Assuming the `LHCbApp` reference isn't needed by other things, then 👍

rob-c:
@drmarkwslater Yes it's not this needed elsewhere and has a very limited scope/use thankfully

rob-c:
@mesmith75 any objections to merging?

mesmith75:
@rob-c No objections if it works. I agree it needs to go into 6.1.25.

PR #740, opened 2016-08-19
Ensure that the VO is set properly for LCG tests

drmarkwslater:
This sets the config setting of `VirtualOrganisation` for the LCG tests because otherwise GLITE is disabled and the tests fail.

### DIFF
--- a/python/Ganga/test/GPI/LCG/TestLCG.py
+++ b/python/Ganga/test/GPI/LCG/TestLCG.py
@@ -7,6 +7,7 @@ except ImportError:
 
 from Ganga.testlib.mark import external
 from Ganga.testlib.monitoring import run_until_completed
+from Ganga.testlib.decorators import add_config
 
 
 @external
@@ -29,6 +30,7 @@ def test_job_kill(gpi):
     j.kill()
 
 
+@add_config([('LCG', 'VirtualOrganisation', "gridpp")])
 def test_submit_kill_resubmit(gpi):
     """
     Test that a simple submit-kill-resubmit-kill cycle works
@@ -57,6 +59,7 @@ def test_submit_kill_resubmit(gpi):
         j.kill()
 
 
+@add_config([('LCG', 'VirtualOrganisation', "gridpp")])
 def test_submit_monitor(gpi):
     """
     Test that an LCG job can be monitored


rob-c:
@drmarkwslater Can we change the monitor test to be something else? it has a history of timeouts

milliams:
This should be being set by the site config file created by Jenkins. https://ganga-ci.cern.ch/jenkins/job/Nightly/configure has a step called "Execute Managed Script" which runs https://ganga-ci.cern.ch/jenkins/configfiles/editConfig?id=org.jenkinsci.plugins.managedscripts.ScriptConfig1465297771272 which should set this stuff. I did it this way to avoid hard-coding any user-specific stuff like VOs into the tests themselves.

Also, for testing things like this, you might find https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/ useful.

drmarkwslater:
@milliams Apologies - I forgot that you set this up. I'll have a closer look as to why this is failing then. Maybe a version number problem looking at the managed script...

@rob-c There is only one that can timeout (`test_job_complete`) and I think it was discussed previously to have these tests not fail on timeout though I don't think that's been implemented yet.

rob-c:
@drmarkwslater Can we consider dropping this test from the nightlies in this case?
relying on whatever the state of LCG or DIRAC is could causes tests to fail quite a lot I think.
I think we can safely try to submit a job but monitoring/completing a job is best done against a mocked backend I think.
@milliams Did you have a solution for this? I forget which issue looked into this

drmarkwslater:
So I've found the problem. A couple of months ago @rob-c wanted to add running the Core tests in different environments, e.g. Core tests with the LHCb or Dirac environment setup. Unfortunately, simply adding the tests in another call was doing something very screwy (at least overwriting) to the test results. This was something @milliams said could happen after I mentioned the change but I forgot to look at it more deeply or change it back. I have now done this and the LCG tests are passing as they should be. Unfortunately, we now have a load of external Dirac failures that I'm not sure are legitimate or are just temporary. I'll keep looking but this PR is getting closed whatever 😄

drmarkwslater:
Seems they were just temporary as they all ran fine just now after a retest 😄

milliams:
Thanks for looking into that Mark. It makes sense that the way the LHCb tests were being re-run would cause problems. If we want to re-introduce that style of test I suggest it is done as a separate Jenkins job.

PR #742, opened 2016-08-19
Changing config setups

rob-c:
This PR guarantees that all of the config changes are applied before any jobs involved in any of the tests are touched by the monitoring code.

This should effect no tests but it does dramatically mess with the new MassStorage ones #735 so I'd advise that __all__ future tests avoid the use of `setConfigOption` over `extra_opts`.

This change in behaviour of the Ganga codebase can be traced to when the changes went into the GangaUnitTest around the handling of the configuration parameters. Setting them back to the default value led to some extremely strange behaviour where code would fail for 0.5s then succeed for the next 10sec due to the monitoring code starting up before the `setUp` methods had finished executing.


rob-c:
retest this please

python/Ganga/test/GPI/Bugs/TestSavannah9638.py
```diff
@@ -12,13 +12,11 @@ def wipe_temp_dir():
...
 class TestSavannah9638(GangaUnitTest):
     def setUp(self):
-        super(TestSavannah9638, self).setUp()
+        extra_opts = [('Configuration', 'gangadir', '/tmp/ganga_topdir-$USER')]
+        super(TestSavannah9638, self).setUp(extra_opts=extra_opts)
 
         wipe_temp_dir()
 
-        from Ganga.Utility.Config import setConfigOption
-        setConfigOption('Configuration', 'gangadir', '/tmp/ganga_topdir-$USER')
```
drmarkwslater: For this to work, you need to shift the `wipe_temp_dir` call as well. That's what's causing the test failures.

rob-c:
this could be revealing of an uncaught bug:
```
>       from Ganga.GPI import jobs
E       ImportError: cannot import name jobs
```

rob-c:
OK, this actually revealed a small bug in the flow of 1 test. Thankfully this doesn't mean that there are any problems with what was mainly being tested and only a few small changes are needed.

rob-c:
OK, ready to merge, can someone have a look at this?

python/Ganga/test/GPI/newXMLTest/TestSJXMLCorruption.py
```diff
@@ -63,8 +61,7 @@ def test_a_JobConstruction(self):
 
     def test_b_TestRemoveSJXML(self):
         # Remove XML force to use backup
-        from Ganga.GPI import jobs
-        XMLFileName = getSJXMLFile(jobs(0).subjobs(0))
+        XMLFileName = getSJXMLFile((0, 0))
```
milliams: Is this change supposed to be a part of this PR?

rob-c: yes because the test is broken without this

milliams:
Go for merge from me.

PR #744, opened 2016-08-22
SharedFile class

### LINKED ISSUE #39: Add a file type which writes to a shared filesystem

rob-c:
This builds upon #735 with the logical conclusion of lightly modifying the `MassStorageFile` object such that a `SharedFile` object can inherit from it.

The main changes are that the `MassStorageFile` now checks against it's `self._name` and `self.__class__()` in order to use the correctly overloaded values from any `SharedFile` configuration.

This means that very little new code is added atop #735 in order to add this new functionality.

I think the more obvious pull is to hold off on merging this until after 6.2.0 but I see no reason not to add it on top unless anyone is slightly concerned about the changes.

Again, this builds upon #735 with some minor changes on top but is a PR against `develop`.

Fixes #39 
Needs #803


rob-c:
Can someone have a look at this PR with a view to merging.

This adds a working implementation of `SharedFile` which is basically a `MassStorageFile` behind the scenes. Also this adds tests for the new file type and it behaves as you would expect.

python/Ganga/__init__.py
```diff
@@ -727,6 +727,15 @@ def filefilter(fn):
                         'defaultProtocol': defaultMassStorageProto},
                        massStorageFileExt)
 
+sharedFileBackendPost = {'LSF': 'client', 'LCG': 'client', 'ARC': 'client', 'Dirac': 'client', 'PBS': 'client', 'Interactive': 'client', 'Local': 'client', 'CREAM': 'client'}
+
+output_config.addOption('SharedFile',
+                       {'fileExtensions': [''],
+                        'backendPostprocess': sharedFileBackendPost,
+                        'uploadOptions': {'path': '/tmp/sharedStorage', 'cp_cmd': 'cp', 'ls_cmd': 'ls', 'mkdir_cmd': 'mkdir -p'},
```
milliams: I don't think it makes sense to hardcode `/tmp/sharedStorage` as the path. This is something which we should enforce being configured by the user.

python/Ganga/test/GPI/FileTests/TestSharedFileClient.py
```diff
@@ -0,0 +1,114 @@
+from __future__ import absolute_import
+from Ganga.testlib.GangaUnitTest import GangaUnitTest
+from Ganga.testlib.file_utils import generate_unique_temp_file
+
+import datetime
+import time
+import os
+import shutil
+
+class TestSharedClient(GangaUnitTest):
```
milliams: These tests seem to be copy-pasted from `MassStorageFile`. It's not surprising that they are so similar but in that case they should use a single, parametrised test.

milliams:
One thing we discussed before (and I can't tell whether this PR addresses this) is that by default, `MassStorageFile` should be unconfigured and not work until the user has provided some default arguments. Likewise, it is likely that `SharedFile` should require some configuration before it can be used.

python/Ganga/__init__.py
rob-c: How best can that be achieved?

python/Ganga/test/GPI/FileTests/TestSharedFileClient.py
rob-c: Guess I'll go back for a rewrite of that then

python/Ganga/__init__.py
milliams: Probably:

1) Set it to `None` by default
2) Check for it being `None` and inform the user that they need to set it in their `~/.gangarc`.

rob-c:
We could shift the MassStorage config to LHCb/Atlas but this is best for another PR so that we cab track/test that change. I don't want to break EOS access accidentally.

How should we enforce a user configures the configuration before using the class?
Can we come up with some standard way of enforcing this as this keeps cropping up as a concept.

python/Ganga/__init__.py
rob-c: Then how do we disable accidental use of the class in this case?

rob-c: I'm not wanting to spam all users just because we're adding an optional extra feature, that's an annoyance users will not thank us for.

drmarkwslater: But surely if a user tries to use the class (even accidently) and hasn't configured it correctly then it **should** throw an error? It would only need to be checked when the class actually does anything (or is instantiated maybe).

Btw, the `mkdir` command shouldn't have the `-p` option as this will **always** create all directories irrespective of the code in `MassStorageFile`.

drmarkwslater:
Checks could be performed in `__init__` and errors shown if there's a problem? That would be my vote anyway.

rob-c:
OK, I just updated the defaults, changed the integration tests to do more with a lot less code.
I think this addresses most of the comments from people and I'd like to see this go into 6.2.1 if possible

rob-c:
I'm concerned that this broke as a result of updating to the latest develop... I'll add an appropriate check in `MassStorageFile` before calling `mkdir`. I implicitly added `-p` to avoid this but I suppose it needs addressing properly.

rob-c:
retest this please

milliams:
This is a new feature and so will have to wait until 6.3.0. I'll hopefully be able to have a look over the code soon though.

rob-c:
This is needed for LSST work and holding up what is effectively a change of config for 6.3.0 is a bad idea that will slow down development time. In addition the changes here are very isolated so I'm not agreeing that it's worth holding this off for a month.

drmarkwslater:
There's nothing to stop us releasing 6.3.0 early and make 6.4.0 contain the credentials branch if this is an urgent feature request. However, we **really** should stick to this release policy as it makes understanding the version numbering a lot easier.

rob-c:
So can we talk about having a 6.3.0 PR branch so I can put this in and work off this branch and we can just merge 'features' against this PR which will go in ahead of the 6.3.0 release.
(Obviously the PR will then pull off develop which will be bugfixes etc)

drmarkwslater:
@milliams and I have just had a chat about this and think this may be worth giving a go. I'm going to create a 6.3-features branch that you can then rebase these changes off. After that, merge into the features branch and any other features, base off this branch as well. When we want to release 6.3 we can then merge the features branch first and then create the release after the testing has run. Any bug fixes will still be taken from and merged back into develop.

@egede @alexanderrichards We'll probably discuss this in more detail on Thursday but do you guys have any thoughts on this?

drmarkwslater:
@rob-c Rightio - I've created the `6.3-features` branch. I can't create a PR yet as it's the same as the develop at the moment. However, if you rebase this PR from that instead and merge after code review I'll create the PR and we can use that as a basis to put Release Notes in, etc.

rob-c:
Right I'm looking to merge this into the features branch unless there are complaints as I've got other IGangaFile work which will likely build on this and I'd like to test that everything is compatible.

rob-c:
Merged #785 into this PR with the intent that I'll rebase after merging this branch, however this PR had some conflicts to resolve atop that work.

rob-c:
OK, finally resolved all of the conflicts with 6.3-features so now looking to add the new file class to 6.3 features once the tests have passed. I'll leave it until the end of Monday for people to have a look at the code to see if they've any comments.

rob-c:
Any objections to adding the `SharedFile` this way into 6.3?

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -98,7 +101,7 @@ def _setNamePath(self, _namePattern='', _localDir=''):
     def _on_attribute__set__(self, obj_type, attrib_name):
         # This is defining the object as uncopyable from outputfiles... do we want this mechanism still?
         r = copy.deepcopy(self)
-        if getName(obj_type) == 'Job' and attrib_name == 'outputfiles':
+        if _getName(obj_type) == 'Job' and attrib_name == 'outputfiles':
```
drmarkwslater: Could this not be `isinstance`?

rob-c: Yes this probably could be. I'm always hesitant about touching the filters as I'm not certain where the proxy is/was stripped in the process. Given the new tests are working pretty well I'll see that this doesn't break anything and push the change 👍

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
```diff
@@ -30,13 +32,19 @@ def setUp(self):
...
-                    ('Output', 'MassStorageFile', TestMassStorageClient.MassStorageTestConfig),
+                    ('Output', TestMassStorageClient.fileName, TestMassStorageClient.MassStorageTestConfig),
                     ('Output', 'FailJobIfNoOutputMatched', 'True')]
         super(TestMassStorageClient, self).setUp(extra_opts=extra_opts)
 
+    @classmethod
+    def getFileObject(cls):
+        """ Return an instance of the file we're wanting to test """
+        import Ganga.GPI as gpi
+        return getattr(gpi, cls.fileName)
```
milliams: Could this be done as a plugin system lookup rather than a GPI `getattr`?

rob-c: I was wanting something sane, small and quick. I was not wanting to go messing with the Plugin system's internals here. This required things working and I need the proxied version of the class in that case so would likely have to call the proxy layer in addition.
I just didn't want to do in about 7-8 lines what can be done in 2.
I was toying with a clever import method that uses the strings but settled for this which works and is simple.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: OK, nothing appeared to break, now to run the full set of tests.
Can't do a global import of Job I don't think as Job imports this class by name.

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
rob-c: I'll see if I wrote a clever `ProxyObjectLookUp` method or something similar but I don't think we have one atm.

rob-c: Ah, there is `getRuntimeGPIObject` I'll make use of this

milliams: No, don't. That's a bulky slow brute-force solution. Instead it's as simple as:
```python
from Ganga.Utility.Plugin.GangaPlugin import allPlugins
file_class = allPlugins.find('gangafiles', cls.fileName)
```

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -101,7 +101,8 @@ def _setNamePath(self, _namePattern='', _localDir=''):
     def _on_attribute__set__(self, obj_type, attrib_name):
         # This is defining the object as uncopyable from outputfiles... do we want this mechanism still?
         r = copy.deepcopy(self)
-        if _getName(obj_type) == 'Job' and attrib_name == 'outputfiles':
+        from Ganga.GPIDev.Lib.Job.Job import Job
```
milliams: Can this import be put at the top of the file? It should be possible since `Job` shouldn't be importing this file.

rob-c: But it does. I __really__ don't want to go trapsing round fixing the imports with regard to the IGangaFile system. Everything gets pulled in from importing `Ganga.GPIDev.Lib.File` I think still and the work to fix this is better done in another PR as it's a much wider scope than this one.

milliams: You're right it does. That's ridiculous but honestly not that surprising in Ganga. I agree that it's not something to fix in this PR so leave it as is for now.

rob-c:
OK, slightly larger than expected commit at the end. I've expanded a Proxy interface to allow me to extract proxied classes from the GPI using strings.
In addition the MassStorageFile/SharedFile tests now all correctly use a temp directory for the `storage` during the tests which is cleaned up at the end which is safer when running tests in parallel.

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
rob-c: This _should_ be OK, but I'm weary that this has no proxy. I'll re-run with this assuming I can import this OK. (I still need the string form elsewhere but then at least there is _getName)

rob-c: I'm not wanting to pull in the Plugins as I'd like to see this die one day rather than be used/relied-upon in tests.

milliams: If you need a proxied version then using:
```python
fileClass = addProxy(MassStorageFile)
```
should suffice.

As for the plugin system, I used to feel the same but I actually think it's got a lot of use and power. It helps abstract away the interface and reduces the dependencies between classes.

rob-c: Moved to `fileClass`, lacking a proxy isn't a problem as the `stripProxy` is thankfully sensible. Just means that I now get the class name from the class vs the class from the class name.
I still think the plugin system can be redesigned to be better than it is so I'm reluctant to use it somewhere like the tests.

rob-c:
OK, updating the tests exposed a minor bug in the handling of files in this PR, I'll try and fix it today

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
rob-c: see #803 for `fileClass = addProxy(MassStorageFile)`

rob-c: Given this confusion happened it stresses that it would be useful to have `addProxy` be the inverse of `stripProxy` imo but I've reverted the classes to the underlying methods.

rob-c:
OK, this should have addressed all concerns highlighted and should be passing all tests. I'll look to merge this on Monday in the afternoon.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -2091,20 +2089,19 @@ def __setattr__(self, attr, value):
 
             # reduce duplicate values here, leave only duplicates for LCG,
             # where we can have replicas
-            uniqueValuesDict = []
-            uniqueValues = []
+            uniqueValuesDict = oDict()
```
milliams: Rather than making the code use an ordered dict (when outputfiles aren't inherently ordered), perhaps just fix the `testNoFilesSpecifiedAllSame` test to compare sorted output?

rob-c: I've mixed feelings about this given that I've seen some bad user code which expects the order here to be preserved. With this in mind a think an ordered dict is sensible unless we want to start educating users about this when their scripts break?

milliams: In this case why does it need to be a `dict`? Would not a list suffice like the code used to have?

rob-c: because the ordering isn't strongly preserved when using a dict as with a list and the test fails (randomly), hence the ordered dict to stop this happening.

rob-c: I was looking to reduct the total number of objects created and the special test for LCGSEFile (there were lots of lists created which didn't look appropriate. I'll switch back to lists which is imo more confusing but is what we had, and this shouldn't effect the user as in both cases they just get a list back)

rob-c: The full problem is this:
I have 2 `MassStorageFile` objects which are similar but request that the same file at the end of the job be placed in 2 different locations on MassStorage. This test here currently disallows this (not entirely bad) workflow. I was looking to improve the situation and discovered there is a list of keys and a list of values being used to sort everything whilst preserving the order. Given the 2 lists were being kept in sync this looked entirely like an appropriate situation for an ordered dict unless I've misunderstood what the original code is doing?

milliams: It sounds like the root problem is that the `__eq__` method on `IGangaFile`s is not intelligent enough for two separate objects resolving to the same canonical path. If it were then a simple `if f not in file_list: file_list.append(f)` would suffice.

Since the original code is a bit confusing (and I agree about the explicit `'LCGSEFile'` check) we can stick with an ordered dictionary.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -26,6 +25,7 @@
 from Ganga.GPIDev.Schema import ComponentItem, FileItem, GangaFileItem, Schema, SimpleItem, Version
 from Ganga.Runtime.spyware import ganga_job_submitted
 from Ganga.Utility.Config import ConfigError, getConfig
+from Ganga.Utility.external.OrderedDict import OrderedDict as oDict
```
milliams: If this import was simply:
```python
from Ganga.Utility.external.OrderedDict import OrderedDict
```
it would simplify the porting process when we can depend on Python 2.7.

rob-c: There was some convention of using `oDict` elsewhere in the codebase. If this is disliked I'll change back but was just trying to keep it the same when we come to make some edit to rely on this.

python/Ganga/GPIDev/Lib/Job/Job.py
rob-c: Possibly, but this equality check was not being performed until now, all I did was improve what was here but not completely rewrite it.
If we're of the mindset that we should opt for this approach then great, a load of extra noise here reduces away very quickly and I become a lot happier.

rob-c: I think we can drop the dict entirely and just have a check against a list which is what we start with and return to the user as well so opting to trust the equality method is better imo.

python/Ganga/GPIDev/Lib/Job/Job.py
milliams: There was a convention but that started before there was any knowledge of an official ordered dictionary in Python. Since the official one was named `OrderedDict` and our own `OrderedDict.py` is named similarly and is designed to be compatible it makes sense to follow that convention.

rob-c: If we're switching to `not in ...` then this becomes mute but I was hoping to keep to a convention within ganga so finding all of the `oDict` imports becomes easy and something we can _fix_ with a quick sed command or equivalent, this was not about caring for Python convention as it was for future maintenance (although a quick grep reveals this is minimal thankfully). If it stays I'll adopt the future naming scheme

rob-c:
Can someone please review this again please.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -270,6 +270,10 @@ def addProxy(obj):
                 return getattr(obj, proxyObject)
             else:
                 return GPIProxyObjectFactory(obj)
+    elif isclass(obj):
+        if not issubclass(obj, GPIProxyObject):
+            addProxyClass(obj)
+        return getProxyClass(obj)
```
milliams: Didn't the other PR remove these changes in the end?

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -107,7 +111,36 @@ def _on_attribute__set__(self, obj_type, attrib_name):
     def __repr__(self):
         """Get the representation of the file."""
 
-        return "MassStorageFile(namePattern='%s')" % self.namePattern
+        return "%s(namePattern='%s')" % (_getName(self), self.namePattern)
+
+    def mass_line_processor(self, line):
+        lineParts = line.split(' ')
```
alexanderrichards: I think that the default behaviour of split is to split on whitespace so could this simply be

```python
lineParts = line.split()
```

python/Ganga/GPIDev/Base/Proxy.py
rob-c: I'll rebase, thought I had done, probably missed this.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: Will do, this was a wonderful copy/pase :+1:

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -107,7 +111,36 @@ def _on_attribute__set__(self, obj_type, attrib_name):
...
+        pattern = lineParts[1]
+        outputPath = lineParts[2]
+        name = os.path.basename(outputPath).strip('.gz')
```
alexanderrichards: If the purpose of this line is to split out the extension of the filename then using strip can be dangerous. Consider the following example.

```python
>>> '.gg.asjkdhajkh.zz.gz'.strip('.gz')
'asjkdhajkh'
```
strip will remove any combination of the given arg characters at either end. The proper way to achieve what I think that you want is to use `os.path.splitext` as follows:

```python
>>> os.path.splitext('.gg.asjkdhajkh.zz.gz')[0]
'.gg.asjkdhajkh.zz'
```

rob-c: I'm just copying what was there before bugs/warts and all. Is there an easy way to fix this?

rob-c:
Again, should have implemented all comments in the PR, can someone review

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: OK, should be fixed better now

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -107,7 +111,47 @@ def _on_attribute__set__(self, obj_type, attrib_name):
...
+        """ This function splits the input line from the post-processsing system to define where this file is:
+        Args:
+            line(str): This is expected to be in the format of the postprocessor file from jobs transfering files on the WN
+        """
+        lineParts = line.split()
+        pattern = lineParts[1]
+        outputPath = lineParts[2]
+        split_name = os.path.splitext(outputPath)
+        if len(split_name) >= 2:
+            if split_name[1] == 'gz':
```
milliams: This should be `'.gz'`

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -107,7 +111,47 @@ def _on_attribute__set__(self, obj_type, attrib_name):
...
+    def mass_line_processor(self, line):
+        """ This function splits the input line from the post-processsing system to define where this file is:
+        Args:
+            line(str): This is expected to be in the format of the postprocessor file from jobs transfering files on the WN
+        """
+        lineParts = line.split()
+        pattern = lineParts[1]
+        outputPath = lineParts[2]
+        split_name = os.path.splitext(outputPath)
+        if len(split_name) >= 2:
```
milliams: `split_name` will *always* be length 2.

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: fixed

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
rob-c: fixed

rob-c:
OK, I plan to merge this tomorrow morning unless anyone objects I'll give it time to settle

PR #748, opened 2016-08-23
Update release notes for 6.1.25

mesmith75:
Release notes for 6.1.25

### DIFF
--- a/release/ReleaseNotes
+++ b/release/ReleaseNotes
@@ -6,6 +6,8 @@ ganga/python/GPIDev
 --------------------------------------------------------------------------------------------------------------
 * Fix to MassStorageFile. The local area is now successfully cleaned up after files have been copied
 to the mass storage area. Much improvement to the code.#733
+* Fix bug to stop duplication of files when using multiple wildcards.
+* Improved whitespace, special character support for filenames and paths.
 
 --------------------------------------------------------------------------------------------------------------
 ganga/python/GangaLHCb
--- a/release/ReleaseNotes
+++ b/release/ReleaseNotes
@@ -1,31 +1,24 @@
 **************************************************************************************************************
 @VERSION@
 
-
 --------------------------------------------------------------------------------------------------------------
 ganga/python/GPIDev
 --------------------------------------------------------------------------------------------------------------
 * Fix to MassStorageFile. The local area is now successfully cleaned up after files have been copied
-  to the mass storage area. Much improvement to the code.
-
-  --------------------------------------------------------------------------------------------------------------
-ganga/python/test
---------------------------------------------------------------------------------------------------------------
-* Added better testing for MassStorageFile.
+to the mass storage area. Much improvement to the code.#733
 
 --------------------------------------------------------------------------------------------------------------
 ganga/python/GangaLHCb
 --------------------------------------------------------------------------------------------------------------
-* Removing LHCbApp references in the Gauss splitter. This stops the pickled options being overridden.
+* Removing LHCbApp references in the Gauss splitter. This stops the pickled options being overridden. #736
 
 --------------------------------------------------------------------------------------------------------------
 ganga/python/GangaATLAS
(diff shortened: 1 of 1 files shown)


release/ReleaseNotes
```diff
@@ -3,8 +3,29 @@
...
+ganga/python/GPIDev
 --------------------------------------------------------------------------------------------------------------
-* ...
+* Fix to MassStorageFile. The local area is now successfully cleaned up after files have been copied
+  to the mass storage area. Much improvement to the code.
+
+  --------------------------------------------------------------------------------------------------------------
+ganga/python/test
+--------------------------------------------------------------------------------------------------------------
+* Added better testing for MassStorageFile.
```
milliams: I'm not sure that we need to tell the user about testing changes. We should try to keep these as useful information for user as to how things have changed or important bugs that have been fixed.

release/ReleaseNotes
```diff
@@ -3,8 +3,29 @@
...
+* Removing LHCbApp references in the Gauss splitter. This stops the pickled options being overridden.
+
+--------------------------------------------------------------------------------------------------------------
+ganga/python/GangaATLAS
+--------------------------------------------------------------------------------------------------------------
+* Fix to Condor backend. It now waits for the stdout to be populated.
+* Added command_line option to Athena to pass command line options to the job
+* Remove 64-bit config warning
+* Fix for submission to Lxbatch
+* Cleanup of unused TAG code
```
milliams: I don't think they need to know this as it will affect no one.

release/ReleaseNotes
```diff
@@ -3,8 +3,29 @@
 
 
 --------------------------------------------------------------------------------------------------------------
-ganga/python/Ganga
+ganga/python/GPIDev
 --------------------------------------------------------------------------------------------------------------
-* ...
+* Fix to MassStorageFile. The local area is now successfully cleaned up after files have been copied
```
milliams: I think it may be useful (opinions welcome but I was doing it for my release notes) to include the GihHub bug number and/or the PR number. See https://github.com/ganga-devs/ganga/releases/tag/6.1.20 for how this can work.

release/ReleaseNotes
```diff
@@ -3,8 +3,29 @@
...
+  to the mass storage area. Much improvement to the code.
+
+  --------------------------------------------------------------------------------------------------------------
```
milliams: Extraneous spaces at beginning of line.

release/ReleaseNotes
mesmith75: I agree that would be useful.

rob-c: Also fixed bug where there was a duplication of files when looping over job.outputfiles due to the use of multiple wildcards.

rob-c: Not to mention improved whitespace, special character support for filenames and paths.

mesmith75:
If there are no further comments I shall merge this.

rob-c:
Shouldn't this be named ReleaseNotes-6.1.25 or is this moved during the release process?

release/ReleaseNotes
```diff
@@ -1,10 +1,27 @@
...
+ganga/python/GangaLHCb
+--------------------------------------------------------------------------------------------------------------
+* Removing LHCbApp references in the Gauss splitter. This stops the pickled options being overridden. #736
 
 --------------------------------------------------------------------------------------------------------------
-ganga/python/Ganga
+ganga/python/GangaATLAS
 --------------------------------------------------------------------------------------------------------------
-* ...
+* Fix to Condor backend. It now waits for the stdout to be populated. #741
```
rob-c: Should this be in Core as it's a core fix?

mesmith75:
@rob-c The renaming and copying is done during the release.

release/ReleaseNotes
mesmith75: So it is. I've updated that.

PR #752, opened 2016-08-24
Cleanup exception code

drmarkwslater:
This PR cleans up the exception code, removes the import of them at the Core level and adds tests for all the exceptions. This does result in a slight change to the error messages:

```
Ganga In [1]: raise GangaException("KAJNCSJK")
ERROR    Error: GangaException:  ('KAJNCSJK',)
```

To:

```
Ganga In [1]: raise GangaException("KAJNCSJK")
ERROR    Error: KAJNCSJK
```

I feel this is fine but others may have differing opinions 😄


rob-c:
Aside from the obvious 100+ files being touched I'm in favor! :+1: 

I tried auto-flaking the codebase a while ago to see what the state was and there are still a lot of less than optimal imports hanging around. This fixes a huge amount of them so I'm tempted to merge this asap for working on which raises the issue of how we plan to maintain 6.2.1 and should this PR go in against develop or another more-unstable branch?

python/Ganga/test/GPI/TestExceptions.py
```diff
@@ -0,0 +1,45 @@
...
+    """Create all exceptions and make sure they behave correctly"""
+
+    import Ganga.Core.exceptions
+    test_str = "My Test Error"
+
+    def exception_test(err_name):
+        """Run tests on the given exception"""
+        err_type = getattr(Ganga.Core.exceptions, err_name)
+        err_obj = err_type(test_str)
+        assert err_obj.args[0].find(test_str) != -1
```
milliams: Maybe:
```python
assert test_str in err_obj.args[0]
```

drmarkwslater:
Unfortunately, yes, the number of touched files is quite large 😄  Because of this though, I'd regard this as 'feature' level rather than 'bugfix' level so I'd say 6.3.0 is the best bet.

python/Ganga/test/GPI/TestExceptions.py
```diff
@@ -0,0 +1,45 @@
...
+                "GangaValueError", "GangaIOError", "SplitterError", "ProtectedAttributeError", "ReadOnlyObjectError",
+                "TypeMismatchError", "SchemaError", "SchemaVersionError"]
+
+    for e in err_list:
+        exception_test(e)
+
+    # check the BackendError
+    from Ganga.Core.exceptions import BackendError
+    err = BackendError("TestBackend", test_str)
+    assert err.__str__().find("TestBackend") != -1
```
milliams: Similar to comment above:
```python
assert 'TestBackend' in str(err)
```
and likewise for below as well.

rob-c:
@drmarkwslater Yes I agree this should be released in 6.3.0 but where will development for that take place?
Having this and the credentials branch both as PR means there will likely be a few small conflicts which could be fixed sooner if we have branch where 6.3 is going to take shape on.
Is the plan to have several large PR which we'll merge into develop ahead of 6.3.0 or shall we have develop for 6.3 work (after 6.2.0) and have a 6.2 stable branch which just cherry-picks bugfix PR from develop?

drmarkwslater:
@rob-c I'm not sure if @milliams thought of this differently but I had envisioned releasing 6.2.0. Then having a few weeks preparing 6.3.0 features (after previously deciding what was going to go in) but not putting them in. During this time we'd only merge and release bugfixes. Then, when we think the features are ready start merging and testing (if necessary) the new features. I guess we could have a separate 'develop-6.3.0' branch but I would hope we wouldn't need that.

In this case though, because the Credentials is such a big feature, maybe we just put that in and hold off other features for 6.4?

milliams:
I think the best and simplest way to handle it is to only have one long-lived development branch. So `develop` will be used to merge in 6.2.x bugfixes until we decide we want to start 6.3. At that time we start merging in the feature PRs.

rob-c:
@drmarkwslater I just have visions of large PR going in which are alright by themselves but require effort to merge all of them due to conflicts which may or may not be obvious to resolve.

I'm not overly attached to either way of working I'm just wanting to avoid git conflicts which, whilst aren't as bad as svn are a pain imo.

rob-c:
Looks good. But i think we dropped the lhcb old tests do these can be removed

drmarkwslater:
retest this please

drmarkwslater:
retest this please

rob-c:
@drmarkwslater The purpose of the GangaKeyError is to stop the user getting a stack trace when an intentional exception of KeyError type is thrown to the user (i.e. the credentials system and elsewhere).
I suggest either think about turning off the interactive stack traces by default or avoid removing it.

drmarkwslater:
The problem I have with this code at the moment is the duel inheritance of `GangaException` and `KeyError`. It really should be one or the other because otherwise you could `catch` things you weren't expecting. Ideally, I feel it **should** be shifted to `GangaKeyError` but it seems there are currently so many places where `KeyError` is caught instead that I couldn't get this to work after quite a bit of trying. 

I suppose for the moment I'll revert back to the `GangaKeyError` but with a note that this should be fixed in the future.

rob-c:
@drmarkwslater I think we should probably look for the places where this is occuring, catch `KeyErrors` and raise `GangaKeyErrors` which don't inherit from the `KeyError` type. The difficulty then is that I htink there are some parts of the code which raise exceptions a lot would have to be looked at. I somehow feel it's bad in _normal use_ to be raising exceptions but I think that's just personal taste tbh

drmarkwslater:
Actually I agree - I'm also not a fan of using exceptions to decide, e.g. if something is in a dictionary/list or whatever. Having general code in a `try..catch` block looks ugly to me but as you say, maybe just personal taste 😄 

I'd probably much prefer to just have a function to check if a credential is present rather than relying on the exceptions and looking at the code, that seems like it would be a good way to do this. In any case, stuff for another PR one way or another 😄

drmarkwslater:
Finally got the tests working so hopefully this can go in for 6.3.1 or 6.4 (probably 6.4?). Any objections?

egede:
@drmarkwslater At the risk of being unpopular about commenting on a quaestion asked back in August.

```
This does result in a slight change to the error messages:

Ganga In [1]: raise GangaException("KAJNCSJK")
ERROR    Error: GangaException:  ('KAJNCSJK',)

To:

Ganga In [1]: raise GangaException("KAJNCSJK")
ERROR    Error: KAJNCSJK
```

Will this change not make it really hard for the user to figure out which exception was actually raised? So it will make it hard to put it into a `try` -  `except` clause. Python itself always reports the actual exception to the terminal (as stdout or stderr, I don't know).

rob-c:
I think we could change the new error message to:
```
Ganga In [1]: raise SplittingError("KAJNCSJK")
ERROR    SplittingError: KAJNCSJK
```

This should convey what has gone wrong whilst keeping most of the good code cleanup and cleaning up the error message which contains 1 too many statements that something "errored" for my liking

egede:
@rob-c Yes, that proposal is much better.

rob-c:
@drmarkwslater Do you mind if I make a small change to the PR to change the error message in this case?

drmarkwslater:
@rob-c Just done it :smile: The exceptions now print as @rob-c suggested:
```
Ganga In [1]: raise SplittingError("KAJNCSJK")
ERROR    SplittingError: KAJNCSJK 
```

OK to merge now?

egede:
Remember that @mesmith75 asked to refrain from merging until he sorted out the problems with the LHCb release process.

mesmith75:
Let's merge this now.

python/Ganga/Core/GangaRepository/Registry.py
```diff
@@ -3,8 +3,8 @@
 import functools
 from Ganga.Utility.logging import getLogger
 
-from Ganga.Core import GangaException
-from Ganga.Core.GangaRepository import InaccessibleObjectError, RepositoryError
+from Ganga.Core.exceptions import GangaException
+from Ganga.Core.exceptions import InaccessibleObjectError, RepositoryError
```
alexanderrichards: Would these not be better grouped together?

python/Ganga/Utility/Runtime.py
```diff
@@ -304,6 +304,7 @@ def loadPlugins(environment):
         except Exception as err:
             logger.error('problems with loading Plugin %s', n)
             logger.error('Reason: %s' % str(err))
+            raise
             raise PluginError("Failed to load plugin: %s. Ganga will now shutdown to prevent job corruption." % n)
```
alexanderrichards: Two raises one after another?

python/GangaLHCb/Lib/Backends/Dirac.py
```diff
@@ -2,8 +2,8 @@
 from GangaDirac.Lib.Backends.DiracBase import DiracBase
 from GangaDirac.Lib.Backends.DiracUtils import result_ok
 from Ganga.GPIDev.Schema import Schema, Version, ComponentItem
-from Ganga.Core import BackendError
-from Ganga.Core import GangaException
+from Ganga.Core.exceptions import BackendError
+from Ganga.Core.exceptions import GangaException
```
alexanderrichards: Combine these two?

python/GangaSuperB/Lib/SBApp.py
```diff
@@ -6,8 +6,8 @@
 import subprocess
 import sys
 
-from Ganga.Core import ApplicationConfigurationError
-from Ganga.Core import GangaException
+from Ganga.Core.exceptions import ApplicationConfigurationError
+from Ganga.Core.exceptions import GangaException
```
alexanderrichards: Combine?

python/GangaSuperB/Lib/SBInputDataset.py
```diff
@@ -4,8 +4,8 @@
 import os
 from subprocess import *
 
-from Ganga.Core import ApplicationConfigurationError
-from Ganga.Core import GangaException
+from Ganga.Core.exceptions import ApplicationConfigurationError
+from Ganga.Core.exceptions import GangaException
```
alexanderrichards: Combine?

python/GangaSuperB/Lib/SBOutputDataset.py
```diff
@@ -2,8 +2,8 @@
 
 import os
 
-from Ganga.Core import ApplicationConfigurationError
-from Ganga.Core import GangaException
+from Ganga.Core.exceptions import ApplicationConfigurationError
+from Ganga.Core.exceptions import GangaException
```
alexanderrichards: Combine?

rob-c:
NB the changes here now need to be incorporated into the new exceptions added since this branch was started (top of the XML Repo code are the only ones that spring to mind from me)

mesmith75:
Can this go in for 6.3.1?

alexanderrichards:
I think the points I raise in the review need addressing first, if nothing else the double raise should be fixed.

alexanderrichards:
I have prepared the patch to address these issues for you if you would like to review them on https://github.com/ganga-devs/ganga/pull/914 and then I'll merge into here.

mesmith75:
Afraid I am going to bump this.

PR #753, opened 2016-08-24
Update LHCb.ini

rob-c:
Adding a definition of ~~`GLITE_SETUP=/home`~~ `GLITE_SETUP=/bin/echo` so that all LHCb users get a gridProxy no matter how they use Ganga. The code behind this is deprecated and dying in 6.3.0 so no need to patch it now to get it working for users.

The suggested plan is that this is the __ONLY__ PR against 6.2.0 over 6.1.25 and unless there are any major problems with 6.1.25 it will be re-released as 6.2.0 next week.


drmarkwslater:
I would personally suggest using `/bin/echo` as the `GLITE_SETUP` script can be called and consequently, `/home` would raise an error. However, I guess that's moot anyway as for the LHCb setup it's never called. Whatever you decide, 👍

rob-c:
@drmarkwslater We've been wondering how/when this would happen. It would be bad imo if we're calling the wrong script for LHCb as there have been issues in the past with voms extensions when not using the LHCb scripts. It would be useful to know if we ever get an error but LHCb overloads all of the functions which are called: https://github.com/ganga-devs/ganga/pull/753/files#diff-5b3268856a6cde229321e31b0efb13b3R58

Thankfully all of this will be mute soon! :)

drmarkwslater:
@rob-c As I say, I'm guessing this doesn't matter as it's never actually called and it's just checked to be present. If this wasn't set for LHCb before then I guess it wasn't being used directly anyway.

rob-c:
This was defaulting to a file on afs which works file for people on lxplus and on systems with afs present but users on CernVM and in some setups (Not all) were reporting problems because the way that a shell object is constructed is a bit of a black art. It just has to point to something which exists due to: https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/Utility/GridShell.py#L77

rob-c:
I think I'm preferring /bin/echo tbh just because it sounds safer

mesmith75:
Can we merge this now?

rob-c:
I'd rather just fix it in cvmfs I just haven't had time to do this yet. We should contact the lhcb release shifter

mesmith75:
Wouldn't this affect installations on other systems as well without AFS access?

rob-c:
@mesmith75 no, this is a cvmfs only issue. the required default file exists on afs. Given this goes away with the credentials branch we can wait for that save putting in a default which we will remove in 6.3.0.

mesmith75:
I lost track of this - is the CVMFS configuration updated now or we still wanting this to go in?

egede:
@mesmith75 I do not think so. I see nothing new in the area.

```
bash-4.1$ ls /cvmfs/lhcb.cern.ch/lib/lhcb/GANGA 
GANGA_v505r1   GANGA_v505r20  GANGA_v505r4p1  GANGA_v600r44  GANGA_v601r22
GANGA_v505r11  GANGA_v505r21  GANGA_v505r5    GANGA_v601r11  GANGA_v601r25
GANGA_v505r14  GANGA_v505r23  GANGA_v505r6    GANGA_v601r13  GANGA_v601r4
GANGA_v505r15  GANGA_v505r24  GANGA_v505r6p1  GANGA_v601r14  GANGA_v601r6
GANGA_v505r16  GANGA_v505r25  GANGA_v505r7    GANGA_v601r17  GANGA_v601r8
GANGA_v505r17  GANGA_v505r26  GANGA_v505r8    GANGA_v601r18  GANGA_v601r9
GANGA_v505r18  GANGA_v505r3   GANGA_v507r10   GANGA_v601r19  GANGA_v602r0
GANGA_v505r19  GANGA_v505r4   GANGA_v600r12   GANGA_v601r20
```

rob-c:
@egede Apologies this has been on my todo list wrt 6.2.1. I'll make this a priority to get a request in for a new .ini file today

mesmith75:
Ok, the release doesn't depend on it anyway then.

rob-c:
OK, I'm moving to merge this as I'm getting confused with how ganga gets the config files from the env variables for cvmfs an short of setting up a new cernvm to debug this this looks like the easiest solution as everyone on the latest version will just have things working.

egede:
@rob-c The setup comes from the value of the environment variable 'GANGA_SITE_CONFIG_AREA' which defaults to '/cvmfs/lhcb.cern.ch/lib/lhcb/GANGA'. However, local clusters (like Imperial) might override this value in their login files. In that case, they will not get any benefit of changes applied.

rob-c:
@egede I'm assembling the file I want deployed on cvmfs and I think I'm just going to correct for 6-2-0 onwards as we plan to remove the older versions soon anyway.
I was getting confused by the fact Imperial makes use of overloading certain env parameters.

rob-c:
OK, I'm requesting a new config to be deployed at /cvmfs/...config_path.../6-1-20 to have the safe defaults for GaudiExec and LCG setup.
https://sft.its.cern.ch/jira/browse/LHCBDEP-3110

This is a relatively arbitrary choice but should fix cvmfs grid submission for all users using this install of ganga on LHCb.

mesmith75:
Just to check - this doesn't need merging and I can start the release?

rob-c:
Yep, sorry just wanted to wait until people saw this before I delete the branch go for 6.2.1 I think the release notes are ok from what I looked at 👍

PR #754, opened 2016-08-24
Specifiy protocol for accessURL

mesmith75:
This pull request is to allow a user to specify what protocol they want for their accessURL. Currently set the default for DiracFile to be xroot.


python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -463,7 +463,7 @@ def location(self):
                     LFNs.append(this_url)
             return LFNs
 
-    def accessURL(self, thisSE=''):
+    def accessURL(self, thisSE='', protocol='xroot'):
```
rob-c: We don't want this to be the case for GridPP, smaller VOs (non-CERN VOs) don't use ROOT. Can we make this None by default and check against the config system to get the value correct value (for LHCb it is xroot).

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -463,7 +463,7 @@ def location(self):
...
         """
         Attempt to find an accessURL which corresponds to the specified SE. If no SE is specified then
         return a random one from all the replicas.
```
rob-c: Since we're adding a new argument can we add a quick docstr on each. Thanks! :)

python/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -58,9 +58,9 @@ def getReplicas(lfns):
     output(dirac.getReplicas(lfns, active=True, preferDisk = True))
 
 
-def getAccessURL(lfn,SE):
-    ''' Return the access URL for the given LFN and storage element '''
-    output(dirac.getAccessURL(lfn,SE))
+def getAccessURL(lfn, SE, protocol):
+    ''' Return the access URL for the given LFN, storage element and protocol '''
```
drmarkwslater: Could you add appropriate docs for this function as well? As per the Google python style guide:

http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

We're trying to sort this out across the code base as we go 😄

python/GangaDirac/Lib/Files/DiracFile.py
drmarkwslater: I'm assuming that prior to this it defaulted to returning an xrootd protocol anyway? Just wanted to check the default behaviour wasn't being changed!

python/GangaDirac/Lib/Server/DiracCommands.py
rob-c: But be careful we can't use `"""` in this file due to how the file is used. It's a slightly odd limitation which is due to the way we construct Dirac scripts which would be nice to clean up at some point but is very low priority.

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: The default is determined within DIRAC so I'd rather not start requesting it here where we used to leave the default of `False`. I'm weary of moving away from the DIRAC default in-case anything changes.

mesmith75: @rob-c Yes I'll make it so that it is just LHCb that uses xroot. As you say for other collaborations something else may be more appropriate.

mesmith75: Actually, the more I think about it the less I like the idea of setting the default. I think it will just cause problems so I'll leave it blank unless anyone thinks the xroot default is necessary.

rob-c: This is good and useful (and required) mostly for LHCb isn't it?

egede: A few comments. Could the default be a configuration parameter? In that case it can be different for different experiments. It would also be good if the help text gives a feeling of the possible options for this.

mesmith75:
There looks to be a problem with this, although I cannot figure out the problem.
Specifying the protocol is giving pretty random answers. For example requesting xroot for a file on CERN-DST-EOS is giving me something that looks like a dcap URL. Similarly requesting the accessURL for the same file at GRIDKA-DST with xroot returns a failure:
```
myURL:  {'OK': True, 'Value': {'Successful': {}, 'Failed': {'/lhcb/LHCb/Collision11/BHADRON.MDST/00024034/0002/00024034_00024451_1.bhadron.mdst': 'SRM2Storage.__gfal_exec(gfal_turlsfromsurls): Execution failed. [SE][PrepareToGet][SRM_NOT_SUPPORTED] httpg://lhcbsrm-kit.gridka.de:8443/srm/managerv2: Protocol(s) not supported: [xroot]\n'}}}
```
The message seems to suggest that the correct protocol is being requested so I cannot see that it is from my part that it is going wrong.

This works from the command line with LHCbDirac. As far as I can tell this getReplicaAccessURL function from the DiracDataManager should be right. The LHCbDirac code seems to be doing something very similar.

Does anyone have any ideas?

rob-c:
@mesmith75 Are we calling the same code that is used within the dirac command line tool? If not then we should make sure if we're calling the wrong methods to update the Dirac commands.
If this is hit an miss then I think the only thing we can ever do it provide the ability for the user to request an `accessURL` and just use the default returned value from DIRAC otherwise.

mesmith75:
I am using the getReplicaAccessURL from regular Dirac. The LHCb version uses its own thing and does something with the protocols:
```python
   protocol = None
     for switch in Script.getUnprocessedSwitches():
       if switch[0] == 'Protocol':
         protocol = switch[1].lower().split( ',' )
         if 'root' in protocol and 'xroot' not in protocol:
           protocol.insert( protocol.index( 'root' ), 'xroot' )
         elif 'xroot' in protocol and 'root' not in protocol:
           protocol.insert( protocol.index( 'xroot' ) + 1, 'root' )
```
and then eventually it just does
```python
res = StorageElement( se ).getURL( lfns, protocol = protocol )
```
as per regular Dirac. I'll see if adding in a few elements from the LHCb version helps.

rob-c:
I've lost sight now. What is the expected use case of this method?
If it's internal to Ganga then this is looking like something we can't support without somewhere looping through all possible access protocols in case a site doesn't support one protocol and that this command should only provide the ability to pass the protocol to the request.
If this is some user command we can't be expected to protect against them putting in 'foobar' so it's not our responsibility to do anything other than raise an exception if the DIRAC request fails.

mesmith75:
@rob-c So the idea was to have the default protocol for the accessURL be xroot for LHCb due to the odd cases where SEs are set-up with other defaults, including ones where the default doesn't allow outside access. Checking again the twiki I think it is just CNAF that has that problem.

I think I will leave it in the current state without any defaults so if a user really wants to they can specify it. For the most part what comes back automatically is the optimum though.

mesmith75:
This looks to be working -any objections to merging?

mesmith75:
Could someone please review this? I would like it in for 6.3.1

rob-c:
It would be nice to see a nightly test for this incase the api stops working. Bit other than that it would be good to document the user option so people know what it does.

mesmith75:
@rob-c I agree for the tests - I'll put something together into another PR.

Issue #667, opened 2016-07-13
Setting Lbscripts and LHCbDirac versions


mesmith75:
I have added some functionality to specify the version of LHCbDirac to use at runtime. It is just another LHCb config option. If left blank the default is used. It checks that the given version is valid by checking that the release folder exists - if it doesn't there is a warning and the default is used.

https://github.com/ganga-devs/ganga/blob/Lbversions/python/GangaLHCb/__init__.py#L25

``` python
configLHCb.addOption('diracVersion','',"set LHCbDirac version")
```

https://github.com/ganga-devs/ganga/blob/Lbversions/python/GangaLHCb/__init__.py#L83

``` python
def _store_dirac_environment():
    from GangaDirac.Lib.Utilities.DiracUtilities import write_env_cache, get_env
    diracversion = _guess_version('LHCBDIRAC')
    optionVersion = Ganga.Utility.Config.getConfig('LHCb')['diracVersion']
    if os.path.isdir("/afs/cern.ch/lhcb/software/releases/LHCBDIRAC/LHCBDIRAC_"+optionVersion):
        diracversion = optionVersion
    elif optionVersion != '' :
        logger.warning('LHCbDirac version %s does not exist, using default version %s instead',optionVersion,diracversion)
    logger.info("Using LHCbDirac version %s", diracversion)
    platform = os.environ['CMTOPT']
    fdir = os.path.join(os.path.expanduser("~/.cache/Ganga/GangaLHCb"), platform)
    fname = os.path.join(fdir, diracversion)
    if not os.path.exists(fname) or not os.path.getsize(fname):
        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
        cmd = 'lb-run LHCBDIRAC {version} python -c "import os;     print(dict(os.environ))"'.format(version=diracversion)
        env = execute(cmd) # grab the stdout text
        env = eval(env) # env is a string so convert it to a dict
        write_env_cache(env, fname)
    os.environ['GANGADIRACENVIRONMENT'] = fname
```

Not sure how to deal with the nightlies just yet.

rob-c:
@mesmith75 I would prefer to see the the whole `_guess_version` code here go away entirely.

I think we want to declare that we expect a given version of LHCbDirac to be available and if not to warn the user (I'd say exit as well as there is likely a site mis-configuration but people running not at CERN can't all be expected to have LHCbDirac set up correctly).

[...]

PR #755, opened 2016-08-25
Enable the LHCbDirac version to be set when starting Ganga.

mesmith75:
For issue #667 

The LHCbDirac version can be set via a LHCb config option: ```"ganga -o'[LHCb]diracVersion=v8r2p50'"```. The default is set at the start of the LHCb ```__init__.py```. It checks that the supplied version exists by trying to set up the environment and checking the result is a string of length greater than 0.


python/GangaLHCb/__init__.py
```diff
@@ -82,16 +83,30 @@ def _store_root_version():
 
 def _store_dirac_environment():
     from GangaDirac.Lib.Utilities.DiracUtilities import write_env_cache, get_env
-    diracversion = _guess_version('LHCBDIRAC')
```
drmarkwslater: You can also remove the `_guess_version` code completely with this PR as this is the only use of it.

python/GangaLHCb/__init__.py
```diff
@@ -82,16 +83,30 @@ def _store_root_version():
...
     platform = os.environ['CMTOPT']
+    diracversion = Ganga.Utility.Config.getConfig('LHCb')['diracVersion']
     fdir = os.path.join(os.path.expanduser("~/.cache/Ganga/GangaLHCb"), platform)
-    fname = os.path.join(fdir, diracversion)
+    fname = os.path.join(fdir, defaultLHCbDirac)
```
milliams: Why do we care what the 'default' DIRAC version is? Can't we simply use the `diracversion` throughout?

python/GangaLHCb/__init__.py
```diff
@@ -46,7 +46,8 @@
     configLHCb.addOption('SplitByFilesBackend', 'OfflineGangaDiracSplitter',
                      'Possible SplitByFiles backend algorithms to use to split jobs into subjobs,\
                       options are: GangaDiracSplitter, OfflineGangaDiracSplitter, splitInputDataBySize and splitInputData')
-
+    defaultLHCbDirac = 'v8r2p36'
+    configLHCb.addOption('diracVersion',defaultLHCbDirac,"set LHCbDirac version")
```
milliams: Convention for naming config options is usually `UpperCamelCase` (but in some places `snake_case` is used).

python/GangaLHCb/__init__.py
mesmith75: The dirac version can be specified when starting ganga - if you give it a version that doesn't exist you need to fall back on the default. This part checks that the default environment has already been saved in case there is an empty cache and you are trying to use a non-existing version of LHCbDirac.

rob-c: I think it's sensible if that version doesn't exist to exit as something has gone wrong with setting up the environment. I'm not sure we plan to advertise changing this to users as it's mainly a change for us devs

python/GangaLHCb/__init__.py
```diff
@@ -46,7 +46,8 @@
     configLHCb.addOption('SplitByFilesBackend', 'OfflineGangaDiracSplitter',
                      'Possible SplitByFiles backend algorithms to use to split jobs into subjobs,\
                       options are: GangaDiracSplitter, OfflineGangaDiracSplitter, splitInputDataBySize and splitInputData')
-
+    defaultLHCbDirac = 'v8r2p36'
```
rob-c: great I'll modify the lhcb_prepare script to pick up this version from the codebase rather than having it hard-coded in that repo :)

python/GangaLHCb/__init__.py
mesmith75: Fair enough, now you get an error and it quits.

python/GangaLHCb/__init__.py
```diff
@@ -84,28 +65,20 @@ def _store_root_version():
...
-          logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
-        else:
-          logger.warning('LHCbDirac version %s does not exist, using default version %s instead',diracversion,defaultLHCbDirac)
-          diracversion = defaultLHCbDirac
-          fname = os.path.join(fdir, defaultLHCbDirac)
+        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
+      else:
+        msg = 'LHCbDirac version {version} does not exist'.format(version=diracversion)
+        logger.error(msg)
+        sys.exit()
```
rob-c: Doesn't it make sense to raise an exception here as it's being handled through the loading of the plugins?

python/GangaLHCb/__init__.py
```diff
@@ -84,28 +65,20 @@ def _store_root_version():
...
     fdir = os.path.join(os.path.expanduser("~/.cache/Ganga/GangaLHCb"), platform)
-    fname = os.path.join(fdir, defaultLHCbDirac)
+    fname = os.path.join(fdir, diracversion)
     if not os.path.exists(fname) or not os.path.getsize(fname):
-        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(defaultLHCbDirac), str(platform)))
-        cmd = 'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=defaultLHCbDirac)
-        env = execute(cmd)
+      cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
+      env = execute(cmd)
+      if len(env)!=0:
```
rob-c: Does this always return a list of can it return `None` if there is an error of some sort?

mesmith75: I don't think so if the error is with lb-run. The string you get just lists the env, so if there is an error that is not set and the string has no length. I guess if the execute command itself has a problem then you might end up with something that is a None, but I would assume you would find errors elsewhere to let you know?

rob-c: This is unlikely to be can't elsewhere i think, I'm not store the execute command does much to check the command ran correctly atm.

python/GangaLHCb/__init__.py
```diff
@@ -8,7 +9,7 @@
 import tempfile
 import Ganga.Utility.logging
 import Ganga.Utility.Config
-from optparse import OptionParser, OptionValueError
+from optparse import OptionParser, OptionValueError, BadOptionError
```
drmarkwslater: I could be wrong, but I don't think `BadOptionError` is used and so can be removed here.

mesmith75: Yes I was considering the exception options. I'll get rid of it.

python/GangaLHCb/__init__.py
mesmith75: Yes I see. Would it be preferable to put the check into execute and get that to return a failure or just have the check here? I think a check only here would be best in case other things don't mind if the executed command doesn't run perfectly.

rob-c: We could but this being used in very particular ways atm.
It would be nice to fix this but it quickly evolves into a bit of a major task and is best left for another PR imo.
It would be better to check for the returned object being sane until then. It would be nice to replace all calls to `execute` to deal with a tuple containing the rc of a command if run

python/GangaLHCb/__init__.py
mesmith75: I went with this as I didn't want the Traceback etc. I'll change it to an exception.

python/GangaLHCb/__init__.py
```diff
@@ -82,16 +63,22 @@ def _store_root_version():
...
-        cmd = 'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
-        env = execute(cmd)  # grab the stdout text
-        env = eval(env)  # env is a string so convert it to a dict
+      cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
+      env = execute(cmd)
+      try:
+        env = eval(env)
         write_env_cache(env, fname)
+        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
+      except:
```
milliams: Never catch an exception with a bare `except:`. Always catch the specific exception type that you are expecting.

rob-c: Is it clear what the expected exceptions are? I'd count about 3 or 4 here which starts to make a proper except statement that raises an additional exception look reasonable imo.
(Maybe just raise the original exception here for debugging with a `raise` statement?)

mesmith75: I think it is just a SyntaxError from the ```eval()``` . I've added that in.

rob-c: Yes but can't we fail to write/access the cache file and such problems and possibly not being able to parse an un-tested object into JSON format.

python/GangaLHCb/__init__.py
```diff
@@ -82,16 +64,22 @@ def _store_root_version():
...
+      cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
+      env = execute(cmd)
+      try:
+        env = eval(env)
         write_env_cache(env, fname)
+        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
+      except SyntaxError:
+        msg = 'LHCbDirac version {version} does not exist'.format(version=diracversion)
+        logger.error(msg)
+        sys.exit()
```
rob-c: Can we at least return a -ve value if we have to use a sys.exit here so that ganga in any scripts bails out in the correct way.

mesmith75: Yes sure. Would it be better to raise the OptionValueError here? I'm not really sure...

rob-c:
I'm slightly bugged by that latest error. Could probably do with some investigation on Monday.

python/GangaLHCb/__init__.py
```diff
@@ -82,16 +64,22 @@ def _store_root_version():
...
-        cmd = 'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
-        env = execute(cmd)  # grab the stdout text
-        env = eval(env)  # env is a string so convert it to a dict
+      cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
+      env = execute(cmd)
+      try:
+        env = eval(env)
         write_env_cache(env, fname)
+        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
+      except SyntaxError:
```
rob-c: I'm failing to see why this shouldn't be a catch all. If this fails due to disk full or bad permissins and failing to write the file to disk (a different exception) Ganga will continue to load. (That's even if it gets this far in the first place).

mesmith75: Good point - I had only thought about the version error. I'll add in and ```except Exception``` for generic ones.

milliams: Generic exception handlers are a bad idea because it will catch all sorts of things you don't expect and will make debugging much more difficult.
```python
try:
	var_1  = list()
	var_1.append(4)
	var_2 = [5,6,7]
	var_3 = var_l + var_2
except Exception:
	print 'something broke'
else:
	print 'everything worked'
```
will catch the `NameError` which is a programmer error, not a runtime one. This *has* happened in Ganga code in recent time due to typos and broad exception handlers.

Unless you can do something useful with an exception, you should not be catching it.

rob-c: I'd really rather see the error printed if something broke (the string rep of the exception being the obvious one) so we know what happened but if this is requiring an exit it's clear that this was invoked at some point at Ganga startup.

milliams: An `exit` here is putting the responsibility in the wrong place. The plugin system should run each plugin-load in a `try`/`except` statement, catch any thrown exceptions and either exit (if the plugin is essential) or continue (if the plugin can just be disabled). It is not up to the plugin to decide that the whole of Ganga must shutdown because it failed to load.

To put it another way, there's a difference between library and application code. Library code (e.g. the LHCb plugin) should never call things like `exit` because you never know how the application (the Ganga runtime) is using it.

rob-c: I think the vote is in favor of not exiting here as that is not the correct thing to do :+1:

mesmith75:
Made some improvements to the exceptions. Catch the SyntaxError from giving it the wrong DiracVersion so the ```eval``` fails. Put the other possible ones from writing the environment together.

python/GangaLHCb/__init__.py
```diff
@@ -1,5 +1,6 @@
 import os
 import os.path
+import sys
```
rob-c: I don't think this is needed anymore?

python/GangaLHCb/__init__.py
```diff
@@ -81,17 +63,28 @@ def _store_root_version():
...
+        msg = 'LHCbDirac version {version} does not exist'.format(version=diracversion)
+        logger.error(msg)
+        raise OptionValueError()
+      try:
         write_env_cache(env, fname)
+        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
+      except (OSError, IOError, TypeError):
+        msg = 'Unable to store LHCbDirac environment'
+        logger.error(msg)
+        raise
```
rob-c: Maybe some form `GangaException` of some sort would be nice but __not__ essential. As long as exceptions get passed up Ganga should stop loading correctly due to the plugin's requirements not being met.

mesmith75: I'm not sure how that would help? The three there are the ones that could possibly be thrown. What does the GangaException do?

drmarkwslater: I'd actually argue that keeping the original exception gives you more info about what the problem was. Though you could of course wrap it in a `GangaException` to be sure. As @rob-c says though, I don't think it's essential.

python/GangaLHCb/__init__.py
```diff
@@ -81,17 +63,28 @@ def _store_root_version():
...
+      except SyntaxError:
+        msg = 'LHCbDirac version {version} does not exist'.format(version=diracversion)
+        logger.error(msg)
+        raise OptionValueError()
+      try:
         write_env_cache(env, fname)
+        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
+      except (OSError, IOError, TypeError):
+        msg = 'Unable to store LHCbDirac environment'
+        logger.error(msg)
```
drmarkwslater: Don't bother creating the `msg` variable - just put the string in the `error` call and save one line 😄

rob-c: If we're catching so much and just passing it up, why are we catching things here?

drmarkwslater: I thought the same but actually I think it makes some sense here as if you didn't have the message, you would just have the uncaught `IOError` or whatever and that would be quite hard to track down without a stack trace (which I assume there wouldn't be?).

drmarkwslater: Just noticed a similar issue with a `msg` variable above - you can remove that as well 😄

python/GangaLHCb/__init__.py
```diff
@@ -46,27 +46,8 @@
...
-    p = re.compile(r'^\s*%s\s+%s_(\S+)\s+' % (name, name))
-    for line in tmp:
-        m = p.match(line)
-        if m:
-            version = m.group(1)
-            return version
-    msg = 'Failed to identify %s version that Ganga depends on' % name
-    raise OptionValueError(msg)
+    defaultLHCbDirac = 'v8r2p36'
+    configLHCb.addOption('LHCbDiracVersion',defaultLHCbDirac,"set LHCbDirac version")
```
milliams: Minor issue: There should be spaces after the commas in the argument list
Even more minor issue: There's an inconsistent use of types of quotation marks on this line. Perhaps just use `'` here?

mesmith75: yes, sorted

python/GangaLHCb/__init__.py
mesmith75: yes

python/GangaLHCb/__init__.py
```diff
@@ -81,17 +62,26 @@ def _store_root_version():
...
-        cmd = 'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
-        env = execute(cmd)  # grab the stdout text
-        env = eval(env)  # env is a string so convert it to a dict
+      cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
+      env = execute(cmd)
+      try:
+        env = eval(env)
+      except SyntaxError:
+        logger.error("LHCbDirac version {version} does not exist".format(version=diracversion))
+        raise OptionValueError()
```
milliams: This should probably be raised with some message like:
```python
raise OptionValueError('LHCbDirac version {version} does not exist'.format(version=diracversion))
```
so that when it's caught it can be displayed usefully.

mesmith75:
Do we want this in for 6.2.1? The LHCbDirac version needs updating so it would seem to be an apposite time, though it is more feature than bug.

rob-c:
ah, good point, if this is the case we need another PR to port this for 6.3

rob-c:
@mesmith75 Is there a good chance of getting this in for 6.2.1 or should I update the `lhcb-prepare` script with the LHCbDirac 8.3 deps and install it in the same way as 6.2.0?
Either way I plan to fix the lhcb install scripts today/this-week to make use of this PR.

mesmith75:
@rob-c I think this is finished and ready to go in. I think I implemented all the requested changes - except for the GangaException, which I don't think is necessary.

rob-c:
@mesmith75 I'd say merge this after #816 then I'll update the lhcb-prepare script today.

mesmith75:
If there are no objections I'll merge this imminently.

mesmith75:
This concludes part of #667 . Still need to take care of the Lbscripts part.

PR #759, opened 2016-08-31
Switching to getFQDN

### LINKED ISSUE #758: hostname resolution broken
nearly all LHCb users  (and probably many more not LHCb) (regardless of site etc) see their hostname stored as `localhost`.

This is due to the following exception being caught in a catch all exception handler:
``` WARNING  NameError: global name 'hostname_tmp' is not defined ```

This is thrown from:
https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/Utility/util.py#L160

I suspect this is because the object simply isn't defined.

Why we're simply not using `socket.getfqdn` I don't know but I'll prepare a PR for this.

This has ruined the monitoring data and messes up the jobs summary such that #383 is an issue.

This should probably have been caught a _long_ time ago.

This isn't serious but I will get this into 6.2.1 as this is slightly bad. If there are any more PR before 6.2.0 I'd like to bump this for a few reasons.

### PR DESCRIPTION
**Pull request by rob-c**
Fixes #758 moves to use a different method to get the fqdn of the site we're running on. Used to identify the host machine in various sections of the code.

I've had a look at the docs and I'm not sure if we need to protect against any other form of exception. But if this is called and cached I don't see too much risk in having the code just check for a socket.error

### DIFF
--- a/python/Ganga/Utility/util.py
+++ b/python/Ganga/Utility/util.py
@@ -157,7 +157,7 @@ def hostname():
         # [bugfix #20333]:
         # while working offline and with an improper /etc/hosts configuration
         # the localhost cannot be resolved
-        except socket.error as err:
+        except socket.error:
             hostname._hostname_cache = 'localhost'
 
     return hostname._hostname_cache
--- a/python/Ganga/Utility/util.py
+++ b/python/Ganga/Utility/util.py
@@ -153,11 +153,11 @@ def hostname():
     if hostname._hostname_cache == '':
         import socket
         try:
-            hostname._hostname_cache = socket.gethostbyaddr(hostname_tmp)[0]
+            hostname._hostname_cache = socket.getfqdn()
         # [bugfix #20333]:
         # while working offline and with an improper /etc/hosts configuration
         # the localhost cannot be resolved
-        except:
+        except socket.error as err:
             hostname._hostname_cache = 'localhost'
 
     return hostname._hostname_cache

### CODE REVIEW

File: python/Ganga/Utility/util.py [hunk 1]
```diff
@@ -153,11 +153,11 @@ def hostname():
     if hostname._hostname_cache == '':
         import socket
         try:
-            hostname._hostname_cache = socket.gethostbyaddr(hostname_tmp)[0]
+            hostname._hostname_cache = socket.getfqdn()
         # [bugfix #20333]:
         # while working offline and with an improper /etc/hosts configuration
         # the localhost cannot be resolved
-        except:
+        except socket.error as err:
```
milliams: There's no need to bind the exception as `err`, simply `except socket.error:` would suffice.

rob-c: The plan was to make a warning here, I'll add that in a future commit, wanted to check nothing in the tests broke for non-LHCb due to this.

rob-c:
Would there be an objection to bumping this for 6.2.0 so we can track users migrating to the new version in a better way?

milliams:
Personally I'm ok with putting this into 6.2.0 (since the current state is unambiguously broken) but I'd welcome other opinions.

mesmith75:
Looks fine for 6.2.0 for me. I'll start once this is merged.

alexanderrichards:
Yes lets do it.

PR #760, opened 2016-09-01
Make sure the --trf option is added to Jedi jobs

drmarkwslater:
Jedi needs the `--trf` flag set for TRF jobs. This makes sure this is set correctly.


drmarkwslater:
I **think** this is one of the only things to add for 6.2.0 so if everyone's happy I'll merge it and we can go with 6.2.0 😄

PR #771, opened 2016-09-05
Updating setup.cfg

rob-c:
Updating based upon warnings from Jenkins box


rob-c:
Planning to merge this after the tests finish, it's intended to remove a warning from Jenkins about using an old py.test config style.

PR #772, opened 2016-09-05
Update TestDiracCommands.py

rob-c:
Adding some assertion xfail errors around file manipulation with DIRAC. This shouldn't cause our nightly tests to fail when this is having issues.

It's currently impossible to get past these tests despite the commands running without error and the commands working in normal use at least on LHCbDirac.

I think it's better to mark these tests as able to fail and move on from there.

### DIFF
--- a/python/GangaDirac/test/Unit/DiracAPI/TestDiracCommands.py
+++ b/python/GangaDirac/test/Unit/DiracAPI/TestDiracCommands.py
@@ -227,12 +227,14 @@ class TestDiracCommands(object):
         logger.info(confirm)
         assert confirm['OK'], 'Command not executed successfully'
 
+    @pytest.mark.xfail(raises=AssertionError)
     def test_a_replicateFile(self, dirac_job):
         new_location = 'UKI-SOUTHGRID-OX-HEP-disk'
         confirm = execute('replicateFile("%s","%s","")' % (dirac_job.get_file_lfn, new_location))
         logger.info(confirm)
         assert confirm['OK'], 'Command not executed successfully'
 
+    @pytest.mark.xfail(raises=AssertionError)
     def test_b_removeReplica(self, dirac_job):
         new_location = 'UKI-SOUTHGRID-OX-HEP-disk'
         confirm = execute('removeReplica("%s","%s")' % (dirac_job.get_file_lfn, new_location))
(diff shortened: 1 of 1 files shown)


milliams:
How/why are the tests failing?

rob-c:
Keeps failing with:
https://ganga-ci.cern.ch/jenkins/job/Nightly/424/testReport/python.GangaDirac.test.Unit.DiracAPI.TestDiracCommands/TestDiracCommands/test_addFile/
```
#x1B[0;0mINFO     {'Message': 'Failed to put file to Storage Element. file:/tmp/pytest-of-jenkins/pytest-302/test_addFile0/add_file: zero number of replicas\n', 'OK': False}#x1B[0;0m
```

which may or may not happen to pretty much any of the tests.

These should be testing the API and have the potential to bring down all of the tests is say RAL goes into downtime. This isn't necessarily our problem we just want to test the API here.

milliams:
I think for the integration tests we want to try reasonably hard to make sure the test runs. If that means trying a list of sites until one works or querying some API then so be it. We want to be sure that we can make a call and the correct change is effected. We're not just testing our API, we're also testing that it calls the DIRAC server correctly.

When we just want to "test the API", that's what the mocked unit tests are for.

rob-c:
In this case there is something inconsistently broken with the DIRAC API and it's making it impossible for us to test the externals of the other plugins (including Core) correctly.
We shouldn't be failing all of the nightly tests because DIRAC is having troubles imo.
I've run the tests multiple times and get different results which either means the tests are bad and do inconsistent things with DIRAC or the DIRAC server is having troubles.

That being said most of the test code looks sensible.

rob-c:
@milliams Is there a better way of getting the nightly tests working short of disabling these tests?

milliams:
Yes, it looks like (as you say) that there is something wrong with the site (writing to `UKI-NORTHGRID-LANCS-HEP-disk` works fine for example). It's possible (and I'm putting together a PR for this now) to try a set of sites until one works.

rob-c:
@milliams I'm keen to try and get the nightlies running, is your PR near ready or shall I put in something which loops through a list of 4 or 5 sites and fails only if they all fail?

milliams:
@rob-c I'm pretty much ready, just preparing it now.

rob-c:
Superseded by #773

PR #773, opened 2016-09-06
Fix DIRAC test by trying each SE in order until one works

milliams:
Otherwise the hard-coded SE could be down and cause a false negative. This is an alternative fix for #772


rob-c:
Looks good, I'm trusting the test has been run, if it's working for you please feel free to merge and I'll remove the old PR and have a look at getting the nightlies up and running. cheers!

milliams:
Indeed I ran the test at https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/57/ and it passed (as well as locally a few times).

PR #777, opened 2016-09-06
Adding skip test around LCG tests

rob-c:
This step adds a small step to load the config and to not run the LCG tests when they're known to fail.

This adds <1 sec to the discovery of the tests so I'm planning to merge this once everything works to run some nightly tests. I'll revert if this causes problems.

### DIFF
--- a/python/Ganga/test/GPI/LCG/TestCREAM.py
+++ b/python/Ganga/test/GPI/LCG/TestCREAM.py
@@ -1,12 +1,19 @@
 import re
 import subprocess
+import pytest
 
 from Ganga.Utility.Config import getConfig
 
 from Ganga.testlib.mark import external
 
+from Ganga.Utility.Config.Config import getConfig
+from Ganga.testlib.GangaUnitTest import load_config_files, clear_config
+load_config_files()
+glite_status = getConfig('LCG')['GLITE_ENABLE']
+clear_config()
 
 @external
+@pytest.mark.skipif(glite_status == False, reason="Cannot run test GLITE not enabled")
 def test_job_kill(gpi):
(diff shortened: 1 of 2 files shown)


milliams:
This should not have gone in without a review, that's not how we do things.

There's some unnecessary repetition here, going against the DRY principle. It would be better off adding a decorator to `Ganga.testlib.mark` called something like `skipif_config` doing something like:
```python
def skipif_config(function):
    def _skip_if_config(section, item, value, reason)
        load_config_files()
        test = getConfig(section)[item] == value
        clear_config()
        if test:
            return pytest.mark.skip(function)(reason=reason)
        return function
    return _skip_if_config
```
which would then be used like:
```python
@skipif_config('LCG', 'GLITE_ENABLE', False, reason='GLITE not enabled')
def test_job_kill(gpi):
    ...
```

Of course, this doesn't change *when* the code runs (still at import) but it's wrapped up nicely in a reusable component and usable in a Pythonic way.

rob-c:
OK, Apologies I had assumed a statement of I plan to merge this and that it was 10:30 was sufficient. It was this or change half of the settings in the nightly to run a test of this which is the intention.

Given the feelings about this I'll revert as the original comment said.

milliams:
I understand but that only gave 1.5 hours of working hours to review this PR which, given that I (and I imagine others as well) have many other responsibilities beyond Ganga is not enough time.

drmarkwslater:
You can just setup your own Jenkins tests to check particular branches. That's what I've done.

milliams:
Or use [Run a test](https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/) and give it a branch and a test path.

PR #780, opened 2016-09-07
Fail on failed split

rob-c:
Adding a SplitterError if a spitter fails to split a job.

This should only pick up the case where we have 1 master job which has failed to have subjobs associated with it from a splitter but I'm keen to hear feedback.


python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -1472,6 +1471,9 @@ def submit(self, keep_going=None, keep_on_fail=None, prepare=False):
             # split into subjobs
             rjobs = self._doSplitting()
 
+            if not rjobs or len(rjobs) == 1 and rjobs[0] is self:
```
milliams: Why do we care if there is only one subjob? Is it ever the case that we have a splitter which simply sets itself as `rjobs[0]`?

rob-c: Yes, as per the code in `_doSplitting` it may return `[self]` when there are no subjobs which means that no splitting occured but we have a splitter.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -1472,6 +1471,9 @@ def submit(self, keep_going=None, keep_on_fail=None, prepare=False):
             # split into subjobs
             rjobs = self._doSplitting()
 
+            if not rjobs or len(rjobs) == 1 and rjobs[0] is self and self.splitter is not None:
```
drmarkwslater: I think it makes more sense to just check against `len(self.subjobs)` and if there is a splitter. I actually think this `rjobs` stuff probably isn't needed (or at least can be greatly simplified) but for that's probably for another PR.

rob-c: In a lot of cases you will get a list containing only a reference to the single master job and in this case the exception is not thrown as you would expect.
We could fix the rjobs stuff but frankly I'd rather not go near something in-case I break it.

drmarkwslater: I mean a conditional like:
```
if self.splitter and len(self.subjobs) == 0:
    raise ...
```
If you have a splitter then subjobs should have been filled by this point. If not it's an error.

rob-c:
retest this please

rob-c:
I strongly feel we should __not__ have DIRAC tests which can timeout in the integration tests. I'll bump this to an external in another PR as we want the integration tests to fail when and only when there is a problem with the PR imo.

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -1472,6 +1471,10 @@ def submit(self, keep_going=None, keep_on_fail=None, prepare=False):
             # split into subjobs
             rjobs = self._doSplitting()
 
+            # Some old jobs may still contain None assigned to the self.subjobs so be careful when checking length
+            if self.subjobs is not None and len(self.subjobs) == 0 and self.splitter is not None:
```
drmarkwslater: Minor: You should just need the `len` check as `subjobs` defaults to a `GangaList`. Other than that 👍

rob-c: There are still potentially jobs before the default was changed. Minor number of jobs from minor number of users but I'm erring on the side of caution.

milliams: Even before the default was changed it was set to `[]` I think which is safe to do a `len()` on? Regardless, as this code is run on `submit()`, the chance of someone having a job in `new` from that long ago which they have not submitted yet is very small.

It should be as simple as:
```python
if self.splitter is not None and not self.subjobs:
    raise ...
```

rob-c:
well replacing the original condition with what has been suggested here seems to cause more problems. I'm going to return back to the original check condition unless anyone can understand why the other suggestions here fail?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -1472,6 +1471,10 @@ def submit(self, keep_going=None, keep_on_fail=None, prepare=False):
             # split into subjobs
             rjobs = self._doSplitting()
 
+            # Some old jobs may still contain None assigned to the self.subjobs so be careful when checking length
+            if self.subjobs is not None and not self.subjobs:
```
drmarkwslater: This should be:
```
if self.splitter....
```

That's why it fails.

rob-c: Ah nvm, missed that :+1:

rob-c:
OK this should be working now. I plan to merge this if nobody objects?

PR #782, opened 2016-09-08
Removing old LHCb tests

### LINKED ISSUE #781: Remove the legacy LHCb tests
The legacy LHCb tests seems no longer to fulfil any purpose. Many have been migrated and for the remaining the number of "normal fails" is so large that we do not discover anything new.

rob-c:
Fixes #781

If there are no objections we can merge this PR which just removes the old LHCb tests.


drmarkwslater:
Assuming people are sure we don't care about these tests then 👍

milliams:
As @drmarkwslater says, I've no problem with removing these if no one is looking at them. If we want to reinstate them in a more modern form at some later date, they'll always be available in the Git history.

PR #770, opened 2016-09-05
Cleaning up IGangaFile implementations

rob-c:
I'm working on cleaning up the IGangaFile implementations specifically removing a lot of `LocalFile` special cases.

I'm wanting to be able to use the IGangaFile objects in a more generic way moving forwards and this are some of the changes I'm looking to put in.

~~I'm making the PR to run the tests on Jenkins atm, I'll ask for a review once it's finished but I want to be sure nothing has broken as a result of what I've done so far.~~
OK this should be working and roughly complete, see comments below.

PR #785, opened 2016-09-09
Cleanup xml/sessionlock implementation

rob-c:
OK, this was based off #770 perhaps incorrectly but this will ultimately go into a features branch ahead of 6.3.
There is no point reviewing anything outside of `Ganga/Core/GangaRepository` as they're ultimately not relevant to this PR.

I've cleaned up and fixed a few issues around the implementation of the XML (minor changes).

Also, I've cleaned up the `SessionLock` code which was mainly full of code which retries to to ops that were failing (I suspect 99%+ of this was actually threading/race issues). This makes the file a lot more readable again and adds a new decorator around functions requiring the global file-lock to be held.


rob-c:
OK, clearly requires some work in the case of non-AFS systems. I'll look into this at some point as it's a nicety PR not essential.

rob-c:
retest this please

rob-c:
retest this please

rob-c:
This runs fine by itself but it will fail to play nice and wait for a global file lock acquisition. This PR needs to add some simplified wait and retry logic on the global lock file just to make sure that it doesn't fight with other running PRs.

@milliams I'll look on Jenkins but has this been configured so that independent builds share a common workspace or is this correctly isolated to each build is in an isolated directory? If it's the 2nd I think we should probably consider adding a timestamp to the repo locations so that Ganga doesn't fight about global lock acquisition between tests.

rob-c:
OK so this relatively short PR removes some unused code from the XML Repo but it mainly focused on the `SessionLock` code.
I've introduced a `global_disk_lock` wrapper in addition to removing some old delay methods which would delay the opening of a file for a few seconds.

I'd like to merge this into 6.3 once someone has reviewed this, it's not urgent but it does fix a few cases where there could be file objects left open. Again, I'm happy to change things based upon input but I think this makes the resulting code a little easier to read.

rob-c:
Can someone review this if they get a chance please

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -286,6 +270,28 @@ def decorated(self, *args, **kwargs):
             return f(self, *args, **kwargs)
     return decorated
 
+def global_disk_lock(f):
+    @functools.wraps(f)
+    def decorated_global(self, *args, **kwds):
+        with global_disk_lock.global_lock:
+            i=0
+            while i< 10:
+                i+=1
```
milliams: These three line should be:
```python
for i in range(10):
```

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -286,6 +270,28 @@ def decorated(self, *args, **kwargs):
...
+                try:
+                    self.global_lock_acquire()
+                    break
+                except:
```
milliams: This must be a more precise exception handler.

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -369,48 +361,49 @@ def startup(self):
...
+    def finish_startup(self):
+        # setup counter file if it does not exist, read it if it does
+        if not os.path.exists(self.cntfn):
+            fd = None
             try:
-                fd = self.delay_init_open(self.fn)
-                try:
-                    os.write(fd, pickle.dumps(set()))
-                finally:
+                fd = os.open(self.cntfn, os.O_EXCL | os.O_CREAT | os.O_WRONLY)
```
milliams: Why does this not just use a standard `open()` call with a context manager?

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -565,8 +523,9 @@ def session_read(self, fn):
             Raises RepositoryError if severe access problems occur (corruption otherwise!) """
         try:
             # This can fail (thats OK, file deleted in the meantime)
-            fd = self.delay_session_open(fn)
+            fd = None
             try:
+                fd = os.open(fn, os.O_RDONLY)
```
milliams: Again, why not a standard `open()` with a context manager?

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -959,3 +841,4 @@ def session_to_info(self, session):
         except Exception, err:
             logger.debug( "Session Info Exception: %s" % err)
             return session
+
```
milliams: No need for this bank line.

rob-c: W298 according to pep8 says it should exist and I've editors which expect it to be there as good practice. Not functionally needed but recommended.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: you're probably correct

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: I'll rewrite the global lock acquire to sanitize the exceptions it throws then. Only  way to know whats coming from there.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: Because opening the file with a `with open` presents issues. I tried this it didn't work I didn't have the time to invest fixing it. This is what was done before. is it a requirement or an issue? Frankly I'm more interested in cleaning up other code here in this PR

python/Ganga/Core/GangaRepository/SessionLock.py
milliams: What's W292? PEP 8 says nothing about blank lines at the end of files. However, there is a general principle when working with text files in version control that they should end with a `\n`. This file already seemed to end with one and you've added a second one.

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -274,16 +274,15 @@ def global_disk_lock(f):
     @functools.wraps(f)
     def decorated_global(self, *args, **kwds):
         with global_disk_lock.global_lock:
-            i=0
-            while i< 10:
-                i+=1
+            for i in range(10):
```
milliams: Actually, this should be `range(11)` or the check below should be `if i == 9`. My mistake.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: Apologies this is the warning number from a common piece of software for checking the code, I forget which. I was having issues with a certain editor adding whitelines to files in the past. obv not the case here it's an extra line but this wasn't clear from the github interface.

rob-c:
I'm under the impression this is good to go unless there is anything pressing which needs to be fixed?

milliams:
It looks like this PR may be causing some problems on the Jenkins machine in some of the parallel tests as seen at https://ganga-ci.cern.ch/jenkins/job/PR%20builder/2114/#showFailuresLink It might be worth reverting this and taking a closer look unless you have a good idea as to the cause?

alexanderrichards:
@milliams did this test show up problems prior to merging?

milliams:
@alexanderrichards It didn't no. Given that it's the parallel tests that are failing, they're a bit more temperamental. For a reproducible local check, run:
```bash
$ python/Ganga/test/GPI/Parallel/TestQueuedSJSubmit.py
```
The test will likely *pass* but it will also print
```
close failed in file object destructor:
IOError: [Errno 9] Bad file descriptor
```
Losing track of these file handles is easy to do which I why I was encouraging the use of standard `open()` using context managers rather than `os.open()` using file descriptors. In its current state (and apparently exacerbated by this PR) it is not exception safe and can lead to leaking handles which rather breaks the purpose of a session lock file.

rob-c:
Reverting now, I'll take a look. Strange that it didn't show up when running on 3 test machines with 100 threads and >1000 jobs...
I've also seen problems with the logger having issues due to being overly full and trying to write to teh disks during the file being cycled. (I should really have opened an issue for this a long time ago but I plan to disable disk-based logging in tests by default and test it in a dedicated test)

I'll see if I can drop the open commands to something simpler and I'll make a new PR for that. I'm happy to drop this to 6.4.0 or later it was just rewriting some fairly difficult to parse code so I was keen to get it in as it doesn't seem to break things for me :/

rob-c:
Looks more like it's possibly related to: `IOError on unlock` looking for the first mistake. I'll do some digging into simplifying the code as we do want to communicate actual lock states so we may want to lock the file then open it for writing or something similar.

rob-c:
I'm concerned that this test is now no longer submitting the jobs, leaving ganga and then monitoring them. There are plenty of errors not caught by shutting down and then starting the monitoring up again... I'll do check why this happening because this wasn't what I was trying to test in the first iteration of this test.

PR #788, opened 2016-09-14
work to stabilize GangaDirac API

### LINKED ISSUE #786: More checks needed in DIRACBase
Looking through the DIRAC Base I think we should look into checking the output from DIRAC a lot more. The monitoring code in particular still falls over in bad ways when DIRAC returns something unexpected due to various reasons. This would help for people with potentially flaky connections or when DIRAC is under heavy load for instance.

rob-c:
This PR addresses several different things.
Fixes #786 

 1. This class fixes a bug in the `Ganga/Utility/execute.py` a deepcopy is needed to avoid a random difficult to track down bug
 2. I've added `GangDiracException` which is thrown from `GangaDirac/Lib/Utility/DiracUtilities.py`
 3. I've added a wrapper within the DIRAC server code which ensures that a dictionary object is always returned through the streams to Ganga.
 4. By default I've made sure that the ~~`DiracUtility`~~ `execute` function (`GangaDirac/Lib/Utilities/DiracUtilities.py`) now returns the `Value` or raises an appropriate `GangaDiracException` as this is the safest and imo best default implementation.
We can change this to return the dict by default and only inspect the dict on request but this is a point for discussion.
 5. I've separated tests which require a DIRAC proxy from tests which don't require a DIRAC proxy.
All tests requiring something to be returned via the stdout require a proxy otherwise DIRAC pollutes the stdout. If the test wants to test the executable method only it should do so explicitly.
This is not exposed to users and users are required to have a DIRAC proxy to talk to DIRAC within Ganga.
 6. Updated the tests to check for the dict always being returned from the `execute` method. This allows for better test failures.
 7. Updated the appropriate code in `GangaLHCb`, `DiracBase` and `DiracFile` to work correctly with these changes.

I'm not looking to merge this now but the tests should work.
If they don't I've made a mistake in rebasing this and will do so again later today.

Feel free to take a look through the code and as always.

Edit:
Also:
 1. Migrated the execute tests to Core tests. Fixed some bugs exposed by using this over the Dirac wrapper to this.


milliams:
(before looking at the code) What's the `DiracUtility` function you mention? I can find no reference to that string anywhere in the Ganga codebase.

rob-c:
typo `execute` in `GangaDirac/Lib/Utilities/DiracUtilities.py`

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -365,7 +369,7 @@ def kill(self):
         if not self.id:
             return None
         dirac_cmd = 'kill(%d)' % self.id
-        result = execute(dirac_cmd)
+        result = execute(dirac_cmd, return_raw_dict=True)
```
milliams: In these cases where you do
```python
result = execute(dirac_cmd, return_raw_dict=True)
if not result_ok(result):
    raise BackendError('Dirac', 'Could not kill job: %s' % str(result))
return result['OK']
```
could it not be instead
```python
try:
    result = execute(dirac_cmd)
except GangaDiracException as err:
    raise BackendError('Dirac', 'Could not kill job: %s' % err)
return True
```
which in this case (and I'd expect most all others) could remove the need for the `return_raw_dict` parameter for `execute` entirely. This PR seems to have added the ability to do automatic error checking by default for all `execute` calls but has then turned it off for almost all cases.

python/GangaDirac/test/Unit/Utilities/TestDiracUtilitiesExternals.py
```diff
@@ -0,0 +1,78 @@
+import os
```
milliams: There's no need to split these out into a separate file. It probably makes more sense to keep them all in one place (I do agree though that it makes sense to split out some of these tests into parts since some of it calls DIRAC and other parts don't).

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: Almost all calls within `GangaDirac` itself yes turn this off as they have code explicitly written to deal with this and I'm wanting to make _minimal_ changes elsewhere in the code. Touching things breaks things and I've just managed to get this stable and reasonably performant.

As I've said we cab change policy about if we want to explicitly request this, but imo it makes more sense to have this turned on as people don't think to do error checking and to pick up the cases where people are using Ganga to talk to DIRAC (aka LHCb expert users!).

This does however break backwards compatibility and there was a desire to have a `GangaDiracException` thrown from this level. If this needs to change I'd like to see some discussion on #786 about whether we infact want `GangaDiracException`.

python/GangaDirac/test/Unit/Utilities/TestDiracUtilitiesExternals.py
rob-c: Yes but this then requires manipulating globals and renaming all of the tests so that they expliclty run after a global has been changed. See the top of the previous file for what has to be so the tests can run without a proxy. Putting them in a separate file makes things a lot clearer and removes the need for some overly complex wrapper or extension to allow some methods to be called with/without a proxy.

milliams: Really I think that *all* these tests should be marked as external since they are using the `GangaDirac` version of `execute`. If there are any tests which are not testing DIRAC they should be moved to `Ganga` and test against the code version of `execute` in `Ganga/Utility/execute.py`. We then end up with:

1. `GangaDirac/test/Unit/Utilities/TestDiracUtilities.py` which are *all* marked as `@external` and only contains the tests/asserts which relate to DIRAC
2. `Ganga/test/Unit/Utility/TestExecute.py` which are *not* marked as `@external` and contain all the timeout/cwd/etc. tests.

rob-c: By that logic all of these should be moved to Ganga as they do nothing unique to DIRAC and worse require special treatment due to the fact that DIRAC has extra requirements external to these tests. I don't know why they've been placed here I was assuming there was a good reason.

milliams: You're right, they don't do anything DIRAC-specific (I was thrown off by the `assert timeit.timeit(` below).

The reason they are here in `GangaDirac` is a historical one that `execute` was originally written by Alex (I think) as part of `GangaDirac` and so the tests were placed alongside it. At some later point, the DIRAC-agnostic parts were factored out and moved to `Ganga` core but for some reason the tests remained. here. I'd be in favour of simply moving `GangaDirac/test/Unit/Utilities/TestDiracUtilities.py` to `Ganga/test/Unit/Utility/TestExecute.py`.

rob-c: I'm moving these tests as they don't contribute anything DIRAC specific here but these tests highlight again the need for testing that the tests will execute without a proxy present.

rob-c:
OK, migrated the `execute` tests to the Core. This required fixing some minor problems in the tests and `execute` so I think everything is the healthier for this.

rob-c:
OK added a `GangaDiracException` for timed-out commands which wasn't present before.
With this in mind I'm going to go through the code and allow for GangaDiracExceptions to be caught in more palces so that there is some level of consistency. This is also the first stage to allowing the automatic dictionary inspection across most of the DIRAC codebase.

rob-c:
OK, I've migrated a lot of the internal code over to use `try/except GangaDiracException` patterns.
Once the tests pass I'll run through some checks with DIRAC jobs and then look to merge this PR.

As I side note I would like to point out there are some oddities with indentation in the code, I'm not sure if this is due to different editors but we should all be careful in the future.

rob-c:
OK a complete rewrite of a lot of the DiracFile tests are needed. The cost of making sure the interface is stable I suppose.

rob-c:
OK, I've updated the tests to reflect that a `GangaDiracException` is thrown when the interface identifies that something bad happened and the function calling it can't proceed.

We probably need a separate PR to address files returning `True/False` based upon the `get/put` methods.
(In addition to needing a PR to add something like `uploadTo` to protect the logic from `put`).

This being said I believe this PR is good for inclusion as it stabilizes a lot of the DIRAC code and prevents from stupid things like trying to expand a tuple out of a string which users were seeing in production!

rob-c:
The last bug here just highlights that we need #794 to stop Ganga starting up when a plugin is bad, it hides far too many errors in it's current implementation.

rob-c:
retest this please

rob-c:
I've changed the code in `GangaDirac` to rely on the auto-inspection of the returned dictionary from the `Dirac` interface. I've updated the tests and have re-run the nightly tests locally and everything is working as expected and I've submitted and completed a few test jobs.

Is there anything else worth addressing from a stability perspective when communicating with DIRAC?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -213,6 +212,14 @@ def _checkProxy( delay=60, renew = True, shouldRaise = True, force = False ):
             _dirac_check_proxy( renew, shouldRaise )
             last_modified_time = time.time()
 
+class GangaDiracException(GangaException):
```
milliams: Most exceptions types are named `FooError` rather than `FooException` both here in Ganga and in Python in general. Therefore I think naming this `GangaDiracError` would be better.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -272,11 +281,23 @@ def execute(command,
...
 
-    return returnable
+    if isinstance(returnable, dict) and not return_raw_dict:
+        # If the output is a dictionary allow for automatic error detection
+        if returnable['OK']:
+            return returnable['Value']
+        else:
+            raise GangaDiracException(returnable['Message'])
+    else:
+        # If the output is NOT a dictionary return it
```
milliams: Do we have any situations where the returned item is not a dictionary?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -145,7 +145,10 @@ def _common_submit(self, dirac_script):
         self.statusInfo = ''
         j.been_queued = False
         dirac_cmd = """execfile(\'%s\')""" % dirac_script
-        result = execute(dirac_cmd)
+
+        try:
+            result = execute(dirac_cmd)
+        except GangaDiracException as err:
```
milliams: Could we indent the comment lines following this one too?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -974,8 +970,13 @@ def updateMonitoringInformation(jobs_):
...
 
-        DiracBase.requeue_dirac_finished_jobs(requeue_jobs, finalised_statuses)
-        DiracBase.monitor_dirac_running_jobs(monitor_jobs, finalised_statuses)
+        try:
+            DiracBase.requeue_dirac_finished_jobs(requeue_jobs, finalised_statuses)
+            DiracBase.monitor_dirac_running_jobs(monitor_jobs, finalised_statuses)
+        except GangaDiracException as err:
+            logger.warning("Error in Monitoring Loop, jobs on the DIRAC backend may not update")
+            logger.debug(err)
+            return
```
milliams: There's no need for this bare `return` statement.

python/GangaDirac/Lib/Backends/DiracUtils.py
```diff
@@ -30,27 +30,17 @@ def get_result(command,
...
             logger.debug("Sleeping for 5 additional seconds to reduce possible overloading")
             time.sleep(5.)
             if retries == retry_limit - 1:
                 raise
             retries = retries + 1
-            logger.error("An Error Occured: %s" % str(x))
+            logger.error("An Error Occured: %s" % str(err))
             logger.error("Retrying: %s / %s " % (str(retries + 1), str(retry_limit)))
 
+        return result
```
milliams: This `return result` should probably be inside the `try` block. Otherwise, if an exception is raised, `result` will not be defined.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -317,21 +319,20 @@ def getMetadata(self):
             self._optionallyUploadLocalFile()
 
         # eval again here as datatime not included in dirac_ganga_server
-        r = execute('getMetadata("%s")' % self.lfn)
+
         try:
-            ret = eval(r)
+            ret = execute('getMetadata("%s")' % self.lfn)
+        except GangaDiracException as err:
+            raise
```
milliams: These 4 lines are equivalent to just:
```python
ret = execute('getMetadata("%s")' % self.lfn)
```
which is much simpler.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -317,21 +319,20 @@ def getMetadata(self):
...
+                self.guid = ret['Successful'][self.lfn]['GUID']
         except:
-            ret = r
-        if isinstance(ret, dict) and ret.get('OK', False) and self.lfn in ret.get('Value', {'Successful': {}})['Successful']:
-            try:
-                if self.guid != ret['Value']['Successful'][self.lfn]['GUID']:
-                    self.guid = ret['Value']['Successful'][self.lfn]['GUID']
-            except:
-                pass
+            pass
```
milliams: Simply `pass`ing on a bare caught exception is a bad idea. What is this catching and what can we do about it?

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -399,16 +400,19 @@ def getReplicas(self, forceRefresh=False):
...
                     logger.error("Couldn't find replicas for: %s" % str(self.lfn))
-                    raise GangaException("Couldn't find replicas for: %s" % str(self.lfn))
+                    self._storedReplicas = {}
+                    raise
+
+                try:
+                    self._storedReplicas = self._storedReplicas['Successful']
+                except Exception as err:
+                    logger.error("Unknown Error: %s from %s" % (str(err), self._storedReplicas))
+                    raise err
```
milliams: Plain `raise` will keep more information about the exception I believe.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -469,24 +473,30 @@ def accessURL(self, thisSE=''):
...
+                this_SE = thisSE
+            # If the specified SE doesn't have a replica then return another one at random.
+            else:
+                logger.warning('No replica at specified SE for the LFN %s, here is a URL for another replica' % self.lfn)
+                this_SE = random.choice(self.locations)
+
+            try:
+                myurl = execute('getAccessURL("%s" , "%s")' % (self.lfn, this_SE))
+            except GangaDiracException as err:
+                raise GangaException("Error requesting URL: %s" % err)
```
milliams: Here (and elsewhere in this PR) we should avoid raising plain `GangaException`s. That class is intended simply as the root of the inheritance hierarchy, not as something that is actively used. Try finding an appropriate exception type (e.g. `GangaIOError`) and if there isn't one create it.

python/GangaDirac/Lib/Server/DiracDefinition.py
```diff
@@ -4,10 +4,49 @@
...
+    The intention is that all functions will now return a dict which is used to identify failures
+    Args:
+        f(function): Function we are wrapping
+    '''
+    @wraps(f)
+    def diracWrapper(*args, **kwargs):
+        ''' This method does the parsing of the wrapped function and it's output '''
+
+        # When pipe_out == False this function is being called internally and shouldn't pipe the output to the streams
+        if kwargs and 'pipe_out' in kwargs and kwargs['pipe_out'] is False:
```
milliams:
```python
if not kwargs.get('pipe_out', True):
```

python/GangaDirac/test/Unit/Utilities/TestDiracUtilities.py
```diff
@@ -12,6 +9,7 @@
 def config_files():
     """
     Load the config files in a way similar to a full Ganga session
+    NB: This is taken to be only within the context of __this__ file!
```
milliams: I don't understand this comment.

python/GangaLHCb/Lib/Server/DiracLHCbDefinition.py
```diff
@@ -4,10 +4,49 @@
...
 import pickle
+from funcutils import wraps
 ## NB parseCommandLine first then import Dirac!!
 from DIRAC.Core.Base.Script import parseCommandLine
 parseCommandLine()
 from LHCbDIRAC.Interfaces.API.DiracLHCb import DiracLHCb
 from DIRAC.Interfaces.API.DiracAdmin import DiracAdmin
 dirac = DiracLHCb()
 
+def diracCommand(f):
```
milliams: Does this need to be repeated from the `GangaDirac` version?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
rob-c: OK, so a quick `sed` to fix that

python/GangaDirac/Lib/Utilities/DiracUtilities.py
rob-c: Yes, if the user is talking to DIRAC or `print('foo')` or `output('foo')` as in the tests.

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: I've no idea the origin of these and tbh I'm tempted to remove it entirely as it's not used atm and hasn't been for a _long_ time.

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: They were in the code for catching/testing during development, will remove

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: I suspect this is trying to access a potentially missing element in a dictionary, I'll go and add some exceptions here which is fixing bugs already present in develop

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: frankly I'm just going to throw the GangaDiracException in this case, problem solved :)

python/GangaDirac/Lib/Server/DiracDefinition.py
rob-c: good catch

python/GangaDirac/test/Unit/Utilities/TestDiracUtilities.py
rob-c: It's not obvious what the scope of the loaded config was when debugging the other tests in this file. The config is (I believe) loaded here and remains valid while the tests in this file are executed.

python/GangaLHCb/Lib/Server/DiracLHCbDefinition.py
rob-c: Unless we move the definition into the Commands which I'm not sure are (or are needed to be) loaded when performing a BKQuery operation.

python/GangaDirac/test/Unit/Utilities/TestDiracUtilities.py
milliams: This is a standard pytest fixture and the pytest framework defines the scope of it. I suggest reading up on pytest as their docs will explain it better than I can.

rob-c: When I get the time, yes.

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: In future can we make sure the tests don't expect this behavior, if it's bad there should at least be a fixme in the test expressing some desire to change this.

rob-c:
These reviews really don't help work out what has or has not been addressed by future commits or am I missing something?

rob-c:
OK, currently re-running the externals but this should have addressed everything, again difficult to see as the reviews are against a fixed commit

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -298,14 +298,16 @@ def remove(self):
...
+        try:
+            stdout = execute('removeFile("%s")' % self.lfn)
+        except GangaDiracError as err:
+            logger.error("Error in removing file '%s' : %s" % (self.lfn, err))
+            raise
+
+        self.lfn = ""
+        self.locations = []
+        self.guid = ''
+        return
```
milliams: Should this be `return stdout`?

milliams:
Apart from the `return stdout` comment above, I'm happy for this to be merged. Thanks @rob-c

milliams:
Actually, I'm just running the external tests for DIRAC at https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/58/ and it looks like there might be a problem. We'll see when it finishes.

rob-c:
OK, I fixed the tests to not expect a `GangaException` to be thrown and had to add a 2nd `GangaFileError` class (I'm happy to rename to whatever but I don't think `GangaDiracError` is appropriate unless people strongly thing the 1 class should only throw 1 exception type).

@milliams  I'm not seeing the above error and at least one test for you went into a strange state. I'll look once mine has finished but it doesn't appear to be a problem for me.

python/GangaDirac/Lib/Files/DiracFile.py
rob-c: Actually I think this should be True as we expressed an interest that the remove command as well should return True/False depending on whether it succeeded or not.

rob-c:
I see different problems to do with file transfers which I think are DIRAC only problems but I'm looking into them.
@milliams Can you have a look at the changes around the thrown exceptions in this case as I've made an attempt to remove the `GangaException` from the tests as it's not directly expected to be thrown to the user.

milliams:
The error in `test_getOutputData` is being caused by `DiracBase.getOutputData` raising a `GangaException` on line 426.

The error I'm seeing in my 'Run a test' above do indeed seem to be limited to the Jenkins machine so I'm looking into that now.

milliams:
I can reproduce the errors seen in [Run a test 58](https://ganga-ci.cern.ch/jenkins/job/Run%20a%20test/58/) by running with a space in my gangadir. To test this for yourself locally, edit `python/Ganga/testlib/fixtures.py` line 22 to change `'gangadir_testing'` to something like `'gangadir_testing foo'`

rob-c:
@milliams OK, sounds like we need to quote the path of the script name being passed to the execute.
I'll have a quick look I think I know where this needs to be done.

rob-c:
@milliams Should we consider changing this directory name to force a whitespace?
Fixing bugs like this are helpful and annoying to users to ask them to rename their environments just because we're not expanding things correctly?
(Obviously if 50% or the tests fail due to this then no, but I'm re-running over core using `ganagdir testing` rather than `gangadir_testing` to see what happens)

rob-c:
@milliams Reading the error that you see, could this be a limitation in DIRAC? In which case what should be the best way of solving this I wonder?

rob-c:
OK, updating this branch has exposed how out of date some things were. I'll give this a re-read as I suspect there are now a few small inconsistencies in the tests

milliams:
I imagine that most of our tests will work with a gangadir with a space in it as I've gone and fixed this sort of issue before. Perhaps you're right though that we should change the default test gangadir to one with a space so we can catch these earlier in the future.

About the specific error I saw, it's very possible that DIRAC doesn't support paths with spaces in but I would be surprised. If we do find any then that should be reported as a bug in DIRAC.

rob-c:
@milliams The particular error you see (and I see as well) appears because the file doesn't seem to exist (I think). I'm doing some investigating now

rob-c:
OK, DIRAC really doesn't like sandboxes containing whitespace. I suspect this could be something to do with the file being uploaded to a sql database behind the scenes.
I've added some logic to copy the file to a tmpdir for submit in this case and cleanup the tmpdir after the submit has succeeded/failed.

I'll re-run the whole set of unittests (inc externals) again and hopefully everything should now be working.
If it is I'd like to look to merge this PR unless there is anything I've missed?

milliams:
Could we put together a bug report for DIRAC to reproduce this with a vanilla DIRAC script to make sure that it's definitely a bug on their part and not something we're missing at our end?

rob-c:
I'll try and collect together a working example, easier said than done when the DIRAC command line scripts seem to be unresponsive for me.
I'll submit it against the ic-hep setup as we may be able to get some help in looking into this.

rob-c:
Still finding a few tests which need fixing before I'll say I'm happy to merge this but once everything is in this PR I'll indicate.

rob-c:
OK, I'm waiting for the tests to re-run again now, had a few tests but no luck due to timeouts.
I plan to merge this if it all looks sensible as it makes the DIRAC backend much more robust.
I've collected together a minimal example showing problems within DIRAC and I'm submitting a bug now against the ic-hep instance.

rob-c:
OK, all of the tests passed (including cycling through multiple SE until the replicate functionality worked).

I'd like to look to merge this unless there is anything else to work on?

drmarkwslater:
I've got a few comments - I'll try to have the review done in a few hours. It's pretty much just minor stuff but I'm a bit concerned about the need for deepcopy. As I say, give a me a bit longer, and I'll post it.

rob-c:
@drmarkwslater 
The need to make a deepcopy was due to some problems I was seeing that the execute method was not returning a full dictionary object when it should have been. I could clearly demonstrate that without the deepcopy in I got an erroneous dictionary which was missing most of the information in question.
(running on lxplus with LHCb/LSST dev env that I have setup).

Right now I seem to be able to remove the requirement that a deepcopy be made now which doesn't go any further to explaining what I saw, and I've seen this multiple times now. When the problem exposes itself I can reliably reproduce it every single time I run the GangaDirac tests which suggests that I'm seeing something.

This has occurred for me in the past in different setups and I'm lost for explanation as to what is happening (corruption of the stream perhaps? or could it potentially be due to the .pyc/.pyo files in the project getting out of sync? could it even be thread-safety?).

Given I can't reproduce this on demand I'm happy to revert the deepcopy as it appears not to be needed, all I will say is that there is a bug somewhere which I can't track down.

rob-c:
OK, I moved the LHCb DIRAC tests over the GangaLHCb plugin and have verified that everything is working there.
Would we want to duplicate all of the standard DIRAC commands on GangaLHCb as well or just the GangaLHCb specific stuff?

python/Ganga/Utility/execute.py
```diff
@@ -92,7 +93,7 @@ def __reader(pipes, output_ns, output_var, require_output):
     os.close(pipes[1])
     with os.fdopen(pipes[0], 'rb') as read_file:
         try:
-            output_ns.update({output_var: pickle.load(read_file)})
+            output_ns[output_var] = deepcopy(pickle.load(read_file))
```
drmarkwslater: So from the comment in the main thread, I would say we go for two options:

1. Remove deepcopy and see if the bug happens again
2. Keep it in but put a comment saying this fixed a bug that couldn't be found, remove at your own risk 😄

python/GangaDirac/test/Unit/Backends/TestDiracBase.py
```diff
@@ -1,4 +1,5 @@
 import tempfile
+
```
drmarkwslater: Random blank line

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -131,8 +132,11 @@ def getReplicas(self):
         'Returns the replicas for all files in the dataset.'
         lfns = self.getLFNs()
         cmd = 'getReplicas(%s)' % str(lfns)
-        result = get_result(cmd, 'LFC query error', 'Could not get replicas.')
-        return result['Value']['Successful']
+        try:
+            result = get_result(cmd, 'LFC query error. Could not get replicas.')
+        except GangaDiracError:
+            raise
```
drmarkwslater: `try` block not needed if we're just `raise`ing

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -293,7 +297,10 @@ def getCatalog(self, site=''):
         tmp_xml = tempfile.NamedTemporaryFile(suffix='.xml')
         cmd = 'getLHCbInputDataCatalog(%s,%d,"%s","%s")' \
               % (str(lfns), depth, site, tmp_xml.name)
-        result = get_result(cmd, 'LFN->PFN error', 'XML catalog error.')
+        try:
+            result = get_result(cmd, 'LFN->PFN error. XML catalog error.')
+        except GangaDiracError:
+            raise
```
drmarkwslater: As above

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -424,7 +431,10 @@ def bkMetadata(self):
         'Returns the bookkeeping metadata for all LFNs. '
         logger.info("Using BKQuery(bkpath).getDatasetMetadata() with bkpath=the bookkeeping path, will yeild more metadata such as 'TCK' info...")
         cmd = 'bkMetaData(%s)' % self.getLFNs()
-        b = get_result(cmd, 'Error removing replica', 'Replica rm error.')
+        try:
+            b = get_result(cmd, 'Error removing replica. Replica rm error.')
+        except GangaDiracError:
+            raise
```
drmarkwslater: And again 😄

python/Ganga/Utility/execute.py
rob-c: I'd prefer to add a comment in this case if people are happy with this :)

python/Ganga/Utility/execute.py
```diff
@@ -266,30 +267,40 @@ def execute(command,
...
         else:
             logger.debug("Execute Err: %s", err)
-        local_ns = {}
+
+    if not stdout_temp:
+        local_ns = locals()
         if isinstance(eval_includes, str):
             try:
-                exec(eval_includes, {}, local_ns)
+                exec(eval_includes, local_ns)
```
rob-c: you need to have the `eval includes` in order to eval say a `datetime` object and then parse the `repr(datetime(...))` back into a real object which is expected to be returned from this function.
I could knock this back to use the local_ns as the local_ns I was just wanting to just work with the 1 namespace save confusion.
Should I reset the global ns to be {} here?

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -222,7 +224,11 @@ def submit(self, subjobconfig, master_input_sandbox):
         f = open(dirac_script_filename, 'w')
         f.write(dirac_script)
         f.close()
-        return self._common_submit(dirac_script_filename)
+        try:
+            return self._common_submit(dirac_script_filename)
+        except GangaDiracException as err:
+            logger.warning("Error during job submission: %s" % err)
+            raise GangaException(err)
```
rob-c: What version are we reviewing this was renamed to Error a while ago, does it still exist in the code in the latest form? If so I'll just let the command raise itself (bar the `finally:` method which has now been added).

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -365,21 +371,22 @@ def kill(self):
...
             return None
         dirac_cmd = 'kill(%d)' % self.id
-        result = execute(dirac_cmd)
-        if not result_ok(result):
-            raise BackendError('Dirac', 'Could not kill job: %s' % str(result))
-        return result['OK']
+        try:
+            result = execute(dirac_cmd)
+        except GangaDiracException as err:
+            raise BackendError('Dirac', 'Could not kill job: %s' % str(err))
```
rob-c: `BackendError` is expected from this interface is a `GangaDiracError` has been thrown the rest of Ganga probably does not expect to see a `GangaDiracError` here so I'd rather submit a sensible `BackendError` which could be raised from any backend.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -527,8 +524,7 @@ def _bulk_updateStateTime(jobStateDict, bulk_time_lookup={} ):
         """
         for this_state, these_jobs in jobStateDict.iteritems():
             if bulk_time_lookup == {} or this_state not in bulk_time_lookup:
-                bulk_result = execute("getBulkStateTime(%s,\'%s\')" %
-                                        (repr([j.backend.id for j in these_jobs]), this_state))
+                bulk_result = execute("getBulkStateTime(%s,\'%s\')" % (repr([j.backend.id for j in these_jobs]), this_state))
```
rob-c: yes it should expect to see a `GangaDiracError` I'll check the latest code.

python/Ganga/Utility/execute.py
```diff
@@ -266,30 +267,40 @@ def execute(command,
...
     try:
+        # If output
         if stdout:
-            stdout = pickle.loads(stdout)
+            stdout_temp = pickle.loads(stdout)
     except pickle.UnpicklingError as err:
+        # If here, output likely unpickled
         if not shell:
-            logger.error("Execute Err: %s", err)
+            logger.debug("Execute Err: %s", err)
```
rob-c: this is an error in the latest code

python/GangaDirac/Lib/Backends/DiracBase.py
rob-c: This should throw actually looking at the code. All a try/except would do here is forward the exception up the stack

rob-c:
OK, so I've removed a load of un-needed try/except blocks as well as some other small changes. This should have addressed everything raised I think. I'll look to review tomorrow and will try and check everything again before then.

egede:
> Would we want to duplicate all of the standard DIRAC commands on GangaLHCb as well or just the GangaLHCb specific stuff?

Given that it is actually a different external Dirac package we load, then it would indeed be helpful to repeat the tests.

python/Ganga/Utility/execute.py
drmarkwslater: I have no strong feelings really as long as it's clear 😄

python/Ganga/Utility/execute.py
drmarkwslater: I doubt it makes a big difference but I'd have a mild preference to doing that simply because otherwise it looks like there's a specific reason to set the global ns as the local ns.

python/GangaDirac/Lib/Backends/DiracBase.py
drmarkwslater: Apologies - I started the review end of last week and I guess there have been changes since then.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -400,16 +401,19 @@ def getReplicas(self, forceRefresh=False):
...
                     logger.error("Couldn't find replicas for: %s" % str(self.lfn))
-                    raise GangaException("Couldn't find replicas for: %s" % str(self.lfn))
+                    self._storedReplicas = {}
+                    raise
+
+                try:
+                    self._storedReplicas = self._storedReplicas['Successful']
+                except Exception as err:
+                    logger.error("Unknown Error: %s from %s" % (str(err), self._storedReplicas))
+                    raise err
```
drmarkwslater: Actually scratch this comment as the error is providing more info.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -691,38 +704,29 @@ def put(self, lfn='', force=False, uploadSE="", replicate=False):
...
-                # return ## WHY?
             else:
-                failureReason = "Error in uploading file %s : %s" % (os.path.basename(name), str(stdout))
-                logger.error(failureReason)
-                if regex.search(self.namePattern) is not None:
-                    d.failureReason = failureReason
-                    outputFiles.append(GPIProxyObjectFactory(d))
-                    continue
-                self.failureReason = failureReason
-                return str(stdout)
```
drmarkwslater: I think this is the only thing that hasn't been addressed. After that, 👍

python/GangaLHCb/test/Unit/DiracAPI/TestDiracCommands.py
```diff
@@ -0,0 +1,368 @@
...
+
+from Ganga.Utility.logging import getLogger
+from GangaDirac.Lib.Utilities.DiracUtilities import execute
+
+from Ganga.testlib.mark import external
+from Ganga.testlib.GangaUnitTest import load_config_files, clear_config
+
+logger = getLogger(modulename=True)
+
+random.seed()
```
milliams: Why is this needed? I thought that by default Python seeded the random number generator to something 'random' automatically?

rob-c: I don't know/remember I think I added some quick and dirty code for making random file contents a while back and it seems to have been copied verbatim. I'll replace this with an appropriate uuid call.

rob-c: OK, switched to `uuid.uuid4` which should provide unique checksums when uploading a file to the DFC.

python/GangaDirac/test/Unit/DiracAPI/TestDiracCommands.py
```diff
@@ -65,11 +49,11 @@ def load_config():
...
 def dirac_job(load_config):
 
-    sandbox_str = random_str()
+    sandbox_str = uuid.uuid4()
     time.sleep(0.5)
-    get_file_str = random_str()
+    get_file_str = uuid.uuid4()
     time.sleep(0.5)
-    remove_file_str = random_str()
+    remove_file_str = uuid.uuid4()
```
milliams: There's now no need to `sleep` between these calls.

rob-c: done

rob-c:
I've some minor code cleanups to commit which are not functional changes. I'll update the PR and look to merge after todays meeting hopefully :)

rob-c:
If there are no objections I'll merge this at lunchtime.

PR #789, opened 2016-09-14
Missing if statement

datty:
Missing if from jobwrapper section, causes jobs to fail immediately with ARC backend.


rob-c:
retest this please

rob-c:
I'd love to merge this but github won't let me. What is required to get a commit in is seeming like a moving target for some reason. Hopefully it'll be in by the end of today after someone else has a look at what's going on.

rob-c:
Why is it now possible to merge this. What changed?

rob-c:
@datty Can you run `update branch` on this PR? I assumed one of us on the Ganga team could do it but it appears to be greyed out (at least for me). I'll look to merge the PR if everything is good after this). Thanks!

alexanderrichards:
`update branch` is greyed out because the PR is pulling in changes from a downstream branch owned by @datty which we do not have write access to.

@datty as @rob-c says you will need to update your branch before it can be merged. A simple rebase on our develop should suffice

rob-c:
@alexanderrichards I assumed it was something like this, thanks for confirming that this is expected behavior :+1:

datty:
Think that should be all done, sorry for the delay

rob-c:
retest this please

PR #792, opened 2016-09-15
Adding lock around GaudiExec build step

### LINKED ISSUE #790: GaudiExec not thread safe on master submit
Just a holder issue to the effect that I have to re-add a global lock around the GaudiExec master_prepare which was removed in code review. I'll update this with more details once I've finished the work and have come up with a testing script (Not an integration test for jenkins as this requires a lot more work).

rob-c:
Adding a lock around the build process and adding some minor formatting changes in GaudiExec.
Fixes #790


egede:
While this is safe in all cases, it is also a bit annoying. Two build processes can happily be going on at the same time of the directories are not the same. On the other hand, two simultaneous build requests on the same directory is almost for sure an error from the user (our a highly inefficient work pattern). So should build on different directories at the same time be allowed, and a build request for one that is already in progress result in an error rather than just waiting for the lock to be released?

rob-c:
@egede Would it be sensible to give an error when this conflict is hit that it may be better for a user to first prepare and application and then submit it.
The difficulty is that users may wish to use the same application on disk but run with different configurations on the backend which does necessitate a rebuild of the application as the target is removed and the application should be read-only after a prepare (I think we still need a test for this!)

mesmith75:
Is this ready to merge?

rob-c:
I think I would like to add some sort of error if the lock is hit by multiple threads but this is secondary to having a fix like this in imo.

rob-c:
so I'm keen to merge

PR #794, opened 2016-09-16
Safely shut down ganga to prevent bad things

### LINKED ISSUE #793: Ganga should exit if it can't load a plugin
Ganga should be exiting if it can't load a plugin to prevent job corruption.

If a user has a plugin which is failing to load (for whatever reason) Then Ganga should stop the startup process and shutdown rather than startup without the plugin as this can lead to confusion and potentially risk stored Job corruption.

rob-c:
Fixes #793 

This replaces the erroneous template error and raises a new `GangaException` when a plugin fails to launch.

This correctly triggers a stop to the Ganga startup process which hopefully should allow the user a chance to fix whatever in their system/configuration is causing the problems. (or to report a bug).


python/Ganga/Utility/Runtime.py
```diff
@@ -257,12 +257,14 @@ def loadPlugins(environment):
             logger.error('problems with loading Named Templates for %s', n)
             logger.error('Reason: %s' % str(err))
 
-    for r in allRuntimes.values():
+    for n, r in zip(allRuntimes.keys(), allRuntimes.values()):
```
milliams: Why not:
```python
for n, r in allRuntimes.items():
```

python/Ganga/Utility/Runtime.py
```diff
@@ -257,12 +257,14 @@ def loadPlugins(environment):
...
         try:
             r.loadPlugins()
         except Exception as err:
-            logger.error('problems with loading Named Templates for %s', n)
+            from Ganga.Core.exceptions import GangaException
+            logger.error('problems with loading Plugin %s', n)
             logger.error('Reason: %s' % str(err))
+            raise GangaException("Failed to load plugin: %s. Ganga will now shutdown to prevent job corruption." % n)
```
milliams: We should probably avoid raising bare `GangaException`s. Instead a new exception type (`GangaPluginError` or similar) should be used so that when it's caught you know what happened.

python/Ganga/Utility/Runtime.py
```diff
@@ -257,12 +257,14 @@ def loadPlugins(environment):
             logger.error('problems with loading Named Templates for %s', n)
             logger.error('Reason: %s' % str(err))
 
-    for r in allRuntimes.values():
+    for n, r in zip(allRuntimes.keys(), allRuntimes.values()):
         try:
             r.loadPlugins()
         except Exception as err:
-            logger.error('problems with loading Named Templates for %s', n)
+            from Ganga.Core.exceptions import GangaException
```
milliams: This import should be at the top of the file if possible.

python/Ganga/Utility/Runtime.py
rob-c: Following coding style above so I'd rather stick to what's already used for sake of readability.

python/Ganga/Utility/Runtime.py
rob-c: `GangaPluginError` it is then.

rob-c: Which doesn't exist and pulling in Core code here is against the point of splitting Core and Utlity... I'll create a new Exception type for the sake of this then.

rob-c: This just makes a mess of things frankly. I don't want to global import a file from Core here as that is clearly bad and I can't create a class deriving from GangaException here which we want to avoid throwing an un-caught exception to the user. (Which will arrive from some plugins!) I'm opting to keep the import here.

python/Ganga/Utility/Runtime.py
rob-c: Now defunct since latest commit but github is no longer showing this.

python/Ganga/Utility/Runtime.py
rob-c: No, you pull in too much accidental stuff by going near Ganga.Core

rob-c:
What is the status of this? I'm looking to merge as we need this for #755 to raise correctly.

python/Ganga/Utility/Runtime.py
drmarkwslater: I'm confused - I think @milliams was just saying it would be good to have a new exception class for this type of exception and import that from the exceptions in Core instead of `GangaException`, i.e. just create a new Exception class in Core.exceptions and import+raise that instead.

python/Ganga/Utility/Runtime.py
drmarkwslater: I may be missing something but looking at the code you don't pull in anything from the `__init__` of `Ganga.Core` and in most other places in the code base I've seen the exception imports are at the top of the code. Is there a circular import problem?

python/Ganga/Utility/Runtime.py
milliams: Why can't we 'global import a file from Core here'? This file should be able to depend on any Core module just fine.

rob-c: 1. I'm reluctant to start going around making new exceptions for all the different situations, we just end up with a dictionary of possible classes in the exception which doesn't really shed much light imho.
 2. Fine, I'll try a global import again, but once again I strongly feel this is the wrong thing to be doing here as we shouldn't really have Core classes being imported into Utility imo as well as the amount of extra stuff pulled in out of a sane order or safe way when we do this.

rob-c:
Changes made, this should be ready for merging I think

python/Ganga/Utility/Runtime.py
drmarkwslater: 1. I would personally argue that this is a big enough collection of errors (i.e. any failure to load a plugin) that it's worth it's own exception. However I agree that in general we would want to avoid a proliferation of these.

2. But we need to do the Core import at some point to use the exception at all and therefore it's just a question of where it goes. The import itself won't pull anything in outside of the exceptions looking at the code and we (or at least me) have been trying to make these imports as safe as possible so there is an absolute minimum of stuff done at import time for just this reason.

rob-c: I've fixed this to what is requested as per the latest commit but this whole file (in fact a _lot_ of files) need to have the import situation sorted out.
Fixing 1 or 2 methods ad-hoc just makes the files messy imo and instead of each method importing what it needs where it needs it (yes a bad thing) we now have a horrible mix of global and local imports which should probably be 90%+ global imports due to a handful of circular dependencies.

python/Ganga/Core/exceptions.py
```diff
@@ -42,6 +42,22 @@ def __str__(self):
...
+        """ We only need a message here to be constructed
+        Args:
+            msg(str): Message of what caused this exception
+        """
+        self.msg = msg
+
+    def __str__(self):
+        """ Return a string explainng our message, just like GangaException """
+        return "PluginError: %s" % self.msg
+
```
milliams: This could be as simple as:
```python
class PluginError(GangaException):
    pass
```
and inheritance will take care of the rest.

drmarkwslater: This would also make it fit in nicely with #752 actually. I should really look in to getting that merged...

rob-c:
And now?

PR #796, opened 2016-09-16
adding uploadTo to IGangaFile

rob-c:
This is in an early alpha stage of what I'm proposing to do to add `uploadTo` to the `IGangaFile` GPI to be the `put` equivalent to the `copyTo`.

Edit:
I've also added:
  `makeStorageDir` which is expected to be called from within the put method when there is a parent job.
  `getOutputFilename` which is expected to be called from whithin the put method and returns the filename structure we want to use within this `IGangaFile`

github should resolve that this is small set of changes atop `6.3-feature`s right now and I'll update this with other files working with the new API shortly but I invite people to have a look and comment on the API.


rob-c:
Does anyone have any useful input on how to crush the history and diffs picked up by github on this branch?
This branch is isolated to ~8 files atop `6.3-features` but is showing all of the diffs as-if this is a PR against develop.

I could condense the work into another branch and start again but I'd rather not as I think this is a github issue.

rob-c:
OK nvm, pulling from the commit on the branch I suspect. I'll try and fix this now.

egede:
@rob-c can we get an overview of the changes at the dev meeting on Thursday. Both Alex and I are very confused and fear that users will be the same.

rob-c:
@egede I was hoping to _not_ expose the new methods (as with `copyTo` I wanted this private as my use was internal). This is mainly to allow us to write code that doesn't care what implementation of the `IGangaFile` we're using.
My problem is if you call `get` or `put` on a file the whole where the file comes-from/goes-to varies between `IGangaFile` objects in a difficult to manage way.

The desire from this PR us to be able to upload a `DiracFile`(`GoogleFile/LCGSEFile`) with a different name per jid/sjid whilst keeping the same interface and not just duplicate code from `MassStorageFile` which adds overhead.

I'm prototyping most of the work needed for this here (It can land 6.3 or 6.4 I don't mind) and we can change the API and expose what we want but I'm just wanting to get to a stage where we know what the minimum requirements are for each filetype.

rob-c:
Would be good to clean up the interface here to have a unified method to put as well as get, however this code is unlikely to be useful as there were large scale changes to get all of the IGangaFile objects passing a consistent set of tests,

PR #798, opened 2016-09-19
Adding whitespace to all test dirs

rob-c:
Just a quick PR to highlight the effect of adding whitespace to the test directory name.

This shouldn't effect tests run through the `PR Builder` but will highlight problems in the `nightly` builds.

Depending on how people feel I propose changing `gangadir_testing` to `gangadir testing` in the test-suite to catch problems with whitespaces which crop up atm under some tests but not others.

I'm happy to revert if there is opposition, but I thought this might be a good idea.

### DIFF
--- a/python/Ganga/testlib/GangaUnitTest.py
+++ b/python/Ganga/testlib/GangaUnitTest.py
@@ -1,4 +1,4 @@
-from __future__ import print_function, absolute_import
+from __future__ import print_function
 
 import sys
 import shutil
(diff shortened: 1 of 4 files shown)


rob-c:
Actually doing this highlighted some bugs on AFS. I'll make a PR for this soon.

python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -8,6 +8,7 @@
 except ImportError:
     import unittest
 
+ganga_test_dir = 'gangadir testing'
```
milliams: I would call this `ganga_test_dir_name` so make it clear that it's not the full directory but rather only a fragment.

rob-c:
retest this please

rob-c:
@milliams Am I ok to merge this?

PR #800, opened 2016-09-21
Making sure ~/.gangarc file isn't parsed during startup of tests

### LINKED ISSUE #799: Should tests respect the .gangarc?
Hi all,

Just a quick question, should we let the test configuration as used by py.test make use of the `.gangarc` configuration file?

I only ask this as I am surprised to see settings in my .gangarc file _not_ overloaded by the test configuration.
(In particular because I've been experimenting with my gangadir in different locations I ended up with multiple tests running on the same repo which is less than ideal).

Should we bump the setting of options within the tests to be the same as a a user-set option rather than what it is now so that the test configuration always takes priority?
If the test knows it needs a certain configuration parameter having a different one from .gangarc would just lead to false errors I think an could lead to confusion.

I'm keen to hear people's thoughts on this,

Rob

rob-c:
Fixes #799 

I've changed the default config file to point to a non-existent path so that changes to `~/.gangarc` on dev machines (and Jenkins) are isolated from the running of the tests.

This should hopefully reduce confusion and will mean anyone running a test doesn't run the risk of torching a production repo while they do so.

This appears to work without problem and I've verified it solved my problem with my tests taking forever to shutdown while they torched my 2k Job test repo.

Can someone review this as I'd like to merge this for other dev work I'm doing. Thanks!

### DIFF
--- a/python/Ganga/testlib/GangaUnitTest.py
+++ b/python/Ganga/testlib/GangaUnitTest.py
@@ -9,6 +9,7 @@ except ImportError:
     import unittest
 
 ganga_test_dir_name = 'gangadir testing'
+ganga_config_file = '/not/a/file'
 
 def _getGangaPath():
     """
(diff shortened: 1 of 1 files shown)


rob-c:
If there are no objections I'll merge this after lunch

Issue #738, opened 2016-08-18
Fast accessURL on many Dirac files


rmatev:
`DiracFile.accessURL` is great, but way too slow (~8 s). I often have jobs with few hundred subjobs. If I call `.accessURL()` on the output file of each subjob, it will take O(hour) to get all the urls, which is unacceptable.

My solution to this was to make one Dirac call for many files:

``` python
from GangaDirac.Lib.Utilities.DiracUtilities import execute
opts = '--Protocol xroot,root'
cmd = 'dirac-dms-lfn-accessURL {} {}'.format(','.join(lfns), opts)
output = execute(cmd, shell=True)
```

Unfortunately, this is broken in 6.1.22 (and 6.1.24), MWE:

``` python
[22:42:41]
Ganga In [6]: from GangaDirac.Lib.Utilities.DiracUtilities import execute

[22:42:46]
Ganga In [7]: execute('dirac-dms-lfn-accessURL /lhcb/user/r/rmatev/2016_08/136797/136797530/ntuples.root', shell=True)
ERROR    Error: insecure string pickle
ERROR    Unknown/Unexpected ERROR!!
...
ValueError: insecure string pickle
```

I'd appreciate a lot any workarounds as well as a proper longer term solution.

rmatev:
And of course I don't need ganga's execute in this case. It can be replaced by
```python
subprocess.check_output(['lb-run', 'LHCbDirac/latest', 'dirac-dms-lfn-accessURL', '/lhcb/user/r/rmatev/2016_08/136797/136797530/ntuples.root'])
```

The question about a proper solution still holds.

mesmith75:
You make a good point about the time. I think it is because ganga looks up the list of locations for the DiracFile and then gets the accessURL for a specified location picked from the list. Both of these things take a little time.

We can certainly come up with a function in ganga to give you the accessURLs from a list of LFNs.
As you point out that functionality exists with LHCbDirac. You can download a list of LFNs into a text file from Ganga:
```python
J = 1017
j=jobs(J)
thefile = open(str(J)+"_LFNs.txt",'w')
ds = []
for js in j.subjobs:
     if js.status=='completed':
         thefile.write("%s\n" % js.backend.getOutputDataLFNs()[0])
thefile.close()
```
And then point dirac-dms-lfn-accessURL to it:
```
dirac-dms-lfn-accessURL < 1017_LFNs.txt
```

[...]

PR #801, opened 2016-09-21
Get a list of accessURLs from a Dirac job

mesmith75:
Implementation of #738  This allows a user to get a list of accessURLs from a Dirac job with `jobs(1).backend.getOutputDataAccessURLs()`. There is a function in Dirac utilities to get the URLs from a list of LFNs. The function in DiracBase uses `getOutputDataLFNs` and passes that list.

I may need to add in some checks and exceptions in case of missing files or SEs.


python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -214,6 +214,34 @@ def _checkProxy( delay=60, renew = True, shouldRaise = True, force = False ):
...
+        else:
+          SEs.append(thisSE)
+    myURLs = []
+    # Loop over the possible SEs and get the URLs of the files stored there.
+    # Remove the successfully found ones from the list and move on to the next SE.
+    for SE in SEs:
+      thisSEFiles = execute('getAccessURL(%s, "%s")' % (lfns , SE))['Value']['Successful']
+      for lfn in thisSEFiles.keys():
+        myURLs.append(thisSEFiles[lfn])
+        lfns.remove(lfn)
```
drmarkwslater: This will change the list that has been sent into the function which is almost certainly **not** what's expected. Either do a `copy` of the list or just add to a `done_lfn_list` and compare the two.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -214,6 +214,34 @@ def _checkProxy( delay=60, renew = True, shouldRaise = True, force = False ):
...
+      # If we gotten to the end of the list then break
+      if len(lfns)==0:
```
drmarkwslater: Spaces around the operator '=='

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -214,6 +214,34 @@ def _checkProxy( delay=60, renew = True, shouldRaise = True, force = False ):
             last_modified_time = time.time()
 
 
+def getAccessURLs(lfns):
```
rob-c: Can we introduce some code quality tests on Jenkins to highlight these things? I've seen this come up before somewhere.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
rob-c: `if not lfns` ?

rob-c:
Can we add some level of checks around the execute rather than just calling dictionary elements which may not exist.

rob-c:
Should this be 6.3 or 6.2.1 as it's not really a bugfix?

drmarkwslater:
But doesn't execute now raise an exception if it fails? And therefore if 
it's got to this point, then the dictionaries should be valid? Maybe I'm 
missing something though...

python/GangaDirac/Lib/Utilities/DiracUtilities.py
drmarkwslater: Fair point 😄

drmarkwslater:
Definitely 6.3 :)

rob-c:
@drmarkwslater Not if this is going into 6.2.1. In which case I'll look to merge my branch and then this can be changed to make use of the new features.
I think I agree with 6.3.0 😄

mesmith75:
@rob-c Yes I think it would be a good idea to add some checks for whether files or SEs are properly available and wait for 6.3.

mesmith75:
@drmarkwslater I would be fine with making the function to get the URLs from a list of LFNs global.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -474,6 +474,11 @@ def getOutputDataLFNs(self):
             lfns.extend([f.lfn for f in outputfiles_iterator(j, DiracFile) if f.lfn != ''])
         return lfns
 
+    def getOutputDataAccessURLs(self):
+        """Retrieve the list of accessURLs assigned to outputdata for a job"""
+        urls = getAccessURLs(self.getOutputDataLFNs())
+        return urls
```
rob-c: why not just return this and save a not strictly needed assignment?

drmarkwslater:
Actually,I was suggesting making the global function take a list of 
*DiracFiles* rather than LFNs, then this backend function would just 
give all the DiracFiles for all the jobs/subjobs to this function. It 
could then be exposed in the GPI.

Actually, thinking more about it, maybe a 'static' function in 
DiracFile? Not sure - anyone else care to comment?

rob-c:
I think I like the approach which is taken here tbh.

The method could easily take an iter-able object containing either `DiracFile` or `LFN` if there is an `isinstance` check to parse the input. (It would be nice to use this with a `DataSet` object for instance.
(This or we add a short helper function in the `GangaDataset` which _should_ inherit correctly)

It would be nice for users save the constant extraction of `file.lfn` values if we expose this at the `GPI`. I think this has been effectively requested at the `GPI` from users.

I think it makes sense to have this method used by the `DiracFile.getAccessURL` method which should call the static method in the same way to save the interfaces becoming inconsistent.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -90,7 +90,7 @@ class DiracBase(IBackend):
                                doc='Settings for DIRAC job (e.g. CPUTime, BannedSites, etc.)')
     })
     _exportmethods = ['getOutputData', 'getOutputSandbox', 'removeOutputData',
-                      'getOutputDataLFNs', 'peek', 'reset', 'debug']
+                      'getOutputDataLFNs','getOutputDataAccessURLs', 'peek', 'reset', 'debug']
```
milliams: Needs a space after the comma before `'getOutputDataAccessURLs'`

python/GangaDirac/Lib/Utilities/DiracUtilities.py
milliams: I'm not sure that `DiracUtilities.py` is the right place for this. This file is used for functions which support the DIRAC subsystem, not for miscellaneous DIRAC functions.

Given that this function is only used in one place (in a one-liner no less) inside `DiracBase`, it probably makes more sense to put it in inline inside `getOutputDataAccessURLs`.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -214,6 +214,34 @@ def _checkProxy( delay=60, renew = True, shouldRaise = True, force = False ):
...
+    for a provided list of lfns.
+    """
+    # Get all the replicas
+    reps = execute('getReplicas(%s)' % lfns)
+    # Get the SEs
+    SEs = []
+    for lf in reps['Value']['Successful']:
+        for thisSE in reps['Value']['Successful'][lf].keys():
+            if thisSE not in SEs:
+                continue
```
milliams: I'm guessing there was a mistake made here when editing as this set of loops does nothing.

milliams:
Also, the new function will need a test or two.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
rob-c: I would have said `DiracUtils` within the `Backend` would have been better, there's already functions like that there.
I would __really__ like to see the `DiracFile` making use of this to reduce the overhead of duplication of functionality.

rob-c:
Testing, again, should be covered by a test working with the `DiracFile` imo. But obviously with the mocks passed down to the execute calls.

python/GangaDirac/Lib/Utilities/DiracUtilities.py
mesmith75: Yes, that was a mistake

mesmith75:
Ok, I'll move it into the backend DiracUtils. I'll also export it to the GPI as that would be useful. I'll add an ```isinstance``` and grab the LFNs from DiracFiles if they are passed.

mesmith75:
@rob-c There was a reason I didn't put this into the DiracFile implementation. For a single file it is possible to specify the preferred SE. Here is is more of an issue as files will be spread all over the place. I would quite like to keep the option to specify the SE for a single file.

rob-c:
@mesmith75 I would argue it still makes sense to allow for selecting a `defaultSE` for a list of LFN if you know they are replicated in multiple sites (and in the case of 1 experiment I know this would be useful!).
Just keeping the sensible defaults or not advertising this feature in the GPI interface would be fine when processing a list I would argue, but I'd be keen to make sure that these 2 different methods don't get out of sync. If we're duplicating functionality I'd say at least try to keep it in 1 file so that keeping track is easier.

mesmith75:
Ok then that seems reasonable. We can make sure it tries the specified SE first before moving on to the others.

mesmith75:
The last commit puts the main function in Backends/DiracUtils.py . This is called from DiracFile, and in DiracBase . I also exported it to the GPI. It takes a list of all strings, or extracts the DiracFiles from a mixed list (i.e. from job.outputfiles).

For a specified SE I just move it to the front of the list that gets looped over so it goes first, otherwise it just takes them in the order of the SEs that is returned by getReplicas.

Is there a better way to check for a DiracFile? Currently it is the MetaClassObject type.

rob-c:
@mesmith75 first can you rebase on 6.3-features to pull in the changes wrt execute which have been committed to this branch. This removes the need to find the Value dict element.

An isinstance(object, Diracfile) check should identify the class as you would expect. If you're expiring the method correctly the proxy layer should remove the proxy around the file objects correctly. If not a stipProxy call is needed and i think you've found a bug

mesmith75:
retest this please

mesmith75:
Any objections to merging this, pending the tests?

python/GangaDirac/Lib/Backends/DiracUtils.py
```diff
@@ -4,6 +4,9 @@
 #from GangaDirac.BOOT       import dirac_ganga_server
 from GangaDirac.Lib.Utilities.DiracUtilities import execute, GangaDiracError
 from Ganga.Utility.logging import getLogger
+from GangaDirac.Lib.Files.DiracFile import DiracFile
+from Ganga.GPIDev.Lib.File.File import File
```
alexanderrichards: This import is never used in the code that you have added to the file in this PR. Is this a missing dependency from before or is it redundant?

python/GangaDirac/Lib/Backends/DiracUtils.py
```diff
@@ -120,3 +123,53 @@ def for_each(func, *iterables, **kwargs):
...
+    for lf in reps['Successful']:
+        for thisSE in reps['Successful'][lf].keys():
+            if thisSE not in SEs:
+                SEs.append(thisSE)
+    myURLs = []
+    # If an SE is specified, move it to be the first element in the list to be processed.
+    if defaultSE != '':
+        if defaultSE in SEs:
+            SEs.remove(defaultSE)
+            SEs.insert(0,defaultSE)
```
alexanderrichards: PEP8: requires a space between the args

python/GangaDirac/Lib/Backends/DiracUtils.py
mesmith75: It is redundant.

PR #803, opened 2016-09-23
Some proxy improvements, docs and handling of objects through 'addProxy'

rob-c:
This PR fixes a shortcoming of the `addProxy` method which means that using this around a class would simply return the class and not add a proxy to it. You would unlikely notice until your class starts getting the raw kwds from the GPI which is a problem when the class hasn't been correctly modified to pass this down to the underlying `GangaObject` or to handle them itself.

I've also added a switch to `getRuntimeGPIObject` as well as a few doc changes.

I'm keen to merge this asap into `6.3-features` as this is required for another PR I'm working on.


python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -229,24 +244,36 @@ def isType(_obj, type_or_seq):
...
+    """
     if isType(obj, GangaObject):
         if not isProxy(obj):
             if hasattr(obj, proxyObject):
                 return getattr(obj, proxyObject)
             else:
                 return GPIProxyObjectFactory(obj)
+    elif isclass(obj):
+        if not issubclass(obj, GPIProxyObject):
+            addProxyClass(obj)
```
milliams: Does this need the `addProxyClass()` call. Is that not done automatically by `getProxyClass()` anyway.

python/Ganga/GPIDev/Base/Proxy.py
```diff
@@ -229,24 +244,36 @@ def isType(_obj, type_or_seq):
...
+        obj (object): This may be an instance or a class
+    """
     return getattr(obj, implRef, obj)
 
 
 def addProxy(obj):
-    """Adds a proxy to a GangaObject"""
+    """Adds a proxy to a GangaObject instance or class
+    Args:
+        obj (GangaObject, class): This may be a Ganga object or class which you're wanting to add a proxy to
```
milliams: When I made the suggestion in the SharedFile PR to use `addProxy()` on a class name to get the proxied version I didn't realise it didn't currently support it. I'd rather keep this function simple and only use it on instances, not classes. If you need to get the proxy class version of a `GangaObject` then there's `getProxyClass()`. There's no need for a single function to be able to cover all bases, different situations need different solutions.

rob-c: I think it is used and advertised to be used in this way in the core. I see no downside to being able to add a wrapper to any and all objects especially when the logic is effectively this simple.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: good point.

python/Ganga/GPIDev/Base/Proxy.py
rob-c: OK, it's not actively used this was but it was intended to be when I cleaned up the methods around this a while ago.
I'll drop this down to the other minor features and code cleanup in this PR.

I don't like that we're inconsistent wit the naming of `addProxy` and `getProxyClass` and had coped we could make this simpler so that people really don't have to care about the technical aspects of whats going on here.

milliams: The reason that they're apparently inconsistent is that they do slightly different things. `getProxyClass()` will only return a proxy class if one has already been created for a type (i.e. it's a `GangaObject`) whereas `addProxy()` will most likely be creating a new proxy from scratch (and proxy reuse is primarily for caching).

I think it's unrealistic to not expect Ganga developers to have to understand the proxy system. I've worked to reduce the need to use it as much as possible but it can't be escaped entirely. Given that there's only a small handful of Ganga developers and the proxy system is such a core part of it I don't feel it's a big ask.

I agree that the naming isn't completely clear but I think it's certainly good enough for now.

rob-c: I think we disagree about this so I've reverted back to not have this feature.
I'm assuming the rest of the PR is good to go?
I also removed some unused imports which were highlighted as immediately obvious from grepping the codebase.

PR #804, opened 2016-09-23
Adding return True for a credential which didn't throw an exception

rob-c:
This adds a `return True` for the case that a credential is valid and an exception isn't raised.
If this is not here AFS users get an error that the AFS token is invalid and that the monitoring will stop for it.
@milliams Does this look sensible compared to the design of the new credentials?

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -6,6 +6,7 @@ from contextlib import contextmanager
 
 from Ganga.Core.GangaThread import GangaThread
 from Ganga.Core.GangaRepository import RegistryKeyError, RegistryLockError
+from Ganga.GPIDev.Credentials2.exceptions import CredentialRenewalError
 
 from Ganga.Utility.threads import SynchronisedObject
 
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -1159,12 +1159,14 @@ class JobRegistry_Monitor(GangaThread):
             log.debug("Checking %s." % getName(credObj))
             try:
                 s = credObj.renew()
-            except Exception as msg:
+            except CredentialRenewalError:
+                return False
+            except Exception as err:
+                logger.error("Unexpected exception!")
+                logger.error("Error: %s" % err)
                 return False
             else:
-                if s is None:
-                    return True
-                return s
+                return True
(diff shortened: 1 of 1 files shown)


python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -1162,6 +1162,8 @@ def credChecker():
             except Exception as msg:
                 return False
             else:
+                if s is None:
+                    return True
```
milliams: This whole section could be
```python
try:
    credObj.renew()
except CredentialRenewalError:
    return False
else:
    return True
```
since `renew()` will never return a value any more.

rob-c: Thanks for confirming the design. I'm assuming the old renew methods are being removed in this case?

milliams: Well the entire classes that they were previously inside are being removed so yes :smile:

rob-c:
@milliams How about now?

python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -1158,11 +1158,15 @@ def makeCredChecker(self, credObj):
         def credChecker():
             log.debug("Checking %s." % getName(credObj))
             try:
-                s = credObj.renew()
-            except Exception as msg:
+                credObj.renew()
+            except CredentialRenewalError:
```
milliams: This will need an appropriate `import`.

rob-c: I was expecting to find this in `Ganga/Core/exceptions` with the rest of the definitions, is this class ultimately going to be placed here?

milliams: It's at `Ganga.GPIDev.Credentials2.exceptions.CredentialRenewalError`. I'm unsure whether it should be move to be alongside the core exceptions, I feel like it's violating the modularisation a bit?

rob-c: Added import, good to go?

rob-c: The codebase has been moving to be better laid out imo. We already put all of the interfaces in `Ganga.GPIDev.Adapters` and exceptions in `Ganga.Core.exception`. This has been the way we've been heading for a while by moving other files around so that it's easy to know where things are.

PR #806, opened 2016-09-26
Bugfix/fix post boot strap safety

rob-c:
This PR is fairly short but it addresses 2 issues.

I wrote a `postBootstrapHook` for `GangaDirac` to ensure that the user has correctly set a `DiracProxy` `group`.

When I tried to throw an exception I realised that even though `GangaDirac` knows there is potentially something wrong it can't stop Ganga attempting to load old jobs and run the monitoring on them with potentially disastrous consequences.

In order to fix this I moved the post-bootstrap hooks to run before the jobs are loaded and the monitoring has started. This allows Ganga to correctly exit before the jobs are loaded from disk in a way which can cause problems.

This is atop the credentials branch where the problem can occur but this prevents job corruption on starting up Ganga with a bad config.

I've made this against the credentials branch as this is where it's most applicable and where I first spotted the problem.


milliams:
Instead of moving the post-bootstrap to happen before bootstrapping has finished, perhaps put the `getConfig('defaults_DiracProxy')['group']` check in the module scope (just after all the `configDirac.addOption` but within the `if not _after_bootstrap` block). That way the exception will be raised early enough to stop corrupting jobs but avoids changing the meaning of *post*bootstrap.

rob-c:
@milliams This still leaves:
1. The problem that a post-bootstrap check will find or have a problem which could effect user data and will not be able to avoid data corruption.
2. Does this correctly take into account the fact we can't check for the value being populated until after `.gangarc` is loaded? I think this _could_ also work but I'm less of a fan of this as it feels slightly hacky and I'm not sure when the exception would then be thrown.

I also have the strong opinion that loading the jobs and monitoring them is happening __after__ bootstrap anyway. This is something which I know was never documented but I somehow feel calling this part of the bootstrap when both of these things are optional for loading a ganga session is incorrect.

milliams:
1. The simple answer is to check for these things earlier than post-boostrap. Any consistency checks can be performed *during* bootstrap, while the plugin is being loaded and safely exit there.
2. The config files (including `~/.gangarc`) are loaded during `configure()`. Any entries which are not understood (due to them having not been loaded in a pugin yet) are cached until the appropriate `addOption()` call is made. So by line 85 in `GangaDirac/__init__.py` you can check config values and correctly see any config file overrides.

rob-c:
1. Where is the best place in this case? If your suggestion above holds great we should check and throw an exception on plugin load which woks fine for me :+1:  I still don't like the fact we can get corruption of jobs and data so I'm keen to fix this.
2. Your going to have to provide an example for me to get this. I'm not overly familiar with the subtleties with the config system and I will get this check wrong here, I have no idea what version of the configurable I get when during the bootstrap process and always tend to resort to checking values after as this is a lot safer when I don't know whats going on.

milliams:
In the credentials branch I am about to put this in `DiracProxy.py` at the bottom of the file. At this point the `defaults_DiracProxy` config section has been created but nothing has started running yet. From the point of view of `bootstrap.py`, this is happening on the `loadPlugins()` call inside `bootstrap()`.

As for how config files work, there a clever bit of caching going on. The config file is read quite early during the startup of Ganga. Importantly it happens before any plugins are loaded (in fact it *must* be since we need to know the value of `RUNTIME_PATH`). While the file is being read, if a particular setting is known about (i.e. we've had a relevant `addOption()` call already) then the value is simply set accordingly. However, if there is no matching config section defined then the parser (inside the function `setSessionValue()`) puts the config option inside a dictionary called `allConfigFileValues`. Later, when the config option is defined (via `addOption()`) the code checks `allConfigFileValues` to see if it has a value there waiting for it.

So, if you're inside a plugin then as soon as `addOption` is called, you will have access to whatever value was set via the config file. If you're inside `python/Ganga` at module level then most likely the config file has not yet been read and so you must wait until after `configure()` in `bootstrap.py` has run.

rob-c:
(I'm going to have to re-read the config description once I've got some time I've never particularly liked the way it handles things and have tried to avoid it where possible).

OK, if we can have some sort of sane test that will prevent Ganga from loading during the import of GangaDirac until a user has taken action over this I'm very happy. I don't mind as to the implementation as long as it works and prevents the user being greeted with XML errors during startup.

I've no idea why the history of this PR has now become full of conflicts from files I've not touched I'm imagining this is from some conflict elsewhere. Once this is working on the credentials branch I'll close the PR 👍

milliams:
The conflict happened because I had to rebase the `feature/credentials` branch onto the 6.3 staging branch. I've just pushed the change as "Add single global check for the DIRAC group setting" and it worked in my testing.

PR #807, opened 2016-09-26
Bugfix/forward compatibility with 6.3

rob-c:
This PR adds dummy Voms and Dirac proxy classes to allow us to load jobs from 6.3.x within Ganga 6.2.x. This is needed and prevents data corruption when moving between versions (as I'm sure some users will inevitably try to do)

I'm happy to move the classes around/rename them/completely rewrite them and I can look at if they're exposed to the GPI and try to prevent this. However the classes need to be present within Ganga in some form to prevent data corruption.
The intention was to hide these in a file which is being removed in the transition to 6.3.x so this will not effect us when this is merged.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -936,6 +936,8 @@ class GangaRepositoryLocal(GangaRepository):
 
         if len(errs) > 0:
             logger.error("#%s Error(s) Loading File: %s" % (len(errs), fobj.name))
+            for err in errs:
+                logger.error("err: %s" % err)
             raise InaccessibleObjectError(self, this_id, errs[0])
 
         logger.debug("Checking children: %s" % str(this_id))
--- a/python/Ganga/GPIDev/Credentials/ICredential.py
+++ b/python/Ganga/GPIDev/Credentials/ICredential.py
@@ -81,16 +81,6 @@ logTimeStamp = 0.0
 logRepeatDuration = 120.0
 
 
-class VomsProxyInfo(GangaObject):
-    """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
-    _schema = Schema(Version(1, 0), {})
-    _category = 'CredentialRequirement'
-
-class DiracProxyInfo(GangaObject):
-    """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
-    _schema = Schema(Version(1, 0), {})
-    _category = 'CredentialRequirement'
-
 def registerCommandSet(commandClass=None):
 
     try:
@@ -534,3 +524,4 @@ class ICredential(GangaObject):
         """
         logger.warning("Dummy method used - no information returned")
         return ""
+
--- a/python/Ganga/GPIDev/Credentials/ICredential.py
+++ b/python/Ganga/GPIDev/Credentials/ICredential.py
@@ -81,6 +81,16 @@ logTimeStamp = 0.0
 logRepeatDuration = 120.0
 
 
+class VomsProxyInfo(GangaObject):
+
+    """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
+    _category = 'CredentialRequirement'
+
+class DiracProxyInfo(GangaObject):
+    
+    """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
+    _category = 'CredentialRequirement'
+
 def registerCommandSet(commandClass=None):
 
     try:
--- a/python/Ganga/GPIDev/Credentials/ICredential.py
+++ b/python/Ganga/GPIDev/Credentials/ICredential.py
@@ -82,13 +82,13 @@ logRepeatDuration = 120.0
 
 
 class VomsProxyInfo(GangaObject):
-
     """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
+    _schema = Schema(Version(1, 0), {})
     _category = 'CredentialRequirement'
 
 class DiracProxyInfo(GangaObject):
-    
     """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
+    _schema = Schema(Version(1, 0), {})
     _category = 'CredentialRequirement'
 
 def registerCommandSet(commandClass=None):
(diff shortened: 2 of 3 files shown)


rob-c:
Can someone review this

egede:
@rob-c So I guess this require that we release a 6.2.X release with this in and then remove all 6.2.Y releases with Y<X?

rob-c:
@egede in a word yes, if we want to support switching to/from 6.3/6.2. I think we can try and add a fix which negates this within the XML but that would be a very large rewrite of the XML parser I fear.

rob-c:
@milliams what are your thoughts on this?

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -936,6 +936,8 @@ def _actually_load_xml(self, fobj, fn, this_id, load_backup):
 
         if len(errs) > 0:
             logger.error("#%s Error(s) Loading File: %s" % (len(errs), fobj.name))
+            for err in errs:
+                logger.error("err: %s" % err)
```
milliams: This doesn't seem related to the proposed change.

python/Ganga/GPIDev/Credentials/ICredential.py
```diff
@@ -524,3 +524,4 @@ def timeleftInHMS(self, force_check=False):
         """
         logger.warning("Dummy method used - no information returned")
         return ""
+
```
milliams: Extra blank line?

python/Ganga/GPIDev/Credentials/__init__.py
```diff
@@ -92,3 +94,16 @@ def getCredential(name="", create=True):
         credential = False
 
     return credential
+
+
+class VomsProxy(GangaObject):
+    """ dummy class to allow for loading of 6.3.x jobs inside 6.2.x. PLEASE NEVER USE THIS! """
+    _schema = Schema(Version(1, 0), {})
```
milliams: Can be slightly simpler as `_schema = Schema(Version(1, 0))`

python/Ganga/GPIDev/Credentials/__init__.py
```diff
@@ -92,3 +94,16 @@ def getCredential(name="", create=True):
...
+    _category = 'CredentialRequirement'
+    _name = 'VomsProxy'
```
milliams: There's no need for a `_name` when it's identical to the given class name.

milliams:
Firstly, I'm not I see how this can cause data corruption? A job failing to load due to missing classes should create an `IncompleteObject`. This will never be flushed so there will be no damage to the data on disk.

Secondly, we've never done this sort of retroactive compatibility fix in the past. To first order, this is no different to when we added `LocalFile` or `GaudiExec`. Loading a job with those as attributes would fail in exactly the same way. We make no promise to our users that it is safe to run new jobs on a old version of Ganga. There are very few pieces of software that allow this sort of forward-compatibility.

There is some small special case here though in that a 6.2 job which is running, when loaded in 6.3 could automatically get these attributes added without the user's knowledge. Re-loading in 6.2 would then fail as you described. It would still be completely safe though to load it again in 6.3 as nothing should be lost in that cycle. The only reason we might want to make an exception in this one special case is that the credentials branch has a larger than usual chance of breaking users' workflow or introducing bugs and so to be kind to them we should perhaps make it possible to temporarily revert to 6.2.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: Without this change in loading the job from XML it's impossible to see why the job has been marked as corrupt. it's indexes removed and it's primary XML file destroyed as the backup was attempted to be loaded.

rob-c:
@milliams
1. no. The job will first attempt to load it's history from disk as is expected. If it's history is from some earlier state that did __not__ contain one of these classes there can be job corruption. This can happen with jobs which are being created and submitted on different versions or jobs having to be monitored again which would lead to problems such as the file already existing on some remote storeage when we try to put it there.
Plus it's just wrong for a user to get hundreds of errors about corrupt indexes and XML when moving back from 6.3 to 6.2 which will happen.
2. There will be _no_ support in the mid-term future on the project and this gives some level of breathing room other than simply pulling the 6.3 release until it's working. It makes for a dangerous precedent and if the release if pulled then the jobs it's touched are lost from a user's perspective.
We've had schema tests for this sort of stuff in the past and we have checks that we can still open LogicalFile and such around the 6.1 transition so I think there is a __lot__ of precedent for this sort of move.
3. No loading a job in 6.2 causes the removal of the last saved state in the XML _as we all agreed_ during discussion and development of the tests. There is no way ganga can tell what caused an XML to be corrupt at the moment (yes this is a shortcoming but we always want to revert the users job to the last 'good' state was the agreed fallback).

python/Ganga/GPIDev/Credentials/ICredential.py
rob-c: W298

python/Ganga/GPIDev/Credentials/__init__.py
rob-c: yes but there are numerous classes where we have this still.

python/Ganga/GPIDev/Credentials/ICredential.py
milliams: ?

python/Ganga/GPIDev/Credentials/__init__.py
milliams: Sure, but that doesn't mean it's needed to add to a new one.

python/Ganga/GPIDev/Credentials/ICredential.py
rob-c: sorry, for a start new lines are picked up as `W292` warning suggesting they should be there and are added by some editors by default as good practice. I'll remove as not needed for this PR.

python/Ganga/GPIDev/Credentials/__init__.py
rob-c: It's in a class that's being removed but I'll remove it

rob-c:
I believe this should be mergable?

python/Ganga/GPIDev/Credentials/ICredential.py
milliams: Both this file and `python/Ganga/GPIDev/Credentials/__init__.py ` seem to still have the double new line at the end of the file.

rob-c: This has to be a problem between editors as it doesn't appear in my vim, I'll revert the change as it's not relevant to what's going on

milliams: It seems that vim has a different definition of `\n` than most other editors I've used. vim treats it as an end-of-line character rather than a line-separator character. This means that while most other editors will display an empty blank line at the end of the file, vim will not.

The important thing for Git is that there is a `\n` at the end of the file. Vim will inform you of its lack by putting a `[noeol]` in the status bar at the bottom. GitHub will display a missing terminating `\n` with a red `⃠↵`.

rob-c:
OK I plan to merge this in an hour or so as all should be well

PR #809, opened 2016-09-27
Fixing unique test for outputfiles

rob-c:
Changing the Job.py outputfiles unique test to just use an append, extending a list from a set from the original list breaks the order ad the set function sorts the input.
This doesn't effect too many tests but does interfere with the File tests when running on different machines.

I'll merge this once the tests pass as this fixes a minor bug which has already been reviewed in a better way on top of the 6.3 branch.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -2088,9 +2088,9 @@ class Job(GangaObject):
 
             # reduce duplicate values here
             uniqueValues = GangaList()
-	     for val in values:
-		if val not in uniqueValues:
-			uniqueValues.append(val)
+            for val in values:
+                if val not in uniqueValues:
+                    uniqueValues.append(val)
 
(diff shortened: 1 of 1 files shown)


python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -2088,7 +2088,9 @@ def __setattr__(self, attr, value):
 
             # reduce duplicate values here
             uniqueValues = GangaList()
-            uniqueValues.extend(set(value))
+	    for val in values:
```
milliams: The indentation here is off by 1.

milliams:
Should this not target `develop` as it is a bugfix?

rob-c:
@milliams Not really. It's not a bugfix over anything over the 6.3 branch. The code in `develop` _works_ and is untested (besides there are plenty of bugs around the file objects in develop I don't want to accidentally expose with a change like this so I'm puttng it into 6.3

PR #814, opened 2016-09-30
Don't quote arguments passed to non-shell Popen

milliams:
At present the `checkDiskQuota()` function is printing errors like:
```
df: `\'/disk/homedisk/home/mpw/ganga/ganga/python/gangadir testing/TestSJSubmit/repository/testframework/LocalXML\'': No such file or directory
df: no file systems processed
```
`Popen` will do the quoting automatically when passed a list of arguments in non-shell mode and so the presence of the `quote` function is causing spurious double quoting.


rob-c:
Go for merge from me I've got this in a few private PRs also just haven't had time to back-port it.

Issue #783, opened 2016-09-08
Inputfiles not copied to WN for GaudiExec


mesmith75:
I think this is just a GaudiExec issue. The files listed under ```j.inputfiles = ['somefile.py']``` are not being copied over to the worker node. They are put into the tar in the input dir ```_input_sandbox_13_master.tgz ``` but I can't find that on the WN.

rob-c:
OK, which backend is this for?
(I should really get the gangamon server running again and restore the report functionality)
I'm curious to see the full trace from a job as this should be extracted at the start of the job running and I think I left the extract process fairly verbose to know what is extracted where.

If this is on DIRAC it's possible that the file is not extracted due to it already being inside one of the many levels of tar files being used in the process of getting files to the WN.
If DIRAC could you point me to an example job so I can have a look at the sandboxes.

mesmith75:
@rob-c This is on Dirac. The job setup is very generic:
```python
 j = Job(name="test", application = GaudiExec(), backend = Dirac())
j.application.directory = '/afs/cern.ch/work/m/masmith/GangaTest/test/DaVinciDev_v40r1'
j.application.options =  ['cocktail_options.py']
j.inputfiles = ['tuplebranchinfos.py']
j.submit()
```
where the tuplebranchinfos.py is just generic python file. I'll look a bit further if I can find some time.

rob-c:
OK, I'll try and submit a test job tomorrow and have a look at the job logging as a workaround may be as simple as calling multiple decompression runs over the files on the WN.

rob-c:
(Or at least re-building the index after extracting each file and doing a better job of remembering what files we've already de-compressed, should be possible with some simple set logic)

egede:
We should be careful in allowing a user to submit a compressed tarball as an inputfile (that then should stay compressed).

rob-c:
@egede This gets difficult as if the user requests it as a `DiracFile` there is no way of stopping Dirac's own auto-decompression on the file.
We could try to have a list of files we want to decompress and in what order but this seems fragile. @mesmith75 Can you provide a jobID in this case to help with the debugging?

mesmith75:
@rob-c a jobID is 139522694 .

rmatev:
Any news here? I am also affected by this bug.

mesmith75:
[...]

PR #815, opened 2016-09-30
Fix to add the input files to the prepared sandbox

mesmith75:
Fix for #783 . I added a line to the collectPreparedFiles function to adding the job.inputfiles into the input_files to be uploaded. In my test the file gets uploaded to the WN successfully and used.

I am not sure if this is how @rob-c intended the logic to work though so he had best check it.

### DIFF
--- a/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
+++ b/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
@@ -112,6 +112,9 @@ def collectPreparedFiles(app):
         for name in dirs:
             input_folders.append(os.path.join(root, name))
 
+    for file_ in app.getJobObject().inputfiles :
+        input_files.append(os.path.join(file_.localDir, os.path.basename(file_.namePattern)))
+
     return input_files, input_folders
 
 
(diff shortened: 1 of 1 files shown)


rmatev:
After staring a bit at the code, I doubt this is the intention.
The `inputfiles` seem to be packed already (see `Job.createPackedInputSandbox`) and put into e.g. `gangadir/workspace/rmatev/LocalXML/4/input/_input_sandbox_4_master.tgz`. You have to ship this file to the grid somehow and make sure it's untared.

rob-c:
@mesmith75 I think your fix is mostly correct. However I would insist on a check that this is a `LocalFile` for the moment. And to raise an error an on everything not `LocalFile` or `DiracFile`, we don't for the moment support `MassStorageFile` objects with the `GaudiExec` this is a known shortcoming.
We can support `MassStorageFile` etc in 6.3.x but this relies on several changes having been made to fix problems with the `IGangaFile` system.

mesmith75:
I am not sure we can easily use DiracFiles with the current setup.

rob-c:
@mesmith75 They should just be trivially added to the correct part of the jdl. This is fully supported in DIRAC

rob-c:
@mesmith75 would you like me to make a small PR with these changes and this fix?

mesmith75:
@rob-c Thanks for that. As this has been overtaken I shall close and delete the branch.

Issue #764, opened 2016-09-02
GaudiExec enhancements


rob-c:
GaudiExec loses a few features which it would be good to re-implement:
- [x] Post-processing metadata per subjob
- [ ] Pickling payload before submitting several hundred subjobs
- [x] XML lumi data could be added back in

Edit:
- [ ] Inspect the submitted scripts for `DaVinci().TupleFile`

This list may change but these are the biggest todo items for best support I can think of.

rmatev:
Do you think it makes sense to use `GaudiExec` with released applications, too? I guess you would need to add two properties -- `project` and `version`, and make them mutually exclusive with `directory`. You'd then use `lb-run ${project} ${version} ...` instead of `./run`.

The advantage is that  everything becomes uniform, and you can drop all applications like DaVinci, Moore, etc.
I can submit another issue if you like this proposal and think it deserves it.

rob-c:
@rmatev I've written a function which I was hoping to test/advertise which does a `lb-dev` locally to setup the local environment of a released application.
https://github.com/ganga-devs/ganga/blob/develop/python/GangaLHCb/Lib/Applications/GaudiExecUtils.py#L89
If this does what you want great, if it has bugs, less great but can you file an issue so we can fix it 😄 

The older application within Ganga are there for backwards compatibility in loading old jobs from Ganga 6.1.x and prior so we can't remove these although we have de-facto frozen them and have no intention to support any PR against these moving forwards.

rob-c:
I'm extremely reluctant to add any knowledge of the application name or type to `GaudiExec`. This is information which strictly does not belong there and is due to the new ability to build a reliable target for all LHCb apps.

rmatev:
I'm afraid it does not work with 6.2.0:
```python
Ganga In [26]: prepare_cmake_app('DaVinci', 'v41r2p1')
ERROR    Error: global name 'expandfilename' is not defined
```
(looking at the code, there are some other imports missing, too).

Anyway, it seems to me a bit too much to make an lb-dev project when I don't add code, but I think see your worry. Then, how about doing `directory=$LHCBRELEASES/DAVINCI/DAVINCI_v41r2p1`?

I don't know exactly what magic is needed to run on Dirac, since `$LHCBRELEASES` will resolve to `/afs` by default on LXPLUS, but then again I don't know how lb-dev projects prepared on LXPLUS are able to work on the grid.

rob-c:
[...]

PR #816, opened 2016-10-01
Fix old LHCb RT handlers to propagate parsed xml summary

### LINKED ISSUE #140: Metadata not populated for subjobs
I have a checker that fails with 6.1.14 because the metadata per job is missing. The strange thing is that the combined metadata for the master job is there. 
```
[09:33:40]
Ganga In [1]: j=jobs[-1]

[09:33:45]
Ganga In [2]: j.subjobs[0].metadata
Ganga Out [2]: <Ganga.GPIDev.Lib.Job.MetadataDict.MetadataDict at 0x7f19501e3b10>

[09:33:59]
Ganga In [3]: j.subjobs[0].metadata.data
Ganga Out [3]: {}

[09:34:07]
Ganga In [4]: j.metadata
Ganga Out [4]:
{
     xmldatafiles = {'full': set(['LFN:/lhcb/data/2015/RAW/LUMI/LHCb/LEAD15/169504/169504_0000000088.raw', ...,  'LFN:/lhcb/data/2015/RAW/LUMI/LHCb/LEAD15/169503/169503_0000000405.raw'])}
     xmldatanumbers = {'full': 49}
     lumi = Not Available.
     events = {'output': 0, 'input': 90383787L}
     xmlskippedfiles = set([])
    }
```
On a side note, subjobs are not returned as proxy objects, so it is annoying to work interactively with them.

rmatev:
This fixes #140 
If it can make it to 6.2.1 I would be very happy.

### DIFF
--- a/python/GangaLHCb/Lib/RTHandlers/LHCbGaudiDiracRunTimeHandler.py
+++ b/python/GangaLHCb/Lib/RTHandlers/LHCbGaudiDiracRunTimeHandler.py
@@ -103,7 +103,7 @@ class LHCbGaudiDiracRunTimeHandler(GaudiDiracRunTimeHandler):
             outbox_files = [fileTransform(this_file, None) for this_file in outbox if not FileUtils.doesFileExist(this_file, job.outputfiles)]
             job.non_copyable_outputfiles.extend([outbox_file for outbox_file in outbox_files if not isType(outbox_file, DiracFile)])
 
-            outputsandbox = [f.namePattern for f in job.non_copyable_outputfiles]
+            outputsandbox.extend([f.namePattern for f in job.non_copyable_outputfiles])
 
             outputsandbox.extend([f.namePattern for f in job.outputfiles if not isType(f, DiracFile)])
             outputsandbox = unique(outputsandbox)  # + outbox[:])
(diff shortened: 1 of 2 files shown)


milliams:
@ganag-ci test this please

rob-c:
@rmatev Thanks for the contribution. I will make the comment here that we technically __no longer__ support the old application objects. That being said given this is a small fix I'm very happy to merge this into 6.2.1
@egede Do we want to stop accepting fixes for the old apps at some point?

rmatev:
@rob-c I am aware since you mentioned this already, but I think `GaudiExec` is a bit shy in features to drop standard applications just yet.

PS I've started looking into implementing the summary.xml / metadata support in GaudiExec, but it is not exactly trivial due to (I think) a deeper issue with GaudiExec.

rob-c:
@rmatev What are the features which `GaudiExec` requires in this case? I will try and find some time to get these implemented or templated for 6.3.0.

I see no deep issue with the `GaudiExec` class. If you're needing the XML summary to be parsed from the jobs I'm certain there is a solution for this that doesn't involve too much work. To what issue(s) are you referring?

rmatev:
@rob-c, let's move the discussion to #764 if you don't mind

rob-c:
@rmatev If you can update your branch to match ganga develop I'll merge this in for 6.2.1

rmatev:
@rob-c thanks, it's done

rob-c:
@ganag-ci test this please

egede:
@rob-c Small user supplied bug fixes to the old application handlers will be accepted as long as they do not have side effects or will require significant developer time to implement.

PR #820, opened 2016-10-04
Fixing defaults in GaudiExec

rob-c:
This fixes some display issues with the GaudiExec app. The defaults were bad, but not too harmful.

### DIFF
--- a/python/GangaLHCb/Lib/Applications/GaudiExec.py
+++ b/python/GangaLHCb/Lib/Applications/GaudiExec.py
@@ -93,8 +93,8 @@ class GaudiExec(IPrepareApp):
     """
     _schema = Schema(Version(1, 0), {
         # Options created for constructing/submitting this app
-        'directory':    SimpleItem(defvalue=None, typelist=[None, str], comparable=1, doc='A path to the project that you\'re wanting to run.'),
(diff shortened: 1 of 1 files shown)


python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -93,8 +93,8 @@ class GaudiExec(IPrepareApp):
     """
     _schema = Schema(Version(1, 0), {
         # Options created for constructing/submitting this app
-        'directory':    SimpleItem(defvalue=None, typelist=[None, str], comparable=1, doc='A path to the project that you\'re wanting to run.'),
-        'options':       GangaFileItem(defvalue=None, sequence=1, doc='List of files which contain the options I want to pass to gaudirun.py'),
+        'directory':    SimpleItem(defvalue='', typelist=[None, str], comparable=1, doc='A path to the project that you\'re wanting to run.'),
+        'options':       GangaFileItem(defvalue=[], typelist=[str], sequence=1, doc='List of files which contain the options I want to pass to gaudirun.py'),
```
milliams: A `GangaFileItem` is a `ComponentItem`. That means it does not use `typelist` and can only have a `defvalue` of `None` or a string which names a plugin i.e. `'LocalFile'`.

PR #821, opened 2016-10-04
Clean up exceptions

rob-c:
This just stops the user being confronted with a stack trace when a proxy(credential) doesn't exist.


milliams:
This makes sense :+1:

PR #822, opened 2016-10-04
Fixing adding of inputfiles to the GaudiExec when submitting to Dirac

rob-c:
This should supersede #815. I'm checking for unsupported filetypes (we don't correctly support moving files in a nice way and I don't want to add botch code to work around what has been fixed in 6.3)
This should add inputfiles which vary between subjobs to the inputsandbox as there is no way in principle of collecting these before the subjob is being prepared so might as well send these along with the sandbox which slows submission but this is only for very complex job configurations.

I'll try and test this now.

### DIFF
--- a/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
+++ b/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
@@ -112,6 +112,13 @@ def collectPreparedFiles(app):
         for name in dirs:
             input_folders.append(os.path.join(root, name))
 
+    for file_ in app.getJobObject().inputfiles:
+        if isinstance(file_, LocalFile):
+            shutil.copy(os.path.join(file_.localDir, os.path.basename(file_.namePattern)), shared_dir)
+            input_files.append(file._namePattern)
+        elif not isinstance(file_, DiracFile):
+            raise ApplicationConfigurationError(None, "File type: %s Not _yet_ supported in GaudiExec" % type(file_))
+
     return input_files, input_folders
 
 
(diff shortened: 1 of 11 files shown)


rob-c:
I've also fixed `prepare_cmake_app` and added `prepareGaudiExec`. The latter allows for:
```j=Job(application=prepareGaudiExec('DaVinci','v40r2'))```

**Comment by rob-c** [comment]
@mesmith75 What are your thoughts on this?

**Comment by mesmith75** [comment]
@rob-c If this works then great. The PR I made is not very good. I guess we really want this to start supporting DiracFiles in the near future but for now just LocalFiles is pretty necessary so it should go in for 6.2.1.

**Comment by rob-c** [comment]
@mesmith75 We _already_ support `DiracFiles` elsewhere in the logic in assembling the JDL. The lack of support for the `LocalFile` here is more of an oversight. The problem is `MassStorageFile` we have no (in 6.2.1) nice way of copying the file to the correct location before they're submitted.

**Comment by rob-c** [comment]
I'll check the code works as expected in a few min, I'm debugging a few different things at once now.

**Comment by mesmith75** [comment]
The code for this looks ok. I would like this for 6.2.1 as it is quite an important fix.

**Comment by rob-c** [comment]
I'm trying to check now and facing the usual grid problems... I'll decompose the inputsandbox manually.

File: python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py [hunk 1]
```diff
@@ -112,6 +112,13 @@ def collectPreparedFiles(app):
         for name in dirs:
             input_folders.append(os.path.join(root, name))
 
+    for file_ in app.getJobObject().inputfiles:
+        if isinstance(file_, LocalFile):
+            shutil.copy(os.path.join(file_.localDir, os.path.basename(file_.namePattern)), shared_dir)
+            input_files.append(file._namePattern)
```
mesmith75: Should this be ```file_.namePattern``` ?

rob-c: yes I think it should, god catch!

rob-c: it does in my testing version of the PR, uploading now

rob-c:
I'm looking to merge today, everything appears working from a test job. I'll add something about this to the release notes when this gets merged.

PR #823, opened 2016-10-04
Change default LHCbDirac version to v8r3p7

mesmith75:
Updating the LHCbDirac version


rob-c:
Thanks, I'm updating the lhcb-prepare script once this goes in. We should be weary we still rely on SetupProject to deploy ganga to LHCb and this could blow up in our faces as LHCbDirac technically no longer supports SetupProject for the new cmake tools. Unfortunately I suspect we won't know the implication for this until we request the version be deployed for testing.

mesmith75:
Good point - would it be best to go back to the last v8r2 release? If not then I'll merge this.

rob-c:
@mesmith75 Imo we can't really dodge the bullet on it and we should look to move to LHCbDirac 8.3 sooner imo so if there are problems they're found sooner rather than later.
This only really effects LHCb and we can push out 6.2.2 if we reach a point that we discover that some serious work needs doing.

mesmith75:
Ok, I'll merge this.

PR #824, opened 2016-10-05
Extra info in the error failing to upload GaudiExec

mesmith75:
An extra line to the error suggesting using the ```DiracBasLFN``` config if the tarball cannot be uploaded.

### DIFF
--- a/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
+++ b/python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
@@ -359,7 +359,7 @@ class GaudiExecDiracRTHandler(IRuntimeHandler):
 
         if not isinstance(app.uploadedInput, DiracFile):
             generateDiracInput(app)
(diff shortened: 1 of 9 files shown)


python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -359,7 +359,7 @@ def master_prepare(self, app, appmasterconfig):
 
         if not isinstance(app.uploadedInput, DiracFile):
             generateDiracInput(app)
-            assert isinstance(app.uploadedInput, DiracFile), "Failed to upload needed file, aborting submit"
+            assert isinstance(app.uploadedInput, DiracFile), "Failed to upload needed file, aborting submit\nIf your Ganga installation is not at CERN your username may be trying to create a non-existent LFN. Try setting the 'DIRAC' configuration 'DiracLFNBase' to your grid user path.\n"
```
mesmith75: This is suggesting using the DiracLFNBase configurable explicitly, rather than getting Ganga to piece it together from the username and VO. I thought it would be better to avoid possible conflicts elsewhere with the username config.

rob-c: @egede I think this should be in the doc for the option so I think if we put something like:
```("Tried to upload to: %s" % DiracFile.LFNBase)``` this should be enough to show if this is an issue I would have thought

mesmith75:

PR #797, opened 2016-09-19
6.3 features

rob-c:
This PR is a collection of all of the things which have gone into the `6.3-features` branch over `develop`.

The main changes are (broadly):
(please feel free to edit if more features get merged into the branch)
1. Adding `copyTo` (internal) and changing the `get` methods for all `IGangaFile` Objects
2. Renamed the template script files to fix problems with code checker
3. Better handling of Dirac errors within Ganga added `GangaDiracError`
4. Added support for `SharedFile`
5. Cleaned up code for handling of file locks
6. Initial implementation of new credentials system for LCG/Dirac
7. tbd...

I'm not looking to merge this PR I'm just opening this so that I can update this branch with the changes from develop and make sure that the code merge for 6.3.0 is hopefully pain free.

PR #825, opened 2016-10-06
Bugfix/6.3 Credentials fixes for LHCb

rob-c:
This PR addresses several problems I've been having with the credentials code from an LHCb perspective.

 1. Added `DiracProxy` tests
 2. Added the ability for a `DiracProxy` to _not_ add the default group extension to a proxy filename as it may or may not be on disk already. This fixes the problems for lhcb users as they will likely already have or need to use a 'normal' proxy with other tools. (Other groups correctly get a new proxy made it's only the default proxy which 90%+ of users will want or care about which we change here and this is configurable using a flag in the DiracProxy itself).
 3. Moved the remaining interface classes to `Ganga.GPIDev.Adapters`

I've discovered that you can create/destroy proxies using something like: `credential_manager.create(DiracProxy())` and `credentialManager.destroy(DiracProxy())`.

I'll add a further PR shortly with some improved documentation around this and I think it may be worth adding a wrapper for something like `gridProxy.renew()` for LHCb users to smooth the transition to 6.3.

Can someone review this so I can look to merge. Thanks


rob-c:
After some more experimenting with this i think i need to add another proxy layer to the credential manager or at least expand what we have as ours possible to get access to unproxied objects from the cli. I'll look into adding this into this pr.

rob-c:
OK, I've added some auto-detection of DIRAC proxies on disk during Ganga startup (This avoids the proxy not looking like it's there when running with `--no-mon` or other situations.)
I've improved the string rep of the proxy objects and added a pretty repr hook to make the output look nicer for users on IPython.

rob-c:
OK, I've added a dummy `gridProxy` class for `GangaLHCb`. All this does is smooth the transition to 6.3 and has some simple logic to handle creating and destroying a grid proxy using the old terminology which is used by LHCb users. The LHCb community is special in this regard as we know the default config up front.

rob-c:
Can someone review this so I can look to merge as this should fix most of the paper-cuts I've found for LHCb users. Thanks.

mesmith75:
@rob-c I'll test some jobs on this branch

mesmith75:
So I am not sure this is working. Starting Ganga gives
```python
ERROR    unknown options [defaults_AfsToken]minValidity
ERROR    unknown options [LCG]EDG_ENABLE,GLITE_ENABLE
ERROR    unknown configuration section: [defaults_AfsCommand]
WARNING  exit status [1] of command kinit -R
WARNING  <first 4096 bytes of output>
kinit: KDC can't fulfill requested option while renewing credentials

WARNING  <end of first 4096 bytes of output>
Kerberos password: 
[16:19:01]
Ganga In [1]:
```
Here it becomes rather unresponsive.

rob-c:
@mesmith75 i have big problems on lxplus with this. What exact system config are you using?
I suspect we may need better detection of needing Afs credentials of you're not on lxplus of you are then this is locking up as you cast perform an any disk i/o which is 'expected'. We still need to handle bad passwords and authentication retries which aren't yet in the framework. I'll look into that tomorrow.

rob-c:
Apologies that should be I'm having bio problems with this afs credential checking on lxplus. I will have a look tomorrow on another test system to see what could be causing this.

mesmith75:
Ok, it works for me this morning.

rob-c:
@mesmith75 I'm trying to work out where to add a more appropriate error for this as it's a bit difficult to work out for a user what to do in this situation.

rob-c:
@mesmith75 when you saw this error did you get any info output that you created a token successfully or had there been a raised exception due to maybe mis-typing a password. I would have assumed that you should be seeing some info telling you your new tokens validity.
I suspect this can be improved by re-trying to get the user authentication when it doesn't work a few times, this is something needing to be added to all of the new credential objects I think so I'm trying to add it in a generic way.

rob-c:
OK, I'm busy adding the credential requirements to make the book-keeping commands work properly rather than 'some of the time'.

rob-c:
OK, I've fixed some outstanding bugs in the 6.3 branch with respect to splitting jobs in LHCb, bookkeeping queries, and some minor bugs in Core which I've exposed at different stages.
In addition to this I've managed to get the integration tests down to ~7min which is a drop of about 10min in the runtime.
This is due to some very minor optimizations in some methods which are called O(10Billion) times during submitting a large complex job  and the fixing of minor bugs in the session lock handling and simplifying the global lock acquisition.

I'm keen to try and merge this branch as it's become a very large collection of bugfixes against the `6.3-features` branch now. I'll work on some minor doc improvements ahead of merging.

mesmith75:
@rob-c I'll run some test jobs on it now then.

mesmith75:
Starting the session for the first time still gives issues with Kerberos:
```python
WARNING  exit status [1] of command kinit -R
Kerberos password: 
[12:40:57]
Ganga In [1]: 
```
and the prompt is unresponsive. I have to suspend Ganga and then kill it. I notice that there is a defunct kinit process when I do so.
The next time I start Ganga it works OK though.

rob-c:
@mesmith75 What system are you running this on? I'm going to need more information as I can't reproduce this on my lxplus.

mesmith75:
This is on lxplus with a fresh tcsh shell. The login just takes the latest LbLogin script. My kerberos tokens are forwarded from my local computer.

rob-c:
@mesmith75 is this a common use case? I've never heard of this being done tbh. I'll investigate the use of kinit commands in the `AfsToken` but I suspect it wasn't designed with this in mind. I may need you to dump some info about your token but I'll see if anything obvious crops up.

rob-c:
@mesmith75 Are the tokens stored in /tmp when they're forwarded, what mechanism are you using for this?

rob-c:
@mesmith75 is the environmental variable `KRB5CCNAME` correctly setup for your use case?
If not I'll modify the `AfsToken` to try and make use of the output from `klist` which will require a little work.

mesmith75:
@rob-c I don't know how many people use tcsh. I am pretty sure more than just me forward their kerberos tokens.
On lxplus the environment variable seems to be set and points to the /tmp folder.
To forward them I just set property in my ssh config:
```
HOST lxplus*
    GSSAPIDelegateCredentials yes
```
and set ```allow_weak_crypto = true``` in the krb5.conf .

rob-c:
@mesmith75 What value does your `KRB5CCNAME` have and can you share your output from `klist`? (You may want to move over to email as we're bordering on security related question here).

rob-c:
@mesmith75 Can you try updating to the latest code here?
I've refactored the AFS tokens to take the soonest ending kerberos token to determine when the AfsToken is actually valid. I suspect this is caused by the difference in authority behind the scenes.

I believe this should address all issues that we know about atm for 6.3 and it's possible to move silently between 6.2.2<->6.3 and I've tested. 6.2.0 isn't 6.3.x compatible and this was a known bug as the credentials branch hadn't stabilised before 6.2 was released.

rob-c:
retest this please

mesmith75:
@rob-c I still get the same problem, although when I suspend Ganga the kinit process is not defunct this time. I think the problem is that it is trying to create a new token which isn't necessary.

rob-c:
@mesmith75 I'm not sure I understand what you're saying there.
The changes I put in should detect the forwarded token correctly if you've said that `KRB5CCNAME` and the file it points to are correct and existing.
The old code was setup to detect only tokens from `afs@cern.ch` and now it should be detecting any tokens picked up from the `tokens` command regardless.
I think I'm going to have to ask if you can point me to what is required in forwarding the kerberos token in this case as I'm likely going to have to debug which function call(s) are failing to be parsed correctly.

mesmith75:
@rob-c  I think I figured out the problem. The tokens I create on my computer don't have a renewable lifetime. Does Ganga try and automatically renew tickets?

rob-c:
@mesmith75 Ganga should only renew a token which have expired from the summary of 'tokens'.
I'm not an expert in kerberos tokens so don't know how different the tokens look when they're regenerated as renewable or not (I just assumed that all of them were renewable).
I'll have a quick recheck of the code and see if there's anything obvious I can do to fix this.
I'm assuming the afs tokens are only constructed during bootstrap but there may be some code in the monitoring too.

rob-c:
@mesmith75 If you add a renewable lifetime to your `krb5.conf` file does this fix things?
If so this is a relatively easy fix for people wanting to forward their kerberos tokens.

If not I'll have to evaluate maybe switching the code in `AfsToken` to use `klist` over `tokens` but this is a pain to do and if we can instruct users to update their configuration if they have made one themselves this seems sensible to me.

mesmith75:
@rob-c Yes its fine if I set the configuration to have a renewable lifetime. I guess this is more a problem with my configuration than Ganga. Sorry for the noise.

mesmith75:
I think a more pressing issue is that job submission fails if the grid proxy hasn't already been created; it just raises an exception instead of asking for a password to create a proxy. The failure is in the getLFNReplicas part:
```python
ERROR    Failed to Get Replica Info: [0:3] of 3
ERROR    Exception raised executing 'getLFNReplicas' in Thread 'None':
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 114, in __worker_thread           
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py", line 100, in getLFNReplicas               
    output = wrapped_execute('getReplicas(%s)' % str(allLFNs[this_min:this_max]), dict)                                                                                             
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py", line 35, in wrapped_execute               
    raise SplittingError("Error from Dirac: %s" % err)                                                                                                                              
SplittingError: Error from Dirac: GangaDiracError:  ('',)
```
Otherwise it is fine if the proxy already exists.

rob-c:
@mesmith75 I think it's probably worth adding something to this effect into the release notes for 6.3 when we get to release to minimize the problems. Users will still see the release notes on launch and will then have to do something when it stalls.

Can you send me a MWE as I can submit a simple hello world to the grid and the proxy is automatically required and generated.
I'll look through the trace and try and spot any missing decorators now.

mesmith75:
@rob-c I've put the ganga script and an options file for the job in
/afs/cern.ch/user/m/masmith/public/forRob/

rob-c:
OK this looks like I missed this because my test jobs all perform BKQuery before submitting the job so I forgot to add a credential requirement for the splitter which calls DIRAC...

At this point I've 3 potential places where the credential is being saved per job. This is fine until the user wants to change the credential from the default then they have to add it everywhere... I'm going to look into changing the code to correctly find a master job and get the credential always from the backend rather than having 1 credential for each DiracFile and 1 credential for each splitter...

rob-c:
@mesmith75 Can you try this fix. I've added a credential_store to the splitter object.

I don't think there is a nice solution for changing the credential_store for many objects which are children of 1 job so I think we'll have to accept if a user wants to change the credential then they need to change it manually everywhere which is slightly annoying but not difficult.

mesmith75:
@rob-c Thanks, that looks to have sorted it.

rob-c:
@mesmith75 I've done some further testing of this PR and I'm ready to merge it into the `6.3-features` staging area. Has anything else cropped up in testing I may have missed?

mesmith75:
@rob-c I've been submitting some jobs with it and it seems fine. Give me a couple of hours to test out some more involved GaudiExec things - otherwise it looks good.

rob-c:
@mesmith75 I'm keen to merge this PR so do you have any objections to rebasing to #797 and testing from there?

mesmith75:
Yes that is fine. It seems pretty fine so I assume any issues will be small bugfixes.

PR #826, opened 2016-10-06
Fixing LHCb init code

rob-c:
The latest LHCb code is not safe for me and bails out in a bad way.

I've updated the code and I think this needs integrating into 6.2.1.

### DIFF
--- a/python/GangaLHCb/__init__.py
+++ b/python/GangaLHCb/__init__.py
@@ -68,19 +68,22 @@ def _store_dirac_environment():
     fdir = os.path.join(os.path.expanduser("~/.cache/Ganga/GangaLHCb"), platform)
     fname = os.path.join(fdir, diracversion)
     if not os.path.exists(fname) or not os.path.getsize(fname):
-      cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
-      env = execute(cmd)
-      try:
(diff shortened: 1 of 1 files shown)


rob-c:
I still need to update the lhcb-prepare script. I'll finish this now

rob-c:
@mesmith75 Any objections to merging this? which branch is best?

mesmith75:
It looks ok. It could probably go into the release/6.2.1 branch. Then I can rerun the release tests.

python/GangaLHCb/__init__.py
```diff
@@ -68,19 +68,22 @@ def _store_dirac_environment():
...
-        logger.info("Storing new LHCbDirac environment (%s:%s)" % (str(diracversion), str(platform)))
-      except (OSError, IOError, TypeError):
-        logger.error("Unable to store LHCbDirac environment")
-        raise
+        cmd =  'lb-run LHCBDIRAC {version} python -c "import os; print(dict(os.environ))"'.format(version=diracversion)
+        env = execute(cmd)
+        if isinstance(env, str):
+            try:
+               env_temp = eval(env)
+               env = env_temp
```
milliams: Why was this split on to two separate lines?

rob-c: So when debugging or returning what the env was its quickly clear what the value of env should be.

PR #827, opened 2016-10-06
Fix config on start up

rob-c:
This fixes a bug in the config system that running the new user code without a .gangarc would lead to a crash. This fixes it. This is present in the 6.2.1 testing branch.


rob-c:
@mesmith75 Any objections to merging?

mesmith75:
Sure, go ahead

Issue #830, opened 2016-10-08
GaudiExec doesn't work in 6.2.1


mesmith75:
Submitting GaudiExec jobs doesn't work in 6.2.1. The prepare stage throws an exception:
```python
ERROR    Error: buildGangaTarget() takes exactly 1 argument (0 given)
ERROR    !!Unknown/Unexpected ERROR!!
ERROR    If you're able to reproduce this please report this to the Ganga developers!
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r1/install/ganga/python/Ganga/GPI/__init__.py in <module>()
----> 1 j.prepare()

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r1/install/ganga/python/Ganga/GPIDev/Base/Proxy.py in proxy_wrapped(*args, **kwargs)
    666         s_args = [stripProxy(a) for a in args]
    667         s_kwargs = dict((name, stripProxy(a)) for name, a in kwargs.items())
--> 668         r = f(*s_args, **s_kwargs)
    669         return addProxy(r)
    670 

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r1/install/ganga/python/Ganga/GPIDev/Lib/Job/Job.py in prepare(self, force)
   1119             raise JobError(msg)
   1120         if (self.application.is_prepared is None):
-> 1121             add_to_inputsandbox = addProxy(self.application).prepare()
   1122             if isType(add_to_inputsandbox, (list, tuple, GangaList)):
   1123                 self.inputsandbox.extend(add_to_inputsandbox)

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r1/install/ganga/python/Ganga/GPIDev/Base/Proxy.py in proxy_wrapped(*args, **kwargs)
    666         s_args = [stripProxy(a) for a in args]
    667         s_kwargs = dict((name, stripProxy(a)) for name, a in kwargs.items())
--> 668         r = f(*s_args, **s_kwargs)
    669         return addProxy(r)
    670 

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r1/install/ganga/python/GangaLHCb/Lib/Applications/GaudiExec.py in prepare(self, force)
    196         logger.info('Created shared directory: %s' % (self.is_prepared.name))
    197 
--> 198         this_build_target = self.buildGangaTarget()
[...]

PR #831, opened 2016-10-08
A fix for the thread lock in the GaudiExec prepare

mesmith75:
I think this fixes the problem #830  - could someone else check. I also fixed something in the DiracRTHUtils.

### DIFF
--- a/python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
+++ b/python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
@@ -87,7 +87,7 @@ def dirac_outputfile_jdl(output_files, empty_SE_check):
             if empty_SE_check:
                 ## If true check, if not false check
                 raise BackendError("Dirac", "Can't submit a DIRAC job with DiracFile outputfile without setting a defaultSE.")
-            config = getConfig('Dirac')
+            config = getConfig('DIRAC')
             myLine = myLine.replace('###OUTPUT_SE###', str([]))
 
         total_JDL += myLine + "\n"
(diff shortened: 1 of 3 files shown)


egede:
Looks good.

mesmith75:
If there are no objections I shall merge this and release 6.2.2

python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
```diff
@@ -87,7 +87,7 @@ def dirac_outputfile_jdl(output_files, empty_SE_check):
             if empty_SE_check:
                 ## If true check, if not false check
                 raise BackendError("Dirac", "Can't submit a DIRAC job with DiracFile outputfile without setting a defaultSE.")
-            config = getConfig('Dirac')
+            config = getConfig('DIRAC')
```
drmarkwslater: Actually - config isn't used so this line can be deleted :smile:

mesmith75:
restest this please

python/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py
mesmith75: Somehow this has broken the integration test...

drmarkwslater: If you look at the console output of the test, at the end it has:
```
Project Ganga v602r1 is not available for platform x86_64-slc6-gcc49-opt
Build step 'Virtualenv Builder' marked build as failure
```

I'm guessing this is because 6.2.1 was removed for LHCb?

PR #832, opened 2016-10-10
Cleanup xml

rob-c:
This is a slightly updated form of the PR Cleaning up the lock files for 6.3 I've changed the code to not throw exceptions when the global lock timeouts and now it keeps trying. This solves the problems that appeared on my testing box. Im making a new PR to run multiple tests on the Jenkins server at CERN to see if this is now working or if i need to do any more debugging.

Of this is now working I'll look to merge layer this week into the 6.3 features.


rob-c:
Retest this please

rob-c:
Retest this please

rob-c:

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -286,6 +268,25 @@ def decorated(self, *args, **kwargs):
             return f(self, *args, **kwargs)
     return decorated
 
+def global_disk_lock(f):
+    @functools.wraps(f)
+    def decorated_global(self, *args, **kwds):
+        with global_disk_lock.global_lock:
+            while True:
+                try:
+                    self.global_lock_acquire()
```
alexanderrichards: This looks like it's missing a timeout parameter unless this is non-blocking?

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -286,6 +268,25 @@ def decorated(self, *args, **kwargs):
...
+                    time.sleep(1.)
+
+            self.safe_LockCheck()
+            try:
+                return f(self, *args, **kwds)
+            finally:
+                self.global_lock_release()
+    return decorated_global
+
+global_disk_lock.global_lock = threading.Lock()
```
alexanderrichards: personally I don't like having the lock attached to the decorator like this. However if there is a good reason for it then fine.

python/Ganga/test/GPI/Parallel/TestQueuedSJSubmit.py
```diff
@@ -7,8 +7,8 @@
 from Ganga.testlib.decorators import add_config
 from Ganga.testlib.monitoring import run_until_completed
 
-global_num_threads = 5
-global_num_jobs = global_num_threads*5
+global_num_threads = 10
+global_num_jobs = global_num_threads*10
```
alexanderrichards: pep8: spaces around operator

python/Ganga/test/GPI/Parallel/TestQueuedSubmit.py
```diff
@@ -8,7 +8,7 @@
 from Ganga.testlib.monitoring import run_until_completed
 
 global_num_threads = 10
-global_num_jobs = global_num_threads*5
+global_num_jobs = global_num_threads*10
```
alexanderrichards: pep8: spaces around operator

python/Ganga/Core/GangaRepository/SessionLock.py
alexanderrichards: more of a comment really :wink:

rob-c: Having a global lock here avoids the same ganga session torturing the disk to get the disk lock as the thread knows that it can't access it ahead of checking the status of the disk lock. It's a reduction of I/O

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: This is not intended to have a timeout. If you can't access the global file lock you can't do anything. This doesn't lead to deadlocks as the global lock (within 1 session) is released correctly once it's no longer needed. Adding a timeout here caused errors to be thrown when a thread can't access the global file lock vs the thread waiting for access which is what you'd expect.

python/Ganga/Core/GangaRepository/SessionLock.py
rob-c: I was happy to make this a global variable but several people have expressed opinions against global objects.

rob-c:
I think this may lead to a slight reduction in the time it takes the tests to run but the change is just a bit larger than the normal variation which can happen.

PR #836, opened 2016-10-19
BUGFIX: Fixing error with backend missing when raising BackendError

henryiii:
The signature for one of the 10 BackendError calls is missing the first argument ("Dirac"). (one line fix) This causes the following incorrect error message when a job fails:

```
raise BackendError('Problem retrieving outputsandbox: %s' % str(getSandboxResult))
TypeError: __init__() takes exactly 3 arguments (2 given)
```


alexanderrichards:
test this please

rob-c:
I'm planning to merge this unless there are any objections? @henryiii Thanks for the contribution

PR #839, opened 2016-10-25
Faster xml load

rob-c:
This PR contains:

 1. Some minor test fixes (obviously bad code if the test was run interactively!)
 2. Minor speed improvements (Pieces of code which turn up in benchmarks way too much)
 3. *Fix for slow reading in of XML files*

3. consists of adding a new class factory method to `GangaObject` which makes correct use of the `__new__` method as well as fully initializing the created class instance without populating the internal dictionary with the default values from the schema. This is useful in situations where this _expensive_ operation is a waste as the code intends to replace these objects in the next step.
This factory is not a good solution to initializing a normal class for normal use but for the specific corner cases such as a deep copy or constructing a class for populating from XML this offers a real demonstrable performance boost.

This has the advantage that it doesn't require fixing `__init__` to be correct all across the codebase. It works as expected. It doesn't require special types to get it's manipulations done and it doesn't add significant overhead into the `GangaObject` or `MetaClass` objects.


rob-c:
@mesmith75 I would like to have this considered for 6.3 (if not 6.4 feels OK).

This addresses a specific problem which LHCb users are still afflicted with.
When a user has a large inputdata set (10k+ files) writing this to disk and reading it in is a very expensive OP.

This PR addresses the XML problem by reducing the need to fully construct objects which don't need to be fully qualified and reduces the number of I/O ops for the writing of the XML to disk by using a String buffer which allows for the data to be written in a single bust after being constructed.

Testing with a dummy Job containing 10k DiracFile in the inputfile field:

Without this PR: 70sec +  37sec  = 107sec
With this PR:       54sec + 17sec  =  71sec

~30% performance improvement in the basic test and I claim some general speed improvements in casual use using this PR.

Code used for testing:

Construction:
```
%time j=Job(inputfiles=[DiracFile(str(_)+'.root') for _ in range(10000)])
```
restart ganga with `--no-mon`
```
%time print jobs[-1].backend
```

mesmith75:
@rob-c I agree this looks nice to have in for 6.3. Just let me give it a quick try and then we can merge it.

mesmith75:
Ok this is good. If there are no objections I'll merge this and start the release.

PR #846, opened 2016-10-26
Adding a more complex fixture to protect LCG tests

rob-c:
This PR adds a more complex fixture which allows us to skip a test if we know that the config will always cause the test to fail.

I'd like to merge this into `develop` to get the nightly tests working for all backends.
I know this can be viewed as an enhancement but I'm keen to get the nightlies running with core tests for LHCb so I'm viewing this as a bugfix as the LCG tests will incorrectly fail when inside an LHCb environment.

Does anyone have any objections to me merging this?


rob-c:
retest this please

PR #849, opened 2016-10-27
Adding test for initial startup

rob-c:
This adds a small test which actually ensures that exceptions caused by launching Ganga in a clean env are caught. Most devs (myself included) don't often startup in a clean env regularly so it's possible to break things and for them to go un-noticed as we don't remove our .gangarc and similar files very often.


rob-c:
OK, I've improved the test to not break things. I've also moved some common code out so that the code being tested is the code being used in the ganga executable during startup.

rob-c:
I've also added some checks to see if the `.gangarc`, `.ganga.log` and `gangadir\...` are all created correctly as appropriate.

mesmith75:
retest this please

bin/ganga
```diff
@@ -55,17 +56,9 @@ def log(level, err):
...
-    Ganga.Runtime._prog = Ganga.Runtime.GangaProgram()
-    Ganga.Runtime._prog.parseOptions()
-    Ganga.Runtime._prog.configure()
-    Ganga.Runtime._prog.initEnvironment()
-    Ganga.Runtime._prog.bootstrap(Ganga.Runtime._prog.interactive)
-    Ganga.Runtime._prog.new_user_wizard()
     # Import GPI and run Ganga
     from Ganga.GPI import *
+    setupGanga()
     Ganga.Runtime._prog.run()
```
mesmith75: This didn't work for me without bringing back the ```import Ganga.Runtime```

rob-c: How so?

mesmith75:
```python
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/bin/ganga", line 62, in <module>
    Ganga.Runtime._prog.run()
NameError: name 'Ganga' is not defined
```

rob-c: Hmm, looks like the `standardSetup` doesn't seem to take correctly and the test is over-compensating for this by inserting ganga into the python path. I'll see if I can make some changes to the standard startup script to fix this.
(I think we should __really__ be using the `__file__` property rather than the argv to determine where this file is...)

rob-c: @mesmith75 How about now?

mesmith75: Still the same problem

rob-c: Hmm, I'll try importing Ganga.Runtime then, strange how I can't seem to be able to reproduce this and the tests pass...

rob-c: Does it work now? I decided to be explicit and import the `_prog` after it's been initialized

mesmith75: Yes that works.

mesmith75:
Spoke too soon - it has messed up something for creating jobs:
```python
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/bin/ganga", line 63, in <module>
    _prog.run()
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/Ganga/Runtime/bootstrap.py", line 1058, in run
    execfile(script, local_ns)
  File "job.py", line 20, in <module>
    j = Job()
NameError: name 'Job' is not defined

```

rob-c:
@mesmith75 OK now I need to know how you reproduced this? I am able to launch an IPython prompt are you trying to do something different than use the interactive mode?

egede:
@mesmith75 @rob-c This works for me. It starts Ganga and I can create a job. Tested both in a plain environment and in an LHCb environment.

rob-c:
@egede I discovered this effects running scripts so I'm looking into it.

rob-c:
OK, fixed that and solved the mystery of how user scripts were run from within the GPI.
I spent nearly 3 years trying to work out where the GPI was imported for them and it turns out it was imported in the `bin/ganga` script's global scope and was then pulled from this elsewhere.

mesmith75:
I'm afraid I still get errors, starting ganga with a job script: ```ganga job.py```
```python
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/bin/ganga", line 65, in <module>
    log(Ganga.Runtime._prog.options.force_loglevel, err)
NameError: name 'Ganga' is not defined
````
It is fine to start ganga without the script argument

**Comment by rob-c** [comment]
@mesmith75 I get no problems running with a test job `j=Job();j.submit()`
I've fixed the `Ganga` bug you've mentioned above, can you let me know if there are any other issues?

**Comment by mesmith75** [comment]
Thanks, that works fine.

**Comment by rob-c** [comment]
Any objections to merging this for 6.3.1?

File: bin/ganga [hunk 2]
```diff
@@ -55,19 +56,13 @@ def log(level, err):
...
     # Import GPI and run Ganga
-    from Ganga.GPI import *
-    Ganga.Runtime._prog.run()
+    setupGanga()
+    from Ganga.Runtime import _prog
+    from Ganga import GPI
+    _prog.run(GPI.__dict__)
 except GangaException as err:
-    log(Ganga.Runtime._prog.options.force_loglevel, err)
+    from Ganga.Runtime import _prog
```
alexanderrichards: Why do we import this in two locations, within the try and the except rather than at the top?

python/Ganga/Utility/logging/__init__.py
```diff
@@ -159,8 +159,9 @@ def _make_file_handler(logfile, logfile_size):
     global file_handler
     if logfile:
         try:
-            new_file_handler = logging.handlers.RotatingFileHandler(
-                logfile, maxBytes=logfile_size, backupCount=1)
+            with open(logfile, 'w'):
+                pass
+            new_file_handler = logging.handlers.RotatingFileHandler(logfile, maxBytes=logfile_size, backupCount=1)
```
alexanderrichards: Just a question, why do we need to open the file here and immediately close it?

mesmith75:
Do we still want this in for 6.3.1?

rob-c:
@mesmith75 Yes please if we can

mesmith75:
ok. If you could reply to @alexanderrichards review then we can stick it in.

bin/ganga
rob-c: Hmm, I think your correct, this should be a single import

rob-c:
updated

mesmith75:
If we are happy, let's merge this and release the next version.

Issue #828, opened 2016-10-07
Add a check for available SEs


mesmith75:
We need to add a check for which SEs are available for uploads in case one goes into downtime. This might be best done at startup.

egede:
@mesmith75 We should try to minimise checks at startup as they delay everything else in Ganga and might in reality never be needed. In this example it would most likely require a GridProxy at startup which might otherwise not be required. Can we delay it until the first time that it is needed?

rob-c:
@mesmith there is simply no api I'm aware of in DIRAC which will return on request a some lost which we would need. Adding this is complex and the complexity is the reason I've held off going near it. We could add redundancy checks to attempt to upload to sites of an upload fails in Ganga but there is little we can do to find which sites are currently in a writable state for a given credential unfortunately.

egede:
So should we rather just have a mechanism where it picks one and if it fails, it goes onto try the next. If 1 of the 6 are out, it means that on average one in every six uploads will get to see this delay.

mesmith75:
There are functions in Dirac to do it, so it can be done, but maybe it is simpler just to cycle through the list in case of a failure.

mesmith75:
Looking again, the function in the Dirac DataManager I wanted to use is not accessible outside. Its a bit annoying as it would be really useful to have. In which case I'll change it to try and cycle through the SEs.

rob-c:
@mesmith75 It's worth noting that there have been problems with site configurations not being 100% reliable when automatically generated as per the GridPP Dirac. We can roughly trust the SE in LHCbDirac but I'm weary to add any checks for vanilla Dirac based on some previous conversations with the Imperial experts.
At what stage are you planning on introducing these checks?

rob-c:
This is looking like a duplicate of #416 which one should I close?

egede:
@egede. I suggest to close #416

mesmith75:
They are slightly different things. I think this is the more pressing issue. I'll try and get it done for tomorrow.

mesmith75:
Closed by #852 for the GaudiExec upload. The function is now in the DiracAPI if we decide we want to check SEs elsewhere as well.

PR #852, opened 2016-10-28
Cycle through available SEs for GaudiExec upload

mesmith75:
For #828 .This checks that the upload of the GaudiExec tarballs has not failed. In case of failure it tries another SE.

This is just a stop-gap. I have requested an SE status check function from DIRAC which should be in a release in the near future, although I don't know when LHCbDirac will pick it up.

I am on board with Phillipe's suggestion to create at least one replica of these files to save from hammering a single SE if there are several hundred subjobs. I'll add this in.


rob-c:
@mesmith75 Can we wrap the `put` call in a try/except for the moment so that we can make the exception catch the `GangaDiracError` in 6.3?

egede:
@mesmith75 I discussed the general approach of this with Raja and he is very supportive. One thing is maybe to consider if we get a sensible error if the WN can't download the user code.

mesmith75:
@rob-c I'm not sure. The put doesn't raise an exception if the upload fails, it just puts the ouput into the stdout.

rob-c:
@mesmith75 but it potentially will in 6.3 and it will be very easy to catch and we know what to expect. Failing that we'll have to update the 6.3 branch for this anyway.
@egede The handling of moving files around is all done through the DIRAC API so DIRAC will get it's own sane errors if a file transfer fails. As far as I've understood this we're not responsible for anything other than asking DIRAC to provide a given LFN for a given job. Whatever DIRAC does to achieve this is out of our responsibility I think.

rob-c:
@mesmith75 It could also fail if something times out or returns a value we're not expecting the code is all very fragile in 6.2 with not every call to DIRAC correctly checking the output before proceeding to do things like expand a tuple/dict.

rob-c:
Also can we raise an exception here or is an exception raised if the upload fails? I suspect the GaudiExec class probably proceeds and gets into a bad state later on rather than throwing a `RTHandlerError` or equivalent if the upload has failed?

mesmith75:
@rob-c Ok I can put in a try, although I'm not sure what exceptions could be caught. If the upload fails then it isn't caught until the uploadedInput is checked.

rob-c:
@mesmith75 I would be expecting that either a raised exception or a `failureReason` should trigger moving onto the next SE as DIRAC could potentially fail in transferring the file to a site for whatever reason and we end up throwing an exception (by mistake in 6.2 and by design in 6.3).
I would just do a catch-all `except:` here as this is a stop-gap for `except GangaDiracError` which will come in 6.3.

mesmith75:
Ok, I added in a general exception. At the moment it won't catch a failed upload due to Dirac issues (unavailable SE etc) but I guess it will a timeout or something worse. I'll open an issue to remind me to update this properly for 6.3.

Any objections to merging?

python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -330,10 +331,19 @@ def uploadLocalFile(job, namePattern, localDir, should_del=True):
...
-    random_SE = random.choice(getConfig('DIRAC')['allDiracSE'])
+    trySEs = getConfig('DIRAC')['allDiracSE']
+    random.shuffle(trySEs)
     new_lfn = os.path.join(getInputFileDir(job), namePattern)
-    returnable = new_df.put(force=True, uploadSE=random_SE, lfn=new_lfn)[0]
-
+    for SE in trySEs:
+        try:
+            returnable = new_df.put(force=True, uploadSE=SE, lfn=new_lfn)[0]
+        except:
```
egede: @mesmith75 Not sure that I understand the logic here. As it is, a `put` call raising an (unknown) will terminate the loop and not try the next SE. Is that the intention? Also, when we catch the exception (and subsequently raise the generic GangaException, could we forward some information about the original exception? At the moment it will be completely hidden what went wrong.

python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -330,10 +331,19 @@ def uploadLocalFile(job, namePattern, localDir, should_del=True):
...
+            raise GangaException("Upload of LFN %s to SE %s failed" % (new_lfn, SE)) 
+        if new_df.failureReason == '':
+            break
+        else:
+            logger.warning("Unable to upload LFN %s to SE %s. Trying another SE." % (new_lfn, SE))
+            new_df.failureReason = ''
```
egede: @mesmith75 IF the upload fails (to any SE), i.e. it just runs to the end of the loop and give `failureReason` different from the empty string for all SEs, then this code will not raise an exception. Is that the intention?

mesmith75: Currently the ```put``` doesn't raise an exception if the file upload fails. So the idea is it gets to the end of the list and then the function that called the uploadLocalFile checks the stdout that is returned to it and raises an exception if that doesn't show success.
If for some reason ```put``` does raise an exception then I assume something bad has gone wrong so I raise an exception here. In 6.3 I think this is taken care of rather better.

mesmith75:
I don't think this is urgent for 6.2.3 so I'll bump this to 6.3 where an exception can be raised and dealt with properly.

mesmith75:
I would like this in for 6.3, could someone review please?

Issue #855, opened 2016-11-04
Taking different between two LHCbDataset objects not working


egede:
Patrick Koppenburg reports a bug on subtracting two LHCbDataset objects. 

```
[16:21:15]
Ganga In [36]: ld=ldo.difference(success)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/Ganga/GPI/__init__.py in <module>()
----> 1 ld=ldo.difference(success)

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/Ganga/GPIDev/Base/Proxy.py in proxy_wrapped(*args, **kwargs)
    666         s_args = [stripProxy(a) for a in args]
    667         s_kwargs = dict((name, stripProxy(a)) for name, a in kwargs.items())
--> 668         r = f(*s_args, **s_kwargs)
    669         return addProxy(r)
    670

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py in difference(self, other)
    379         files = set(self.getFullFileNames()).difference(other_files)
    380         data = LHCbDataset()
--> 381         data.extend([list(files)])
    382         data.depth = self.depth
    383         return data

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py in extend(self, files, unique)
    219             if _file is None:
    220                 _file = this_f
--> 221             myName = _file.namePattern
    222             from GangaDirac.Lib.Files.DiracFile import DiracFile
    223             if isType(_file, DiracFile):

AttributeError: 'list' object has no attribute 'namePattern'

[16:20:54]
Ganga In [34]: ldo
Ganga Out [34]:
 LHCbDataset (
   depth = 0,
   treat_as_inputfiles = False,
   persistency = None,
   files = [14632 Entries of type 'DiracFile'] ,
   XMLCatalogueSlice =    LocalFile (
     namePattern = ,
     compressed = False,
     localDir =
   )
 )

[16:21:10]
Ganga In [35]: success
Ganga Out [35]:
 LHCbDataset (
   depth = 0,
   treat_as_inputfiles = False,
   persistency = None,
[...]

PR #856, opened 2016-11-05
Some fixes to LHCbDataset

mesmith75:
This fixes some issues in LHCbDataset #855 . The ```difference```, ```symmetricDifference``` and ```union``` functions were calling ```extend()``` with a ```[list(set)]```. So the set was being changed to a list but then put inside another list which was causing problems.

I think this is just a hangover from when the code was changed to use ```extend``` instead of the now-defunct ```construct``` function.


mesmith75:
If there are no objections I'll merge this...

PR #861, opened 2016-11-12
Fix for prepare in old style apps

### LINKED ISSUE #860: prepare() of prepared jobs suggests using keyword force which unprepare cannot handle
Hi,

this is probably trivial.

I had to re-prepare a job (actually only trying if it works to save time instead of deleting, recreating,...) and saw the following output

```
[00:55:58]
Ganga In [15]: jobs(2699).prepare()
ERROR    Error: JobError: The application associated with job 2699 has already been prepared. To force the operation, call prepare(force=True)
```

okay, so i tried

```
[00:57:26]
Ganga In [16]: jobs(2699).prepare(force=True)
ERROR    Error: unprepare() got an unexpected keyword argument 'force'
TypeError                                 Traceback (most recent call last)
/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/Ganga/GPI/__init__.py in <module>()
----> 1 jobs(2699).prepare(force=True)

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/Ganga/GPIDev/Base/Proxy.py in proxy_wrapped(*args, **kwargs)
    666         s_args = [stripProxy(a) for a in args]
    667         s_kwargs = dict((name, stripProxy(a)) for name, a in kwargs.items())
--> 668         r = f(*s_args, **s_kwargs)
    669         return addProxy(r)
    670 

/afs/cern.ch/lhcb/software/releases/GANGA/GANGA_v602r2/install/ganga/python/Ganga/GPIDev/Lib/Job/Job.py in prepare(self, force)
   1123                 self.inputsandbox.extend(add_to_inputsandbox)
   1124         elif (self.application.is_prepared is not None) and (force is True):
-> 1125             self.application.unprepare(force=True)
   1126             addProxy(self.application).prepare(force=True)
   1127 

TypeError: unprepare() got an unexpected keyword argument 'force'


```
So, I just called `unprepare()` and then `prepare()` (w/o force) - I naively thought `force` would just call `unprepare`. This seems to work fine so far.

This is ganga v602r2. I quickly grepped through the source and saw that most `unprepare` methods actually have a keyword argument `force`. Only the one in python/GangaGaudi/Lib/Applications/GaudiBase.py:193 doesn't. (but i didn't have a closer look to understand if that's where my error stems from).

Cheers,
Paul

PS: this is a `DaVinci` job (not `GaudiExec`).

mesmith75:
Fixes #860. Added in the ```force``` argument for the ```unprepare``` function in the GaudiBase. Doesn't actually do anything but other places call ```unprepare``` with a force flag so I added it in.

### DIFF
--- a/python/GangaGaudi/Lib/Applications/GaudiBase.py
+++ b/python/GangaGaudi/Lib/Applications/GaudiBase.py
@@ -190,7 +190,7 @@ class GaudiBase(IPrepareApp):
                 env=self.getenv(False),
                 cwd=self.user_release_area)
 
-    def unprepare(self):
+    def unprepare(self, force=False):
         self._unregister()
 
     def _unregister(self):


mesmith75:
retest this please

python/GangaGaudi/Lib/Applications/GaudiBase.py
```diff
@@ -190,7 +190,7 @@ def cmt(self, command):
                 env=self.getenv(False),
                 cwd=self.user_release_area)
 
-    def unprepare(self):
+    def unprepare(self, force=False):
```
alexanderrichards: Thanks @mesmith75 It should be there anyway as it is defined in the interface:

[IPrepareApp#L61](https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/GPIDev/Adapters/IPrepareApp.py#L61)

This method looks (from quick glance) like it's doing similar stuff to the base class method so I wonder if we ought not be calling that directly. It also has the `self.decrementShareCounter(self.is_prepared.name)` which *used* to be needed to ensure that the share ref counter properly decremented when you unprepared an app. I have a gut feeling that this was streamlined and that maybe we don't need to make this call directly.

However this PR is still valid as it forces the GaudiBase to conform to the interface.

rob-c: I think the calls to move the counter up and down are still needed from memory but the interface should be respected yes so i think this is the minimal change needed.

rob-c:
Retest this please

mesmith75:
retest this please

mesmith75:
If there are no objections I will merge this...

PR #864, opened 2016-11-15
Fix for GaussSplitter with GaudiExec

### LINKED ISSUE #840: GaussSplitter not overriding LHCbApp for GaudiExec
I paste the email below. We had a problem before with the GaussSplitter overriding the LHCbApp options but it seems here to still not interacting with them properly.

I have 2 different jobs with 2 subjobs each. The former uses a private version of Gauss with GaudiExec, the latter a default Gauss version. Both use GaussSplitter

 GaussSplitter (
   numberOfJobs = 2,
   eventsPerJob = 50,
   firstEventNumber = 0
 )

as splitter. In the option file I setup the number of events to generate with

LHCbApp().EvtMax = 100

If I run the default version of Gauss then I get 50 events per subjobs correctly uncorrelated --> Don't know why the NextNext event (spillover event) seems to have the same seeds between subjobs, but the main event have different seeds so it is fine
If I run the same option file with GaudiExec then I get 100 events per subjobs, 50 of each are totally correlated among the 2 subjobs since they have the same seeds.
I know I can easily avoid the problem if I change

LHCbApp().EvtMax = 100 --> LHCbApp().EvtMax = 50

when I run with GaudiExec, but I'm wondering if this is a bug or a known behaviour.

mesmith75:
Fixes #840. Quick fix for the number of events in GaussSplitter for GaudiExec.

### DIFF
--- a/python/GangaLHCb/Lib/Splitters/GaussSplitter.py
+++ b/python/GangaLHCb/Lib/Splitters/GaussSplitter.py
@@ -75,9 +75,10 @@ class GaussSplitter(ISplitter):
             j = self._create_subjob(job, inputdata)
             # FIXME this starts from the 1st event and not zero, is it clear why?
             first = self.firstEventNumber + i * self.eventsPerJob + 1
-            opts = 'from Gaudi.Configuration import * \n'
-            opts += 'from Configurables import GenInit \n'
+#            opts = 'from Gaudi.Configuration import * \n'
+            opts = 'from Configurables import GenInit \n'
             if isinstance(job.application, Gauss):
+                opts += 'from Gaudi.Configuration import * \n'
                 opts += 'ApplicationMgr().EvtMax = %d\n' % self.eventsPerJob
             else:
                 opts += 'from Configurables import LHCbApp \n'
--- a/python/GangaLHCb/Lib/Splitters/GaussSplitter.py
+++ b/python/GangaLHCb/Lib/Splitters/GaussSplitter.py
@@ -75,7 +75,6 @@ class GaussSplitter(ISplitter):
             j = self._create_subjob(job, inputdata)
             # FIXME this starts from the 1st event and not zero, is it clear why?
             first = self.firstEventNumber + i * self.eventsPerJob + 1
-#            opts = 'from Gaudi.Configuration import * \n'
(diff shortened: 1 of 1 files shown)


python/GangaLHCb/Lib/Splitters/GaussSplitter.py
```diff
@@ -77,7 +77,11 @@ def split(self, job):
             first = self.firstEventNumber + i * self.eventsPerJob + 1
             opts = 'from Gaudi.Configuration import * \n'
             opts += 'from Configurables import GenInit \n'
-            opts += 'ApplicationMgr().EvtMax = %d\n' % self.eventsPerJob
+            if isinstance(job.application, GaudiExec):
```
mesmith75: @egede Unfortunately not. It seems as if for the old code the options are pickled, so subsequent changes to LHCbApp undo any options for it in the options files (eg. ```LHCbApp().DDDBtag```). For GaudiExec though, the ```LHCbApp().EvtMax``` takes precedence over ```ApplicationMgr().EvtMax``` so we need to use LHCbApp for the GaudiExec.

Ok, I can switch it around to test for Gauss application instead.

mesmith75:
If there are no objections I will merge this and start the release for 6.2.3

python/GangaLHCb/Lib/Splitters/GaussSplitter.py
```diff
@@ -77,7 +77,11 @@ def split(self, job):
             first = self.firstEventNumber + i * self.eventsPerJob + 1
             opts = 'from Gaudi.Configuration import * \n'
```
rob-c: Is this line still needed? The `GenInit` is used but I don't see anything other than the `ApplicationMgr` which may be imported here, or am I mis-reading the code?

mesmith75: Yes you need to ```import ApplicationMgr``` - I guess this can be moved to the relevant part.

mesmith75: It is needed to import ```ApplicationMgr``` - I guess I could move this to be specifically under the Gauss application.

PR #866, opened 2016-11-16
Moving some code around and allowing for 'ganga' to load plugins 

### LINKED ISSUE #863: importing little ganga doesn't import plugins
```python
>>> import ganga
INFO     reading config file /home/local/ganga/.gangarc
>>> d=ganga.Dirac()
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'Dirac'
```

Note I have the plugin defined in my config

```ini
RUNTIME_PATH = GangaDirac
```

and it works interactively
```python
*** Welcome to Ganga ***
Version: 6.2.2
Type help() or help('index') for online help.

This is free software (GPL), and you are welcome to redistribute it
under certain conditions; type license() for details.


INFO     reading config file /home/local/ganga/.gangarc
[16:11:59]
Ganga In [1]: d=Dirac()

[16:12:02]
Ganga In [2]:
```

rob-c:
Fixes #863  This moves more code out of the bootstrap into the correct utility and calls it from both `ganga` and during normal startup.

Rough and ready so feel free to fix/improve/edit/take-inspiration from this PR.

### DIFF
--- a/python/Ganga/Utility/Runtime.py
+++ b/python/Ganga/Utility/Runtime.py
@@ -122,7 +122,7 @@ class RuntimePackage(object):
             if allRuntimes[self.name].path != self.path:
                 logger.warning('possible clash: runtime "%s" already exists at path "%s"', self.name, allRuntimes[self.name].path)
 
-        allRuntimes[self.name] = self
+        allRuntimes[self.path] = self
 
         if self.syspath:
             # FIXME: not sure if I really want to modify sys.path (side effects!!)
--- a/python/Ganga/Utility/Runtime.py
+++ b/python/Ganga/Utility/Runtime.py
@@ -196,11 +196,11 @@ class RuntimePackage(object):
             template_pathname = os.path.join(self.modpath, 'templates')
             if os.path.isdir(template_pathname):
                 establishNamedTemplates(template_registryname,
-                                        template_pathname,
-                                        "Registry for '%s' NamedTemplates" % self.name.strip(
-                                            'Ganga'),
-                                        file_ext=file_ext,
-                                        pickle_files=pickle_files)
+       template_pathname,
+       "Registry for '%s' NamedTemplates" % self.name.strip(
+           'Ganga'),
+       file_ext=file_ext,
+       pickle_files=pickle_files)
         except:
             logger.debug('failed to load named template registry')
             raise
@@ -236,6 +236,42 @@ class RuntimePackage(object):
             logger.debug("no postBootstrapHook() in runtime package %s", self.name)
 
 
+def initSetupRuntimePackages():
+    """ Short wrapper function to init and setup Runtime Plugins """
+    initRuntimePackages()
+
+    # perform any setup of runtime packages
+    logger.debug('Setting up Runtime Packages')
+    for r in allRuntimes.values():
+        r.standardSetup()
+
+
+def initRuntimePackages():
+    """
+    initialize runtime packages, they are registered in allRuntimes
+    dictionary automatically
+    """
+    import Ganga.Utility.files
+    from Ganga.Utility.Config.Config import getConfig
+    config = getConfig('Configuration')
+
+    #if config['IgnoreRuntimeWarnings']:
+    #    import warnings
+    #    warnings.filterwarnings(action="ignore", category=RuntimeWarning)
+
+
+    import inspect
+    import os.path
+
+    GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
+    def transform(x):
+        return os.path.normpath(Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath,x)))
+
+    paths = map(transform, filter(None, map(lambda x: os.path.expandvars(os.path.expanduser(x)), config['RUNTIME_PATH'].split(':'))))
+
+    for path in paths:
+        r = RuntimePackage(path)
+
 
 def loadPlugins(environment):
     """
@@ -245,7 +281,9 @@ def loadPlugins(environment):
     from Ganga.Utility.logging import getLogger
     logger = getLogger()
     env_dict = environment.__dict__
+    logger.debug("Loading: %s PLUGINS" % str(allRuntimes.keys()))
     for n, r in allRuntimes.iteritems():
+        logger.debug("Bootstrapping: %s" % n)
         try:
             r.bootstrap(env_dict)
         except Exception as err:
@@ -254,12 +292,13 @@ def loadPlugins(environment):
             raise err
         try:
             r.loadNamedTemplates(env_dict, Ganga.Utility.Config.getConfig('Configuration')['namedTemplates_ext'],
-                                           Ganga.Utility.Config.getConfig('Configuration')['namedTemplates_pickle'])
+          Ganga.Utility.Config.getConfig('Configuration')['namedTemplates_pickle'])
         except Exception as err:
             logger.error('problems with loading Named Templates for %s', n)
             logger.error('Reason: %s' % str(err))
 
     for n, r in allRuntimes.iteritems():
+        logger.debug("Loading: %s" % n)
         try:
             r.loadPlugins()
         except Exception as err:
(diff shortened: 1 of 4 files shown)


egede:
@alexanderrichards Can you take charge of getting this merged in?

alexanderrichards:
sure np

alexanderrichards:
i'm afraid this didn't solve the problem for me

rob-c:
@alexanderrichards Can you try defining the plugins via environmental variables rather than the .gangarc I think it's still a bit of a mystery how `ganga` should load the config files.

python/Ganga/Utility/Runtime.py
```diff
@@ -236,6 +236,42 @@ def postBootstrapHook(self):
...
+
+
+    import inspect
+    import os.path
+
+    GangaRootPath = os.path.normpath(os.path.join(os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe()))), '../..'))
+    def transform(x):
+        return os.path.normpath(Ganga.Utility.files.expandfilename(os.path.join(GangaRootPath,x)))
+
+    paths = map(transform, filter(None, map(lambda x: os.path.expandvars(os.path.expanduser(x)), config['RUNTIME_PATH'].split(':'))))
```
alexanderrichards: it seems that
```python
config['RUNTIME_PATH']
```
is empty here even though it's set in my gangarc. I assume it hasn't read it in properly yet

alexanderrichards:
> @alexanderrichards Can you try defining the plugins via environmental variables rather than the .gangarc I think it's still a bit of a mystery how ganga should load the config files.

This ties in with what I just noted. I'll try and let you know

alexanderrichards:
```
$ export RUNTIME_PATH=GangaDirac
$ python
Python 2.7.5 (default, Sep 15 2016, 22:37:39) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ganga
INFO     LCG config is not set. Please set [LCG]VirtualOrganisation to your VO (e.g. "gridpp") or disable LCG with [LCG]GLITE_ENABLE=False
config= ['']
paths= []
>>> ganga.Dirac()
  File "<stdin>", line 1, in <module>
AttributeError: 'module' object has no attribute 'Dirac'
>>> 
alas no dice
```

alexanderrichards:
it seems it's not reading the environment properly either. Here I simply changed your debugging log message to info on line: https://github.com/ganga-devs/ganga/pull/866/files#diff-ee788bbe32ed1c5d501ba3fdcd5b9b93R284

```bash
$ echo $RUNTIME_PATH
GangaDirac
$ python
```
```python
Python 2.7.5 (default, Sep 15 2016, 22:37:39) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import ganga
INFO     LCG config is not set. Please set [LCG]VirtualOrganisation to your VO (e.g. "gridpp") or disable LCG with [LCG]GLITE_ENABLE=False
INFO     Loading: [] PLUGINS
```

rob-c:
@alexanderrichards and now?

python/Ganga/Utility/Runtime.py
```diff
@@ -122,7 +122,7 @@ def __init__(self, path):
             if allRuntimes[self.name].path != self.path:
                 logger.warning('possible clash: runtime "%s" already exists at path "%s"', self.name, allRuntimes[self.name].path)
 
-        allRuntimes[self.name] = self
+        allRuntimes[self.path] = self
```
alexanderrichards: does this need to be changed on line 122 just above?

rob-c:
Hmm it's working for me, but only if I define it in my site config. I suspect this could be related to the bug that Mark saw that the options aren't being loaded/set with the correct ranking.

alexanderrichards:
I found that the following patch fixes the problem and it reads from the config file properly. The strange thing is that it has the right code there just commented out so somebody commented it out on purpose for some reason. Without knowing what that reason was it's hard to know what the consequences of putting it back in are.

```diff
--- a/python/Ganga/Utility/Config/__init__.py
+++ b/python/Ganga/Utility/Config/__init__.py
@@ -31,7 +31,8 @@ def load_config_files():
     system_vars = {}
     for opt in getConfig('System'):
         system_vars[opt] = getConfig('System')[opt]
-        config_files = GangaProgram.get_config_files('')#os.path.expanduser('~/.gangarc'))
+        config_files = GangaProgram.get_config_files(os.path.expanduser('~/.gangarc'))
+        print "load_config_files=", config_files
         setSessionValuesFromFiles(config_files, system_vars)
```

alexanderrichards:
Ah I just figured out that that bit of code was new with this PR. Can you check @rob-c that with my patch added on it still works for you

alexanderrichards:
seems that the tests are passing @rob-c did you get a chance to look at the patch I provided. If all is well I say we should be able to merge this in for the next release whether that's a bugfix for 6.2 or 6.3

rob-c:
@alexanderrichards Yes this works, apologies I commented out the eval of `~/.gangarc` during testing and plumb forgot.

All works for me still I can do LHCb stuff with it which is good :+1: 
<Aside>
We should probably try to consolidate the amount of times that `~/.ganga` appears in the code as it's in quite a few places now.

egede:
One of the options to Ganga at start-up is

```
  --config=FILE         read user configuration from FILE, overrides the
                        GANGA_CONFIG_FILE environment variable. Default:
                        ~/.gangarc
```
Does the code above somehow hard-wire where the `.gangarc` is located as thus ignoring this command line option?

rob-c:
@egede I think this just means we need to update: https://github.com/ganga-devs/ganga/pull/866/commits/11281ca7eaf3f05bb19466521c8214dbee2007f5

Atm this code is only called from `ganga` during import. It would be good to get the normal ganga startup to call this but atm that's too difficult. And this is duplicated code from the tests and we should de-duplicate this code to allow us to recycle the same code between tests and core.

alexanderrichards:
@egede the command line option is not that important because as @rob-c says this code is only executed for importing little ganga hence you won't be passing it and command line options. However it is ignoring the `GANGA_CONFIG_FILE` env variable. A fix for this would be the following.

```diff
--- a/python/Ganga/Utility/Config/__init__.py
+++ b/python/Ganga/Utility/Config/__init__.py
@@ -31,7 +31,8 @@ def load_config_files():
     system_vars = {}
     for opt in getConfig('System'):
         system_vars[opt] = getConfig('System')[opt]
-        config_files = GangaProgram.get_config_files(os.path.expanduser('~/.gangarc'))
+        user_config = os.environ.get('GANGA_CONFIG_FILE') or os.path.expanduser('~/.gangarc')
+        config_files = GangaProgram.get_config_files(user_config)
         setSessionValuesFromFiles(config_files, system_vars)
```

I can make this fix in order to progress this ticket but to me it feels like the system is creaking a bit and we should possibly return to look at the config system again.

alexanderrichards:
@rob-c assuming this doesn't break the tests, can you still do "LHCB Stuff"?

rob-c:
The point of moving this into `Config` is that we then have a very clear single method call to load the config in the import of `ganga` and this (bar a minor correction) should be the code we're using in the tests.
(This doesn't parse command line options and there should be an open issue about the fact these are broken atm as was discovered by mark).

This should load __both__ files in order rather then 1 or the other surely, I imagine this would cause issues for me trying to load LSST @ CERN where the site config expects to load LHCb?

alexanderrichards:
the site configs use a separate env variable GANGA_CONFIG_PATH. The one in this patch is purely a replacement for the default user config file so an or is appropriate here i think.

rob-c:
@alexanderrichards Ah yes your correct, nmv me I mis-read the code at a glance.
I can't imagine this causes problems, I'll try to launch an LHCb env in a min to test

alexanderrichards:
I'm going to merge this branch in the next few minutes unless there are objections from anyone?

rob-c:
@alexanderrichards Don't see any problems, can construct some LHCb/LSST stuff :)

PR #869, opened 2016-11-21
Adding simplest lock to Ganga

rob-c:
This was something I was playing with over the weekend mainly.

This PR shows how we can have the simplest lock possible which only prevents Ganga from running multiple sessions. The shutdown is relatively friendly and doesn't have a stack trace.

This would likely be useful for non-POSIX envs where a user would be able to launch similar to:

```
IPYTHONDIR=/tmp/somePath ganga -o'[Configuration]lockingStrategy=Fixed'
```

This may be useful for unusual installs or debugging (or even in testing as this reduces a lot of I/O).

I'm not sure if someone else has an idea of how they would like to solve this but this would be what I suggest.

NB:
filelock in python still relies on posix locking (I think) if it's there so migrating SessionLock to this is only a cleaning up exercise and doesn't help if we can't posix lock a file.

### DIFF
--- a/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
+++ b/python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
@@ -12,7 +12,7 @@ import errno
 import copy
 import threading
 
-from Ganga.Core.GangaRepository.SessionLock import SessionLockManager
+from Ganga.Core.GangaRepository.SessionLock import SessionLockManager, test_unix_locks
 from Ganga.Core.GangaRepository.FixedLock import FixedLockManager
 
 import Ganga.Utility.logging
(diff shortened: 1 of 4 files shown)


alexanderrichards:
> filelock in python still relies on posix locking (I think) if it's there so migrating SessionLock to this is only a cleaning up exercise and doesn't help if we can't posix lock a file.

I'm not sure that this is true. filelock doesn't wrap POSIX commands, instead it writes a lockfile to disk. This does mean however that it may well suffer strange issues of replication on shared file systems like AFS that we have seem before.

rob-c:
@alexanderrichards Could we be mixing up `lockfile` and `filelock` (probably my mistake). The latter https://github.com/benediktschmitt/py-filelock/blob/master/filelock.py uses posix file locks when the module is available, whilst I suspect the latter may as you say uses file presence on disk which potentially has problems over {a,n}fs systems.

alexanderrichards:
I was indeed talking about filelock which as I said uses files on disk. It looks like lockfile is deprecated in favour of fastners.
[lockfile PyPI](https://pypi.python.org/pypi/lockfile/0.12.2)

rob-c:
Are there any objections to me merging this? I'm keen to play with possible performance improvements we may get from this. It reduces a lot of I/O ops for situations where want to explicitly plan to only allow 1 instance of Ganga to run.
(I think a review of a locking strategy when supporting parallel sessions is probably needed but I am foreseeing use cases for these changes)

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -225,7 +228,10 @@ def startup(self):
             self.from_file = pickle_from_file
         else:
             raise RepositoryError(self.repo, "Unknown Repository type: %s" % self.registry.type)
-        self.sessionlock = SessionLockManager(self, self.lockroot, self.registry.name)
+        if getConfig('Configuration')['lockingStrategy'] == "UNIX":
+            self.sessionlock = SessionLockManager(self, self.lockroot, self.registry.name)
+        else:
```
egede: Should check for the value "Fixed" rather than just an elese statement (and any other message should lead to a complete fail).

rob-c:
Hmm, I'll add a test for file locking not working and have an appropriate exception raised. This would require a little effort so I'll get around to this soon in this case.

rob-c:
@egede I've now added an explicit test for the locking strategy and added a check to see if UNIX style file locking is supported before trying to construct the old SessionLock class objects.
I'd like to push for 6.3.1 with this as I know of a few situations where people are wanting to sacrifice the ability to have multiple sessions for this.

mesmith75:
Merge?

PR #872, opened 2016-11-24
Add 'new' to possible subjob status

### LINKED ISSUE #862: Job status shown as "completed" despite having subjobs with "new" state
Ganga appears to prioritise the "completed" state above "new" when computing the status of a job containing subjobs. I would expect the parent job to only be shown as "completed" when all subjobs are also completed.

Output of `jobs`:
screen shot 2016-11-13 at 12 15 27

Output of `jobs(58).subjobs`:
screen shot 2016-11-13 at 12 15 57

mesmith75:
Fixes #862. Add new to possible subjob status for setting the masterjob status.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -743,7 +743,7 @@ class Job(GangaObject):
 
         new_stat = None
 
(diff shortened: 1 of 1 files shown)


egede:
@rob-c comment in #862 makes me get cold feet on this PR. Not quite clear to me if the Jenkins failures are due to the problems mentioned there.

mesmith75:
OK, I'll scrap this and we'll do something better.

PR #881, opened 2016-11-28
LHCbDataset uses full path for LocalFile

### LINKED ISSUE #874: LocalFile inputdata
It looks as if LocalFiles cannot be used for inputdata for local testing. The directory path gets stripped off so only the namePattern is used. Is this a deliberate choice or something we should fix?

mesmith75:
Fixes #874 . Changed the optionsString of the LHCbDataset to use the accessURL rather than namePattern for non-DiracFile s

### DIFF
--- a/python/Ganga/Core/GangaRepository/VStreamer.py
+++ b/python/Ganga/Core/GangaRepository/VStreamer.py
@@ -1,7 +1,7 @@
 from __future__ import print_function, absolute_import
 from Ganga.Core.exceptions import GangaException
 from Ganga.Utility.logging import getLogger
-from Ganga.GPIDev.Base.Proxy import stripProxy, isType, getName
+from Ganga.GPIDev.Base.Proxy import addProxy, stripProxy, isType, getName
 
 from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList, makeGangaListByRef
 
@@ -10,7 +10,7 @@ from Ganga.Utility.Config import config_scope
 
 from Ganga.Utility.Plugin import PluginManagerError, allPlugins
 
-from Ganga.GPIDev.Base.Objects import GangaObject
+from Ganga.GPIDev.Base.Objects import GangaObject, ObjectMetaclass
 from Ganga.GPIDev.Schema import Schema, Version
 from Ganga.GPIDev.Lib.GangaList.GangaList import makeGangaList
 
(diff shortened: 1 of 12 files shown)


mesmith75:
Any objections to merging?

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -300,6 +300,7 @@ def getCatalog(self, site=''):
         return xml_catalog
 
     def optionsString(self, file=None):
+        print 'Options STRING!'
```
rob-c: This is a typo surely?

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -339,8 +340,8 @@ def optionsString(self, file=None):
                 sdatasetsnew += """ \"LFN:%s\",""" % f.lfn
                 sdatasetsold += """ \"DATAFILE='LFN:%s' %s\",""" % (f.lfn, dtype_str)
             else:
-                sdatasetsnew += """ \"PFN:%s\",""" % f.namePattern
-                sdatasetsold += """ \"DATAFILE='PFN:%s' %s\",""" % (f.namePattern, dtype_str)
+                sdatasetsnew += """ \"%s\",""" % f.accessURL()[0]
+                sdatasetsold += """ \"DATAFILE='%s' %s\",""" % (f.accessURL()[0], dtype_str)
```
rob-c: Does the `accessURL` work for all file types? (I think I remember finishing this but I'm not sure I ever tested it)

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: yes

python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: It does for LocalFile. I'll check for MassStorageFile as well.

mesmith75: Yes, accessURL works for MassStorageFile - I can't think of any others for LHCb.

mesmith75:
Any objections to merging?

PR #884, opened 2016-11-29
Nightly Cred req fix

### LINKED ISSUE #882: Nightly build issues
How are we getting a certificate for the nightly tests? Credential problems seem to be the reason for their continued failure.

rob-c:
Potentially fixes #882 I don't know if we _want_ these external but in my mind they should be as they rely on commands/servers outside of Ganga.


drmarkwslater:
👍

mesmith75:
Can we merge this?

PR #888, opened 2016-12-01
Cmake folder exception exception

### LINKED ISSUE #886: prepareGaudiExec needs warning for pre-existing folder
Reminder to add a warning for prepareGaudiExec in case a checked out version is already there.

mesmith75:
Fixes #886 . For the prepare_cmake function.
Check whether the Dev folder already exists. If it does, raise a GangaException rather than continuing with the job submission.

### DIFF
--- a/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
+++ b/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
@@ -300,7 +300,6 @@ class LHCbDataset(GangaDataset):
         return xml_catalog
 
     def optionsString(self, file=None):
-        print 'Options STRING!'
         'Returns the Gaudi-style options string for the dataset (if a filename' \
             ' is given, the file is created and output is written there).'
         if not self or len(self) == 0:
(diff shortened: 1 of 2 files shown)


python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -339,8 +339,8 @@ def optionsString(self, file=None):
                 sdatasetsnew += """ \"LFN:%s\",""" % f.lfn
                 sdatasetsold += """ \"DATAFILE='LFN:%s' %s\",""" % (f.lfn, dtype_str)
             else:
-                sdatasetsnew += """ \"PFN:%s\",""" % f.namePattern
-                sdatasetsold += """ \"DATAFILE='PFN:%s' %s\",""" % (f.namePattern, dtype_str)
+                sdatasetsnew += """ \"%s\",""" % f.accessURL()[0]
```
rob-c: I'm not complaining about adding it in, but it seems unrelated to the issue unless I'm missing something?

mesmith75:
Any objections to merging?

PR #891, opened 2016-12-02
Fix prepare of old LHCb applications with new login scripts

### LINKED ISSUE #880: Problem when using the DaVinci application in Ganga with LbScripts v9r0p1
Hi,

There was long thread in the distributed analysis mailing list because we cannot use Ganga with LbScripts v9r0p1.

The error was:
RROR    Command ". SetupProject.sh   DaVinci v39r0" failed to properly setup environment.
ERROR    Prepare Error:
ApplicationConfigurationError: Command ". SetupProject.sh   DaVinci v39r0" failed to properly setup environment. 

in:
EnvironFunctions.py

This is because the check implemented matches:
DaVinci_release_area /afs/cern.ch/lhcb/software/releases

And that var does not exist in the new LbScripts.
The check is in any way ineffective as that is a variable set by LbLogin, not by a successful SetupProject...

This is maybe not a problem as I understand that people should use GaudiExec instead of DaVinci.
(Maybe DaVinci should be a thin wrapper around GaudiExec...)

egede:
Fixes #880

### DIFF
--- a/python/GangaLHCb/Lib/Applications/EnvironFunctions.py
+++ b/python/GangaLHCb/Lib/Applications/EnvironFunctions.py
@@ -2,6 +2,7 @@ from Ganga.Core.exceptions import ApplicationConfigurationError
 
 import copy
 import tempfile
+import string
 
 from Ganga.Utility.Shell import Shell
 from Ganga.Utility.logging import getLogger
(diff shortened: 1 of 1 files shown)


egede:
Can I go ahead and merge?

mesmith75:
Please delay, see #892 .

mesmith75:
Ok, this can be merged if there are no objections

PR #896, opened 2016-12-11
Fix for user configurations

### LINKED ISSUE #829: LHCb plugin overrides some configs
When the LHCb plugin loads it overrides some configurations that the user may have set (such as the list of SEs to use). Do we want to take the user configuration over that?

mesmith75:
Fixes #829 . Now the config files and command line options are set as UserValues rather than SessionValues so they take precedence over the later plugin loaded options. The command line options are loaded after the config ones, so they are still top.

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -697,13 +697,15 @@ under certain conditions; type license() for details.
 
         def set_cmdline_config_options(sects=None):
             try:
+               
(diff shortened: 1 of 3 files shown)


rob-c:
I know it's a bad example class to start with but some docs around the methods would be helpful when going back to this in the future if you get a chance before merging. Looks like it solves the problem however 👍

mesmith75:
Sure

mesmith75:
Can I merge this?

PR #901, opened 2016-12-14
Fix LHCb job migration

rob-c:
This fixes a bug where on loading a job with a Dirac backend from Ganga 6.2.x in Ganga 6.3.x

This bug crept in between the credentials branch being fixed and the 6.3 release so I've added a test to make sure that this doesn't happen again.

The consequence is any DIRAC job from Ganga 6.2 can't be loaded in 6.3.
(There won't be too many of these as DIRAC lost a lot of sandboxes but users still on Ganga 6.2.x not moving to 6.3 may create new jobs in the future that can't be read in without this fix).

The fix is small in DiracBase and was due to some shortcoming to check for classes in the Schema. The schema is clearly not working correctly in the case of classes as defaults and this should be examined by someone in more detail in a future PR!

This should be working and I think may prompt a quick release of 6.3.1

### DIFF
--- a/python/Ganga/GPIDev/Base/Proxy.py
+++ b/python/Ganga/GPIDev/Base/Proxy.py
@@ -270,6 +270,8 @@ def addProxy(obj):
                 return getattr(obj, proxyObject)
             else:
                 return GPIProxyObjectFactory(obj)
+    elif isclass(obj) and issubclass(obj, GangaObject):
+        return getProxyClass(obj)
     return obj
 
 
--- a/python/Ganga/GPIDev/Schema/Schema.py
+++ b/python/Ganga/GPIDev/Schema/Schema.py
@@ -21,6 +21,8 @@ import types
 
 from Ganga.GPIDev.TypeCheck import _valueTypeAllowed
 
+from inspect import isclass
+
 logger = Ganga.Utility.logging.getLogger()
 
 # Dictionary for storing data from the Config system which takes a while to lookup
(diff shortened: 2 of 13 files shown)


mesmith75:
:+1: That looks to have sorted it. The tests look fine as well. I agree we should release 6.3.1 promptly (before Christmas break) - there is already a lot to go in.

rob-c:
@mesmith75 Can you verify you see this bug in this PR this is the bug I have tried to fix here

rob-c:
To test I'm submitting a job from 6.2.3 then launch this PR (without monitoring) and run ```runMonitoring(jobs.select(jobs[-1].id))``` and this "just works".

mesmith75:
@rob-c The fix to the ICredential errors works. However if I start ganga and do ```jobs(8)```, where 8 was submitted from 6.2.3 then I get that exception. It is however fine for jobs submitted from 6.3

rob-c:
@mesmith75 I cannot reproduce this. The error your reporting is the same bug, calling a method function on a class object rather than an instance can you make sure your using the correct ganga version from a clean shell because I claim this has been fixed

rob-c:
``jobs(x).anything_not_backend`` will work due to the nature of this bug btw!

mesmith75:
Hmm - I must have messed up my environment somehow.

egede:
@mesmith75 So does your last comment mean that it now works for you?

mesmith75:
Ok, I didn't mess it up - the same problem is in (LHCb) SplitByFiles which that job was using.

mesmith75:
A quick grep suggests also BKQuery and DiracFile?

rob-c:
OK. I decided to revert the change in `DiracBase` and to fix the handling of classes around the proxy in Core and the Schema so that this doesn't just turn into a case of fire-fighting whenever good looking code is written as the errors from Python are cryptic to say the least in this regard.

rob-c:
This now directly addresses #902

rob-c:
retest this please

mesmith75:
Looks good to me

mesmith75:
If there are no objections lets merge this.

PR #907, opened 2016-12-16
Fixing file construction

### LINKED ISSUE #903: Old DiracFile not accessible
Due to some changes within the class initialization framework DiracFile old classes which don't have `defaultSE` will fail to load from XML.
This prevents the jobs from loading.

I think I've understood how this happens and the fix should be fairly quick and effective once I get a chance and it would be good to see in 6.3.1 so we can load old jobs properly (only in 6.3.x).

rob-c:
This should fix #903 whilst improving the exceptions thrown from the XML parser which until now are somewhat cryptic.

### DIFF
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -137,7 +137,7 @@ class DiracFile(IGangaFile):
 
     def _setLFNnamePattern(self, lfn="", namePattern=""):
 
-        if self.defaultSE != "":
+        if hasattr(self, 'defaultSE') and self.defaultSE != "":
             ## TODO REPLACE THIS WITH IN LIST OF VONAMES KNOWN
             # Check for /lhcb/some/path or /gridpp/some/path
             if namePattern.split(os.pathsep)[0] == self.defaultSE \
--- a/python/Ganga/Core/GangaRepository/VStreamer.py
+++ b/python/Ganga/Core/GangaRepository/VStreamer.py
@@ -359,35 +359,44 @@ class Loader(object):
                 aname = self.stack.pop()
                 obj = self.stack[-1]
                 # update the object's attribute
-                obj.setSchemaAttribute(aname, value)
+                try:
+                    obj.setSchemaAttribute(aname, value)
+                except:
+                    raise GangaException("ERROR in loading XML, failed to set attribute %s for class %s" % (aname, _getName(obj)))
                 #logger.info("Setting: %s = %s" % (aname, value))
 
             # when </value> is seen the value_construct buffer (CDATA) should
             # be a python expression (e.g. quoted string)
             if name == 'value':
-                # unescape the special characters
-                s = unescape(self.value_construct)
-                #logger.debug('string value: %s',s)
-                if s not in _cached_eval_strings:
-                    # This is ugly and classes which use this are bad, but this needs to be fixed in another PR
-                    # TODO Make the scope of objects a lot better than whatever is in the config
-                    # This is a dictionary constructed from eval-ing things in the Config. Why does should it do this?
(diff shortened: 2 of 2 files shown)


python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -137,7 +137,7 @@ def locations(self):
 
     def _setLFNnamePattern(self, lfn="", namePattern=""):
 
-        if self.defaultSE != "":
+        if hasattr(self, 'defaultSE') and self.defaultSE != "":
```
alexanderrichards: Given that it's part of the schema why is this change necessary? wouldn't it be guaranteed to have this attribute?

rob-c: No, it's not there when you're reading in from an XML where the schema hasn't been written with the `defaultSE` as the first attribute within the file. Also this could happen during object construction as the Schema is _not_ an ordered dict and so the order in which the parameters arrive here is effectively random (we're just lucky that in 99% of cases it's sequential)

rob-c: Or even an XML where this parameter didn't exist as this will only be assigned to the object after the other XML attributes have been read in.

rob-c: The object has a defined slot in the class as per `Objects.py`.
The attribute is only correctly set via the class calling `__init__` and initializing the attributes to the default value.
(I don't know if hasattr will pass when only the slot has been assigned in the metaclass)

However, when a new class is constructed via `__new__` the class doesn't fully assign the attributes when constructing the object as they're not required.

Users are seeing:
```
ERROR    XML from-file error for file:
Could not find attribute defaultSE in DiracFile(namePattern='', lfn='', localDir='None')
```

When attempting to load old jobs which don't have this attribute stored in the XML.

By checking for the attribute before calling it I think this should be enough to fix this bug and make old jobs accessible again.

Wrt the order of objects in the Schema I think the order that objects are constructed within the `Objects.py` is such that the `defaultSE` is assigned to the class first before the other attributes look for it, although I could be wrong.

Ideally things like this would go away if we made Schema objects handled via a defaultdict.

alexanderrichards: hmmm this strikes me as a failing of the `GangaObject` and `ObjectMeta` classes as something like this is perfectly possible, see the simple example below.
```python
class ObjectMeta(type):
    def __new__(meta, name, bases, dct):
        dct['__slots__'] = dct['schema'].keys()
        return super(ObjectMeta, meta).__new__(meta, name, bases, dct)

class Object(object):
    __metaclass__ = ObjectMeta
    schema = {}

    def __getattr__(self, attr):
        if attr in self.__class__.schema:
            return self.__class__.schema.get(attr)
        return super(Object, self).__getattr__(attr)

class bob(Object):
    schema = {'frank':'default'}

    def __init__(self):
        print self.frank


>>> b=bob()
default
```
__Note__: I realise that this will not properly work with multiple instances, it's a proof that these things can be pre-setup. My schema is just a `dict` and so has no notion of instances whereas a more complicated Schema object could. Of course the proper way to do this is with properties which is what I did in the stuff that I was playing with

----
OK since this is a quick fix for a problem that is annoying for users otherwise I think it should be fine but it certainly highlights to me that I need to redouble my efforts to look at the schema.

You might like to think about reducing the conditional to the smaller but equivalent (I think) version below
```diff
-        if hasattr(self, 'defaultSE') and self.defaultSE != "":
+        if getattr(self, 'defaultSE', '') != '':
```

rob-c:
@mesmith75 also I'd like to see this in 6.3.1 as it solves some problems users have with loading really old jobs.

PR #917, opened 2016-12-20
Optimise ganga

rob-c:
This is to explore the effect (if any) migrating more classes to use `__slots__` has on the performance of Ganga.


rob-c:
@egede I found some things (such as creating large lists of objects) are a bit faster in this branch with some additional changes. I'll work on finalizing this when I get a chance at some point soon.

rob-c:
retest this please

rob-c:
OK, this is a largely working PR which addresses mainly the bug in creating many GangaObject files in series as well as addressing some performance issues to do with memory.
In this PR I see Ganga able to construct many GangaObject class instances with RAM consumed at ~20Mb/s. The overall footprint in Ganga remains un-effected but the shutdown appears to be slightly faster from some tests due to the use of `__slots__`.
Within this PR I am able to construct 10k DiracFile objects within ~7sec where as on develop this can take ~30+sec. I believe the gains here are enough to push for inclusion of the branch as also loading a job with ~1000subjobs only takes 10sec on this branch vs 30-40sec on develop.

The most significant changes have been some minor changes to the constructor within the Proxy tool which was previously fully-constructing then modifying an underlying object instance. As the `__set__` method on GangaObject is known to be extremely heavy this turns out to be rather costly so I've changed it to construct an empty class and then populate it on construction.

The tests are largely uneffected by this but I plan to add a test to check that this doesn't get re-introduced.

In the interest of full disclosure this PR has turned up some minor bugs elsewhere which have been very quick to fix. This is because the order in which attributes are assigned has varied subtly from before. There isn't anything I see too alarming here but it's possible this PR could expose bugs to communities other than Core/LHCb that I simply don't know about.

I plan to go through and document a lot of the changes (probably next week due to other work) but I'd welcome input at this stage. I would like to push for these changes to be merged for 6.5 or 6.6 whichever is best.

egede:
@rob-c Thanks for the long explanation. Can you for my education explain what a ```slot``` is.

Just request a review when you feel that it is ready.

rob-c:
@egede a `__slot__` is a Python feature which instructs Python in advance of constructing a class what attributes it will have. If an attribute isn't named here then you can't assign that attribute to a class. (The disadvantage of the class instance is then not fully dynamic and there are some other gotchas if this is mis-used). 
This has advantages of speed from Python but it comes with a cost of knowing up front what you want the class to look like and I think this behaviour is different to how for example the Config system works.

We've included `__dict__` within the `__slots__` for `GangaObject` so a user/developer can still do something at runtime as they please whilst the Core has some definite memory reduction that we've seen since a while ago (6.1.xx) where we first started using this.

Formally:
https://docs.python.org/2/reference/datamodel.html#slots

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -749,7 +755,7 @@ def check(self):
                 continue
             fd = None
             try:
-                sf = os.path.join(self.sdir, session)
+                sf = self._path_helper()
```
mesmith75: Does this need `session` inside the parentheses?

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -782,7 +788,7 @@ def get_lock_session(self, id):
         for session in sessions:
             fd = None
             try:
-                sf = os.path.join(self.sdir, session)
+                sf = self._path_helper()
```
mesmith75: Also here?

python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -830,7 +836,7 @@ def reap_locks(self):
 
         for session in sessions:
             try:
-                sf = os.path.join(self.sdir, session)
+                sf = self._path_helper()
```
mesmith75: And here?

python/Ganga/GPIDev/Base/Objects.py
```diff
@@ -388,8 +392,6 @@ def cloneObject(v, obj, name):
         else:
             if not isinstance(v, Node) and isinstance(v, (list, tuple)):
                 try:
-                    # must import here as will fail at the top
-                    from Ganga.GPIDev.Lib.GangaList.GangaList import GangaList
```
mesmith75: I may have missed it but where is the GangaList being imported now?

mesmith75:
I'll check this out and run some tests with it. If we are happy we should merge it.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
```diff
@@ -248,7 +248,7 @@ def startup(self):
         self.sessionlock.startup()
         # Load the list of files, this time be verbose and print out a summary
         # of errors
-        self.update_index(verbose=True, firstRun=True)
+        self.update_index(True, True)
```
mesmith75: @rob-c For my own information, what is the benefit here?

rob-c:
There was something I wanted to check re: XML parsing but if you're happy to accept it as is then it addresses some extremely annoying LHCb problems which I suspect are also hitting Durham from what was mentioned recently in another meeting.

python/Ganga/Core/GangaRepository/GangaRepositoryXML.py
rob-c: @mesmith75 A lot of the work in this PR was either to try and keep memory usage as small as possible or to reduce a lot of different entries found by kcachegrind when profiling the  under certain tasks which were showing to take a while.
(There is an extremely minor overhead to using the attribute names here due to (un)packing dictionaries which was showing up in kcachegrind and I was trying to explore what was causing some methods to be called an obscene number of times. This does normally only exhibit when doing tests of calling the same method of the order of 100k+ times.)
I was seeing a large number of dictionary lookups which, whilst not bad (because the code is highly efficient), was still such a large number as to be concerning. (of the order of 1M calls to lookup something during constructing ~1k objects)

Exploring why some methods (such as a dict lookup) were being called a lot eventually lead to finding a highly nested trace just for assigning a default attribute which I was able to fix within the Proxy layer which accounts for (I'm guessing) ~90% of the perceived user slowness when dealing with large numbers of `GangaObject` type objects.

I suppose there are arguments for/against micro-optimisations but I prefer to try and keep the total number of calls to a minimum if required within a kcachegrind profile as it at least makes some performance profiles easier to read (imo).

Ultimately the change here has a negligible performance impact so if the code is clearer without this I'm happy to revert, but it's better to work without the argument names where possible I'd argue.

python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -470,14 +472,13 @@ def accessURL(self, thisSE='', protocol=''):
...
         """
-        from GangaDirac.Lib.Backends.DiracUtils import getAccessURLs
         lfns = []
         if len(self.subfiles) == 0:
             lfns.append(self.lfn)
         else:
             for i in self.subfiles:
                 lfns.append(i.lfn)
-        return getAccessURLs(lfns, thisSE, protocol)
+        return DiracUtils.getAccessURLs(lfns, thisSE, protocol)
```
mesmith75: This doesn't work `NameError: global name 'DiracUtils' is not defined` . What is the reasoning for this change?

rob-c:
I'm happy to do some work on this if it's needed but I'd like to push for inclusion of this as it fixes several bottlenecks for performance users (I'm almost certain Durham is hitting this on a regular basis based upon my discussions with them a while ago).

mesmith75:
Ok, I'll check this out again and git it a test.

rob-c:
retest this please

mesmith75:
Not sure this works for submitting a GaudiExec job:
```
ERROR    ApplicationConfigurationError: Prepared folders not supported yet, please fix this in future
```

rob-c:
@mesmith75 What script did you use to get this error?

mesmith75:
Actually after pulling in the latest changes I have a different problem:
```
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/bin/ganga", line 63, in <module>
    Ganga.Runtime._prog.run(GPI.__dict__)
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/Ganga/Runtime/bootstrap.py", line 1057, in run
    execfile(script, local_ns)
  File "job.py", line 41, in <module>
    ds.extend(BKQuery(modes[mode][polarity]).getDataset())
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/Ganga/GPIDev/Base/Proxy.py", line 695, in proxy_wrapped
    r = f(*s_args, **s_kwargs)
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/Ganga/GPIDev/Credentials/__init__.py", line 44, in cred_wrapped_method
    return method(self, *args, **kwargs)
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/GangaLHCb/Lib/LHCbDataset/BKQuery.py", line 167, in getDataset
    result = get_result(cmd, 'BK query error.')
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/GangaDirac/Lib/Backends/DiracUtils.py", line 34, in get_result
    return execute(command, eval_includes=eval_includes)
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/python/GangaDirac/Lib/Utilities/DiracUtilities.py", line 194, in execute
    env = getDiracEnv(cred_req.dirac_env)
AttributeError: 'NoneType' object has no attribute 'dirac_env'
```

rob-c:
@mesmith75 I'll fix that bug later, it just needs a check for a None object or to pass the proxy correctly from the BKQuery to the dirac call. (the latter I suspect is better)

rob-c:
@mesmith75 I should have fixed the bug in the `BKQuery` but there may problems with the `GaudiExec` if you can check this still I've no check other than the unittests for this class.
I've added a small change which is to make the monitoring follow a fork/merge type approach to talking to the backend. I plan to add a configurable to allow moving back to the old style behaviour as this has only been tested against the Core and LHCb plugin so other plugins may not be correctly thread-safe with regard to monitoring.
To the end user this shouldn't have much of a change but should hopefully improve the speed of the monitoring whilst stopping it from potentially overloading a backend if Ganga is busy with other worker threads.

rob-c:
@mesmith75 Are there any outstanding issues with this branch?

rob-c:
I had to fix a problem with the caching the credentials code not being thread safe. This should alleviate a lot of issues and seems to be related to memory usage going down based upon some tests (1k subjobs in 5 master jobs to the localhost doesn't go above 50Mb on my test system).

mesmith75:
@rob-c Seems to be fine so far from my usage.

rob-c:
I won't push for 6.6 as I'm away next week and have no free time to really test/debug this but I'd like to consider it for 6.7 as it fixes (amongst other things) an annoying bottleneck due to a threading problem in the credentials system.

python/Ganga/GPIDev/Credentials/CredentialStore.py
```diff
@@ -83,6 +83,11 @@ def create(self, query, create=True, check_file=False):
             The newly created ICredentialInfo object
         """
 
+        try:
+            assert isinstance(query, ICredentialRequirement)
+        except AssertionError:
+            raise CredentialsError("Object: '%s' is NOT an ICredentialRequirement instance" % str(query))
```
milliams: Why not simply:
```python
if not isinstance(query, ICredentialRequirement):
    raise CredentialsError("Object: '%s' is NOT an ICredentialRequirement instance" % query)
```

rob-c:
OK, in summary the main changes in this branch are:

 * Using `__slots__` in a lot more classes.
This at the very least improves garbage collection but also means we have a feeling for what objects each class has at a class level more quickly.
 * Added the `_additional_slots` attribute.
This allows for inherited classes to add additional objects to the `__slots__` from the base class easily.
 * Added the `getNew` method to GangaObject.
This allows to get an object of an inherited type which has been fully instantiated correctly without having to fully define the attributes of a class from the schema. This is useful when creating an empty object to be populated with attributes from storage such as an XML file.
* Fixed  a serious performance regression in the construction of simple objects through the Proxy.
There was a loop where objects were being constructed and overloaded multiple times within the proxy construction layer which was slowing the construction of new class objects from within the string parser and IPython interface.
* Fixed a race condition in the credentials caching.
Creating a new cache was not thread safe and appeared to be at times sub-optimal. I've added a global thread lock around cache operations as well as some additional logic to ensure information read from disk is correctly cached and returned from memory where applicable.
* Added the ability for generic backends to parse multiple jobs in parallel using queues.
The code required to monitor/update a job has been demanded to be thread-safe through this interface for a while so I've allowed for parsing multiple jobs in parallel to speed up monitoring. This is good on backends where monitoring many jobs at once causes performance problems.
This expands on the existing logic to prevent overloading the backends with too many large serial queries and allows for multiple parallel short number of jobs to be monitored whilst not consuming all worker threads within the queues system.

These were all problems highlighted from trying to fix profiling tests where various pieces of code were performing poorly. In particular the issues I was able to address were:
* Monitoring would capture a job for monitoring but then be very slow to proceed to monitor further jobs or taking a long time to monitor many subjobs. (fixed by multi-threaded monitoring)
* Reading in a large job from disk took >60sec (This is now better but is still a few sec). (fixed by `getNew`)
* Creating many new objects from in front of the GPI took orders of magnitude longer than from behind the proxy. (fixed by bug in proxy layer)

As well as this I've fixed some annoying papercuts:
* Making better errors rather than throwing from asserts hidden deep in the codebase to the user
* Fixing annoying bug with credentials system where if you pass the wrong object you get a stack trace
* Fixing bug where credentials weren't satisfied ahead of a parallel submit from the baseclass. (This meant the submit fails due to not being able to construct a credential from a queue).

I appreciate it's a rather large single commit so if people have the time to review the changes I'd like to push for 6.7 with this.

python/Ganga/GPIDev/Credentials/CredentialStore.py
rob-c: I was planning to add the other parameters here and have a more general error at the assertion error but I haven't got around to adding checks for the other parameters yet.

mesmith75:
@rob-c Is this the final version? Are you looking to merge this now? If so I'll look at the code and use it for analysis.

rob-c:
@mesmith75 I'm happy to see this merged. If you want to test please let me know if there are any bugs but I'd be happy to have a look 👍

mesmith75:
I think something has gone awry with job submission:
```python
ERROR    ApplicationConfigurationError: Prepared folders not supported yet, please fix this in future
ERROR    ApplicationConfigurationError: Prepared folders not supported yet, please fix this in future ... reverting job 1924 to the new status
```
when submitting a standard job.

rob-c:
@mesmith75 What is the standard job you refer to here? An out of the box `j=Job();j.submit()` works fully as expected.

mesmith75:
Sorry I should have been clear - a GaudiExec job.

rob-c:
Yeah, I'm not seeing this at all.

Can you upload your script of what you're doing because I suspect the problem is elsewhere.
I run the external GaudiExec test and it submits a DaVinci job that passses a string correctly without issue.

The problem you see should be because you're trying to add a directory to your inputfiles to the GaudiExec application. Presently this is not supported (or has been ever!) due to not having the code/support to ensure that the files are extracted to the correct directory on the WN for different backends which are support. Is this the case?

mesmith75:
I have inputfiles but they are not a directory:
```python
j = Job()
j.name = 'prod_Boole_newAttempt_currentMetal_cap_%s_diff_%s' % (cap, diff)
print 'submitting job: ', j.name
myApp = GaudiExec()
myApp.directory = "/afs/cern.ch/user/m/masmith/BooleWspline/constants/BooleDev_v31r0"
myApp.options = ["/afs/cern.ch/work/m/masmith/Boole/xDigi/prodTest/Boole-MC10.py"]
j.application = myApp
j.application.extraOpts = "from Gaudi.Configuration import importOptions\nfrom Configurables import Boole\nfrom Configurables import VeloSim\nVeloSim().sigmaScale = %s\nVeloSim().CapacitiveCoupling = %s" % (diff, cap)
j.inputdata = BKQuery("/MC/Dev/Beam6500GeV-2015-MagUp-Nu1.6-25ns-Pythia8/Sim09a/Digi14a-VeloRDRun2/30000000/XDIGI").getDataset()
j.splitter = SplitByFiles( filesPerJob = 2, maxFiles = 2000)
j.outputfiles = [DiracFile('*.xdigi'), DiracFile('*.digi')]
j.inputfiles = ["/afs/cern.ch/user/m/masmith/VeloClusterSizeFix/radDam-new.db"]
j.backend = Dirac()
j.submit()
```

rob-c:
@mesmith75 I've added another external test to GaudiExec which doesn't show this problem when submitting locally. Is this unique to using the Dirac backend?
Added some configs here to address #1025 I have made a few changes to the credentials here so adding an extra step seemed sensible.

mesmith75:
@rob-c I was submitting to Dirac - I'll try again to figure exactly what the problem is. Might it be the .db file?

mesmith75:
@rob-c So it is just a problem with the Dirac backend. The error occurs just for submitting to Dirac, even without inputfiles.

rob-c:
@mesmith75 I have added an extra test to test submitting a GaudiExec job with and without inputfiles to the LHCb Dirac backend.
It works as expected.

Can you have a look at the following test and check that I'm not doing something different to what you're doing above?
`GangaLHCb/test/GPI/TestExternalGaudiExec.py::TestExternalGaudiExec::testSubmitJobDiracWithInput`

I suspect this could be due to extraOpts as I've not yet added this to the external GaudiExec tests. I'll try and do that and see if it reveals anything.

mesmith75:
@rob-c I don't think it is the `extraOpts` - removing them doesn't help and nor does getting rid of the inputfiles.

rob-c:
@mesmith75 In that case can you run the new test I have put into this branch? If this doesn't go away for you with inputfiles can you verify what absolute minimal example required to reproduce this is?
Does it occur for you if you go away and start again with a clean project and try to submit it?

I suspect this could be a bug I'm sometimes seeing in calculating the path of the shared folder when preparing the app but I've never seen this outside of testing so wrote it off as a bug in the way a test is configured.

Do you see this problem if you prepare the app without submitting it?

mesmith75:
@rob-c It looks as if the preparing is fine.
I tried this with a minimal script using prepareGaudiExec to check out a clean project with no `extraOpts` or `inputfiles` and the error still occurs.

I'm not quite sure how the testing is working then - I've been unable to get them to work on my local copy but I think for other reasons.

rob-c:
@mesmith Why are the external tests failing for you? Does the simplest one of constructing and submitting a simple DaVinci job work?

If this is the case one of the remaining differences I'm seeing is that my tests all work with DaVinci but you're using Boole in your example. I'll rewrite the test (shortly) to duplicate this for some common LHCb apps and see if it works for all as expected.

mesmith75:
@rob-c I tried with DaVInci as well and the problem persisted.

mesmith75:
I think the issue is with the setting of the `shared_dir`. For this branch it is not getting set properly. For example for a job submitted with develop:

```python
[18:37:43]
Ganga In [4]: jobs(30).application.is_prepared
Ganga Out [4]: 
 ShareDir (
   name = conf-ee4b7bcd-c004-4b81-873d-e07258e2521a,
   subdir = .
 )
```

and a job that failed to submit from this branch:
```python
[18:38:07]
Ganga In [5]: jobs(29).application.is_prepared
Ganga Out [5]: 
 ShareDir (
   name = ,
   subdir = .
 )
```

rob-c:
@mesmith75 yes this is what I spotted last week. Working on a fix over the next week or so.

rob-c:
@mesmith75 If all of the tests are passing now then I've solved this.
These changes exposed another _very_ old bug in the prep.metadata repo which I've no idea how it hasn't cropped up before now but have fixed.

mesmith75:
@rob-c That seems to be working

rob-c:
OK, just added a final step to speed generating Ganga objects within the Schema.

At this point I'm happy to call this branch stable and I'm keen to get any feedback and would like to push for this to be merged.

The result of the optimisation work is to bring submitting a test job on my test system down to 0.5s from about 3 as well as speeding up the loading of objects from disk and the construction of many complex objects. (As well as a bunch of bugfixes and papercuts along the way)

mesmith75:
So the jobs submit but I don't think they are running. From the stdout:
```
<<<<<<<<<< jobScript/masmith_GaudiExec_Job_35.0_1b108df3-f446-4ab1-b7fe-dee924f0f7b6_script.py Standard Output >>>>>>>>>>

 <<<<<<<<<< jobScript/masmith_GaudiExec_Job_35.0_1b108df3-f446-4ab1-b7fe-dee924f0f7b6_script.py Standard Error >>>>>>>>>>

/cvmfs/lhcb.cern.ch/lib/lcg/releases/LCG_88/Python/2.7.13/x86_64-slc6-gcc49-opt/bin/python: can't open file 'jobScript/masmith_GaudiExec_Job_35.0_1b108df3-f446-4ab1-b7fe-dee924f0f7b6_script.py': [Errno 2] No such file or directory
```

rob-c:
@mesmith75 Thanks for letting me know, I'll give it a quick look at some point soon, although this may be next week due to other work

mesmith75:
@rob-c As this might still be a little while, would you mind if I picked out the configurable proxy part for a separate PR? I think it needs to be done sooner rather than later.

rob-c:
@mesmith75 I'm working on this anyway as part of bring on dast, should I break the changes out from this PR?

mesmith75:
@rob-c Yes please. I think it would be best to have that separate. Then we can put it in for the bugfix release for tomorrow.

rob-c:
@mesmith75 OK, I'll look into breaking it out, any objections to having any DIRAC interface related fixes in with this or are we wanting to keep this minimal?

mesmith75:
@rob-c In this instance it might be best to keep it minimal - this is a fix for a particular issue.

rob-c:
@mesmith75 on a side note I believe I've fixed this branch if you can test, there was a minor bug in the DIRAC RT for GaudiExec. I don't know of any other bugs in this branch but I'll keep looking.

rob-c:
@egede Shall I break out this PR? I'm happy to do it but it will likely take a bit of time to do.

With putting fixes in I'm finding it difficult to go back to stock Ganga because debugging dirac related things there is more painful due to a race condition and performance issues. but yes in future small fixes should go into additional branches but I was reluctant to start committing branches of branches into a spider-web of dependencies.

With regard to fixing/changing the LFN structure this is something I've been asked about from 2 different communities now, lhcb and lsst is that people want to manage their jobs on the grid but the mapping of job input/output to an LFN is non trivial especially once the DIRAC job has been removed from the UI (which I've seen people resorting to over the `job.backend.id` parameter).
Is there some better naming convention or some better structure which can be adopted if the user wants files to be automatically sorted according to the job destination?

egede:
@rob-c Yes, please break out the PR related to the naming convention. I encourage that we have flexibility in the naming convention, but we have to be careful about changing defaults. We could also consider to change the default, but then keep the "LHCb default" as the old value through use of the configuration system. We have a way of deciding on the directory and file structure of output for mass storage files. Should we review the syntax used there and make it similar for DiracFile objects?

rob-c:
There is some bleeding between external and integration tests as this failed with:
```

PEM file has expired /tmp/x509up_u494:gridpp_user is not valid after 2017-03-02 13:24:45
PEM file has expired /tmp/x509up_u494:gridpp_user is not valid after 2017-03-02 13:24:45
PEM file has expired /tmp/x509up_u494:gridpp_user is not valid after 2017-03-02 13:24:45
Can't contact DIRAC CS: Reason(s):
	PEM file /tmp/x509up_u494:gridpp_user has expired, not valid after 2017-03-02 13:24:45
Generating proxy...
```
where this wasn't expected.

Also, does the certificate need updating on a test node if this error is happening?

rob-c:
Ok, I've reverted code relating to #1040 and I would like to push for this to be merged if it passes testing by devs as it fixes a load of bottlenecks, cpu, disk and memory related in some cases.

egede:
retest this please

rob-c:
@mesmith75 I'm very happy to be nice and slow given the summer is coming up and people have holidays. That being said I'm treating this as finished 👍 I found and squashed some of the bugs which were annoying me here.

I'll start some work on #1072 to fix #1055 in the same PR

alexanderrichards:

mesmith75:
retest this please

mesmith75:
@rob-c The current release seems to be ok. So if everyone is happy let's merge this?

Then can you advertise on the mailing list that the DEV version is available for testing tomorrow (it gets updated at midnight)?

rob-c:
@mesmith75 OK I'll merge this and write a short email to the mailing lists advertising this as available for testing

PR #919, opened 2016-12-21
Create a new application for executing Jupyter notebooks

egede:


egede:
retest please

mesmith75:
restest this please

egede:
@rob-c Can you provide any insight to the failing integration tests. I added a new application type "Notebook" but now tests that seems completely unrelated are talking about errors. Any ideas.

egede:
For some reason it seems like the GPI tests have decided that the Notebook application is the default application for a job??

alexanderrichards:
**Note** For reference, this PR contains commit 6640f33ba6669d9a22be7fd306cb044edb596d91 from PR https://github.com/ganga-devs/ganga/pull/918

rob-c:
@egede My initial assumption is that you're breaking the Plugin system by adding a new application object. (and all of the _bad_ tests which rely on the defaults not changing).
I think some serious attention still needs to be paid to the import problems which is involved in the startup of Ganga. The situation in 6.3 is a lot better than previous releases but how classes end up imported and exposed to the GPI still needs some cleaning up imo.

rob-c:
After looking at this it appears the stage at which the class is imported is wrong as this will trigger a new set of config options to be constructed. The new class should only be imported during startup before bootstrap finishes.

egede:
@rob-c  The application is added into the system in `python/Ganga/Runtime/plugins.py` in what appears to be an identical way to the ROOT application.

rob-c:
@egede I'd have to do more digging if I get the time but this is overloading the plugins as I'm seeing:

```
gpi = <module 'Ganga.GPI' from '/home/ci-01/jenkins/workspace/PR builder/python/Ganga/GPI/__init__.py'>

    def test_job_copy(gpi):
        """Test that a job copy copies everything properly"""
    
        j = gpi.Job()
>       j.application.exe = "sleep"

python/Ganga/test/GPI/TestJob.py:70: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <<Ganga.Lib.Notebook.Notebook.Notebook object at 0x38f8f50> PROXY at 0x3436550>
x = 'exe', v = 'sleep'

    def _setattr(self, x, v):
        'This is the setter method in the Proxied Objects'
        #logger.debug("_setattr")
        # need to know about the types that require metadata attribute checking
        # this allows derived types to get same behaviour for free.
        raw_self = stripProxy(self)
        p_Ref = raw_self
        if p_Ref is not None:
            if not isclass(p_Ref):
                class_type = type(p_Ref)
            else:
                class_type = p_Ref
        else:
            class_type = p_Ref
    
        if x == implRef and not isinstance(v, class_type):
            raise AttributeError("Internal implementation object '%s' cannot be reassigned" % implRef)
    
        elif not raw_self._schema.hasAttribute(x):
            from Ganga.GPIDev.Lib.Job.MetadataDict import MetadataDict
            if hasattr(raw_self, 'metadata') and isType(raw_self.metadata, MetadataDict):
                if x in raw_self.metadata.data:
                    raise GangaAttributeError("Metadata item '%s' cannot be modified" % x)
    
            if x != implRef:
>               raise GangaAttributeError("Can't assign '%s' as it does NOT appear in the object schema for class '%s'" % (x, getName(self)))
E               GangaAttributeError: GangaAttributeError:  ("Can't assign 'exe' as it does NOT appear in the object schema for class 'Notebook'",)
```

This is because the default application object in this branch is not an `Executable`.
(I'd suggest trying to import this either much later or much earlier and see if the situation gets better).

There is now a horrible failover between that if something brings down Ganga in a horrible way in a test the bootstrap isn't cleaned up.
This I think is responsible for 90% of the failed tests due to the config system not being editable as the bootstrap has `finished`.
This comes from us playing around with the global state too much. It would be _much_ nicer to completely destroy all of the config system between tests and fire up a new one each time a test is run.
There are some difficulties with this but this imo is how we should have rebuilt the config system to behave.
I know there is a performance penalty in doing this but I'm convinced it's more than worth paying to keeps tests completely isolated.

Do the tests run fine locally or just struggle on the Jenkins box?

egede:
@rob-c The tests run fine locally as well as on the `travis` setup.

PR #922, opened 2017-01-08
Manifest update

mesmith75:
Update to the manifest template to include ```.py.template``` files.


mesmith75:
made a mess of this - I'll try again.

PR #923, opened 2017-01-08
Update the manifest template

### LINKED ISSUE #920: py.templates not installed
Release 6.3.1 installed via pip.

```
j = Job()
j.submit()
```

fails with

```
INFO     submitting job 2 to Local backend
ERROR    GangaException:  ('Error Finding script file: somewhere/.local/lib/python2.7/site-packages/Ganga/Lib/Localhost/LocalHostExec.py.template',)
ERROR    JobManagerError: error during submit
```

There are no `py.template` files installed at all.

mesmith75:
Fixes #920 . Update to include ```.py.template``` files when making the distribution for release.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include README.rst
+recursive-include python *.py.template


rob-c:
Is it worth including a readme as part of the release in the source?

mesmith75:
There is a README.rst which is already included. Do you mean something else as well? a regular README would be included automatically if it were there.

rob-c:
@mesmith75 Ah, never mind me, first day back :P was more of a question but if we're including one already then great :+1:

mesmith75:
In which case I shall merge this.

PR #924, opened 2017-01-09
Change policy for LHCb to allow even jobs with no input data to run a…

egede:
…t all Tier 1 sites. Leave machinery in place for the moment to allow change to be reverted through a configuration change.


mesmith75:
Do we want to unban these sites for the unit tests as well? In any case the unit lists in the tests will need updating as well.

egede:
@mesmith75 Yes, they should be unbanned for everything. Where is that list defined?

mesmith75:
I found three tests that explicitly ban sites, though I think the first is now defunct:

python/GangaDirac/old_test/Lib/RTHandlers/TestExeDiracRTHandler.py#L200

PR #926, opened 2017-01-20
Whitespace in interactive arguments

mesmith75:
Turns whitespaces in the arguments of interactive jobs to ```\ ``` so they are read correctly in the command line.


mesmith75:
On reflection this is not worth it - it isn't too difficult for the user to put their own slashes in.

Issue #883, opened 2016-11-29
LZ Dirac Job now working


alexanderrichards:
Am I doing something stupidly wrong with the new credentials system? I try the following from a fresh
```bash
$ pip install ganga
$ ganga
```

```python
[09:56:42]
Ganga In [2]: Job(backend=Dirac()).submit()
INFO     submitting job 6
INFO     job 6 status changed to "submitting"
INFO     Preparing Executable application.
INFO     Created shared directory: conf-47741432-9c2d-402b-afe6-081859985dc9
INFO     Preparing subjobs
INFO     Setting Command to be: '['echo', 'Hello World']'
INFO     submitting job 6 to Dirac backend
WARNING  Required credential [DiracProxy(group='lz', encodeDefaultProxyFileName=True)] not found in store
ERROR    Configuration of GLITE for LCG:
ERROR    File not found: /afs/cern.ch/sw/ganga/install/config/grid_env_auto.sh
WARNING  ---------- error in user/extension code ----------
WARNING  Traceback (most recent call last):
WARNING    File "/home/local/ganga/ganga/python/Ganga/GPIDev/Adapters/IBackend.py", line 184, in master_submit
WARNING      if b.submit(sc, master_input_sandbox):
WARNING    File "/home/local/ganga/ganga/python/GangaDirac/Lib/Backends/DiracBase.py", line 256, in submit
WARNING      return self._common_submit(dirac_script_filename)
WARNING    File "/home/local/ganga/ganga/python/Ganga/GPIDev/Credentials/__init__.py", line 33, in cred_wrapped_method
WARNING      cred = credential_store.create(cred_req, create=True)
WARNING    File "/home/local/ganga/ganga/python/Ganga/GPIDev/Credentials/CredentialStore.py", line 78, in create
WARNING      cred = query.info_class(query, check_file=check_file, create=create)
WARNING    File "/home/local/ganga/ganga/python/GangaDirac/Lib/Credentials/DiracProxy.py", line 37, in __init__
WARNING      super(DiracProxyInfo, self).__init__(requirements, check_file, create)
WARNING    File "/home/local/ganga/ganga/python/Ganga/GPIDev/Credentials/VomsProxy.py", line 35, in __init__
WARNING      super(VomsProxyInfo, self).__init__(requirements, check_file, create)
WARNING    File "/home/local/ganga/ganga/python/Ganga/GPIDev/Adapters/ICredentialInfo.py", line 104, in __init__
WARNING      self.create()
[...]

PR #928, opened 2017-01-23
Fix DiracProxy for Vanilla Dirac.

alexanderrichards:
This fixes https://github.com/ganga-devs/ganga/issues/883.
The solution is as suggested by @rob-c i.e. just attach a new Shell object. I don't know the implications of this in the wider credentials universe (although it looks fairly innocuous) but it seems to fix the issue for me. Lets see what the tests reveal.


mesmith75:
Can we merge this?

PR #934, opened 2017-01-27
Make diracLFNBase take the username from the DiracProxy

### LINKED ISSUE #931: Take username for Dirac LFNs from Proxy rather than local username
If users don't specify `DiracLFNBase` specifically then it is taken from the local username. However, this is unlikely to be the same username specified in the Voms Proxy attributes and what should be used through Dirac. It should be fairly easy to use this instead.

(Note that this **may** just be an LHCb problem - not sure!)

mesmith75:
Fixes #931 . This takes the `nickname` from the VomsProxy instead of the local username when making the diracLFNBase.
Should clear up problems for people running ganga on local systems with different usernames.

### DIFF
--- a/python/Ganga/GPIDev/Credentials/VomsProxy.py
+++ b/python/Ganga/GPIDev/Credentials/VomsProxy.py
@@ -110,19 +110,6 @@ class VomsProxyInfo(ICredentialInfo):
 
     @property
     @cache
-    def username(self):
-        """
-        This returns the username associated with a voms proxy on disk
-        """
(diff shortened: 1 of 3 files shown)


milliams:
`DiracProxyInfo` provides a `field()` method which is intended for exactly this situation. You should be able to add a method to `DiracProxyInfo` like:
```python
@property
@cache
def nickname(self):
    """
    Returns the nickname associated with the proxy
    """
    return self.field('nickname')
```

mesmith75:
Thanks, that is somewhat better!

mesmith75:
Any objections to merging?

PR #936, opened 2017-01-30
Change LHCb Backends import to its own Dirac

### LINKED ISSUE #930: Change of behaviour for Dirac.getOutputDataLFNS?
From 
https://twiki.cern.ch/twiki/bin/view/LHCb/FAQ/GangaLHCbFAQ#How_do_I_use_the_outputdata_from

It seems that this used to work:
```
ds = LHCbDataset()
for sj in j.subjobs.select(status="completed"): 
     ds.extend(sj.backend.getOutputDataLFNs())
new_job.inputdata = ds
```

but now doesn't as `getOutputDataLFNs` just returns a string rather than (at least) 'LFN:'+string and therefore the default for constructing a `DiracFile` doesn't set the LFN properly/at all. I'm not sure if this is a change in API or a bug but it should be easy to fix (one way or another) if it's the latter!

mesmith75:
Fixes #930. Sets the LHCb plugin to load its own version of Dirac, rather than vanilla.

### DIFF
--- a/python/GangaLHCb/Lib/Backends/__init__.py
+++ b/python/GangaLHCb/Lib/Backends/__init__.py
@@ -1,3 +1,3 @@
 from __future__ import absolute_import
-from GangaDirac.Lib.Backends.Dirac import Dirac
 from .Bookkeeping import Bookkeeping
+from .Dirac import Dirac


mesmith75:
retest this please

mesmith75:
Can I merge this?

PR #947, opened 2017-02-13
Fix exception in executable script

### LINKED ISSUE #946: Bug in generated exe-script.py file
I am submitting jobs using the `Executable()` application and the `Dirac()` backend.
For some (unknown) reason these are failing at one grid site.
Ganga is not reporting whatever the real error is, because there is a bug in the script ganga generates:
https://github.com/ganga-devs/ganga/blob/master/python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py#L163

This means the reported error is simply
```
> Traceback (most recent call last):
>   File "/scratch/plt/153162074/exe-script.py", line 28, in <module>
>     print('Exception occured in running process: ' + exe_cmd)
> TypeError: cannot concatenate 'str' and 'list' objects
```

mesmith75:
Fixes #946 
Put the executed command inside a repr so it prints out. Have the raised exception printed out instead of None. Also changed the subprocess.call to work properly.

### DIFF
--- a/python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py
+++ b/python/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py
@@ -160,9 +160,9 @@ if __name__ == '__main__':
         rc = subprocess.call(exe_cmd, env=my_env, shell=False)
     except Exception as x:
         rc = -9999
-        print('Exception occured in running process: ' + exe_cmd)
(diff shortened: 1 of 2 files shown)


rob-c:
Do we have a test job which attempts to launch a failing job where we can catch problems like this?

mesmith75:
We do not - I'll add one in

PR #948, opened 2017-02-13
Add DiracSplitter for using DiracFiles locally

### LINKED ISSUE #945: Local() backend doesn't cope with inaccessible LFNs
If my job has splitter `SplitByFiles(ignoremissing=False)` and uses the `Dirac()` backend, then if some of the input `DiracFile`s don't have any accessible replicas (because a site is down) I get an error.

If I use the `Local()` backend for the same job then there is no error, the subjobs corresponding to LFNs without accessible replicas are just run with no inputdata, i.e. `__GangaInputData.txt__` is empty.

It would be better if the `Local()` backend threw an error as the `Dirac()` one does.

mesmith75:
Fixes #945 
This adds a check for DiracFiles in the dataset when splitting and not using the Dirac backend. If they are all DiracFiles then it uses the `OfflineGangaDiracSplitter' in order to check for the availability of all the files.

I also found a bug in LHCbDataset  - I wasn't sure of the logic that the code was after but I think this fixes it.

### DIFF
--- a/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
+++ b/python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
@@ -83,7 +83,8 @@ class LHCbDataset(GangaDataset):
                 for this_file in files:
                     self.files.append(deepcopy(this_file))
             elif isType(files, IGangaFile):
-                self.files.append(deepcopy(this_file))
+                for this_file in files:
+                    self.files.append(deepcopy(this_file))
             elif isType(files, (list, tuple, GangaList)):
                 new_list = []
                 for this_file in files:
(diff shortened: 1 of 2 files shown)


python/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -83,7 +83,8 @@ def isFileTest(_file):
                 for this_file in files:
                     self.files.append(deepcopy(this_file))
             elif isType(files, IGangaFile):
-                self.files.append(deepcopy(this_file))
+                for this_file in files:
```
rob-c: Are we sure this does what we want for all IGangaFiles without subfiles?

mesmith75: ah, I think I may have misinterpreted what it was trying to do here.

mesmith75: I guess IGangaFiles will be in the GangaList below - so this is just for a single file?

mesmith75:
retest this please

mesmith75:
retest this please

mesmith75:
retest this please

PR #950, opened 2017-02-13
Fix for killing jobs and subjobs

### LINKED ISSUE #949: Killing a subjob updates the master status
Killing a single subjob updates the status of the master jobs and all the other subjobs to killed.

Only the statuses are updated - the other subjobs still run and complete.

mesmith75:
Fixes #949

### DIFF
--- a/python/Ganga/GPIDev/Lib/Job/Job.py
+++ b/python/Ganga/GPIDev/Lib/Job/Job.py
@@ -554,13 +554,13 @@ class Job(GangaObject):
         Transitions from to the current state are allowed by default (so you can updateStatus('running') if job is 'running').
         Such default transitions do not have hooks.
         """
-
         # For debugging to trace Failures and such
 
         fqid = self.getFQID('.')
         initial_status = self.status
         logger.debug('attempt to change job %s status from "%s" to "%s"', fqid, initial_status, newstatus)
-
+        print 'caller name:', inspect.stack()
+        print 'initial_status: ', initial_status, ' newstatus: ', newstatus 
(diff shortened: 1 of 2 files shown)


rob-c:
Would it be worth adding a quick test around this the code around the job state transitions is still a mystery to me.

mesmith75:
yes, I'll add some in.

mesmith75:
Something has gone very wrong with this. Killing the master job now fails with and error:
```python
Ganga In [3]: jobs(25).kill()
INFO     job 25 status changed to "running"
INFO     killing job 25
INFO     removing: /tmp/masmith/tmpO2z_RR
INFO     removing: /tmp/masmith/tmp4ZwtOW
INFO     removing: /tmp/masmith/tmpleuUBT
INFO     removing: /tmp/masmith/tmphZkNzT
INFO     removing: /tmp/masmith/tmpeyRj7k
INFO     job 25 status changed to "killed"
ERROR    failed to kill job 25: JobStatusError: forbidden status transition of job 25 from "killed" to "running"
ERROR    JobError: failed to kill job 25: JobStatusError: forbidden status transition of job 25 from "killed" to "running"
```

I assume it is the monitoring that is trying to return the job to running - do we need to put a lock around the killing?

mesmith75:
Updating the master status after the subjob statuses fixes the issue,

mesmith75:
I added a slight improvement to the test - just to make sure the job is running before killing it.

mesmith75:
If this is still approved I'll merge it.

Issue #959, opened 2017-02-27
Ganga fails to lock files on NFSv4 mount


stdietrich:
Hello,

I am currently debugging an issue with a user, who tries to run Ganga v602r3 on our batch farm BIRD/NAF at DESY.
Ganga tries to store the output file on an NFSv4 exported GPFS filesystem, however it fails with the following error message:

```
ERROR Error: [Errno 9] Bad file descriptor
ERROR A severe error occurred in the Repository 'prep.metadata':

Unable to launch due to underlying filesystem not working with unix
locks.Please try launching again with
[Configuration]lockingStrategy=FIXED to start Ganga without multiple
session support.
ERROR If you believe the problem has been solved, type "reactivate()"
to re-enable
INFO Ganga is now attempting to shut down all running processes
accessing the repository in a clean manner
INFO ... Please be patient!
ERROR Unable to disable Internal services, they may have already been
disabled!
```

I have traced this error back to the [dry_run_unix_locks](https://github.com/ganga-devs/ganga/blob/6c4a7cfb75a138b14e7f8c364faee3c27413e116/python/Ganga/Core/GangaRepository/SessionLock.py#L258) function in SessionLock.py.
The problematic part seems to be the following line:

`with open(test_file) as f_lock:`

The file is opened read-only, however this behaves different on NFSv3 and NFSv4.
On NFSv3 exports, opening the file as RO is working, while on NFSv4 the Bad File Descriptor error message is logged.
According to a post on the [NFS mailing list](http://www.spinics.net/lists/linux-nfs/msg18594.html), this is an expected behavior.

If the lock file is opened with Read/Write, it works on NFSv4.
Please let me know, if you need more information.

Regards,
Stefan

egede:
@stdietrich This is interesting. We have had problems on systems before with the `f_lock` but they have usually been related to badly configured systems. I have not come across this discussion of if it depends on how the file is opened before. Can you check on the system in question if the `f_lock` actually works if you open the file in read/write mode? As this particular file locking has caused trouble in other places and that there are performance issues related to it, we have to be a bit careful about just changing it.

stdietrich:
What do you mean, if Ganga works with f_lock in read/write or just the function?
The latter I already verified, because I created a sample script based on that particular function.
Running `with open(test_file, 'w') as f_lock:` works and the bad file descriptor error is gone.

For the first part, at least there I have to forward this to the user and let him try again.

egede:
[...]

PR #960, opened 2017-02-27
Try to change the lock files to be opened in read/write mode to have

egede:
Better compatibility with NFSv4.

### DIFF
--- a/python/Ganga/Core/GangaRepository/SessionLock.py
+++ b/python/Ganga/Core/GangaRepository/SessionLock.py
@@ -565,7 +565,7 @@ class SessionLockManager(object):
             # possible)
             fd = None
             try:
-                fd = os.open(self.fn, os.O_WRONLY)
+                fd = os.open(self.fn, os.O_RDWR)
                 if not self.afs:
                     fcntl.lockf(fd, fcntl.LOCK_EX)
                 os.write(fd, pickle.dumps(self.locked))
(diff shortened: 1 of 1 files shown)


python/Ganga/Core/GangaRepository/SessionLock.py
```diff
@@ -533,7 +533,7 @@ def session_read(self, fn):
             # This can fail (thats OK, file deleted in the meantime)
             fd = None
             try:
-                fd = os.open(fn, os.O_RDONLY)
+                fd = os.open(fn, os.O_RDWR)
```
rob-c: @egede I'm assuming this is needed here because we are touching the file to lock it on NFS?
Should `session_write` be modified to have read access as well? A user may not have seen this as a startup error as we don't write to lock files unless we're creating/saving a job.

egede: @rob-c This is an attempt to fix the problems mentioned in #959. I have indeed changed more places to open the file in r/w mode instead. Is there any performance issue related to opening a file in r/w mode relative to just read mode or just write mode?

rob-c: @egede I'm just curious why we have a problem opening a file for reading when we touch it compared to opening a file for write only when we read the lock state.
If there are known problems on filesystems we support I think it's probably cleaner to just open _everything_ r/w throughout the file as it reduces the questions going forward of 'why was this one not r/w'. I'm not sure there are is a performance difference in opening a file for r/w against ro but I've never looked tbh.

egede: @rob-c Let us first see if this solves the problem for the user. My guess is that a write operation is reported back to the server (and thus can be seen by another client using the same file system). A read operation might stay local to the client. This is in fact a bit how afs works like.

Issue #961, opened 2017-02-28
missing template


baoboa:
Maybe i did miss something but following the install instruction for pip
```python
virtualenv ~/gangaenv
source ~/gangaenv/activate
pip install ganga
```
installation are missing  lib/python2.7/site-packages/Ganga/Lib/Localhost/LocalHostExec.py.template

so the exemple provided
```python
j=Job()
j.submit()
```
fail

    ERROR    GangaException: Error Finding script file: /home/userXXX/SOFTWARE/gangaenv/lib/python2.7/site-packages/Ganga/Lib/Localhost/LocalHostExec.py.template
    ERROR    JobManagerError: error during submit

egede:
@baobao I thought this was fixed by the addition of the *.py.template wildcard in 

https://github.com/ganga-devs/ganga/blob/develop/MANIFEST.in

Are you somehow trying to install an old Ganga release rather than the latest one?

baoboa:
i set up a new virtualenv and i removed ~/.gangarc ~/gangadir

```
Successfully installed ganga-6.4.0 google-api-python-client-1.6.2 httplib2-0.10.3 ipython-1.2.1 oauth2client-4.0.0 pyasn1-0.2.3 pyasn1-modules-0.0.8 python-gflags-3.1.1 rsa-3.4.2 stomp.py-4.1.17 uritemplate-3.0.0
```

from that i got same error , is ganga-6.4.0 not the last version ?

egede:
@baobao Yes, it is indeed. So obviously we have somehow not solved the problem.

egede:
OK, I reproduce problem. There is a bit of an issue here in that we do not use pip for our own testing.

The real problem is that we use `setuptools` to make our pip distribution but the `MANIFEST.in` file is used by `distutils`. Should be fairly easy to fix.

baoboa:
@egede

is it pip specific ?
```bash
    wget https://github.com/ganga-devs/ganga/archive/develop.zip
    unzip develop.zip
    cd ganga-develop/
    python setup.py install
    ls lib/python2.7/site-packages/ganga-6.4.0-py2.7.egg/Ganga/Lib/Localhost/
```
```
    __init__.py  __init__.pyc  Localhost.py  Localhost.pyc
```
i didn't got template either

rob-c:
@baoboa Is there a reason not to use `git clone`?

egede:
[...]

PR #962, opened 2017-02-28
Enable setuptools to use the MANIFEST.in file

egede:


egede:
Fixes #961

egede:
Should be made into a release pretty soon as the pip installation at the moment is partially broken.

PR #967, opened 2017-03-01
 new version (CMAKE) of Bender application

VanyaBelyaev:


mesmith75:
Do we really want to completely remove the old stuff?

rob-c:
I think I agree with @mesmith75 It would be needed to have a new application name at least to ensure we don't torch people's existing `Bender` jobs on disk.
Plus there are issues with building the `GaudiExec` required target for some older `SetupProject` based apps so I would strongly prefer to see the old application(s) remain for a while.

VanyaBelyaev:
@mesmith75 @rob-c  I really see no need to keep old stuff for Bender.  New application works even for so ancient Bender release as v25r3  (dated December 2k+14).  (and anyhow I'll trigger a great cleanup of old released versions of Bender...)

VanyaBelyaev:
@mesmith75 @rob-c  and anyhow, if someone due to some unknown reasons wants to use old application,  he/she can use old Ganga...

rob-c:
@VanyaBelyaev unfortunately no. The old stuff __has__ to stay for backwards compatibility otherwise we will have to maintain a very old version of the Ganga code-base and risk splitting the user-base again like with 6.0.44 and 6.1.xx.
There will be users who open Ganga and expect jobs/scripts from 12months ago (or longer) to still be accessible.
If we suddenly update the latest version of Ganga overnight to change what the `Bender` application is/does there will be fallout from people who's jobs won't load and scripts won't work.

I'd personally much rather see new functionality added with a new application name rather than as a swap-out for old functionality.

If we're looking to completely change the Bender API then we at least should increment the version to something like `Version(3,0)` and then users are prevented from loading old/new jobs when they inevitably switch between Ganga versions to try and get their old ganga scripts working.


Aside: I see a lot of `from Ganga.GPI import ...` in this code. This simply has to go.
It's 100% incompatible with a work which has been done recently to fix the proxy system in Ganga.

egede:
We need some test cases for this. Both some unit tests and some nightly tests that illustrate a job submitted and actually running. Take a look in the `python/GangaLHCb/test` directory and ask for help if required.

VanyaBelyaev:
@egede  I'll make a try to create the test...

VanyaBelyaev:
@rob-c 
Hi Rob,  thank you for hint with Version(3,0)  I'll make a try to use it...
I've not understood the issue with GPI. What is wrong here?  
Initially I've relied here for "local" objects, but  It just fails.   How can I code this function without GPI ?

egede:
@VanyaBelyaev Just to add in my comment, that it is unacceptable to drop the support for the old Bender jobs from one Ganga release to the other. So I would also suggest that you simply create new application types. We can add a depreciation warning on the old types with the aim of eventually removing them.

VanyaBelyaev:
@egede   Hi Ulrik,   I still do not understand.  Bender users will have very clear instructions that if they want to use new version of BENDER Project (CMAKE-based)  they must to use new semantics in Ganga.  New semantics in Ganga is also applicable for all  "old" versions and it runs smoothly.  If someone insists to use  old versions of Bender project with old Ganga-applications, they need to stay to one of current versions of Ganga..  what is wrong here?

egede:
@VanyaBelyaev Remember that `Ganga` has a repository of jobs. When I start up the latest version of `Ganga` - that includes your changes I might have an old `Bender` job from before your changes. This job should load (require that the schema is compatible) and more importantly I should be able to copy/resubmit/change the job and it should still work. Otherwise you risk that a user need to manage parallel repositories with different `Ganga` versions.

VanyaBelyaev:
@egede   Hi Ulrik, I certainly see your point, but I think that  temporary inconvenience is more useful here, since it will speedup the transition period.

rob-c:
@VanyaBelyaev the version should increment if the api changes. All it serves for now is to prevent ganga loading an object which is incompatible. The version should always increase at least rather than risk getting confused about what version of an object is loaded from disk.

Hmm it appears the problem with passing objects back through to the GPI street exporting was never fixed. Using the GPI here should be safe as you're managing and releasing a proxied object. That could be a tricky big to fix.

VanyaBelyaev:
@egede @rob-c @mesmith75   ok, I'll try to find another name  (but I am not good in naming). In a meantime, may I know the recommended replacement for `"from Ganga.GPI import XXX"` ?  I've got the example from `prepareGaudiExec`.

mesmith75:
can we remove this one?

VanyaBelyaev:
@mesmith75  yes. please

PR #969, opened 2017-03-02
New BenderBox application

VanyaBelyaev:
New BenderBox application with three "tools"

- BenderModule
- bender
- ostap

it is corrected/updated version according to Ulrik's comments


python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,521 @@
...
+
+    ======
+    Usage:
+    ======
+    j.application = BenderBox ( tool = ostap( scripts   = ['path_to_script/the_script.py']  ,
+                                              arguments = [ '--no-canvas' ]  ,
+                                              commands  = [ 'print dir()' ]  ) )                                                 
+    """
+    _schema = Schema(Version(1, 0), {
+        'scripts'   : FileItem   (
```
rob-c: Should this be able to accept IGangaFile objects?

VanyaBelyaev: @robc Rob, I am not very familiar with internals of Ganga.  What is the difference?

egede:
test this please

egede:
Does anybody understand why the Jenkins PR builder is not picking this PR up?

egede:
@VanyaBelyaev Why a new pull request for each change. You have a problem with updating the old one?

VanyaBelyaev:
@egede  yeah, I am definitely not an expert in gitting.  I have problmes to update existing pull-requests and thats why I send new requests.  If someone explains me why update fails and how to solve the issue - it will be great.

```
[pclbitep01]~/cmtuser/TESTS/ganga% git push origin new-bender-2
To git@github.com:VanyaBelyaev/ganga.git
 ! [rejected]        new-bender-2 -> new-bender-2 (non-fast-forward)
error: failed to push some refs to 'git@github.com:VanyaBelyaev/ganga.git'
To prevent you from losing history, non-fast-forward updates were rejected
Merge the remote changes before pushing again.  See the 'Note about
fast-forwards' section of 'git push --help' for details.
[pclbitep01]~/cmtuser/TESTS/ganga% git push origin new-bender-3
```

egede:
@VanyaBelyaev The user of lower case classes in the GPI is not acceptable. All the Classes are uppercase while the objects and the functions are lowercase. That there would be both a upper and a lower case `bender`/`Bender` class would make matters even worse.

VanyaBelyaev:
@egede I do not want to introduce fake fancy names...  "B"ender, I hope, will disappear very soon...

python/GangaLHCb/Lib/Applications/BenderBox.py
egede: @VanyaBelyaev If it is an `IGangaFile` object, you would be able to specify this as a `DiracFile` or any other file type. Gives more flexibility to the user. The file will be downloaded to the worker node before your script is starting in any case. IF you need to do some manipulations on the file on the client first, it get a bit more messy though.

egede:
@VanyaBelyaev After I have created the pull request, and I subsequntly want to update it in the central repository, I just do

```
git push -u origin
```

which seems to work.

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @egede  Thank you!   It supposed to be a plain local file/script/code.  Might be due to a lack of imagination, but I can't see what one can gain by converting  python code into  DiracFile.  Properly-written code for Bender&bender&ostap  is usually very compact  [it is a main idea fro all these cases: very compact and well readable code for analysis].

VanyaBelyaev:
@egede Thanks. I'll make a try.     As for lower case names  - I really can't generate  names that are not very different from  ostap/bender. Do you have suggestions?

egede:
@VanyaBelyaev The old `Bender` application will take longer to disappear than you think and the naming conventions are not flexible -- so I am afraid that I insist.

VanyaBelyaev:
@egede  Hi Ulrik,  it is done

alexanderrichards:
> @egede yeah, I am definitely not an expert in gitting. I have problmes to update existing pull-requests and thats why I send new requests. If someone explains me why update fails and how to solve the issue - it will be great.
> 
> [pclbitep01]~/cmtuser/TESTS/ganga% git push origin new-bender-2
> To git@github.com:VanyaBelyaev/ganga.git
>  ! [rejected]        new-bender-2 -> new-bender-2 (non-fast-forward)
> error: failed to push some refs to 'git@github.com:VanyaBelyaev/ganga.git'
> To prevent you from losing history, non-fast-forward updates were rejected
> Merge the remote changes before pushing again.  See the 'Note about
> fast-forwards' section of 'git push --help' for details.
> [pclbitep01]~/cmtuser/TESTS/ganga% git push origin new-bender-3

Looks like the remote branch was updated and your local branch does not reflect this change. you would need to merge the remote changes first before you push.

alexanderrichards:
retest this please

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
+# -*- coding: utf-8 -*-
```
mesmith75: I think this was asked elsewhere - are these necessary?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+# =============================================================================
+## @file BenderBox.py
+#  Bender application for Ganga  (CMAKE-based version)
+#  Actually there are three applications under the single hat
+#  - BenderModule : classic Bender application: use a full power of bender 
+#  - BenderRun    : execute scripts in the context of ``bender'' environment
+#  - OstapRun     : execute scripts in the context of ``ostap'' environment
+#  @author Vladimir ROMANOVSKY Vladimir.Romanovskiy@cern.ch
+#  @author Vanya BELYAEV  Ivan.Belyaev@itep.ru       
+# =============================================================================
```
mesmith75: Do we need the commented part and the docstring?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+
+    ======
+    Usage:
+    ======
+    
+    j.application = BenderBox ( tool = BenderModule ( module = 'the_path/the_module.py' , events = 1000 ) )
+
+    """    
+    _schema = Schema(Version(1, 0), {
+        'module'    : FileItem   (
```
mesmith75: I think this was also commented elsewhere - Should these be GangaFileItem and IGangaFile ?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+USERMODULE.run({events})
+# =============================================================================
+# The END
+# =============================================================================
+"""
+
+    ## Returns file names which should be copyed to shared area
+    def inputfiles(self):
+        "Returns file names which should be copyed to shared area"
+        return [LocalFile(self.module.name),]
```
mesmith75: what is the comma for?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+    > bender [ scripts [ scripts ...  --no-color [ arguments ] --import [ imports [ imports [ ... --no-castor --import=data.py --batch   [ --command  [ commands [ ...
+    
+    ======
+    Usage:
+    ======
+    j.application = BenderBox ( tool = BenderRun ( scripts   = ['path_to_script/the_script.py']  ,
+                                                   commands  = [ 'print ls()' ]  ) )                                                 
+    """
+    _schema = Schema(Version(1, 0), {
+        'scripts'   : FileItem   (
```
mesmith75: Should this be a GangaFileItem ?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def configure(self, masterjobconfig):	
+        self.options = self.tool.inputfiles()
+        return (None,None)
+
+    # =======================================================================
+    def getWNPythonContents(self):
+        """Return the wrapper script which is used to run Bender on the WN
+        """
+        from ..RTHandlers.GaudiExecRTHandlers import GaudiExecDiracRTHandler
```
mesmith75: Why the `..`? Should this rather be `GangaLHCb.Lib.RtHandlers.GaudiExecRTHandlers` ?

python/GangaLHCb/Lib/Applications/BenderBox.py
alexanderrichards: These are not necessary unless the code contains non-ASCII (unicode) characters. This I suspect was put in by default by the IDE

VanyaBelyaev: it is fine.

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: does it destroy Ganga?

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: no need

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: it does not harm

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: no

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @alexanderrichards  for sure it is not nesessary. But if I write my name in comments using my language, this line will be very usefull....  anyhow it does not harm..

alexanderrichards: > anyhow it does not harm..

Agreed, this is kind of what I was getting at by saying I was guessing that your IDE put it in automatically. As you say it doesn't matter either way, I was attempting to reassure @mesmith75 :wink:

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
+# -*- coding: utf-8 -*-
+# =============================================================================
+## @file BenderBox.py
```
alexanderrichards: General comment probably to the other  @ganga-devs/core or @egede  (I only put it at this line as this was the first instance). There are many PEP8 violations in this PR. How strictly are we requiring PEP8 compliance these days?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+        'params'    : SimpleItem (
+        defvalue  = {} ,
+        typelist  = ['dict', 'str', 'int', 'bool', 'float'],
+        doc       = """The dictionary of parameters to be forwarded to ``configure'' method of the supplied Bender module""") , 
+        'events'    : SimpleItem ( defvalue=-1, typelist=['int'], doc= "Number of events to process"),
+        })
+    _category = 'BenderTool'
+    _name     = 'BenderModule'
+    
+    layout    = """#!/usr/bin/env python
```
alexanderrichards: just a suggestion but I would be tempted to use the dedent function to allow this to be indented a bit.

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+#  @endcode
+#  or use the temporary directory 
+#  @code
+#  j = Job ( application = prepareBender ( 'v30r1' , use_tmp = True  ) ) 
+#  @endcode
+#  One can also use the configuration parameters:
+#  @code
+#  j = Job ( application = prepareBender ( 'v30r1' , module = 'the_path/the_module.py' , params = { ...} ) ) 
+#  @endcode
+def prepareBender ( version  , path = '$HOME/cmtuser', use_tmp = False , **kwargs ) :
```
alexanderrichards: I don't understand the need for this function. Maybe this is the key part of the PR? My question would simply be how does this differ from what the prepare step of `Bender` already does or could be made to do? hopefully someone more LHCb minded like @mesmith75 understands this?

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+            scripts   = [ os.path.join ( f.subdir , os.path.basename ( f.name ) ) for f in self.scripts ] , 
+            arguments = self.arguments  ,
+            command   = self.commands    
+            )
+        return the_script
+
+## import the base class 
+from .GaudiExec import GaudiExec
+# =============================================================================
+## @class BenderBox
```
alexanderrichards: In general I agree with @mesmith75 that these commented out lines add clutter, especially since it's mostly repeated in the docstrings anyway. A few lines for authorship credit (of course git will record your contribution anyway) are fine but I fail to see what the rest is adding?

python/GangaLHCb/Lib/Applications/BenderBox.py
alexanderrichards: For the record: [textwrap.dedent](https://docs.python.org/2/library/textwrap.html#textwrap.dedent)

python/GangaLHCb/Lib/Applications/BenderBox.py
mesmith75: I think this is the Bender version of prepareGaudiExec. It checks out a clean version of Bender for you.

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: Yes but not having this a `File` object vs a `LocalFile` object raises more confusions. I would argue making it a `LocalFile` here makes the code more maintainable as we don't have to remember that this is not an `IGangaFile` in places.

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+#  - dictionary of parameters to be forwarder to <code>configure</code> method
+#  - number of event to process
+#
+#  Usage:
+#  @code 
+#  j.application = BenderBox ( tool = BenderModule ( module = 'the_path/the_module.py' , events = 1000 ) )
+#  @endcode 
+#  @author Vladimir ROMANOVSKY Vladimir.Romanovskiy@cern.ch
+#  @author Vanya BELYAEV  Ivan.Belyaev@itep.ru       
+class BenderModule(GangaObject):
```
rob-c: Is this ever used anywhere over being used inside `BenderBox`? Why no just directly implement this functionality there rather than having another class (I may have misunderstood the working here).

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+        param_string =  ',params=%s' % self.params if self.params else ''
+        
+        the_script      = self.layout.format (
+            datafile    = data_file ,
+            modulename  = module_name ,
+            paramstring = param_string ,
+            events      = self.events
+            )
+
+        from GangaLHCb.Lib.RTHandlers.RTHUtils import getXMLSummaryScript
```
rob-c: imports should be done at a global level

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: I disagree I think a `LocalFile` here is preferred.

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: I think it would be easier to just implement a common short API between both `BenderModule` and `BenderRun` that would be something like `BenderOptions` which reduces the total amount of classes and I think could reduce the amount of code. Just a suggestion but I'm not overly convinced breaking this out into many classes helps here.

rob-c: That or `BenderModule` and `BenderRun` could both directly inherit from `GaudiExec` directly and share a common static method to `configure`/`getWNPythonContents`. I just keep feeling there are 3 classes doing the work of 2 here and adding an extra `_category` to the plugins is not the way we probably want to go. (There has been talk of removing this in the past from what I remember).

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c  It allows to write compact and readable code...

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c  Rob, it will just increase number of lines and code duplication....

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c  why??? 

as for LocalFile -  ok - I'll change it to ... but I really do not understand *any* gain here...

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+    j3 = Job()
+    j3.application = BenderBox( directory = '$HOME/cmtuser/BenderDev_v30r1' ) 
+    j3.application.tool = OstapRun ( scripts = [ 'path_to_script/the_script.py' ]  )
+
+    Note:
+    1. The ``scripts'' are executed within ``ostap'' context
+    2. ``--no-color'' and ``--batch'' options are added automatically
+
+    """
+    _schema = Schema(Version(1, 0), {
```
VanyaBelyaev: @rob-c sorry, I do not understand you..

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c why???

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+#  @code
+#  j = Job ( application = prepareBender ( 'v30r1' , module = 'the_path/the_module.py' , params = { ...} ) ) 
+#  @endcode
+def prepareBender ( version  , path = '$HOME/cmtuser', use_tmp = False , **kwargs ) :
+    """Prepare Bender application
+    >>> j = Job ( application = prepareBender ( 'v30r1' ) )
+    One can specify the directory explicitely:
+    >>> j = Job ( application = prepareBender ( 'v30r1' , path = '$HOME/mydir' ) ) 
+    or use temporary directory: 
+    >>> j = Job ( application = prepareBender ( 'v30r1' , use_tmp = True  ) )
```
VanyaBelyaev: @rob-c yes, it is needed, it allows to write much more compact&readable  code...

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,520 @@
...
+    or use temporary directory: 
+    >>> j = Job ( application = prepareBender ( 'v30r1' , use_tmp = True  ) ) 
+    One can also use the configuration parameters:
+    >>> j = Job ( application = prepareBender ( 'v30r1' , module = 'the_path/the_module.py' , params = { ...} ) ) 
+    """
+    from .GaudiExecUtils import prepare_cmake_app
+    if use_tmp :
+        import tempfile
+        path = tempfile.mkdtemp ( prefix = 'GANGA_BENDER_%s_' % version )
+        from Ganga.Utility.logging import getLogger
```
VanyaBelyaev: @rob-c I do not understand. why?

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: I'm really not seeing it. There should be at most some 4 or 5 lines to correctly inherit from the base class and expand it and the methods of `BenderBox` can be kept static. (This or the code in `BenderBox` can at least be cleaned up.

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: `prepareBender(...` vs ` prepareGaudiExec('Bender'...` Given this _should_ be done once I don't think there is much saved in adding a wrapper around doing this.

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: To be consistent with what we've tried to move to with regard to code standards across the project. imports should always be done at the global level in the file unless there's a good reason it can't in python (at least from numerous discussions about standards I've been told this is the case).

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c 
 - it also creates (when it can deduce) BenderModule objects
 - it also allows to goes to temporary directory

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: e.g. 
https://github.com/ganga-devs/ganga/blob/develop/python/GangaLHCb/Lib/Backends/Dirac.py#L15

The schema can be inherited from and then expanded as appropriate there is no need to duplicate all that the base class does here.

python/GangaLHCb/Lib/Applications/BenderBox.py
rob-c: We're trying to get closer to PEP8 (it takes time to cleanup/port existing code) but new code we've been wanting to be better:
https://www.python.org/dev/peps/pep-0008/#imports

Also this file should probably correctly pull in `from __future__ import relative_import` if using `from ..`

python/GangaLHCb/Lib/Applications/BenderBox.py
alexanderrichards: > I think this is the Bender version of prepareGaudiExec. It checks out a clean version of Bender for you.

Ok this is why I left it to you guys, I'm showing my age already here as I remember the good old days when we use to just bind an application instance directly as the input parameter to a job rather than using what to me looks like a factory function. I wasn't even aware of this `prepareGaudiExec`. My question however I think still stands and applies equally to this function as well. Are they doing something that the prepare step, namely 
```python
Bender().prepare()
```
does not do?

python/GangaLHCb/Lib/Applications/BenderBox.py
alexanderrichards: it's a PEP8 thing: [PEP8 imports](https://www.python.org/dev/peps/pep-0008/#imports)

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c  How can I add entries  after inherit_copy ? How can I make some entries as 'hidden' ?

VanyaBelyaev:
will you be happy this time?

alexanderrichards:
retest this please

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,528 @@
...
+# =============================================================================
+import os
+import tempfile 
+from   Ganga.GPIDev.Lib.File.File                       import ShareDir  
+from   Ganga.GPIDev.Schema                              import Schema    , Version , SimpleItem , GangaFileItem
+from   Ganga.Utility.logging                            import getLogger
+from   Ganga.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
+
+from   GangaLHCb.Lib.Applications.GaudiExecUtils        import prepare_cmake_app
+from ..RTHandlers.GaudiExecRTHandlers                   import GaudiExecDiracRTHandler
```
rob-c: this should require `from __future__ import absolute_import`

python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -0,0 +1,528 @@
...
+    if use_tmp or not path :
+        path = tempfile.mkdtemp ( prefix = 'GANGA_' )
+        logger.info ( 'Bender application will be prepared in the temporary directory %s' % path )
+        
+    the_path = prepare_cmake_app ( 'Bender' , version , path )
+    #
+    scripts = kwarsg.pop('scripts', [] )
+    if script : scripts.append ( script )
+    #
+    from Ganga.GPI import BenderRun as _BR  ## black magic
```
rob-c: fyi this _magic_ it's because you're configuring a proxied object which has extra wrappers around the get/set/str and a few other methods over an un-wrapped `GangaObject`.
If you return a properly configured object you can also do an `return addProxy(_BR(...))` but this is slightly less intuitive to what these functions are doing (just reducing the amount of code the user has to type).

rob-c:
It might be nice to have some basic tests against this as well, something similar to: https://github.com/ganga-devs/ganga/blob/develop/python/GangaLHCb/test/GPI/TestGaudiExec.py and https://github.com/ganga-devs/ganga/blob/develop/python/GangaLHCb/test/GPI/TestExternalGaudiExec.py should be enough

python/GangaLHCb/Lib/Applications/BenderBox.py
VanyaBelyaev: @rob-c thank you for clarification

VanyaBelyaev:
hopefully this time it is ok

VanyaBelyaev:
as for tests, I have real examples and applications. I'll make a try to prepare some "tests" , but it is not for this week. Can we postpone tests and make a first release without?

mesmith75:
I think we can add the tests at a later date.
@egede Are you also happy for this to be merged?

mesmith75:
test this please

mesmith75:
test this please

egede:
@mesmith75 Yes, I think you can go ahead and merge.

VanyaBelyaev:
@egede   Hi Ulrik, I'll try to do all my best within a week or two to prepare some tests, but I'll need some help from Gangs experts.  I've tried to follow what 'ganga -h' says about test, but it it still rather unclear for me.

mesmith75:
Great - I'll merge this and release 6.5.0. There is an issue now as a reminder for the tests.

PR #971, opened 2017-03-05
Choose LHCbDIRAC version

egede:


egede:
Will enable #965 to be implemented by setting version to ```prod```.

egede:
We need to agree here if we follow the `LHCbDIRAC` developers recommendation of using `prod` here as the default version.

PR #976, opened 2017-03-07
Get a working version of the nightly tests

egede:


python/GangaLHCb/test/GPI/TestExternalGaudiExec.py
```diff
@@ -110,6 +110,10 @@ def testSubmitJobComplete(self):
         from Ganga.GPI import jobs
 
         j = jobs[-1]
+        j = j.copy()
+        j.unprepare()
+        j.prepare()
+        j.submit()
```
rob-c: This should not be required

egede: @rob-c No, but we have introduced another bug which it is not the job of this test to expose.

python/GangaLHCb/testlib/setup_env.sh
```diff
@@ -1,2 +1,6 @@
 source /cvmfs/lhcb.cern.ch/lib/LbLogin.sh
-SetupProject ganga
+export GANGA_CONFIG_PATH=GangaLHCb/LHCb.ini
+export GANGA_SITE_CONFIG_AREA=/cvmfs/lhcb.cern.ch/lib/GangaConfig/config
+ls -l .globus
+lhcb-proxy-init
+#SetupProject ganga
```
mesmith75: Has something changed that makes this necessary now?

egede: Before there was a ```SeupProject``` command here that effectively was used to get the dependencies included and setting the environment variables. The ```lhcb-proxy-init`` is because I have no idea about how the proxy was created before ...

python/GangaLHCb/test/GPI/TestExternalGaudiExec.py
rob-c: This concerns me a bit as the persistency here is crucial to 90% of the tests I've written. Nothing should have changed in this regard. You're still accessing the job from the repo, is the next step of sleep until hanging? In which case this is likely not due to persistency problems.

rob-c: I wonder what an `if not j.status in ['completed', 'failed']:` check would do to help in this case as I've seen the `wait_until` scripts sometimes struggle in odd ways with already completed jobs.

egede: There seems to be a problem in that the SharedPath object, when read back from disk, is assigned a new directory name. So if I prepare a job, quit Ganga and then start again, the application object points to an empty directory and the job fails. So this is a potentially serious problem (that we need to understand), but not the propose of this test.

rob-c: I'd prefer to note this is a hack I this case. Is this related to why the payload is returned with status 127? It would be good to fix that. Is it really a persistency issue with prepared objects there should be an independent test for that.

egede: @rob-c The PR #980 has a test case implemented that shows two problems with the shared area implementation. Discussion on that topic should continue there.

python/GangaLHCb/testlib/setup_env.sh
egede: Will it be possible to approve this PR?

mesmith75:
@egede I'm a little with Rob with regard the persistency. I agree this is not a test for that - it should be explicitly tested elsewhere but does that mean you would like to get rid of it from all tests?

In any case is it better to create a new job here rather than loading, unpreparing, preparing and submitting?

rob-c:
@alexanderrichards I did some _clever_ stuff to do with the prepared state for this one so I was tying to make sure it works (I'm surprised that the job may not be persisting correctly at a lower level).

egede:
Implemented it as a new job.

PR #977, opened 2017-03-07
Add option to not need an AFS token

### LINKED ISSUE #944: AFS token check
We have been asked to add a configuration to prevent the check for an AFS token. Some sites appear to have a weird setup where they use AFS but don't have tokens.

mesmith75:
Fixes #944 . Config option to not need an AFS token when running on AFS.


egede:
@alexanderrichards At IHEP in Beijing they have some bizarre setup where they use an afs file system without afs tokens. We asked them the question, but they said it was completely outside their control. So this is a hack, but a rather innocent one.

PR #980, opened 2017-03-08
Solve problems with the shared area functionality

egede:


egede:
@rob-c This test demonstrates two problems with the shared area implementation.

- When reloading a job, a `prepared` has the wrong path.
- When a job is `unprepared` and has the last link to a shared area, it is not deleted. 

The first of these is a serious bug.

mesmith75:
The nightlies started failing for build 610 (they failed on 609 but nothing had changed and it looks like that was because of Dirac problems).
The changes were:
#947 Fixing the executable RT script and adding a failing test for it
#956 Fixing getOutputDataAccessURLs in Dirac
#701 The XML parsing for SimpleItem

I would hazard a guess at the last being the problem.

rob-c:
There's an unbalanced getter/setter in the `SharedDir`, I'll post a quick fix in a bit

rob-c:
retest this please

rob-c:
I've merged a copy of this into #981 where the only problem is unprepare and reference counting. Also fyi, there shouldn't be references to `self.attribute` in py.test classes as the class doesn't persist between tests.

egede:
@rob-c I am confused about what you have done with Git here, but can you make sure that both the tests from #980 and the fix from #981 will be merged into `develop`.

rob-c:
Yeah, the shared repository isn't being persisted at all, this is why the reference counting is completely gone... and this means it's likely been destroyed for all users currently using it!

rob-c:
@egede I was trying to rebase and was confused about why the test was still failing (turns out `None` is a valid directory).
This has highlighted a serious shortcoming in the test suite, I thought the prep tests were ported over a long time ago?

rob-c:
The prep.metadata repo is dubbly flushed, the first time the data is sent to disk I see it in `/afs/cern.ch/user/r/rcurrie/gangadir/repository/rcurrie/LocalXML/6.0/prep.metadata/0xxx/0/data~` and the 2nd time it's overwritten with an empty dict `/afs/cern.ch/user/r/rcurrie/gangadir/repository/rcurrie/LocalXML/6.0/prep.metadata/0xxx/0/data`
Something in the shutdown is destroying the prep.metadata dict when it shouldn't.

rob-c:
Also, unused shared dir are only cleaned up on shutdown, we never implemented this as when the count gets to zero destory the data at once

PR #981, opened 2017-03-09
Fixing ShareDir persistency... schema should support setters

rob-c:
This PR touches the `ShareDir` class and does 2 things:

 1) Fixed the mis-use of the `._name` attr -> `._real_name`
 2) Added a global and schema based setter to handle storing the persisted values from XML better

As always feel free to fix/modify but this should address (at least some of) the problems correctly highlighted in #980 .

### DIFF
--- a/python/Ganga/test/GPI/TestShared.py
+++ b/python/Ganga/test/GPI/TestShared.py
@@ -4,6 +4,8 @@ from os import path
 
 from Ganga.testlib.GangaUnitTest import GangaUnitTest
 
+from Ganga.Core.GangaRepository import getRegistry
+
 
 class TestShared(GangaUnitTest):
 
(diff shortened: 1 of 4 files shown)


rob-c:
I've fixed and expanded the tests from #980. I've also removed a load of spaghetti code from the prep shutdown that I've never understood or wanted. (It appears to now be causing errors which are fixed by removing it so it is assigned to oblivion).
I've added some tests which actually look at the counter explicitly as well as checking that a folder is removed on the count reaching zero. This all worked on my lxplus setup although I'm thinking there must be a corner case I've possibly missed.

python/Ganga/test/GPI/TestShared.py
```diff
@@ -0,0 +1,124 @@
...
+        assert(j.application.is_prepared != None)
+
+        TestShared.shared_area_location == j.application.is_prepared.path()
+
+        j.unprepare()
+
+        assert(j.application.is_prepared == None)
+
+
+    def test_D_Cleanup(self):
```
egede: This method seems to have a poor name now.

Issue #963, opened 2017-03-01
Change the default BkQuery dqflag to be ['OK',UNCHECKED']


egede:

mesmith75:
Why do we want this? As a casual user would one not expect the files returned to only be checked ones?

egede:
If I understand it correctly, the label UNCHECKED can persist in principle forever, and it is the BAD data that you do not want. I was explicitly asked to make this change, but will ask on the mailing list to ensure that there is a general agreement.

egede:
Agreed that this change should *not* be made. Would be good to keep new test cases though.

PR #982, opened 2017-03-10
Adjust bk queries

egede:


python/GangaLHCb/test/GPI/TestBookkeeping.py
```diff
@@ -0,0 +1,108 @@
...
+            dqflag=['OK', 'UNCHECKED'],
+            path="/LHCb/Collision16/Beam6500GeV-VeloClosed-MagDown/Real Data/91000000/RAW",
+            type="Path"
+        )
+        dataset = bkq.getDataset()
+
+        assert isinstance(dataset, LHCbDataset)
+        assert len(dataset) > 0
+
+    def test_date_query(self):
```
egede: This test doesn't work. Can somebody help me with getting a working query (or is there a general problem here),

mesmith75: @egede The path you are supplying depends on the BK query type: see the examples in https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/blob/master/LHCbDIRAC/Interfaces/API/DiracLHCb.py#L194

So for this you need to get rid of the beam conditions part of the path:
/LHCb/Collision16/Real Data/91000000/RAW

As you are requesting the whole year's data the query takes a little while - might it be worth restricting the range?

Also, what is this PR doing?

mesmith75: Although having said that I cannot get his to work at all even with LHCbDirac `dirac-bookkeeping-get-files` on the command line so maybe there is an issue there?

egede:
Addresses #963

egede:
Hi Ulrik,
I fixed the method: https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/merge_requests/299
We need a new LHCbDIRAC patch release.

Let me know, if you need more test case:

```
from DIRAC.Core.Base import Script
Script.parseCommandLine( ignoreErrors = True )

d = DiracLHCb()
#res = d.bkQueryRunsByDate('/LHCb/Collision16/Beam6500GeV-VeloClosed-MagDown/Real Data/Reco16/Stripping26/90000000/EW.DST','2016-08-20','2016-08-22',dqFlag='OK',selection='Runs')
res = d.bkQueryRunsByDate('/LHCb/Collision16//Real Data/Reco16/Stripping26/90000000/EW.DST','2016-08-20','2016-08-22',dqFlag='OK',selection='Runs')
#res = d.bkQueryRunsByDate('/LHCb/Collision15/Beam6500GeV-VeloClosed-MagDown/Real Data/Reco15a/90000000/RDST','2015-09-06','2015-09-08',dqFlag='OK',selection='Runs')
print res
```

Cheers,
Zoltan

mesmith75:
@egede Does this need any more alterations?

egede:
Yes. some more work is required here.

python/GangaLHCb/Lib/LHCbDataset/BKQuery.py
```diff
@@ -257,14 +255,14 @@ def getDatasetMetadata(self):
         cmd = 'bkQueryDict(%s)' % self.dict
         try:
             value = get_result(cmd, 'BK query error.')
-        except GangaDiracError as err:
+        except GangaDiracError:
             return {'OK':False, 'Value': {}}
```
rob-c: Would it be better to put the string rep of the error into the value here?

mesmith75:
@egede What is the plan for this?

egede:
Abandoning.

PR #989, opened 2017-04-04
Set root/xroot default protocol for LHCb

### LINKED ISSUE #984: root/xroot default for LHCb
Request xroot/root as the default protocol for accessURL within LHCb. It is a pain for people otherwise with some sites not allowing outside access by default.

mesmith75:
Fixes #984 . I need to add some nightly tests for this for regular Dirac and LHCb Dirac.


mesmith75:
Can I merge this? I'll add some tests in another PR.

PR #990, opened 2017-04-07
Remove banned sites from splitter

### LINKED ISSUE #988: Ganga submits impossible jobs
The OfflineDiracSplitter should remove the SE associated with the banned sites from the possible sites to use in the splitter. Additionally the splitter should abort on submit when there are LFN with no valid replicas available, I'm not sure if there is a proper check for this or not.

This should stop users not knowing why jobs can't run because they've banned a site.
This saves the DIRAC servers wasting time trying to run a job that can't run.

(If this functionality is directly removed from the backend we can still can the extra args for the banned sites attribute)

mesmith75:
Fixes #988 . This removes the sites at the site<->SE mapping and removes the LFNs if they end up with no sites or raises an exception if `ignoremissing=False`

### DIFF
--- a/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
+++ b/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
@@ -158,6 +158,8 @@ def calculateSiteSEMapping(file_replicas, uniqueSE, site_to_SE_mapping, SE_to_si
         file_replicas (dict): This is the dictionary of LFN replicas with LFN as the key
         site_to_SE_mapping (dict): Dict which has sites as keys and SE as values
         SE_to_site_mapping (dict): Dict which has sites as values and SE as keys
+        bannedSites (list) : List which has the sites banned by the job
+        ignoremissing (bool) : Bool for whether to continue if an LFN has no available SEs
 
     Returns:
         site_dict (dict): Dict of {'LFN':set([sites]), ...}
@@ -340,6 +342,7 @@ def OfflineGangaDiracSplitter(_inputs, filesPerJob, maxFiles, ignoremissing, ban
         filesPerJob (int): Max files per jobs as defined by splitter
         maxFiles (int): This is the max number of files per subset(subjob)
         ignoremissing (bool): Should we ignore missing LFN
+        bannedSites (list): List of banned sites of which the SEs will not be used
 
     Yields:
         dataset (list): A list of LFNs for each subset(subjob)
--- a/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
+++ b/python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
@@ -68,7 +68,6 @@ def addToMapping(SE, site_to_SE_mapping):
         SE (str): For this SE we want to determine which sites we can access it
         site_to_SE_mapping (dict): We will add sites which can find this SE to this dict with the key (SE) and value (sites(list))
     """
-
     result = wrapped_execute('getSitesForSE("%s")' % str(SE), list)
     site_to_SE_mapping[SE] = result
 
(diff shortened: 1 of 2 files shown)


python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -145,7 +144,7 @@ def generate_site_selection(input_site, wanted_common_site, uniqueSE, site_to_SE
     return req_sitez
 
 
-def calculateSiteSEMapping(file_replicas, uniqueSE, site_to_SE_mapping, SE_to_site_mapping):
+def calculateSiteSEMapping(file_replicas, uniqueSE, site_to_SE_mapping, SE_to_site_mapping, bannedSites, ignoremissing):
```
rob-c: Can you add what these control to the doc string

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -317,7 +331,7 @@ def badLFNCheck(bad_lfns, allLFNs, LFNdict, ignoremissing, allLFNData):
 
 # Actually Do the work of the splitting
 
-def OfflineGangaDiracSplitter(_inputs, filesPerJob, maxFiles, ignoremissing):
+def OfflineGangaDiracSplitter(_inputs, filesPerJob, maxFiles, ignoremissing, bannedSites=[]):
```
rob-c: please add this to the doc string

rob-c:
Aside from some minor doc changes, looks good 👍

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -189,6 +187,13 @@ def calculateSiteSEMapping(file_replicas, uniqueSE, site_to_SE_mapping, SE_to_si
     while len(site_to_SE_mapping) != maps_size:
         time.sleep(0.1)
 
+    for iSE in site_to_SE_mapping.keys():
```
rob-c: I think you can remove the `.keys` here (?) but not essential

rob-c: Actually removing keys is potentially good for setups with many sites as the lookup is faster

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -207,10 +212,19 @@ def calculateSiteSEMapping(file_replicas, uniqueSE, site_to_SE_mapping, SE_to_si
     for _lfn, sites in SE_dict.iteritems():
         site_dict[_lfn] = set([])
         for _site in sites:
-            for _SE in site_to_SE_mapping[_site]:
-                site_dict[_lfn].add(_SE)
-
-    return site_dict
+            if _site in site_to_SE_mapping.keys():
```
rob-c: Again here as lookup in dict vs flat list has different speeds

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
mesmith75: I think the keys had to be there to have a copy of the iterator, otherwise the iterator gets altered and so there is an error.

rob-c: Ah, I missed the `del` command, OK, it's not likely this set of elements gets too big so might as well use this here

python/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
mesmith75: ok perhaps here I can get rid of it.

mesmith75:
retest this please

mesmith75:
Can I merge this?

PR #992, opened 2017-04-16
Allow submitting GaudiExec with only extraOpts

chrisburr:


rob-c:
@chrisburr I'm not sure I'm immediately seeing the benefit of this or is it equivalent to:
```
        else:
            if not self.extraOpts:
                raise ApplicationConfigurationError("No Opts File has been specified, please provide one!")
```
The former suggests we're treating the options and extraOpts the same which they aren't.

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -347,18 +347,19 @@ def getOptsFiles(self):
...
+            elif isinstance(this_opt, DiracFile):
+                pass
+            else:
+                logger.error("opts: %s" % self.options)
+                raise ApplicationConfigurationError("Opts file type %s not yet supported please contact Ganga devs if you require this support" % getName(this_opt))
+
+        if self.options or self.extraOpts:
             return self.options
         else:
             raise ApplicationConfigurationError("No Opts File has been specified, please provide one!")
```
egede: This line should now say `No options (as options files or extra options) has been specified. Please provide some.`.

chrisburr:
@rob-c I've modified this as @egede suggests. The function returns a list of options files, so it still makes sense to run the loop, even if it's skipped and an empty list is returned.

@egede I'm fairly sure it's okay. I've monkey patched this in to the current LHCb release of ganga and submitted ~40 jobs without any trouble.

python/GangaLHCb/Lib/Applications/GaudiExec.py
chrisburr: Done

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -347,21 +347,22 @@ def getOptsFiles(self):
...
-                    if not path.exists(full_path):
-                        raise ApplicationConfigurationError("Opts File: \'%s\' has been specified but does not exist please check and try again!" % full_path)
-                elif isinstance(this_opt, DiracFile):
-                    pass
-                else:
-                    logger.error("opts: %s" % self.options)
-                    raise ApplicationConfigurationError("Opts file type %s not yet supported please contact Ganga devs if you require this support" % getName(this_opt))
+        for this_opt in self.options:
+            if isinstance(this_opt, LocalFile):
+                ## FIXME LocalFile should return the basename and folder in 2 attibutes so we can piece it together, now it doesn't
```
rob-c: Would have been good to remove this as it no longer applies

rob-c:
@egede This has just bypassed the devlop branch and gone straight into master. This will probably cause problems unless the branches are re-synced

egede:
@rob-c I did not discover this. I reverted the change on master and I am now putting it in as a pull request on develop.

Issue #327, opened 2016-03-18
auto-resubmit broken


rob-c:
An LHCb user is seeing problem with I suspect auto-resubmit as they get errors to the effect of:

```
Ganga.Core.MonitoringComponent     : ERROR    Monitoring Error: local
variable 'try_resubmit' referenced before assignment
```

a few times per minute.

milliams:
This will happen `if len(j.subjobs) > 0` and `if s.info.submit_counter > config['MaxNumResubmits'] or s.status == "killed"` since the variable `try_resubmit` will never be set to anything. That whole function `resubmit_if_required` looks buggy. I'm not sure if it's broken indentation of the `if try_resubmit` and/or some other logic.

rob-c:
@milliams It sounds like you're best handling this issue then, I've not had a chance to look at the code yet.

milliams:
Sure, I'll take a look.

PR #996, opened 2017-04-18
Fix auto resubmission #327

chrisburr:
Applying this locally appears to fix the issue described in #327.

### DIFF
--- a/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -394,11 +394,9 @@ class CallbackHookEntry(object):
         # record the time when this hook has been run
         self._lastRun = 0
 
+
 def resubmit_if_required(jobList_fromset):
-
-    # resubmit if required
     for j in jobList_fromset:
-
         if not j.do_auto_resubmit:
             continue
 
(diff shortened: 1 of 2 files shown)


python/Ganga/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -394,39 +394,34 @@ def __init__(self, argDict, enabled=True, timeout=0):
...
+                continue
 
-            if try_resubmit:
-                if j.backend.check_auto_resubmit():
-                    log.warning('Auto-resubmit job %d...' % j.id)
-                    j.auto_resubmit()
+            # Only submit if jobs have failed and not a single subjob has
+            # exceeded config['MaxNumResubmits']
+            try_resubmit = n_failed > 0 and not any([
+                sj.info.submit_counter > config['MaxNumResubmits']
```
chrisburr: It might make sense to move this check to [here](https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/GPIDev/Lib/Job/Job.py#L1968) so a single rapidly failing subjob doesn't prevent any jobs from being resubmitted.

rob-c: @chrisburr This will stop a whole master job from re-submitting broken subjobs when 1 subjob is having problems will it not? (Apologies if I've mis-read this)
This would be bad if say 1 subjob is trying to access data which is not available at a given site due to downtime but the rest of the job takes a week to process.

chrisburr: I left it like this as it seemed to be the way it was currently intended to work so I wasn't sure if there was a good reason to do it this way. But I agree entirely (hence the comment), I'll move the filtering into `Job.py`.

rob-c: @chrisburr I'm just wondering slight;y if the `Job.py` is the best place to be accessing the config from the monitoring system. But if it works I'm happy 👍 
Good work for looking at this btw, as per #327 I didn't have a clue what the original code was doing and I think it needed a rewrite.

(As an aside)
I'm wondering what the _correct_ behaviour is for this, I simply don't know as I never used this.
Maybe we can introduce a 2nd max resubmit paramter into the configuration (https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/__init__.py#L218), one parameter for the max resubmit per subjob and one for the resubmit for a master job?

chrisburr: I think `Job._resubmit` could be refactored in some way to make this possible. Perhaps by splitting out a function that will submit a list of subjobs or something similar, though I'm not familiar enough with the core to know if this is actually a good idea. I've also just noticed it already accesses `config['resubmitOnlyFailedSubjobs']`.

(for the aside)
I think a `max resubmit` like option should be covered by the `MaxFracForResubmit` check, which also has the benefit of scaling with the number of jobs.

egede:
retest this please

mesmith75:
@chrisburr Thanks, this looks fine.

mesmith75:
retest this please

mesmith75:
If there are no objections let's merge this.

PR #1006, opened 2017-05-04
GaudiExec is_prepared after succesful build

### LINKED ISSUE #1004: Failed GaudiExec builds marked as prepared
It looks as if the GaudiExec jobs where the build has failed are being left in the prepared state so they subsequently can still be submitted. If the build fails they should be unprepared.

mesmith75:
Fixes #1004

### DIFF
--- a/python/GangaLHCb/Lib/Applications/GaudiExec.py
+++ b/python/GangaLHCb/Lib/Applications/GaudiExec.py
@@ -204,11 +204,11 @@ class GaudiExec(IPrepareApp):
         # file is unspecified, has a space or is a relative path
         self.configure(self)
         logger.info('Preparing %s application.' % getName(self))
-        self.is_prepared = ShareDir()
-        logger.info('Created shared directory: %s' % (self.is_prepared.name))
 
         this_build_target = self.buildGangaTarget()
 
+        self.is_prepared = ShareDir()
+        logger.info('Created shared directory: %s' % (self.is_prepared.name))
         try:
             # copy any 'preparable' objects into the shared directory
             send_to_sharedir = self.copyPreparables()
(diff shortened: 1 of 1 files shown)


python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -204,11 +204,11 @@ def prepare(self, force=False):
         # file is unspecified, has a space or is a relative path
         self.configure(self)
         logger.info('Preparing %s application.' % getName(self))
-        self.is_prepared = ShareDir()
-        logger.info('Created shared directory: %s' % (self.is_prepared.name))
 
         this_build_target = self.buildGangaTarget()
 
+        self.is_prepared = ShareDir()
+        logger.info('Created shared directory: %s' % (self.is_prepared.name))
```
egede: Should the lines not be moved to the end of the try clause just below - so that we only declare that it is prepared when the files are actually copied into their final destination.

rob-c: I thought the build target depended on the target dir existing (else you might end up with one being inconsistent, I forget the logic but just moving this seems potentially bad from the watching what file is created when) 
Surely a try around the build target which removes the prepared dir (and from the prep repo) once any exception is thrown is safer?

mesmith75: @rob-c Looking at the code I don't think the buildTarget requires the sharedDir to exist yet. The stuff in the try though does.

I agree that is_prepared should be removed in the exception though.

mesmith75: Although that is already done with the `unprepare` so I think this is sufficient as is.

PR #1007, opened 2017-05-05
Ask questions when copying the new .gangarc

### LINKED ISSUE #1003: Copy config on startup
Reminder to copy the gangarc generation into the startup.

mesmith75:
Fixes #1003 . This just asks for confirmation when copying the gangarc to a new file, both for generating a new gangarc and for version updates.

So far I have used the interactive value which is false for running scripts. Is this the default behaviour we want?


mesmith75:
Ok - if there is no explicit 'y' then it is taken as no

PR #1012, opened 2017-05-11
Add Dockerfile for IC Dirac UI.

alexanderrichards:


alexanderrichards:
yes we can do this or we can also name the dockerfile something else like lhcb_dockerfile etc. we would then have to manually specify the name of the file to build as Dockerfile is the default as you say.

Either way, if we plan to have a few then in makes sense to have them in a directory somewhere. We could have a Dockerfile in the root of the plugin directories i.e.

GangaAtlas/Dockerfile
GangaLHCb/Dockerfile
GangaDirac/Dockerfile

I guess this would be the latter

alexanderrichards:
can I merge this?

rob-c:
@alexanderrichards Would it be worth having some storage persistence and to mount a host directory as gangadir so that a user can use this to submit a job and still come back to it in future and access the jobs they may have submitted to a remote backend?
(apologies if I've mis-read the file but I think all submitted jobs get lost with the container don't they, is this by design? No objections to merging but that's what I've played with in the past locally)

egede:
@alexanderrichards Please go ahead and merge

@rob-c My understanding of Docker is that mounting a directory from the outside world is done via the `-v` option when you start the container. I do not think that there is anything inside the container that can solve this. A command like

```
docker run --privileged=true \
     -e "vo=gridpp" \
     --volume=/home/egede/.globus:/home/dirac/.globus \
     --volume=/home/egede/gangadir:/home/dirac/gangadir \
     -it ganga
```

should work for that.

alexanderrichards:
@rob-c Hi Rob,
I initially had the docker container running nothing but the ganga ui such that anyone could use it to run ganga with IC HEP dirac. In this scenario it could make sense to allow mounting of the gangadir and indeed I did think about it. It would actually be a run time option as @egede says so doesn't need to be in the Dockerfile.

However as you will notice I have changed the file such that we now drop to a shell with ganga setup ready to go. This was done as after discussion with @egede we can use these internally for running the tests with jenkins spawning them on the slaves.

rob-c:
@alexanderrichards Ah, ok, I wasn't sure what this was targeted for, for testing this is ideal.

I was just curious if there was going to be some instruction to people to run ganga within a docker container as way of distributing ganga?
(In which case I figured it would have thought going for having a folder mounted into the container would be better done by default and to allow for it to be changed if required.)

As an aside I did have ganga setup to run with all the persistent files contained within 1 folder a while ago so there was only 1 mount needed to get running in Docker, but I figured at the time since this required the users installing and understanding docker it was potentially more difficult to sell it to people as a solution.

PR #1013, opened 2017-05-15
Add GangaCore Dockerfile.

alexanderrichards:
Add a base docker image for GangaCore and build the GangaDirac image from it.

### DIFF
--- a/python/Ganga/Dockerfile
+++ b/python/Ganga/Dockerfile
@@ -1,0 +1,40 @@
+FROM centos:6
+LABEL maintainer "Alexander Richards <a.richards@imperial.ac.uk>"
+ARG ganga_version=6.5.2
+
+RUN yum install -y wget
+
+# Install Python2.7
+RUN yum groupinstall -y 'development tools' && \
+    yum install -y zlib-dev openssl-devel sqlite-devel bzip2-devel readline-devel && \
+    wget -np -O Python-2.7.13.tar.xz https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tar.xz && \
+    tar -xvf Python-2.7.13.tar.xz && \
+    cd Python-2.7.13 && \
+    ./configure --prefix=/usr/local && \
+    make && \
+    make altinstall && \
+    cd - && \
+    rm -rf Python-2.7.13*
+
+# Add the user UID:1000, GID:1000, home at /home/ganga
+RUN groupadd -r ganga -g 1000 && \
+    useradd -u 1000 -r -g ganga -m -d /home/ganga -s /sbin/nologin -c "Ganga user" ganga && \
+    chmod 755 /home/ganga
+
+# Set the working directory to ganga home directory
+WORKDIR /home/ganga
+
+# Specify the user to execute all commands below
+USER ganga
+
+# Install pip, virtualenv and ganga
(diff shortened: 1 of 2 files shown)


alexanderrichards:
can I merge this?

python/Ganga/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    chmod 755 /home/ganga
+
+# Set the working directory to ganga home directory
+WORKDIR /home/ganga
+
+# Specify the user to execute all commands below
+USER ganga
+
+# Install pip, virtualenv and ganga
+RUN curl https://bootstrap.pypa.io/get-pip.py | /usr/local/bin/python2.7 - --user && \
```
egede: Why wget above and then curl here?

python/GangaDirac/Dockerfile
```diff
@@ -1,34 +1,28 @@
-FROM alexanderrichards/dirac_ui:latest
+# FROM taking build arg requires at least
+#   Docker version 17.05.0-ce, build 89658be
+ARG ganga_version=latest
+FROM ganga-core:$ganga_version
 LABEL maintainer "Alexander Richards <a.richards@imperial.ac.uk>"
-ARG ganga_version=6.5.2
+ARG dirac_version=v6r17p18
```
egede: Is it possible to put some help text here that explains it is essential to mount the directory containing the Grid proxy.

python/GangaDirac/Dockerfile
```diff
@@ -1,34 +1,28 @@
-FROM alexanderrichards/dirac_ui:latest
+# FROM taking build arg requires at least
+#   Docker version 17.05.0-ce, build 89658be
+ARG ganga_version=latest
+FROM ganga-core:$ganga_version
```
egede: When this docker container is built automatically, is there a danger that you get out of sync? So the 6.5.3 tag is created. Could the notification to docker to build a new container now  happen such that the Dirac one is done first (and thus picking the 6.5.2 tag as being the "latest".

python/Ganga/Dockerfile
alexanderrichards: only because I can do it in one line with curl without the intermediary file

python/GangaDirac/Dockerfile
alexanderrichards: sure

python/GangaDirac/Dockerfile
alexanderrichards: have now removed reference to 6.5.2 instead defaulting to the latest. This however should be a build arg such that the default is never used for our purposes.

alexanderrichards: It is true however that this would require the ganga-core to be built before this one is. However I think that we will have to change the build process anyway as the automatic build is not available to organisations on dockerhub, only individuals, hence while it worked when it was under alexanderrichards/ganga it wont not it's under gangadevs/ganga-core etc.

We can use dockercloud to automate the builds in which case we will have to ensure that ganga-core goes first. However it is only free to build on their nodes while dockercloud is in beta. So I suggest that we make this mark of our build system in jenkins or travis this way we control the build order as well. We will of course be able to kick them off from pull requests just as with the testing.

egede: It should be possible to build the images as part of the release process. If we do it in the first step of the release process, then we can actually run the tests against these images.

PR #1014, opened 2017-05-16
Fix accessing key from wrong config

chrisburr:
This fixes a bug reported by Jibo this morning on lhcb-distributed-analysis which was introduced in #996 where I used the wrong config dictionary to access the value of MaxNumResubmits.


mesmith75:
test this please

mesmith75:
restest this please

PR #1018, opened 2017-05-26
Add optional id to GangaDiracError

### LINKED ISSUE #897: DiracError needs to associate to jobID
DiracError should be a bit more helpful and possibly take an optional jobID or filename to make debugging a DIRAC error (sometimes a little too generic) easier.

mesmith75:
Fixes #897 . This adds an optional id to the GangaDiracError. Not sure how useful this really is as there doesn't seem any obvious place at the moment where it is needed.

Also removed unused import in LHCbDataset.

### DIFF
--- a/python/GangaDirac/Lib/Utilities/DiracUtilities.py
+++ b/python/GangaDirac/Lib/Utilities/DiracUtilities.py
@@ -217,8 +217,8 @@ def execute(command,
                                   update_env=update_env)
 
     # If the time 
-    if returnable == 'Command timed out!':
-        raise GangaDiracError("DIRAC command timed out")
+#    if returnable == 'Command timed out!':
+    raise GangaDiracError("DIRAC command timed out")
 
     # TODO we would like some way of working out if the code has been executed correctly
     # Most commands will be OK now that we've added the check for the valid proxy before executing commands here
--- a/python/GangaDirac/Lib/Utilities/DiracUtilities.py
+++ b/python/GangaDirac/Lib/Utilities/DiracUtilities.py
@@ -217,8 +217,8 @@ def execute(command,
                                   update_env=update_env)
 
     # If the time 
-#    if returnable == 'Command timed out!':
-    raise GangaDiracError("DIRAC command timed out")
+    if returnable == 'Command timed out!':
+        raise GangaDiracError("DIRAC command timed out")
 
     # TODO we would like some way of working out if the code has been executed correctly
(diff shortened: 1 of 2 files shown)


python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -26,6 +26,14 @@
 
 class GangaDiracError(GangaException):
     """ Exception type which is thrown from problems executing a command against DIRAC """
+    def __init__(self, message, job_id = None):
```
rob-c: Is it worth adding the Ganga ID as well?

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -26,6 +26,15 @@
 
 class GangaDiracError(GangaException):
     """ Exception type which is thrown from problems executing a command against DIRAC """
+    def __init__(self, message, dirac_id = None, job_id = None):
+        self.dirac_id = dirac_id
+        self.job_id = job_id
+        self.message = message
```
drmarkwslater: I'd suggest using the parent where possible, so maybe:
```
GangaException.__init__(self, message)
```

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -26,6 +26,15 @@
...
+    def __str__(self):
+        if self.job_id and self.dirac_id:
+            return "GangaDiracError, Job %s with Dirac ID %s : %s" % (self.job_id, self.dirac_id, self.message)
+        else:
+            return "GangaDiracError: %s" % (self.message)
```
drmarkwslater: Similarly here:
```
else:
    return super(GangaException, self).__str__()
```

python/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -26,6 +26,16 @@
 
 class GangaDiracError(GangaException):
     """ Exception type which is thrown from problems executing a command against DIRAC """
+    def __init__(self, message, dirac_id = None, job_id = None):
+        GangaException.__init__(self, message)
+        self.dirac_id = dirac_id
+        self.job_id = job_id
+        self.message = message
```
drmarkwslater: This isn't needed as it's handled in the base class 😄

PR #1019, opened 2017-05-29
Add CMake support and Fix minor Credential problem

drmarkwslater:
This PR adds CMake support into the Atlas plugin. It auto detects this using PandaUtils (which also will need an upgrade) and if it is found, runs a **much** simplified wrapper script on the WN. A later PR will perform clean up of the old code that this supersedes.

In addition, this ensures an Atlas proxy is available when running. This is a 'hacky' fix before I try to get this working properly with the new credential system.

Finally, I found an issue specifically at Cambridge that the code to untar the input sandbox in local and batch jobs was failing on a particular tarball. I've switched the relevant code back to what it was (`extractall()`) and this seems to work.

### DIFF
--- a/python/Ganga/Lib/Localhost/LocalHostExec.py.template
+++ b/python/Ganga/Lib/Localhost/LocalHostExec.py.template
@@ -76,6 +76,8 @@ sys.path.insert(0, ###GANGADIR###)
 sys.path.insert(0,os.path.join(os.getcwd(),PYTHON_DIR))
 
 fullenvironment = os.environ.copy()
+for key,value in environment.iteritems():
+    fullenvironment[key] = value
 
 outfile=open('stdout','w')
 errorfile=open('stderr','w')
--- a/python/GangaAtlas/Lib/ATLASDataset/DQ2Dataset.py
+++ b/python/GangaAtlas/Lib/ATLASDataset/DQ2Dataset.py
@@ -2124,6 +2124,9 @@ logger = getLogger()
 # New for DQ2 client 2.3.0
 from Ganga.GPIDev.Credentials_old import GridProxy
 gridProxy = GridProxy()
+if not gridProxy.isValid():
+    gridProxy.create()
+
 username = gridProxy.identity(safe=True)
 nickname = getNickname(allowMissingNickname=False)
 if nickname:
--- a/python/GangaPanda/PACKAGE.py
+++ b/python/GangaPanda/PACKAGE.py
@@ -8,7 +8,7 @@
 """
 
 _external_packages = {
-    'panda-client' : { 'version' : '0.5.69',
+    'panda-client' : { 'version' : '0.5.80',
                     'syspath':['lib/python2.6/site-packages'],
                     'CONFIGEXTRACTOR_PATH':'etc/panda/share',
                     'PANDA_SYS':'.',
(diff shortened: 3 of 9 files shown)


python/GangaPanda/PACKAGE.py
```diff
@@ -8,7 +8,7 @@
 """
 
 _external_packages = {
-    'panda-client' : { 'version' : '0.5.69',
+    'panda-client' : { 'version' : '0.5.80',
                     'syspath':['lib/python2.6/site-packages'],
                     'CONFIGEXTRACTOR_PATH':'etc/panda/share',
                     'PANDA_SYS':'.',
```
mesmith75: I guess this is not the PR for it but it seems strange you do the exact same thing in two different files. Is there a reason?

python/GangaAtlas/Lib/ATLASDataset/DQ2Dataset.py
```diff
@@ -2124,6 +2124,9 @@ def addTask(self, cmd):
 # New for DQ2 client 2.3.0
 from Ganga.GPIDev.Credentials_old import GridProxy
 gridProxy = GridProxy()
+if not gridProxy.isValid():
+    gridProxy.create()
+
```
mesmith75: Isn't there already a decorator for this kind of thing?

python/Ganga/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -76,6 +76,8 @@ sys.path.insert(0, ###GANGADIR###)
 sys.path.insert(0,os.path.join(os.getcwd(),PYTHON_DIR))
 
 fullenvironment = os.environ.copy()
+for key,value in environment.iteritems():
+    fullenvironment[key] = value
```
mesmith75: I don't quite follow this - why do you define `fullenvironment' and then assign values that look to be the same thing? Are those in environment different/additional?

mesmith75:
I don't know about the ATLAS software but this looks fine. Do you want to add some tests for this at a later stage?

python/Ganga/Lib/Localhost/LocalHostExec.py.template
drmarkwslater: `environment` contains the env variables required by the job (i.e. set in the RT Handler of the application). What this does is combine the current environment with these additional variables. Without it, they aren't set and the Local backend will fail for applications that need additional env variables set (that may just be Atlas ones, but best to be consistent!).

python/GangaAtlas/Lib/ATLASDataset/DQ2Dataset.py
drmarkwslater: For the new credential system, yes. However, I haven't had the time to properly sort that out yet so at present the code is still using the old `gridProxy` object to make sure a proxy is present at startup. In a future PR, this will be removed in favour of decorators on particular functions.

python/GangaPanda/PACKAGE.py
drmarkwslater: This is because, historically, we have both GangaAtlas and GangaPanda modules (one for the general Atlas code and one for the grid/Panda backend) and in the dim and distant past you didn't need both loaded to run jobs (thinking about that's technically still true actually 😄 ). However, this is definitely something that is ripe for clean up in the future!

mesmith75:
It would be good to get this merged so we can release the next version.

drmarkwslater:
Great! Thanks Mark! I'll prepare the updated version of the Panda Tools in the next 24 hours and then merge. I'm assuming this will be 6.6?

mesmith75:
Great - yes this will be 6.6.0

drmarkwslater:
I've now created the new tarball so I'm happy for 6.6 😄

PR #1021, opened 2017-06-02
Change the default when making new gangarc

mesmith75:
Changes from no to yes for future versions.

### DIFF
--- a/python/Ganga/GPIDev/Lib/Config/Config.py
+++ b/python/Ganga/GPIDev/Lib/Config/Config.py
@@ -267,14 +267,14 @@ def config_file_as_text(interactive):
                         except AttributeError as err:
                             pass
                     if interactive:
-                        yes = raw_input('The config option %s %s with value %s in your old .gangarc is not the default. Do you want to copy it to the new .gangarc file (y/[n]) ?\n' % (sect.name, o, value))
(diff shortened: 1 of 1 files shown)


python/Ganga/GPIDev/Lib/Config/Config.py
```diff
@@ -267,14 +267,14 @@ def config_file_as_text(interactive):
...
+                        yes = raw_input('The config option %s %s with value %s in your old .gangarc is not the default. Do you want to copy it to the new .gangarc file ([y]/n) ?\n' % (sect.name, o, value))
                     else:
                         yes = 'y'
-                    if yes.lower() in ['y']:
+                    if yes.lower() in ['n']:
                         text += '#%s = %s\n' % (o, def_value)
-                        text += '%s = %s\n\n' % (o, value)
                     else:
                         text += '#%s = %s\n' % (o, def_value)
+                        text += '%s = %s\n\n' % (o, value)
```
drmarkwslater: Nothing really to do with the specific changes (which are fine!), but this could actually be condensed to:
```
text += '#%s = %s\n' % (o, def_value)
if yes.lower() not in ['n']:
    text += '%s = %s\n\n' % (o, value)
```

Knocks a couple of lines off the file 😄

mesmith75: Thanks, that is a bit tidier

mesmith75:
test this please

PR #1044, opened 2017-06-30
Get Atlas working on SL7/Centos 7

drmarkwslater:
Update the rucio version and use the correct path. This should sort out getting Atlas working on SL7.

I've also put in another minor fix for a non-assigned variable and removed some broken new credential code until I get a chance to fix it properly.


drmarkwslater:
@egede So the problem was with the library path for pulling kerberos and other packages from the rucio client. They were hard-coded to python 2.6 dirs but because Centos 7 is python 2.7 by default, these dirs didn't exist in the latest version of rucio and so it failed. Updating the rucio version and making sure the paths were set properly depending on python version fixed the problem.

The `require_credential` change was just removing the 'new' credential system stuff that @milliams put in as a band aid to get it working again. Basically, the python calls to the panda client require the proxy to be in a particular place and for whatever reason the way the code was at the moment it didn't work. I've currently got the Atlas code forcing an old `gridProxy` to be present and that ensures the proxy is where it should be. In the near future I'll revisit this and do it properly as @milliams intended.

P.S. SL7 does exist, it's just that CERN are going for CentOS instead 😄

PR #1047, opened 2017-07-03
Trying to port back DiracProxy changes. I make no guarantees

rob-c:
This allows LHCb users to construct/query proxies with the `lhcb-proxy-` tools which are wrappers to the `dirac-proxy-` tools in vanilla DIRAC.
This is taken largely from #917 without the ability to pass an environment with the proxy. (The intention of which is to get closer to allowing 1 repo to safely manage jobs from multiple DIRAC entities at once).

This breaks at least one nightly test from memory.
These tests are failing due to larger problems last I looked so I'm not too concerned with trying to port a fix to something which doesn't currently work due to external changes.


rob-c:
OK, I clearly need to port over the fixes to the non-nightly test from #917 I'll try and get around to this today

rob-c:
OK, this should be working and is a minimal change to the Proxy required for calling `lhcb-proxy-init`.

python/GangaDirac/Lib/Credentials/DiracProxy.py
```diff
@@ -153,6 +154,11 @@ def default_location(self):
         else:
             return base_proxy_name
 
+    @property
+    def location(self):
+        """
+        """
+        return self.default_location()
```
mesmith75: What is this for?

rob-c: This was orignally to aid in debugging, it just overrides the underlying logic to call the `default_location` directly so I can remove it if required.

mesmith75: Might be nice to remove it if it isn't needed?

alexanderrichards:

alexanderrichards:

mesmith75:
@rob-c  Can we merge this?

rob-c:
@mesmith75 I'm happy or should I remove the location method? I'd forgotten about this due to other work.

mesmith75:
@rob-c  Yes please if it is unused code. Thanks!

mesmith75:
Thanks @rob-c . Let's merge this today for the next release.

rob-c:
@mesmith75 Just for curiosity does this work with nightlies and/or #1087 (I don't foresee problems but I'm not 100% certain I follow the Dirac tests)?

mesmith75:
er - that is a good question. I still haven't got to the bottom of #1087 . I'll try the nightlies with this branch.

mesmith75:
@rob-c So this doesn't fix the test in #1087 - I think it is roughly unrelated. So let's merge this now.

rob-c:
sounds good to me

mesmith75:
@rob-c can you merge this?

PR #1048, opened 2017-07-04
Update gangarc on minor version change

### LINKED ISSUE #1045: Change gangarc questions to minor version
Reminder to change the `.gangarc` questioning to when the minor version is updated.

mesmith75:
Fixes #1045 . 

As an aside would it be better to put the .used_versions file into the home directory? It is kind of annoying to have the gangarc updating when using different gangadirs .

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -322,14 +322,22 @@ under certain conditions; type license() for details.
         if not os.path.exists(versions_filename):
             with open(versions_filename, 'w') as versions_file:
                 versions_file.write(_gangaVersion + '\n')
-            return True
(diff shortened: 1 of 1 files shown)


egede:
Moving the file out of the `gangadir` is a good idea. Now should it go straight in the home directory or should it go into `.cache/Ganga` ?

mesmith75:
As the filename is not obviously ganga related I think the .cache/Ganga is best if its existence is verified earlier in the startup.

mesmith75:
restest this please

mesmith75:
Ok, so the `.used_versions` has been moved to `~/.cache/Ganga/`. If there are no objections I'll merge this for 6.6.2

PR #1050, opened 2017-07-04
Set getMetadata default to true for BenderBox

mesmith75:
This sets the default for `getMetadata` to true for BenderBox. The postprocessor is there without a flag and the necessary options are in the script for the module. This will ensure that the summary.xml is added to the `outputsandbox` . I forgot to put this in earlier.


python/GangaLHCb/Lib/Applications/BenderBox.py
```diff
@@ -160,7 +160,8 @@ class BenderModule(GaudiExec):
     _schema.version.minor += 0
     ## make entries 
     for key,val in _schema.datadict.iteritems() :
-        if key == 'useGaudiRun'  : val._update( { 'defvalue' : False } )  
+        if key == 'useGaudiRun'  : val._update( { 'defvalue' : False } )
```
egede: A bit odd construction to loop over all key,value pairs and then ask for a specific one. Will do nothing in case that `useGaudiRun` doesn't exist, but is that really a good behaviour?

alexanderrichards:

PR #1051, opened 2017-07-05
Check for existence of cache folder

mesmith75:
Bugfix to check that the `~/.cache/Ganga` folder exists at startup.

Could someone please approve this and I'll restart the 6.6.2 release (also the getMetadata PR would be really useful to have in but this one is vital).

### DIFF
--- a/python/Ganga/Runtime/bootstrap.py
+++ b/python/Ganga/Runtime/bootstrap.py
@@ -319,6 +319,8 @@ under certain conditions; type license() for details.
     @staticmethod
     def new_version():
         versions_filename = os.path.expanduser('~/.cache/Ganga/.used_versions')
+        if not os.path.exists(os.path.dirname(versions_filename)):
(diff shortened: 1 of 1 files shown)


python/Ganga/Runtime/bootstrap.py
```diff
@@ -319,6 +319,8 @@ def file_opens(f, message):
     @staticmethod
     def new_version():
         versions_filename = os.path.expanduser('~/.cache/Ganga/.used_versions')
+        if not os.path.exists(os.path.dirname(versions_filename)):
```
rob-c: surely you want to be creating the directory not the whole path looking at the code below?
Also, does this work recursively like `mkdir -p` or just create a single empty folder?

mesmith75: according to the python website `makedirs` is recursive, `mkdir` is not, so this should be fine

PR #1061, opened 2017-07-10
Add EXE support and fix issues with user packages

### LINKED ISSUE #1053: run-athena-new does not pick up user code
Hi,

It turns out the way run-athena-new is coded up, it will not pick up the user's code in the worker. https://github.com/ganga-devs/ganga/blob/3b53affeff744a4cf80b83c94bd54142c567d37d/python/GangaAtlas/Lib/Athena/run-athena-new.sh#L35

This is the problematic part. 

1. If the user hasn't requested compilation, there should not be any build attempt
2. If the user has sent their workarea with the job, then all which needs to happen is the setup script should be sourced i.e. ```source usr/WorkDir/21.2.0/InstallArea/x86_64-slc6-gcc62-opt/setup.sh```

The problem though is that "path" probably will change depending on what release is set up, and I wouldnt want to definitively say how (especially for nightlies, I am not sure what it does properly). But basically you need to get through to the InstallArea and source the setup script then you are done.

Can this be looked into ASAP because at the moment it means ganga cannot be used with cmake releases :-(
Thanks
Will

drmarkwslater:
This PR does the following:

* Adds EXE/ROOT/PYARA support into CMake releases
* Fixes issues with hardcoded architecture path
* Only does compilation if requested

Fixes #1053.

Could this be reviewed ASAP please so it can go out for a tutorial this week? Thanks!

### DIFF
--- a/python/GangaAtlas/Lib/Athena/Athena.py
+++ b/python/GangaAtlas/Lib/Athena/Athena.py
@@ -1154,13 +1154,25 @@ class Athena(IPrepareApp):
         if self.atlas_exetype in ['EXE']: #and not self.athena_compile:  - for EXE, compilation decides what the tarball is called
             maxFileSize = config['EXE_MAXFILESIZE']
             archiveName, archiveFullName = create_tarball(self.userarea, runDir, currentDir, archiveDir, self.append_to_user_area, self.exclude_from_user_area, maxFileSize, self.useAthenaPackages, verbose, self.athena_compile )
-        else:
-            archiveName = ""
+
             if AthenaUtils.useCMake():
                 self.useCMake = True
-                archiveName,archiveFullName = AthenaUtils.archiveWithCpack(True,tmpDir,True)
+        else:
+            # compilation determines whether to send the sources across as well
+            archiveName = ""
+            if self.athena_compile:
+                if AthenaUtils.useCMake():
+                    self.useCMake = True
+                    archiveName,archiveFullName = AthenaUtils.archiveWithCpack(True,tmpDir,True)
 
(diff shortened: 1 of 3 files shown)


python/GangaAtlas/Lib/Athena/Athena.py
```diff
@@ -1154,13 +1154,25 @@ def prepare(self, force=False, **options):
...
-                archiveName,archiveFullName = AthenaUtils.archiveWithCpack(True,tmpDir,True)
+        else:
+            # compilation determines whether to send the sources across as well
+            archiveName = ""
+            if self.athena_compile:
+                if AthenaUtils.useCMake():
+                    self.useCMake = True
+                    archiveName,archiveFullName = AthenaUtils.archiveWithCpack(True,tmpDir,True)
+
+                archiveName, archiveFullName = AthenaUtils.archiveSourceFiles(self.userarea, runDir, currentDir, archiveDir, verbose, self.glue_packages, config['dereferenceSymLinks'], archiveName=archiveName)
```
rob-c: Doesn't this overload the `achiveName, archiveFullName` above if `not AthenaUtils.useCMake()`?
(This could be intended behaviour so apologies if it is :P )

drmarkwslater: Don''t worry - it's not obvious at all 😄 However, this is copied from other Atlas stuff to ensure compatibility so we're somewhat stuck with it! Basically, `archiveSourceFiles` **appends** the additional source on to the already packaged tarball. I'm not sure why it's done like this but as I say - I don't want to fix something that isn't broken 😄

drmarkwslater:
Thanks @rob-c ! 

@mesmith75 Could we start a release ASAP please? Thanks!

PR #1063, opened 2017-07-11
Adding Dockerfile for the testing framework.

alexanderrichards:


alexanderrichards:
> Does the location of the file indicate that this will be a different Docker file compared to the file one level higher in the file system?

yes I think so. This sets up an image that doesn't use pip to install ganga (as we discussed) rather for the testing simply clones from git hub and runs the py.test. You can get test results by running thusly:

```bash
%> docker run --rm <image_name> <location of tests to run>
```
e.g.
```bash
%> docker run --rm test-core /home/ganga/ganga/python/Ganga/tests/Unit/Notebook/*.py
```

Note that this doesn't run the container interactively nor does it drop to a shell both of which the previous image would do. The question is really what we want these docker images for. The previous one, when I was creating it, I envisioned being a complete Ganga setup from pip ready for the user to use. This one however is intended for use in the testing framework where we clearly cannot install from pip as we are testing versions not yet on PyPI.

Does this make sense? could we be doing it a better way?

egede:
@alexanderrichards This all looks good. Could the execution of all tests in the package maybe be made a default?

alexanderrichards:
> @alexanderrichards This all looks good. Could the execution of all tests in the package maybe be made a default?

indeed

alexanderrichards:
so I have made the default all the unit tests. Did you want to run the integration ones by default as well?

PR #1070, opened 2017-07-17
Add config check to MassStorageFile

### LINKED ISSUE #1069: AFS-based EOS clients are deprecated
See [OTG0037759](https://cern.service-now.com/service-portal/view-outage.do?n=OTG0037759).

At least for LHCb, ganga is using 
/afs/cern.ch/project/eos/installation/lhcb/bin/eos.select
The quick fix is to change instances of this in ~/.gangarc to
/afs/cern.ch/project/eos/installation/lhcb.old/bin/eos.select

Longer term, it seems the only option is to require and use a local installation of EOS tools, as per [KB0003846](https://cern.service-now.com/service-portal/article.do?n=KB0003846).

mesmith75:
Fixes #1069 .

Checks that it can make/ls the default MassStoragePath to check the user configs. Also updated the default path for eos to `/usr/bin/eos`

### DIFF
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -64,9 +64,7 @@ class MassStorageFile(IGangaFile):
         super(MassStorageFile, self).__init__()
         self._setNamePath(_namePattern=namePattern, _localDir=localDir)
         self.locations = []
-
         self.shell = Shell.Shell()
-
 
     def __setattr__(self, attr, value):
         """
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -102,7 +102,7 @@ class MassStorageFile(IGangaFile):
 
     def _checkConfig(self):
         """
-        Check that the MassStorageFile configuration is correct by trying to make the default path
+        Check that the MassStorageFile configuration is correct
         """
         if not getConfig('Output')[_getName(self)]['uploadOptions']['path'] :
             raise GangaException('Unable to create MassStorageFile. Check your configuration!')
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -65,6 +65,8 @@ class MassStorageFile(IGangaFile):
         self.locations = []
 
         self.shell = Shell.Shell()
+        self._checkConfig()
+
 
     def __setattr__(self, attr, value):
         """
@@ -98,6 +100,14 @@ class MassStorageFile(IGangaFile):
             self.namePattern = _namePattern
             self.localDir = _localDir
 
+    def _checkConfig(self):
+        """
+        Check that the MassStorageFile configuration is correct by trying to make the default path
+        """
+        try:
+            self._mkdir(getConfig('Output')[_getName(self)]['uploadOptions']['path'], False)
+        except GangaException:
+            raise GangaException('Unable to create MassStorageFile. Check your configuration!')
 
     def _on_attribute__set__(self, obj_type, attrib_name):
         # This is defining the object as uncopyable from outputfiles... do we want this mechanism still?
--- a/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
+++ b/python/Ganga/GPIDev/Lib/File/MassStorageFile.py
@@ -60,12 +60,12 @@ class MassStorageFile(IGangaFile):
             namePattern (str): is the pattern of the output file that has to be written into mass storage
             localDir (str): This is the optional local directory of a file to be uploaded to mass storage
         """
+        self._checkConfig()
         super(MassStorageFile, self).__init__()
         self._setNamePath(_namePattern=namePattern, _localDir=localDir)
         self.locations = []
 
         self.shell = Shell.Shell()
-        self._checkConfig()
 
 
     def __setattr__(self, attr, value):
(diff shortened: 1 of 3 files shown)


mesmith75:
This broke the tests because the Jenkins box doesn't have EOS installed. Is that something we want to do?

rob-c:
Wouldn't it be better to add a test in the `__init__` to allow checking for `eos` as an executable in the `$PATH` and to extract it's location into the config dynamically.

Also for the tests I would either nop any initialization test using the python test packages (e.g. https://github.com/ganga-devs/ganga/blob/b9ef1b26def5daf93b09c7d831eff9cdaf7035da/python/Ganga/test/Unit/Credentials/TestCredentialInfo.py#L51) or add a config parameter to override this and change the Ganga config on test launch https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/testlib/GangaUnitTest.py#L92 .

rob-c:
Also, should the test be really trying to touch the filesystem on init of a class, it feels slightly bad although I suppose it's OK.

mesmith75:
I don't think so - if you don't use eos or MassStorageFile at all then I guess you don't really want these checks to be carried out.

I see your point about testing the file system though and will just check for the executable.

rob-c:
@mesmith75 I'm not talking about throwing an error/warning on startup I'm just thinking of trying to find the `eos` command in the path _once_ at startup and if it's not there, then throw an exception when trying to construct a mass storage file.
Similar to the logic with SharedFile where if something isn't set don't allow it to be constructed.

mesmith75:
Ah I see, yes that is much better. However I don't think we can just check for the installation.

We currently ask for the commands which to me makes sense as they can vary. Eg, for eos it would be `/usr/bin/eos mkdir` but for gfal `/usr/bin/gfal-mkdir` so I think we have to run a command of some sort. Maybe just the `--help` would suffice?

mesmith75:
So I changed it round a bit. If `/usr/bin/eos` doesn't exist then the default path config doesn't get set. That is checked when the thing is created. Compared with before this won't check that the user has put in anything reasonable from the own configuration but I guess that will become apparent.

I wanted to at least keep the whole thing in the configuration so it would be written to the .gangarc and give the user an idea of what needs to be filled in.

mesmith75:
@rob-c sorted

mesmith75:
retest this please

mesmith75:
retest this please

alexanderrichards:

alexanderrichards:

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -60,12 +60,14 @@ def __init__(self, namePattern='', localDir='', **kwds):
...
             localDir (str): This is the optional local directory of a file to be uploaded to mass storage
         """
+        self._checkConfig()
         super(MassStorageFile, self).__init__()
         self._setNamePath(_namePattern=namePattern, _localDir=localDir)
         self.locations = []
 
         self.shell = Shell.Shell()
 
+
```
rob-c: I think this isn't needed due to PEP8 (I'm think I'm correct this time)

python/Ganga/GPIDev/Lib/File/MassStorageFile.py
```diff
@@ -60,12 +60,14 @@ def __init__(self, namePattern='', localDir='', **kwds):
             namePattern (str): is the pattern of the output file that has to be written into mass storage
             localDir (str): This is the optional local directory of a file to be uploaded to mass storage
         """
+        self._checkConfig()
```
rob-c: Adding this means that you will need to update the tests which use the `MassStorageFile`/`SharedFile` but don't configure it.
I would suggest adding some default `path` to https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/testlib/GangaUnitTest.py#L92 is the quickest solution.

mesmith75: Yes - I am not sure why the current tests work though. The old path should be defunct.

rob-c: https://ganga-ci.cern.ch/jenkins/job/PR%20builder/2843/testReport/junit/python.Ganga.test.GPI.FileTests.TestSJIDName/TestSJIDName/test_a_jobSubmit/ is failing as there is no _safe_ default,
https://ganga-ci.cern.ch/jenkins/job/PR%20builder/2843/testReport/python.Ganga.test.GPI.FileTests.TestMassStorageClient/ passes because it's explicitly adding a path into the config by design (https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/test/GPI/FileTests/TestMassStorageClient.py#L30)

Alternatively I suppose you could fix https://github.com/ganga-devs/ganga/blob/develop/python/Ganga/test/GPI/FileTests/TestSJIDName.py as it's the only test failing

mesmith75: So I think I fixed this by removing references to EOS in the config for the unit tests - I realise that the placing of the files would have been failing before but the tests would still have passed as the job would have completed successfully.

I set the path for the MassStorageFile to `/tmp` . Is that a bad idea?

mesmith75:
If there are no objections I will merge this.

python/Ganga/__init__.py
```diff
@@ -732,9 +732,14 @@ def filefilter(fn):
                      'undefined': 'root://eos.cern.ch'}
 defaultMassStorageProto = protoByExperiment[groupname]
 
-prefix = '/afs/cern.ch/project/eos/installation/%s/bin/eos.select ' % groupname
-massStorageUploadOptions = {'mkdir_cmd': prefix + 'mkdir', 'cp_cmd':
-                            prefix + 'cp', 'ls_cmd': prefix + 'ls', 'path': massStoragePath}
+import commands
```
alexanderrichards: Similar to previous comments, does the import need to be here?

python/Ganga/testlib/GangaUnitTest.py
```diff
@@ -89,6 +89,16 @@ def start_ganga(gangadir_for_test, extra_opts=[], extra_args=None):
 
     # These are the default options for all test instances
     # They can be overridden by extra_opts
+
+    #Sort out eos
+
+    from Ganga.Utility.Config import getConfig
```
alexanderrichards: ditto

python/Ganga/__init__.py
mesmith75: No I'll move this to the top.

python/Ganga/testlib/GangaUnitTest.py
mesmith75: Not sure - I was following the rest of the file - I'll check to be sure

PR #1071, opened 2017-07-18
Fixing some bad excpetions thrown by Ganga

### LINKED ISSUE #1067: Ganga doesn't handle string conversion gracefully
Ganga is mishandling string conversions quite badly atm.

I have been working with an LSST user and we accidentally tried to assign a string to a dataset attribute during debugging and ganga threw an exception which generated a full trace.

1) The full trace shouldn't happen so I suspect the plugin directory is possibly not using exceptions of the GangaException type
2) The error string should probably be updated to be a little more human friendly as it's thrown when a new user does something similar to: `j.inputdata = 'myfile.txt'`

rob-c:
This is a short PR showing some tlc to areas of the code which throw full traces via vanilla ganga and make life difficult for lsst users trying to get used to the framework.

This fixes #1067 that I see when working with LSST people.

### DIFF
--- a/python/Ganga/Core/GangaRepository/Registry.py
+++ b/python/Ganga/Core/GangaRepository/Registry.py
@@ -549,7 +549,7 @@ class Registry(object):
             # Cleanup after ourselves if an error occured
             # Didn't load mark as clean so it's not flushed
             if obj_id in self._objects:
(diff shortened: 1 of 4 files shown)


egede:
retest this please

Issue #1055, opened 2017-07-07
remoteDir for DiracFile doesn't work when submitting a job


rob-c:
There is no mechanism that I can see where a DiracFile in the current code can modify the output JDL structure.

This relates to the _quick fix_ put in for #1040 as there is simply no mechanism behind the scenes in ganga to pass this requirement from a DiracFile object to DIRAC even though DIRAC now supports this.
(I plan to break this out into a small PR which is probably the best way of doing this without embarking on a large amount of work).
This is something a few users are asking for.

It would be nice to share the same functionality as MassStorageFile here to allow for powerful outputfile naming but I suspect that this isn't entirely possible without a bit of work and will likely result in us modifying the payload of DIRAC jobs to juggle files before DIRAC picks them up after execution.

mesmith75:
Fixed by #1072

Issue #1040, opened 2017-06-28
Default grid directory names


DanJohnsonCERN:
Hi,

When Ganga submits to the GRID (in my case Dirac) it is very difficult to connect LFNs with Ganga jobs. This makes managing GRID space extremely difficult and time consuming for the user, and many don't bother, leading to enormous wastage of disk space. 

Please can we make a trivial change to include the Ganga job id (not even the subjob, necessarily) in the LFN path? It would be even better to have the name of the ganga job there too, but perhaps that is asking too much. 

It would be vastly preferable if this could happen by default.

Thanks!
Dan

rob-c:
@DanJohnsonCERN I've added some code to #917 to store LFN using the following directory structure:

Eg: for a test job 4 from my repo:

OutputFiles are stored as:
```
/lhcb/user/r/rcurrie/GangaJob_4/OutputFiles/2017_07/171565/171565261/tmpYEqjgq.txt
```
InputFiles are uploaded as:
```
/lhcb/user/r/rcurrie/GangaJob_4/InputFiles/diracInputFiles_df64928a-dccd-455d-87cb-a3bdc69442b1.tgz
```

I've added this to store subjobs under the `GangaJob_4` folder which should make managing LFN from the base dir.

This should make sorting data by which Ganga job the file belongs to easier to parse as the Ganga job now has a folder under the users base lfn and all subjobs for large processings are collected together easier.

I think this should probably be included in 6.7.0 if everything goes well.

rob-c:
Unfortunately we can't remove the extra folders `2017_07/171565/171565261` in the output path as this is required at some level by DIRAC but this solution should give some way of managing things better.

DanJohnsonCERN:
Thanks very much, Rob, that's great, this is exactly what I was hoping for. Much appreciated.

egede:
retest please

DanJohnsonCERN:
I submitted a load of jobs yesterday early morning. When were the changes available?

On 18 Jul 2017 4:45 pm, Ulrik Egede <notifications@github.com> wrote:

retest please

egede:
@DanJohnsonCERN The pull request #1072 is not merged yet. It needs to be merged and then go into a release. Also the default behaviour will stay as it currently us, so look in the release notes (when ready) for how to change it.

DanJohnsonCERN:
Out of interest, why is the default not changed? It would make grid storage management much simpler-identification by short lived Dirac IDs is very difficult...
[...]

PR #1072, opened 2017-07-18
Lfn changes

rob-c:
This PR contains the code to address #1040 . I've been nudged by several people over this for a few years and it's something I would like to see us have a better handle over.

It would be nice to move the code dealing with output file naming out of mass storage file but this is a serious undertaking to do correctly as it will require expanding the API as well as developing the test suite to make sure that this all works correctly.

With that in mind we currently also have #1055 as an open issue and this goes at least part of the way to solving the need to fix this more quickly.

I think a more complete version of this PR may be to clean up the fact this was cherry picked from #917 as well as adding in a solution for #1055 before looking into how to merge functionality between MassStorage and Dirac file objects.


egede:
retest please

rob-c:
@egede Something is up with the build system I assumed someone was working on it.
This hasn't been tested yet but should pass all non DIRAC/LHCb tests as it doesn't touch anything anywhere else. I'll give this another quick look when I get a chance, hopefully before the end of the week, I suspect there is a minor bug in GaudiExec from this branch tough.

egede:
retest this please

mesmith75:
@rob-c I think this is ok. Are you sure you switched all the RTHandlers? Would it be worth putting in a default job number just in case?

rob-c:
@mesmith75 There is a bug here that will cause all GaudiExec jobs to not submit correctly to DIRAC. I'm fixing it atm.

rob-c:
@mesmith75 This goes some way to fixing things.

`DiracFile.remoteDir` is now respected as the path extension to use when placing user-requested/managed LFN on the grid.

I've migrated the functionality of using `{jid}`, `{sjid}` and `{fname}` out from MassStorageFile.
This now means that `DiracFile.remoteDir` supports these constructs. (although I've not fully tested this yet)

Aka users can now do:

```
...
j.outputfile=[DiracFile(remoteDir='MyImportantWork_{jid}/{sjid}'...)]
...
```

This would then lead to an LFNs similar to:

```
/lhcb/user/r/rcurrie/MyImportantWork_4/0/2017_07/171565/171565261/tmpYEqjgq.txt
/lhcb/user/r/rcurrie/MyImportantWork_4/1/2017_07/171565/171565332/tmpYEqjgq.txt
```


There is an automatic naming option which can be used by default to place LFN in a path which is constructed using the jid and sjid (so gives similar to MassStorageFile defaults).
This has to be turned on through the configuration system as it can potentially lead to confusions.

This works as originally described in #1040 and has the advantage that this requires a 1 line change in the config to get the effects.


This now needs testing to submit real DiracFiles so I'll do this soon.

egede:
retest this please

egede:
retest this please

alexanderrichards:

mesmith75:
Yes let's put it into the bugfix release.

rob-c:
👍 Righto, Do I have enough time to write a docstring example or 2 for this? merge by 2pm or so?

rob-c:
Edit I'll get the changes in by 3pm, need to sort a few things out today,

rob-c:
Ok, I'm curious to hear any feedback on the docstr for DiracFile I've tried to cover a few different use-cases rather than just outputfiles.

mesmith75:
@rob-c They look good.

rob-c:
@mesmith75 OK if there's nothing really to add I'm happy to merge this

mesmith75:
Actually can I request a slight change to No 5 in the docstring? It just needs to be clear that the filename has to correspond to the file created by the job (DaVinci etc) in order for it to be uploaded and you can subsequently only change the base directory path.

mesmith75:
@rob-c Thanks. The `remoteDir` has become a little confusing now but maybe it is just me.

To be clear it only affects the location of files uploaded at the completion of a job? It doesn't affect the location of file uploaded with `DiracFile.put()` ?

rob-c:
@mesmith ah... That's the 1 use case I haven't explored. It obviously won't work if you have a file without a job parent and a jid in the name... I can have a look at this tomorrow of needed but it should work fine if you just change the remotedir before a put i think.
I can do some checks tomorrow now if needed.

There is a bit too much bluring of functionality here but a rewrite to handle mass storage and dirac in a more consistent way is difficult :/

mesmith75:
@rob-c I don't think that is an issue. You can already specify the LFN in `put()` but the meaning of `remoteDir` is perhaps a little opaque.

mesmith75:
retest this please

rob-c:
@mesmith75 Think I discovered a minor bug in this PR, I'm checking now

rob-c:
OK, I need to clean up this PR (not sure what has happened)

rob-c:
Not sure what happened there, possibly the result of cherry-picking. The minor bug I spotted was that the put and WN script behaviour was completely inconsistent with the DIRAC upload at the end of the job. These should now be in sync.

rob-c:
I've just run a test job and the LFN renaming appears to be working for me as well as the new config to have ganga put LFN in a different sub-folder

python/Ganga/test/GPI/FileTests/TestMassStorageClient.py
```diff
@@ -113,6 +113,7 @@ def test_b_testClientSideComplete(self):
             # Check that the files were placed in the correct place on storage
             output_dir = os.path.join(self.outputFilePath, str(j.id), str(sj.id))
             for file_ in j.inputfiles:
+                print(os.path.join(output_dir, file_.namePattern))
```
mesmith75: Is that print supposed to remain?

mesmith75:
@rob-c Thanks that looks great.

PR #1088, opened 2017-07-26
Updated configuration on where output files should be treated

egede:
Added the various batch backends into configuration. In the longer term some default action might be preferable to define.

### DIFF
--- a/python/Ganga/__init__.py
+++ b/python/Ganga/__init__.py
@@ -757,7 +757,7 @@ output_config.addOption('MassStorageFile',
                        massStorageFileExt)
 
 sharedFileBackendPost = {'LSF': 'WN',
-                         'LCG': 'WN',
+                         'LCG': 'client',
                          'ARC': 'client',
                          'Dirac': 'client',
                          'PBS': 'WN',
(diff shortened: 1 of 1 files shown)


alexanderrichards:

alexanderrichards:
alex test

alexanderrichards:
retest this please

mesmith75:
Do we want to add SGE to the MassStorageFile and SharedFile configs as well?

python/Ganga/__init__.py
```diff
@@ -715,7 +756,16 @@ def filefilter(fn):
                         'defaultProtocol': defaultMassStorageProto},
                        massStorageFileExt)
 
-sharedFileBackendPost = {'LSF': 'client', 'LCG': 'client', 'ARC': 'client', 'Dirac': 'client', 'PBS': 'client', 'Interactive': 'client', 'Local': 'client', 'CREAM': 'client'}
+sharedFileBackendPost = {'LSF': 'WN',
+                         'LCG': 'WN',
```
mesmith75: Is this LCG one correct?

PR #1093, opened 2017-07-28
Adding gcc to Ganga test Dockerfile.

alexanderrichards:


alexanderrichards:
seems that when pip installing the requirements from ganga's requirements file there are warnings about missing gcc. I think it works anyway so maybe this is not needed but I've put it here to remind me to check

PR #1103, opened 2017-08-07
Making sure proxy correctly descends into simple nested objects

rob-c:
This fixes an issue where the proxy layer doesn't strip and add proxies to ganga objects as they move through the proxy layer.

This fixes an area of mystery problems where plugin code has to do pointless stripping and adding proxies which should be handled by the proxy layer itself.

This fixes a demonstrable bug that `ld=LHcbDataset();ld.extend([DiracFile('blah')]);print ld.files` doesn't work without this fix.

I've tested this on the GangaList object which is the strictest object I know for deciding when an object should be proxied or not and everything passes in testing.

If this creates a rabbit hole of problems then the above demonstrated bug can be fixed in the plugin but imo this is really a bug in core which I've repeatedly been patching away elsewhere.


rob-c:
@mesmith75 Is there anything other than list, tuple or dict that I may have forgotten to add support for?

mesmith75:
@rob-c I was wondering but I couldn't think of anything.

PR #1104, opened 2017-08-08
Remove script files from tmp in GaudiExec

### LINKED ISSUE #1101: GaudiExec leaves files in /tmp
It's been reported that ganga is bleeding files in `/tmp` from GaudiExec.

I suspect this is a minor bug in the RTHandler possibly for Batch but potentially for all backends.

mesmith75:
Fixes #1101


rob-c:
Looks like this should fix it, looks ok to me 👍

mesmith75:
Can I merge this?

mesmith75:
Yes - the `uploadLocalFile` function removes the local copy of the uploaded file.

PR #1106, opened 2017-08-11
GaudiExec cleanup on job remove

### LINKED ISSUE #957: GaudiExec clean up
Reminder to look into having GaudiExec remove its files of the grid if the job is removed.
Also, consider replicating the tarball once to reduce possible failures - may need a postprocessor for cleanup.

mesmith75:
Fixes #957 .

Not sure how much I like this but it works. The `ShareDir` can now have a list of files associated to it. If the `shareref` counter goes to 0 they should be removed along with the `sharedir` when the job is removed. This seemed more generic than specifically checking of there is an `uploadedInput' and 'jobScriptsArchive' attribute of the job's application.

I changed GaudiExec to include both the job scripts and the cmake target in so both get removed. I need to do some more checks about what gets carried over when the job gets copied.


mesmith75:
Alternatively we could just do `j.application.unprepare()` at the start of the remove.

The `unprepare` function in GaudiExec will remove the `jobScripts` file and if the `sharedref` counter is 0 will remove the other file as well.

JinNamkung:
retest this please

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -294,6 +296,13 @@ def isExecutable(self):
         file are checked"""
         return self.executable or is_executable(expandfilename(self.name))
 
+    def removeAssociatedFiles(self):
+        """ Remove the files in the associated file list"""
+        for entry in self.associated_files:
+            if isinstance(entry, IGangaFile):
+                entry.remove()
+                self.associated_files.pop(entry)
```
egede: IS it not dangerous to remove entries from a list that you are iterating over?

python/Ganga/GPIDev/Lib/Job/Job.py
```diff
@@ -1758,6 +1758,8 @@ def doit(f):
                         self.application.decrementShareCounter(self.application.is_prepared.name)
                         for _ in self.subjobs:
                             self.application.decrementShareCounter(self.application.is_prepared.name)
+                    if self.application.getShareCounterVal(self.application.is_prepared.name)==0:
```
egede: Would it not be better to implement this inside the SharedDir object itself. It can just do it when its internal counter goes to zero. That would also prevent a potential race condition that exist with this solution with jobs being removed in parallel threads.

python/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -196,9 +196,13 @@ def unprepare(self, force=False):
         if self.is_prepared is not None:
             self.decrementShareCounter(self.is_prepared.name)
             self.is_prepared = None
+        if self.getShareCounterVal(self.is_prepared.name)==0 and self.uploadedInput:
```
egede: Putting this ability into the SharedDir object would also prevent that the remove operation has to be implemented in multiple places.

python/Ganga/GPIDev/Lib/File/File.py
mesmith75: Yes, good point.

python/Ganga/GPIDev/Lib/Job/Job.py
mesmith75: Yes I agree but the shareRef and the ShareDir objects are somewhat decoupled. I'll try and rewrite it to be more as one would expect.

rob-c:
@mesmith75 I've got a similar fix to this in another PR I've not yet uploaded. This was mainly due to having to fix other bugs around the XML and Parser which have recently gone away.

Just a quick comment:

There is already logic which counts the `ShareRef` in the `ShareRef` and I'm not sure this needs extending.
I know it requires extending the `ShareRef` class (specifically decrement and maybe increment) to take a `ShareDir` rather than the `shareddir` string but this I would argue is much cleaner as we can catch (and allow for) files which may be created without a parent Job.
This would mean you can make a call to `removeAssociatedFiles` within the decrement method https://github.com/ganga-devs/ganga/blob/0126d980d25e2bb95dcc36e4f8b84c9799700a5f/python/Ganga/GPIDev/Lib/Registry/PrepRegistry.py#L153

Also, would it be better to use a `GangaFileItem` in the schema for a list of IGangaFile objects as I don't forsee us wanting to add a `File` object to be managed by a `ShareDir`.

mesmith75:
@rob-c Yes, I think you are right it would be best to extend the shareref properly.

mesmith75:
Ok that is sorted.

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -294,6 +296,20 @@ def isExecutable(self):
         file are checked"""
         return self.executable or is_executable(expandfilename(self.name))
 
+    def removeAssociatedFiles(self):
+        """ Remove the files in the associated file list"""
+        for entry in list(self.associated_files):
```
egede: Should the list in `self.associated_files` not be replaced with an empty list (or `None`) at the end of this loop?

python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -146,10 +147,11 @@ class ShareDir(GangaObject):
 
     """
     _schema = Schema(Version(1, 0), {'name': SimpleItem(defvalue='', getter="_getName", doc='path to the file source'),
-                                     'subdir': SimpleItem(defvalue=os.curdir, doc='destination subdirectory (a relative path)')})
+                                     'subdir': SimpleItem(defvalue=os.curdir, doc='destination subdirectory (a relative path)'),
+                                     'associated_files': GangaFileItem(defvalue=[], typelist = [str, IGangaFile], doc='A list of files associated with the sharedir')})
 
     _category = 'shareddirs'
-    _exportmethods = ['add', 'ls', 'path']
+    _exportmethods = ['add', 'ls', 'path', 'removeAssociatedFiles', 'remove']
```
egede: Is there a reason for the `removeAssociatedFiles` to be exported to the GUI.

python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -269,7 +269,7 @@ def generateDiracInput(app):
     new_df = uploadLocalFile(job, os.path.basename(compressed_file), tmp_dir)
 
     app.uploadedInput = new_df
-
+    app.is_prepared.associated_files.append(DiracFile(lfn = new_df.lfn))
```
rob-c: Does this deprecate the `app.uploadedInput` now?
I think I'd say go for merge on this PR and worry about removing the `uploadedInput` later if it's duplicate into, I've no idea where or how extensively it's used.

python/Ganga/GPIDev/Lib/File/File.py
mesmith75: Not really other than in case anyone finds it useful. Perhaps if you want to remove the uploaded files without deleting the job itself for record keeping.

egede: Which is fine. Gave me another thought though. If the files have been removed remotely (through a manual clean of grid storage by the user), will this just move on silently? IF not, I think it should be changed to do so.

python/Ganga/GPIDev/Lib/File/File.py
mesmith75: yes, good point

python/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
mesmith75: Not quite - the RT handler still uses it.

python/Ganga/GPIDev/Lib/File/File.py
mesmith75: That is a good point - yes it does move on silently. The DiracFile remove just returns True if the file isn't there.

mesmith75:
Thanks for the comments. I've added in a few changes (changes to the `increase` method I missed) as well as some extras to the `TestShared` for this. If there are no objections I'll merge it.

egede:
Looks good. Please go ahead.

PR #1117, opened 2017-09-04
Save associated_files to xml

mesmith75:
This writes the list of `associated_files` from a `ShareDir` to an XML file in the shared directory. This means that if a job is copied and new files added, the original will know about them.

Also added to `GaudiExec` to check the existence of previously uploaded files in case users have been cleaning their grid areas between job submissions.

### DIFF
--- a/python/Ganga/GPIDev/Lib/File/File.py
+++ b/python/Ganga/GPIDev/Lib/File/File.py
@@ -300,10 +300,9 @@ class ShareDir(GangaObject):
 
     def getAssociatedFiles(self):
         if os.path.isfile(os.path.join(self.path(), 'associated_files.xml')):
-            fobj = open(os.path.join(self.path(), 'associated_files.xml'), "r")
-            tmpobj, errs = from_file(fobj)
-            self.associated_files = tmpobj
-            fobj.close()
+            with open(os.path.join(self.path(), 'associated_files.xml'), "r") as fobj:
+                tmpobj, errs = from_file(fobj)
+                self.associated_files = tmpobj
  
     def addAssociatedFile(self, newFile):
         self.getAssociatedFiles()
--- a/python/Ganga/GPIDev/Lib/File/File.py
+++ b/python/Ganga/GPIDev/Lib/File/File.py
@@ -299,12 +299,16 @@ class ShareDir(GangaObject):
         return self.executable or is_executable(expandfilename(self.name))
 
     def getAssociatedFiles(self):
+        """ Load the list of associated files from the saved XML. This is
+        necessary to keep the list consistent when copying jobs. """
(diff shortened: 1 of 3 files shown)


python/Ganga/GPIDev/Lib/File/File.py
```diff
@@ -296,13 +300,30 @@ def isExecutable(self):
         file are checked"""
         return self.executable or is_executable(expandfilename(self.name))
 
+    def getAssociatedFiles(self):
+        if os.path.isfile(os.path.join(self.path(), 'associated_files.xml')):
+            fobj = open(os.path.join(self.path(), 'associated_files.xml'), "r")
```
rob-c: I forget the naming but isn't this safer as `with open( ... ) :` rather than opening and closing the file handler?

mesmith75: good point.

mesmith75:
@egede yes - the `safe_save` has a lock in it but I will add to the other functions. Is the `synchronised` decorator sufficient here?

mesmith75:
I have added in the `@synchronised` decorator which I think should provide the necessary protection.

I have also added to the ShareRef to remove the `sharedir` from the list when the counter gets to 0. Currently old deleted sharedirs hang about in the prep registry.

rob-c:
@mesmith75 I'm only asking as I can't remember how this works now (been a while since I looked in detail at this).
Isn't the list of LFN which have been uploaded for a `SharedDir` saved with the same object on disk now? If that's the case isn't the object copied when the `ShareDir` object is copied?

I'm just left wondering why this is implemented this way but I could well be missing something.

mesmith75:
@rob-c Kind of - but the ShareDir is saved with job. If you copy a job then the new job has a copy of the `ShareDir` object so if you add an LFN to the copy the original doesn't know about it.

rob-c:
@mesmith75 I wonder if there is a mixing of temporary prepared LFN per job and shared LFN which don't change between jobs.
Would a non-copyable list attribute of ShareDir be better as the original job shouldn't know about future jobs files and if there is some prepared data specific to an individual job it shouldn't be exposed to future jobs I would have suspected?

mesmith75:
@rob-c I see your point - I guess technically this is not correct for the `ShareDir`. 

The thinking is that if you copy a GaudiExec job you end up with a new LFN for the job scripts which needs to be removed - but maybe that should be removed with the job rather than the `ShareDir` as it is specific to the job in question.

But also you may upload a new cmake tarball in case the old one has disappeared (which is fine as it is just uploading what was in the local `ShareDir` so the application itself will still be the same). For that it is a bit trickier because the new upload can still be copied for new jobs so it does need to be part of the `ShareDir` object.

PR #1122, opened 2017-09-12
Temporarily remove proxy stuff for Atlas

drmarkwslater:
As Ganga is almost entirely used for local submission in Atlas at the moment and I've been having serious problems getting the proxy stuff working correctly, I've decided to temporarily 'deactivate' the proxy parts in GangaAtlas until I can get a chance to properly fix them. This should allow Ganga to be used efficiently where it's most needed in Atlas.

### DIFF
--- a/python/GangaAtlas/Lib/ATLASDataset/DQ2Dataset.py
+++ b/python/GangaAtlas/Lib/ATLASDataset/DQ2Dataset.py
@@ -6,7 +6,7 @@
 ###############################################################################
 # A DQ2 dataset
 
-import sys, os, re, urllib, commands, imp, threading, time, fnmatch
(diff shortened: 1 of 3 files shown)


mesmith75:
I am not sure how much I like the idea of disabling things for a particular release. Is this absolutely necessary to get Ganga to work at all for ATLAS?

drmarkwslater:
In a short timescale, unfortunately yes. There's been lots of problems with the getting the VOMS commands to work properly with both the Athena setup and the Panda setup and as the main use-case for Ganga is local submission, I've been asked to disable it so at least it can work like that. My longer term plan is to only request a proxy when it's needed but that will take a while and they want to use it now for local submission.

drmarkwslater:
Thanks all!

mesmith75:
Ok. I'll merge this and start the release.

PR #1126, opened 2017-09-15
Release notes 6.7.1

mesmith75:


mesmith75:
As 6.7.0 has only been out for a day I copied across those release notes as well.

PR #1128, opened 2017-09-28
Set parallel_submit to True by default in job

mesmith75:
Seems worthwhile setting this to True by default.

Do we want to change the default in the `IBackend` as well?

### DIFF
--- a/python/Ganga/GPIDev/Adapters/IBackend.py
+++ b/python/Ganga/GPIDev/Adapters/IBackend.py
@@ -72,21 +72,11 @@ class IBackend(GangaObject):
         try:
             sj.updateStatus('submitting')
             if b.submit(sc, master_input_sandbox):
-                sj.updateStatus('submitted', update_master = False)
                 sj.info.increment()
                 return 1
             else:
                 raise IncompleteJobSubmissionError(fqid, 'submission failed')
         except Exception as err:
-            #from Ganga.Utility.logging import log_user_exception
-            sj.updateStatus('failed')
-#            sj.updateStatus('failed', update_master = False)
-            #from Ganga.Core.exceptions import GangaException
-            #if isinstance(err, GangaException):
-            #    logger.error("%s" % err)
-            #    #log_user_exception(logger, debug=True)
-            #else:
-            #    #log_user_exception(logger, debug=False)
             logger.error("Parallel Job Submission Failed: %s" % err)
             return 0
         finally:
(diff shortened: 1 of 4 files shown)


mesmith75:
retest this please

mesmith75:
retest this please

mesmith75:
retest this please

mesmith75:
I think this works now - I had to redo a fair amount of the logic in `IBackend` and `Job`. I also overload the `master_submit` method in the Interactive backend to prevent it submitting in parallel ever.

python/Ganga/GPIDev/Adapters/IBackend.py
```diff
@@ -72,23 +72,20 @@ def _parallel_submit(self, b, sj, sc, master_input_sandbox, fqid, logger):
...
             sj.updateStatus('submitting')
             if b.submit(sc, master_input_sandbox):
-                sj.updateStatus('submitted')
                 sj.info.increment()
+                return 1
             else:
                 raise IncompleteJobSubmissionError(fqid, 'submission failed')
         except Exception as err:
-            #from Ganga.Utility.logging import log_user_exception
-            sj.updateStatus('failed')
```
rob-c: Shouldn't this update the master job to failed if some part of the whole set of jobs fails to submit?

python/Ganga/test/GPI/Bugs/TestSavannah17080.py
```diff
@@ -12,12 +14,36 @@ def test_CondorConfigDefaults(self):
         j.backend= TestSubmitter()
         b = j.splitter.backs[1]
         b.fail = 'submit'
+        j.parallel_submit = False
 
         assert j.status == 'new'
 
-        j.submit(keep_going=False)
-        assert j.status in ['submitted', 'running']
+        with pytest.raises(Exception):
```
rob-c: Can we be a little more specific with the type of exception raised here in a test which is known to fail?

mesmith75: yes

python/Ganga/GPIDev/Adapters/IBackend.py
mesmith75: Yes but not here. That is taken care of in Job.py which catched the exception raised here.

mesmith75:
retest this please

mesmith75:
retest this please

PR #1131, opened 2017-09-29
Speed up Dirac submission

### LINKED ISSUE #1129: Slow submission speed
I have been looking into how long it takes jobs to submit to DIrac.

In total it is about 5 seconds for Ganga, 8 seconds for Dirac on the command line and about 0.1s for the API in a standalone python session.

For us most of the time is taken for the imports of Dirac etc at the start of the `dirac-script.py`, about 4-5s and then dirac.submit(j) is less than a second.

@egede suggested we put several subjobs into one `dirac-script.py`.  Is there anything else we can be doing to speed this up?

mesmith75:
Fixes #1129 .
This just puts each submission into the same process. Currently goes from 130s to 23s (of which it looks like 16s are other overheads so only 7s for the subjob submission) for 23 subjobs.

Needs tidying and potentially new script templates although that will also affect RTHandlers.

### DIFF
--- a/python/GangaDirac/Lib/Backends/DiracBase.py
+++ b/python/GangaDirac/Lib/Backends/DiracBase.py
@@ -281,6 +281,9 @@ class DiracBase(IBackend):
                 sjScript = sjScript.replace("output(result)", "resultdict.update({sjNo : result['Value']})")
                 if nSubjobs !=0 :
                     sjScript = sjScript.replace("from DIRAC.Core.Base.Script import parseCommandLine\nparseCommandLine()\n", "\n")
+                    sjScript = re.sub("from .*DIRAC\.Interfaces\.API.Dirac.* import Dirac.*","",sjScript)
+                    sjScript = re.sub("from .*DIRAC\.Interfaces\.API\..*Job import .*Job","",sjScript)
+                    sjScript = re.sub("dirac = Dirac.*\(\)","",sjScript)
                 masterScript += "\nsjNo=\'%s\'" % fqid
                 masterScript += sjScript
                 nSubjobs +=1
--- a/python/GangaDirac/Lib/Backends/DiracBase.py
+++ b/python/GangaDirac/Lib/Backends/DiracBase.py
@@ -9,6 +9,7 @@ import time
 import datetime
 import shutil
 import tempfile
+import math
 from collections import defaultdict
 from Ganga.GPIDev.Schema import Schema, Version, SimpleItem, ComponentItem
 from Ganga.GPIDev.Adapters.IBackend import IBackend, group_jobs_by_backend_credential
@@ -28,7 +29,7 @@ from Ganga.Core import monitoring_component
 configDirac = getConfig('DIRAC')
 logger = getLogger()
 regex = re.compile('[*?\[\]]')
-
+from Ganga.Utility.logic import implies
 
 class DiracBase(IBackend):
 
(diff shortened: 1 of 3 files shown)


mesmith75:
So I think this is ready. I decided against redoing the RTHandlers and script templates. Currently the job script therefore has a repetition of 
```python
from DIRAC.Interfaces.API.Job import Job
dirac = Dirac()
```
or similar for the LHCb apps. The subsequent loss of time should be minimal but it would be nice to get rid of it.

rob-c:
@mesmith75 Would it be insane to have a method which removes all repeated lines from the dirac script before submitting it as a work around to declaring Dirac over and over? (This could be a dangerous idea but I don't know if it might help)
The speed improvement in submitting jobs this way looks impressive.

Looking forward to trying this in the future 👍, only minor comment is are the `stripProxy` calls needed in code within the backend?

mesmith75:
@rob-c Sure, I've added in a bit to get rid of the repeated imports etc - just needed to discover regex.

This is ready now.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -28,7 +29,7 @@
 configDirac = getConfig('DIRAC')
 logger = getLogger()
 regex = re.compile('[*?\[\]]')
-
+from Ganga.Utility.logic import implies
```
egede: Move import up to other imports

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -95,6 +96,8 @@ class DiracBase(IBackend):
         'settings': SimpleItem(defvalue={'CPUTime': 2 * 86400},
                                doc='Settings for DIRAC job (e.g. CPUTime, BannedSites, etc.)'),
         'credential_requirements': ComponentItem('CredentialRequirement', defvalue=DiracProxy),
+        'blockSubmit' : SimpleItem(defvalue=True, 
+                               doc='Shall we use the experimental block submission?'),
```
egede: To call it experimental and then implement it as the default seems odd. I would just delete "experimental".

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -179,6 +182,196 @@ def _addition_sandbox_content(self, subjobconfig):
...
+            if upperlimit > len(rjobs) :
+                upperlimit = len(rjobs)
+            logger.info("Submitting subjobs %s to %s" % (i*nPerProcess, upperlimit-1))
+
+            #Either do the submission in parallel with threads or sequentially
+            if parallel_submit:
+                getQueues()._monitoring_threadpool.add_function(self._block_submit, (dirac_script_filename, nSubjobs))
+            else:
+                self._block_submit(dirac_script_filename, nSubjobs)
+            def subjob_status_check(rjobs):
```
egede: Defining a new method in the middle of another method, makes code hard to read.

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -179,6 +182,196 @@ def _addition_sandbox_content(self, subjobconfig):
...
+       the dirac-script into one large file that we then execute.
+        """
+        #If you want to go slowly use the regular master_submit:
+        if not self.blockSubmit:
+            return IBackend.master_submit(self, subjobconfigs, masterjobconfig, keep_joing, parallel_submit)
+
+        #Otherwise use the block submit. Much of this is copied from IBackend
+        logger.debug("SubJobConfigs: %s" % len(subjobconfigs))
+        logger.debug("rjobs: %s" % len(rjobs))
+        assert(implies(rjobs, len(subjobconfigs) == len(rjobs)))
```
egede: We should not use assert in the production code.

mesmith75:
Thanks for the review - I have addressed those issues and tidied a bit further. If there are no objections I'll merge this today.

Issue #1054, opened 2017-07-07
Custom platform isn't passed from GaudiExec to DIRAC


rob-c:
I suspect from a quick example job that the correct platform isn't being translated into the jdl for LHCBDIRAC jobs.

I think this is simply because the JDL doesn't have a `###PLATFORM###` substitute string but it is being passed from the `GaudiExecRTHandler`.

egede:
@rob-c I think this is irrelevant. The wrapper we write is setting up the architecture through an `LbLogin` command. Can this just be closed?

rob-c:
@egede I would disagree. We don't know what the LHCbDIRAC pilot does to potentialy prepare the environment before we try to source the correct CMTCONFIG from cvmfs. It's only 2 or 3 lines of well placed code and it means we can rule this out as a potential source of bugs in the future, esp if Marco is thinking of multi-platform payloads.

mesmith75:
Apparently this is still an issue - people can't run the latest Brunel on the grid.

egede:
I thought that the LHCb grid at the moment only allowed a single architecture, `x86_64-slc6-gcc49-opt`, and that anything else would fail.

mesmith75:
This was fixed by a change to LbScripts (from v9r1p10), so I'll close this.

PR #1134, opened 2017-10-12
Hard-code LHCbDirac platform for gcc49

mesmith75:
Currently this is the only platform LHCbDirac is available on. This just allows people to submit jobs with Ganga from any platform.

Note: #1054 has actually been fixed by a change to LbScripts.


mesmith75:
Could someone please approve this so I can start the release?

PR #1138, opened 2017-10-17
Fix Dirac bulkSubmit resubmission

mesmith75:
Some new code is needed to resubmit Dirac jobs that were initially submitted with bulkSubmit. This still submits them one at a time, but users shouldn't be resubmitting too many.

It takes the original submission scripts and extract the necessary parts.

### DIFF
--- a/python/GangaDirac/Lib/Backends/DiracBase.py
+++ b/python/GangaDirac/Lib/Backends/DiracBase.py
@@ -556,6 +556,8 @@ class DiracBase(IBackend):
             if fnmatch.fnmatch(fileName, 'dirac-script-*.py'):
                 diracScriptFiles.append(fileName)
 
+        new_script_filename = ''
+
         for diracScript in diracScriptFiles:
             script_path = os.path.join(scriptDir, diracScript)
             # Check old script
@@ -602,6 +604,9 @@ class DiracBase(IBackend):
                 f.write(new_script)
             # Break the loop now we have written the new script
             break
+
+        if new_script_filename == '':
+            raise BackendError('Dirac', 'Script for job number %s not found. Resubmission failed.' % j.fqid)
 
         return self._block_submit(new_script_filename, 1)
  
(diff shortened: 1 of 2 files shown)


python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,6 +541,70 @@ def _resubmit(self):
...
+                diracScriptFiles.append(fileName)
+
+        for diracScript in diracScriptFiles:
+            script_path = os.path.join(scriptDir, diracScript)
+            # Check old script
+            if not os.path.exists(script_path):
+                raise BackendError('Dirac', 'No "dirac-script.py" found in j.inputdir')
+
+            # Read old script
+            f = open(script_path, 'r')
```
alexanderrichards: can we use open as a context here. ie.
```python
with open(script_path, 'r') as f:
    script = f.read()

This ensures that close is called properly even in the case of an exception.
```

python/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -536,6 +541,70 @@ def _resubmit(self):
...
+                if type(value) is str:
+                    template = '%s.set%s("%s")\n'
+                else:
+                    template = '%s.set%s(%s)\n'
+                new_script += template % (job_ident, str(_key), str(value))
+            new_script += newScript[newScript.find('# user settings -->'):]
+
+            # Save new script
+            new_script_filename = os.path.join(j.getInputWorkspace().getPath(), 'dirac-script.py')
+            f = open(new_script_filename, 'w')
```
alexanderrichards: again open as a context:
```python
with open(new_script_filename, 'w') as f:
    f.write(new_script)
```

I think then flush should be called automatically. In fact isn't it called automatically already in either the write or close. Do we need a separate flush here?

mesmith75: Thanks, good point; I should have updated these when I copied them.

PR #1145, opened 2017-10-24
Improve DIRAC job finalisation

mesmith75:
New option to finalise subjobs when all have completed. Also make the download of the output sandbox optional. Should be quicker.


mesmith75:
This looks to be doing what I want. The `finaliseOnMaster` option will wait for all the subjobs to in some way complete (including failing and being killed) before finalising everything in one go.
There are some helper functions that will finalise all the completed subjobs if there are others still running, as well as a function exported to the GPI to do something similar.

I also made the downloading of the output sandbox optional - I was wrong it isn't required for locating the output data.

mesmith75:
retest this please

mesmith75:
This still needs some work before we can merge.

mesmith75:

egede:
I think this can just go ahead. It will be a fairly large change for LHCb users (even running jobs will be affected as the default is changing in the schema). So it should be advertised well.

mesmith75:
@egede Good point - I have made the defaults configurable. I guess only those who know what they are doing should be trying this, so the default is for no change in behaviour.

Issue #1115, opened 2017-08-30
out of the box dirac troubles


rob-c:
I was hoping to use ganga to test file storage here at ECDF using the standard gridpp tools.
(plan was to test the `DiracFile.put` to Edinburgh to test if the xrootd storage protocol is working/supported)

Unfortunately I'm seeing the following problems:

From: `/cvmfs/ganga.cern.ch/runGanga-dirac.sh`:
```
[11:15:47]
Ganga In [4]: j=Job();j.backend=Dirac();j.submit()
INFO     submitting job 0
INFO     job 0 status changed to "submitting"
INFO     Preparing Executable application.
INFO     Created shared directory: conf-27680420-c610-442b-9d49-627550c0d2c9
INFO     Preparing subjobs
INFO     Setting Command to be: '['echo', 'Hello World']'
INFO     submitting job 0 to Dirac backend
WARNING  Required credential [DiracProxy(group='gridpp_user', encodeDefaultProxyFileName=True)] not found in store
ERROR    'DiracEnvSource' config variable empty
/bin/sh: dirac-proxy-init: command not found
/bin/sh: dirac-proxy-init: command not found
/bin/sh: dirac-proxy-init: command not found
/bin/sh: dirac-proxy-init: command not found
/bin/sh: dirac-proxy-init: command not found
ERROR    CredentialsError: Proxy object cannot satisfy its own requirements
ERROR    JobManagerError: error during submit
ERROR    JobManagerError: error during submit ... reverting job 0 to the new status
INFO     job 0 status changed to "new"
ERROR    JobError: Error: JobManagerError: error during submit
```

Also this fails to launch _out of the box_ when .gangarc doesn't exist due to:
```
-bash-4.1$ /cvmfs/ganga.cern.ch/runGanga-dirac.sh

*** Welcome to Ganga ***
Version: 6.6.4
Documentation and support: http://cern.ch/ganga
Type help() or help('index') for online help.

This is free software (GPL), and you are welcome to redistribute it
under certain conditions; type license() for details.


INFO     reading config file /cvmfs/ganga.cern.ch/Dirac.ini
ERROR    problems with loading Plugin /cvmfs/ganga.cern.ch/Ganga/install/6.6.4/python/GangaDirac
ERROR    Reason: GangaValueError: DIRAC Proxy `group` is not set. Set this in ~/.gangarc in `[defaults_DiracProxy]/group`
[...]

PR #1147, opened 2017-10-27
Vanilla dirac fix

alexanderrichards:
This PR should hopefully simplify the process of using vanilla GangaDirac for new users. It also includes a README file in the root of the GangaDirac package with setup instructions.

Note this is not quite ready to merge yet


python/GangaDirac/Dirac.ini
```diff
@@ -1,6 +1,6 @@
 [Configuration]
 
-RUNTIME_PATH = GangaDirac
+RUNTIME_PATH = GangaGaudi:GangaDirac
```
rob-c: Just out of curiosity what part is pulling this in?

alexanderrichards: This bit I'm not sure I'm going to include in the final PR yet, hence the "Note this is not quite ready to merge yet" however it is true that GangaDirac has a dependence on GangaGaudi which seems like it's only to do with simple functions. This leads me to think ultimately we should sever this link. If I run vanilla in a new cloud instance I see:

```python
Ganga In [1]: Job(backend=Dirac()).submit()
INFO     submitting job 0
INFO     job 0 status changed to "submitting"
INFO     Preparing Executable application.
INFO     Created shared directory: conf-2e87e008-3786-4c8e-a788-00a64ca3fc23
ERROR    ConfigError: attempt to create a configuration section [defaults_GaudiXMLSummary] after bootstrap 
ERROR    ConfigError: attempt to create a configuration section [defaults_GaudiXMLSummary] after bootstrap  ... reverting job 0 to the new status
INFO     job 0 status changed to "new"
ERROR    JobError: Error: ConfigError: attempt to create a configuration section [defaults_GaudiXMLSummary] after bootstrap 
```

So I must need to explicitly include the GangaGaudi plugin to get these config sections created at bootstrap time. However in testing im not sure that this does solve the problem so hence I don't think I'll be keeping this particular commit

alexanderrichards: indeed the LHCb.ini includes:
```ini
[Configuration]
RUNTIME_PATH=GangaDirac:GangaGaudi:GangaLHCb
```

alexanderrichards: whats concerning me is I'm sure I got DIRAC jobs running without this when setting up the docker images so maybe it's my cloud node.

alexanderrichards: I didn't need this in the end.

alexanderrichards:
Right, this seems ok to me and on a clean cloud node allows me to submit vanilla DIRAC jobs following the instructions in the new README.md file.
Let's see how the testing goes but all being well this can be merged.

alexanderrichards:
commit 58745e165e08bf897add5abde263ab1509f11da0 addresses issue https://github.com/ganga-devs/ganga/issues/1115 but seems to break GangaLHCb

rob-c:
@alexanderrichards This exception was intended to be caught at a higher level as nothing in DIRAC can work without knowing the VO...

alexanderrichards:
The VO is not the problem, I set this, see the README file added in this PR. The problem is that while moving the imports in GangaGaudi back to the original location ( and indeed the matching place that GangaDirac and GangaLHCb does it) fixes all issues for vanilla DIRAC users it somehow breaks for LHCb which I suspect is the reason they were moved in the first place. I think our job here is maybe twofold

1) understand the LHCb  error that led to them being moved and find a better fix.
2) as you other PR does move the relevant code from GangaGaudi to GangaDirac thereby breaking the dependence
3) if GangaGaudi is nolonger needed in LHCb then remove the dependence.

While 2) and 3) will surfice by themselves I think that all 3 steps are needed to fully say that we have fixed this.

p.s. can you link the PR mentioned in 2) here so we can keep track of the two in parallel.

rob-c:
@alexanderrichards  #1155 Should fix a lot of the problems that users see when trying to use:
`/cvmfs/ganga.cern.ch/runGanga-dirac.sh` but it appears that the `dirac_ui` on cvmfs is out of date again so I don't know how much use this is to people over just writing a script which installs updates the script locally for users then launches ganga from cvmfs.

alexanderrichards:
thanks @rob-c I'll have a look. The purpose of this PR is to make vanilla dirac easier to set up for everyone so there is no specific depedence on CVMFS

rob-c:
I don't know why this PR should effect LHCb but I can't help you there, I've avoided the whole dependency hell between GangaLHCb/GangaGaudi. I've tried to clean it up in the past but simply gave up as everything there is fairly muddled (not to mention historical GPI related problems when this was changed which I believe have now gone away).

alexanderrichards:
retest this please

alexanderrichards:
retest this please

alexanderrichards:
retest this please

egede:
@alexanderrichards Can you take a look as this and resolve the merge conflicts?

alexanderrichards:
will do

egede:
Still relevant?

alexanderrichards:
lets delete it and I can re-open it if needed.

PR #1150, opened 2017-11-01
DIRAC resubmit bugfix

mesmith75:
There is a slight bug in that jobs submitted to DIRAC with old ganga versions cannot be resubmitted with the latest - the `blockSubmit` attribute gets added with the default `true` to these old jobs so it tries to resubmit with the wrong method.

The fix just checks for the existence of the new-style `dirac-script-0.py` to confirm the job was originally submitted with `blockSubmit`, otherwise reverting to the old resubmit code.


egede:
Does that mean that a `j.backend._impl.blockSubmit=False` before resubmit would work as a workaround for the few that are affected?

mesmith75:
yes

PR #1155, opened 2017-11-09
This should fix problems I see with the vanilla Dirac out of the box from the ganga side

rob-c:
This PR removes the scripts from GangaGaudi and moves them to GangaDirac where they're needed.

I've verified I am able to do the following with this PR:
 1) Submit an old style LHCb job to LHCbDirac
 2) Submit a new style LHCb job to LHCbDirac
 3) Construct a sane looking 'hello world' job which attempts to submit to the Imperial DIRAC
 4) Attempt to correctly create/destroy a proxy using the gridpp VO

Again I can't submit a job through the CVMFS configs (using the config from: `/cvmfs/ganga.cern.ch/runGanga-dirac.sh` because as far as I can tell the voms information on cvmf is out of date and I can't generate an appropriate proxy certificate from the gridpp VO to use with Imperial's DIRAC using the configs in this script.

FYI:

For submitting to Imperial's Dirac I was using:

`.gangarc`:
```
[defaults_DiracProxy]
group=gridpp_user
```
and:
`export GANGA_CONFIG_PATH=/cvmfs/ganga.cern.ch/Dirac.ini`

which used to work.


rob-c:
@alexanderrichards If a short script can be written to get/manage the dirac_ui locally for users ahead of launching ganga then we can drop my changes to ` python/GangaDirac/Lib/Utilities/DiracUtilities.py` as this is only needed for loading now defunct configs from cvmfs.

alexanderrichards:
agreed, I don't like relying on CVMFS in our code as some people simply wont have it and again it makes us more CERN centric. It's actually a pretty simple set of steps and are detailed here:
https://www.gridpp.ac.uk/wiki/Quick_Guide_to_Dirac#Dirac_client_installation

rob-c:
I'm going to merge this this-evening/tomorrow if there aren't objections as this does get in the way of using gridpp dirac without first pulling the audi module which is officially unsupported from the lhcb perspective anyway.

mesmith75:
I think this is a good thing. @alexanderrichards  How does this fit with your PR #1147 ?

twhyntie:
Hello,

Just to add some notes from issue #1176, I've tried using this version of Ganga on my CernVM _with the CVMFS DIRAC script_ and I'm able to submit jobs via the GridPP DIRAC backend with no "GaudiBadness" 😄 

Cheers, Tom

alexanderrichards:
> I think this is a good thing. @alexanderrichards How does this fit with your PR #1147 ?

I suspect that this should be compatible with the changes in #1147 but I'll have to check again

PR #1157, opened 2017-11-13
Block submit fix

### LINKED ISSUE #1156: ArgSplitter uses final argument for all subjobs
When submitting jobs using `ArgSplitter`, all subjobs are submitted with the arguments of the final job (`[109]` in the example).

### Example:
```bash
echo '#!/bin/bash' > example.sh
echo 'date >> Example.${1}.txt' >> example.sh
echo 'echo ${1} >> Example.${1}.txt' >> example.sh
echo 'cat "Example.${1}.txt"' >> example.sh
chmod +x example.sh
```
```python
j = Job()
j.application = Executable(exe=File('example.sh'))
j.splitter = ArgSplitter(args=[[i] for i in range(100, 110)])
j.outputfiles = [DiracFile(namePattern='Example.*.txt')]
j.backend = Dirac()
j.submit()
```

Follow up to the "Failure to upload output data from Executable job" thread on `lhcb-distributed-anaysis`.

mesmith75:
Fixes #1156

### DIFF
--- a/python/GangaDirac/Lib/Backends/DiracBase.py
+++ b/python/GangaDirac/Lib/Backends/DiracBase.py
@@ -269,7 +269,7 @@ class DiracBase(IBackend):
                 sj.updateStatus('submitting')
                 fqid = sj.getFQID('.')
                 #Change the output of the job script for our own ends. This is a bit of a hack but it saves having to rewrite every RTHandler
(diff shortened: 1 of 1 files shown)


mesmith75:
Yes - I think we need to get the nightlies working properly again.

PR #1159, opened 2017-11-16
Raise exception for failed blockSubmit

mesmith75:
Raise a nice exception if something goes wrong with the block submission.


mesmith75:
Taken care of by #1160

PR #1160, opened 2017-11-17
Return Exception if call to Dirac API is not returning a dictionary

mesmith75:
Added a check that the result of the put is a dict - if not raise an exception. This seems to be where issues creep in if Dirac things are going awry so it would be nice to get the output properly.

### DIFF
--- a/python/GangaDirac/Lib/Files/DiracFile.py
+++ b/python/GangaDirac/Lib/Files/DiracFile.py
@@ -63,7 +63,7 @@ class DiracFile(IGangaFile):
         To upload a file and make it available on a workernode:
 
         df = DiracFile('/path/to/some/local/file')
-        df.put('CERN-USER')
+        df.put(uploadSE = 'CERN-USER')
 
         j=Job( ... )
         j.inputfiles = [df]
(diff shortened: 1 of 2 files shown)


python/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -766,6 +766,9 @@ def put(self, lfn='', force=False, uploadSE="", replicate=False):
                 self.failureReason += '\n' + failureReason
                 continue
 
+            if not isinstance(stdout, dict):
```
egede: Would it be better to insert this `raise` into the `execute` method from `GangaDirac.Lib.Utilities.DiracUtilities`? In that way the sanity check will be done for all calls and not just for this specific one.

mesmith75: I don't think so - from the Dirac call you could conceivably expect strings or anything. It is just for this specific command we expect a dict.

mesmith75: Actually I think everything should come back as a dict - I'll change that and close #1159

PR #1168, opened 2017-11-24
Test jenkins

mesmith75:


mesmith75:
@alexanderrichards @egede Would there be any objections to merging this so we can get a new release out and restore the PR tests until I have time to figure out the Jenkins issue?

PR #1171, opened 2017-12-01
Set DIRAC CPU time to 2 weeks

mesmith75:


python/Ganga/__init__.py
```diff
@@ -26,7 +26,7 @@ def getLCGRootPath():
 # ------------------------------------------------
 # store Ganga version based on new git tag for this file
 _gangaVersion = '6.7.3'
-_development = False
+_development = True
```
egede: Spurious change here?

PR #1173, opened 2017-12-02
Replicate GaudiExec files

### LINKED ISSUE #1133: Create replica of uploaded GaudiExec files
We should request a replica of the uploaded GaudiExec files when we submit a job and maybe get rid of one in the postprocessor.

mesmith75:
Fixes #1133 

Also remove replicas in the postprocessor.


rob-c:
I'm curious what's the motivation for this, is it to avoid a T1 going into downtime?
These builds are potentially  a few hundred Mb in size so should we consider if we want to replicate them by default?
They have no effect on where the user job runs as the file is transferred to the WN ahead of the job running.

mesmith75:
It was requested by Dirac people.

PR #1177, opened 2017-12-08
Add proper format check for nightly

drmarkwslater:
Recently, ATLAS have added the AtlasBuildStamp env variable to all setups which makes the check for a nightly fail. This does a proper check for this timestamp format.


drmarkwslater:
This bug is a bit of show stopper in certain instances so could this be merged and released ASAP please? THanks!

PR #1178, opened 2017-12-08
Fix for race condition

rob-c:
User on LHCb dast list reported:

```
ERROR    Exception raised executing 'updateMonitoringInformation' in Thread 'Backend Monitor':
  File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.3/python/Ganga/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 115, in __worker_thread
  File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.3/python/GangaDirac/Lib/Backends/DiracBase.py", line 1242, in updateMonitoringInformation
    DiracBase.monitor_dirac_running_jobs(monitor_jobs_group, finalised_statuses)
  File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.3/python/GangaDirac/Lib/Backends/DiracBase.py", line 1204, in monitor_dirac_running_jobs
    DiracBase.requeue_dirac_finished_jobs(requeue_job_list, finalised_statuses)
  File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.3/python/GangaDirac/Lib/Backends/DiracBase.py", line 1081, in requeue_dirac_finished_jobs
    args=(j, finalised_statuses[j.backend.status]),
KeyError: 'Checking'
```

This should fix it

### DIFF
--- a/python/GangaDirac/Lib/Backends/DiracBase.py
+++ b/python/GangaDirac/Lib/Backends/DiracBase.py
@@ -1076,6 +1076,10 @@ class DiracBase(IBackend):
             if monitoring_component:
                 if monitoring_component.should_stop():
                     break
+            # Job has changed underneath us don't attempt to finalize
+            if j.backend.status not in finalised_statuses:
+                j.been_queued = False
+                continue
             if not configDirac['serializeBackend']:
                 getQueues()._monitoring_threadpool.add_function(DiracBase.job_finalisation,
                                                            args=(j, finalised_statuses[j.backend.status]),


drmarkwslater:
Looks OK - just wondering if the `j.been_queued = False` line is needed? Does this mean that the finalisation call might have been queued/started (thus setting  `j.been_queued = True`) but the job been shifted back to 'Checking'? Would that be a problem?

rob-c:
Tbh the `been_queued` is only used internally to DiracBase to keep track of jobs which have been put into a _to be completed_ state and is used to avoid calling finalize on a job twice.
It's a bit messy but was a quick solution to the problem.
This just avoids that a job has been already resubmitted (potentially in DIRAC?) and we can't work out if it failed or not because it hasn't.

drmarkwslater:
OK - still not 100% sure I understand but I can't see it would do any harm!

PR #1188, opened 2018-01-12
Move folders to python

mesmith75:
This moves the `release`, `templates` and `docs` folders to be under the `python` folder so when ganga is installed with `pip install` everything ends up in the same place relative to each other.

This is necessary as the `GangaCore` etc folders are the python modules so they individually get installed rather than remaining under one `python` or `ganga/python` module. The other fix would be to change all the imports for `from ganga.python.GangaCore` etc but this seems less painful.


mesmith75:
The change to setup.py creates a new `release` package. The package-data is then the release notes.

mesmith75:
`release`->`python/GangaRelease`
`templates`->`python/GangaTemplates`
`doc`->`python/GangaDoc`

These are all now included in the pip installation which should more closely match the actual structure in the repository.

It is not very neat - ideally we would keep the `Ganga/python/Ganga...` structure but that would require either rewriting all the imports or appending to the `syspath` when ganga starts.

drmarkwslater:
Just a few thoughts:

- I'm guessing this will affect how the readthedocs site is constructed? 
- I don't **think** (though someone can correct me) that we need the `templates` folder anymore
- Where do other pip installed projects put their docs/release notes? Is this the usual way? I'm tempted to ask if it's worth doing this at all if you can get the docs and release notes online anyway...

On that last one, I am probably just not understanding everything that's involved 😄

mesmith75:
@drmarkwslater 

- Yes I guess it would, although I don't know how that is done currently
- I think you are right - I couldn't see anything in the code but I kept it just in case.
- I think it is usually done as we do now but the problem is our pip installation doesn't keep the `Ganga/python` folders. So each individual package (`GangaCore`. `GangaAtlas`) etc is put directly into the `site-packages` directory. Therefore to keep the git repository and pip installations the same I figured the release notes should be moved, otherwise one or the other won't be able to find them. Really we should do it properly, install `Ganga/python` and change all the imports to `from Ganga.python.GangaCore` etc but that would be painful.

I think it is nice to keep the release notes in the ganga startup - but perhaps the docs don't need installing as well. I am open to not installing them - but ultimately I think we want all installations to be done via pip.

mesmith75:
Replaced by #1198

PR #1190, opened 2018-01-15
Warning on GaudiExec upload failure

mesmith75:
Change the exception raise to a warning when the upload of the GaudiExec tarball fails so another SE is tried. There is already a check in case no SEs work.


mesmith75:
retest this please

mesmith75:
retest this please

egede (approved):

Issue #1174, opened 2017-12-04
Proposal for improving handling of broken (sub)jobs


rob-c:
Hi all,

There are some users again reporting that the monitoring isn't picking up DIRAC jobs with subjobs in the `submitting` state.

I can't think of any meaningful way that _any_ of the current backends can deal with monitoring a job in the `submitting` state. I could be wrong, but to me if the job is in this state in the XML it indicates that something failed.
If a (sub)job has been loaded from disk and is in this state then the (sub)job is, I suspect by definition in a bad state. To get here something has crashed/failed or ganga was force exited or killed during the job submission so some or all of the monitoring information has probably been lost.

Can we consider force failing a job automatically if it's loaded from disk in a `submittting` state?

These (sub)jobs seem to be confusing the monitoring as some people are still seeing ganga struggling to update jobs with a subjob in the `submitting` state.

I think a short PR could accomplish this from the XML layer but what are people's opinions on this?

Rob

mesmith75:
@rob-c I think that is a good idea - we already force the job to fail if some subjobs failed to submit.

mesmith75:
Fixed by #1193

PR #1193, opened 2018-01-16
Adding fix for submitting jobs which never get picked up

### LINKED ISSUE #1060: Ganga not monitoring 'submitting' jobs
I have a feeling this may be a duplicate bug but I can't find anything open if it is.

Ganga doesn't monitor jobs in a 'submitting' state. We briefly did but now this appears to not be working again. I don't know if this is because we now more strongly enforce job state transitions or not but it would be desirable to monitor large jobs which take a long time to submit.

rob-c:
Fixes #1060 #1174

After the jobs repo has been launched (and all plugins loaded) a quick run through the jobs repo looking for (sub)jobs in a submitting state is done and the jobs are force_failed as the (sub)job is in an incoherent state which we should fail and resubmit by construction.

This should fix some monitoring bugs


egede:
@rob-c Are you confident that this could not somehow trigger the lazy-loading to load a huge numbver of jobs?

rob-c:
I will add some code to only load the last 5 jobs in this state but if there are many jobs in this state in a repo the user will have none of them being picked up by Monitoring best I can tell so this shouldn't cause too much harm. Plus we have lazy loading on the subjobs now and only the to-be failed subjob is ever loaded from disk.

egede:
@rob-c Your reply is a bit mixed here. Do you think a further change is required to this PR or not?

rob-c:
I think a limit of the last 5 `submitting` being loaded on startup is sensible so requires some minor work to get a reverse subset from the jobs list and to perform cleanup over this.

mesmith75:
test

mesmith75:
test this please

mesmith75:
test this please

rob-c:
@mesmith75 The checks should now no longer run during the tests as they do break testable functionality by design.

ganga/GangaCore/GPIDev/Lib/Registry/JobRegistry.py
```diff
@@ -84,6 +87,71 @@ def startup(self):
...
+        """
+            This code checks for jobs which are in the submitting state
+            If a job is 'submitting' an call to force_status('failed') is made
+            This should be enough to fix some future issues with monitoring """
+
+        def _shouldAutoKill(this_job):
+            """ This checks to see if the job is in a transistional state
+                By construction jobs in this state on disk need to be failed
+            """
+            return this_job.status in ["submitting", "completing"]
```
mesmith75, line 100: Do we really want to include `completing` here. If ganga crashed whilst the job is finalising then surely we can push back to `submitted` and it will do the finalising again?

rob-c: If the job gets stuck in `completing` we don't want to start spinning this round in circles trying to manage the job. It's much better to mark it as failed and have the user manually intervene if they know the job ran successfully through the DIRAC web-UI.
This could be made configurable but I think there's a risk of having a job running around in a loop getting stuck not being parsed due to some external problem to ganga.

rob-c: I've spent some time with a user recently and we've confirmed that this seems to be the best approach as it allows ganga to get back into a helpful state whilst signalling to the user there is a potential problem with the subjob.

rob-c:
Can someone review this please

mesmith75:
test this please

mesmith75:
@rob-c Looks fine to me. Can we get this and #1225 merged so users can have a test?

rob-c:
OK, I'm happy to but I'm still seeing:
```
 Review required
At least 1 approving review is required by reviewers with write access. 
```

mesmith75 (approved):

mesmith75:
@rob-c Oops -sorted now.

PR #1194, opened 2018-01-16
Hardening the code against Ctrl-C events

rob-c:
These are relatively minor changes which are designed to harden the codebase against Ctrl-C events.

I think a user should be allowed (but discouraged) from exiting some running code within ganga and they shouldn't be shown an internal stack trace so I've added some catches explicitly for the `KeyboardInterrupt` which makes ganga behave nicer under various Ctrl-C events.


python/GangaCore/Core/InternalServices/ShutdownManager.py
```diff
@@ -31,10 +31,16 @@ def register_exitfunc():
     registereing this  300 times is just bad...
     """
     if not register_exitfunc._has_registered:
-        atexit.register(_ganga_run_exitfuncs)
+        atexit.register(__ganga_run_exitfuncs)
         register_exitfunc._has_registered = True
 register_exitfunc._has_registered = False
 
+def __ganga_run_exitfuncs():
```
egede: So we have a function called `__ganga_run_exitfuncs` and one called `_ganga_run_exitfuncs`. It can't possibly be a good idea to have two functions with the same name apart from the number of underscores to begin with,

egede (commented):
So this means that if a user types ^C they just get a normal print back? My worry is that the user might then assume that everything is OK for continuing - while in effect they have left the session in some inconsistent state.

python/GangaCore/Core/InternalServices/ShutdownManager.py
rob-c: Yeah I could probably name this better, I tried a wrapper but bad errors happened for me

rob-c:
I would suggest I add a big red error that the user hit a Ctrl+C and that we have aborted what we were doing in that case. having a bit error is still better than a stack trace so long it goes past the end of my interactive screen history I think so this should help,
plus it prevents scaring people who go to stop a submission of say many hundreds of jobs which isn't wanted and then they see a huge stack trace which indicates they've broken something (they may well have but I think showing them the guts of ganga doesn't help)

egede:
Did you get time to update this with a warning?

rob-c:
I've not had a chance to do cleanup my developments further yet, I'll get around to this next week after finishing some dev work on another project

mesmith75:
test this please

mesmith75:
test

mesmith75:
test

mesmith75:
test this please

mesmith75 (changes requested):
Fine apart from the `exitfuncs` still needs renaming if possible.

mesmith75:
test this please

mesmith75:
@rob-c Is there any progress on getting these PRs to pass the tests? It would be nice to include them in the next bug-fix release.

mesmith75 (approved):

PR #1195, opened 2018-01-16
Hardening prompt and fixing shutdown policy bug

### LINKED ISSUE #1143: Ganga not closing automatically after running in batch form
I have a simple ganga script which downloads Dirac files from a given job (ganga-getfiles.py - code below). I have been calling this in a bash script which loops over several jobs and performs a few simple operations, like this (code simplified for clarity):

> for j in `seq 200 300` ; do
>     ganga ../common/ganga-getfiles.py $j
>     hadd myoutfile_${j}.root <gangadir>/${j}/*/output/file.root
>     rm <gangadir>/${j}/*/output/file.root
> done 

(where the python script ganga-getfiles.py is just a few ganga commands to grab root files from the grid for this job $j).

I have observed that if the script ganga-getfiles.py fails to download any Dirac files (e.g. if there are no matching files, or the job doesn't exist) then ganga will not automatically exit - it prompts me to answer:

"Job status update or output download still in progress (shutdown not completed after 5 seconds). 6 background thread(s) still running: ['SessionLockRefresher', 'JobRegistryFlusher', 'TaskRegistryFlusher', 'JobRegistryFlusher', 'JobRegistry_Monitor', None]. Do you want to force the exit (y/[n])?" 

which clearly causes problems when trying to perform automatic tasks. Is this expected behaviour? If not, what might be causing it?

Many thanks,
Mark

==== ganga-getfiles.py ====

> from Ganga.GPI import *
> import sys
> import inspect
> import os
> 
> 
> if len(sys.argv) not in (2,3):
>     sys.exit("Script requires the id of the job to use as inputdata and optionally name of a file containing LFNs to process.")
> 
> jobnum = int(sys.argv[1])
> 
> 
> for sj in jobs(jobnum).subjobs.select(status='completed'):
>     		for f in sj.outputfiles.get(DiracFile):
>         		if f.namePattern.endswith('BGun.root'):
>             			if f.lfn:    #ensure file has lfn
>                 			f.localDir = sj.outputdir
>                 			f.get()

rob-c:
Fixes #1143 

This fixes the fact that ganga when running in script (batch) mode doesn't exit cleanly which is a fairly serious bug which prevents people from scripting with ganga as a command line tool.

I think this potentially needs expanding to fix the `ganga -b` option but I don't think anyone uses this overly strongly.


rob-c:
I've just manually tested #1195 #1194 #1193 and had no failures anywhere any objections to me merging these?

rob-c:
I will add a warning to the Ctrl+C events in this case letting the user know something was aborted rather than running correctly.

egede (approved):

rob-c:
This really was the single PR where we should have been warning most stronlgy against a user hitting Ctrl+C. Hitting Ctrl+C here will likely __always__ lead to __repo corruption__ .
I'm still seeing emails where a user has done this after a job has finished submitting and then cry foul that their jobs in ganga are corrupt because it didn't shut down in 30s.

PR #1198, opened 2018-01-24
Rename python to ganga

mesmith75:
Test


egede:
Does 

import ganga

at the python prompt still work after this change (and if this higher level dir is in the PYTHONPATH)?

setup.py
```diff
@@ -79,8 +82,8 @@ def run(self):
       author_email='project-ganga-developers@cern.ch',
       license='GPL v2',
       scripts=['bin/ganga'],
-      package_dir={'': 'python'},
-      packages=find_packages('python'),
+      package_dir={'ganga': 'ganga'},
+      packages=allPackages,
```
alexanderrichards: I believe that these two lines can now be simply:
```python
packages=find_packages()
```
i.e. package_dir is no longer needed

mesmith75: yes

mesmith75:
So I think this about works. It should replace #1188 

Edit: except it breaks `import ganga`

mesmith75:
test this please

mesmith75:
test this please

mesmith75:
test this please

mesmith75:
test this please

mesmith75:
I *think* this is working fine now. The only change for python is `import ganga.ganga` rather than `import ganga`. Subsequently it is as normal, i.e. `print ganga.jobs` is the same.

egede (approved):

PR #1203, opened 2018-02-06
Add getReplicasForJobs for splitter

mesmith75:
We have been asked to use getReplicasForJobs in the DIRAC splitter - it should filter out copies at SEs that are not to be used for user jobs.


mesmith75:
test this please

egede:
In the logs from the Jenkins box I see

```
[GangaCore] docker: Error response from daemon: Conflict. The container name "/GangaCorePR-1203-2" is already in use by container "6b7cf919a89be3e4f816352c7a44afe736ea8261bd4b672b10889e4fa3efc18e". You have to remove (or rename) that container to be able to reuse that name.
```

so it seems like we still have some name conflict issue. I thought that PR number followed by the build number should be unique, but apparently not. Maybe we can add some kind of UUID?

mesmith75:
@egede I think It's because I've created two sets of pipelines. We need to get rid of one.

egede (approved):

Issue #1144, opened 2017-10-24
Un-friendly errors when there is a mis-configuration


cvazquezs:
I tried to run a LHCbTask (script below) and I got the following error:

```
ERROR    Exception occurred in task monitoring loop: <type 'exceptions.AttributeError'> 'NoneType' object has no attribute 'n_tosub'
The offending task was paused.
ERROR    Full traceback:
 Traceback (most recent call last):
   File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.2/python/Ganga/GPIDev/Lib/Tasks/TaskRegistry.py", line 92, in _thread_main
    p.update()
   File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.2/python/Ganga/GPIDev/Lib/Tasks/ITask.py", line 98, in update
    if trf.update() and not self.check_all_trfs:
   File "/cvmfs/ganga.cern.ch/Ganga/install/6.7.2/python/Ganga/GPIDev/Lib/Tasks/ITransform.py", line 224, in update
    unit_status["bad"], self._getParent().n_tosub())
 AttributeError: 'NoneType' object has no attribute 'n_tosub' 
```

At a first glance I had literally no idea on what was going on. I think this kind of 

errors should raise a more friendly and understandable error message which can provide useful feedback for the users.

Thanks!
Carlos

```
'''
Generate a local MC production for a given event type

@author Miguel Ramos Pernas
'''

#
#    L o c a l    M C   p r o d u c t i o n           [sim09b  /  2016 MC  /  Stripping26]
#______________________________________________

# Run 200k - 100k per magnet.

runInAFS = False

EventType     = "43900075"
fileType      = 'DST'
evtsPerJob    = 1000
nJobs         = 1 # 100
magnet        = 'Down'
gaussVersion  = 'v49r8'
booleVersion  = 'v30r2'#p1'
mooreVersion  = 'v25r4'
brunelVersion = 'v50r2'
dvVersion     = 'v42r5p1'

decays = {'43900075' : 'mH125mA10tA10'}

decay = decays[EventType]

if magnet == 'Up':
    magnetconds = 'CondMagUp.py'
    polarity    = 'mu'
else:
    magnetconds = 'CondMagDown.py'
    polarity    = 'md'

nEvents      = evtsPerJob*nJobs
print '***', nEvents, 'events will be generated ***'

task         = LHCbTask()
task.name    = decay + "_Mag" + magnet
task.comment = "Local Production task"
[...]

PR #1216, opened 2018-02-15
set should_init on a GangaObject to true by default

### LINKED ISSUE #1212: ITransform self._parent returns NoneType
Using ganga 6.7.4 I get the following error when running an ND280Task(). Sample code:

<pre>
hastings@kudu ~ 
% work/ganga-debug/install/6.7.4/bin/ganga

*** Welcome to Ganga ***
Version: 6.7.4
Type help() or help('index') for online help.

This is free software (GPL), and you are welcome to redistribute it
under certain conditions; type license() for details.

INFO     reading config file /home/hastings/.gangarc
tas
[10:51:03]

Ganga In [1]: execfile('nd280test.gpi')
INFO     Reading . ...

[10:51:07]
Ganga In [2]: ERROR    Exception occurred in task monitoring loop: <type 'exceptions.AttributeError'> 'NoneType' object has no attribute 'n_tosub'
The offending task was paused.
ERROR    Full traceback:
   File "/home/hastings/work/ganga-debug/install/6.7.4/python/Ganga/GPIDev/Lib/Tasks/TaskRegistry.py", line 92, in _thread_main
    p.update()
   File "/home/hastings/work/ganga-debug/install/6.7.4/python/Ganga/GPIDev/Lib/Tasks/ITask.py", line 98, in update
    if trf.update() and not self.check_all_trfs:
   File "/home/hastings/work/ganga-debug/install/6.7.4/python/Ganga/GPIDev/Lib/Tasks/ITransform.py", line 224, in update
    unit_status["bad"], self._getParent().n_tosub())
 AttributeError: 'NoneType' object has no attribute 'n_tosub'
</pre>
The problem seems to be because self._getParent() in ITransform.update() returns None.
[nd280test.gpi](https://github.com/ganga-devs/ganga/files/1718348/nd280test.gpi.gz) s a very stripped down version of python/GangaND280/examples/runND280RDP_tasks.py 
I place it here inline too:
<pre>t   = ND280Task()
trf = ND280Transform()
trf.application = runND280RDP()
d=ND280LocalDataset()
d.get_dataset('.','fakedata.mid.gz')
trf.addInputData(d)
t.appendTransform(trf)
t.float = 10
t.run()</pre>
Note that the file "fakedata.mid.gz" does not need to exist to reproduce the error.

This error does not occur with ganga 6.6.4 or earlier versions. It seems that perhaps something in the ND280Transform or DataSet needs to be updated to use ganga 6.7.4 or that perhaps something under GPIDev broke. I looked in ITransform.py but don't understand why this is happening.

P.S. This looks like it may be the same problem as reported at https://github.com/ganga-devs/ganga/issues/1144 hopefully the devs can reproduce it now with the nd280test.gpi I provided.

mesmith75:
Fixes #1212 and #1144 .


mesmith75:
Also fixes #1218

rob-c:
Massive shame that there are objects out there breaking lazy init. Ideally there are better fixes than this but it's functional. (even if it burns CPU and caused problems).
I would strongly suggest against this as it adds a non-insignificant cpu and overhead to the proxy so i'd check that the proxy layer correctly explicitly requests that this is false unless that still breaks these classes. (edit: additionally the XML layer should explicitly request this be false to avoid burning CPU for no need)

mesmith75:
test this please

mesmith75:
@rob-c How about that?

rob-c:
The change you made in the proxy layer won't have the effect you're after because we don't correctly implement inheritance with class initialization options almost anywhere.
If this is fixing problems from devs then we might as well revert to fully initialize everything everywhere but it comes at a cost of CPU being burnt in the core functions which creates a feeling of overhead and sluggishness. (I would suggest making this configurable if that's the case for workflows which don't need this as doing things like constructing large lists in front of the proxy/from-xml will potentially be even more painful).
You could move to start flagging applications which are known to be safe/unsafe with lazy init and have a class level flag which is used in the constructor but I would have strongly leaned toward fixing these classes which have strange behavior.

egede (approved):

PR #1227, opened 2018-02-22
Atlas exe use athena setup

drmarkwslater:
This PR fixes some issues of `--useAthenaPackages` not working correctly for EXE jobs. It also adds a quick fix to very recent versions of Athena changing how to assign input files. In addition, the `--useAthenaPackages` option has been made the default and deprecated with an option added (`--noUserPackageSetup`) to disable as requested by Will Buttinger.

### DIFF
--- a/ganga/GangaAtlas/Lib/Athena/Athena.py
+++ b/ganga/GangaAtlas/Lib/Athena/Athena.py
@@ -1163,7 +1163,7 @@ class Athena(IPrepareApp):
 
         # archive sources
         verbose = False
-        if self.atlas_exetype in ['EXE']: #and not self.athena_compile:  - for EXE, compilation decides what the tarball is called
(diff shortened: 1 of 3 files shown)


mesmith75 (approved):

mesmith75:
@drmarkwslater Are these non-python files being wanted in the released version? If so it might be worth checking they are there.

drmarkwslater:
@mesmith75 Not sure what you mean by 'non-python files'? Do you mean the `athena` script? This is a python file, just not a `.py` file. It means you can run Ganga from the command line with:
```
ganga athena ...
```

mesmith75:
@drmarkwslater Yes, I mean files without the `.py` suffix. I am not sure if they are being picked up at the moment.

drmarkwslater:
@mesmith75 It has been up to this point so I'm assuming it's all OK 😄

mesmith75:
@drmarkwslater Can we merge this?

drmarkwslater:
@mesmith75 Not yet - after discussions with the users, they want to change the default behaviour and so I'll try to get this done today/tomorrow.

PR #1228, opened 2018-02-22
Updated Atlas.ini to remove startup messages

drmarkwslater:
This updates the Atlas.ini file to prevent some unnecessary error messages appearing at startup in Ganga 7


mesmith75:
Do you really want to remove those lines or rather change `Ganga` to `GangaCore`?

drmarkwslater:
@mesmith75 You're exactly right 😄 Changed to the correct paths. Still going to remove the GridProxy stuff though.

mesmith75 (approved):

PR #1229, opened 2018-02-26
Use IPython 5

### LINKED ISSUE #1222: Outdated requirements
In our setup.py we require some ancient versions of things - in particular I note ipython 1.2.1 . We should update or get rid of the version specific requirements altogether.

mesmith75:
Fixes #1222 

Updates to IPython 5.X , with the requisite adjustments.


rob-c:
How does the IPython5 history play with AFS is it known?
Also, is the user history forward accessible after the update?

mesmith75:
@rob-c Not sure, although I don't think the second point really matters. Ill try out a fresh installation and see what happens.

This now has the necessary updates for the shell things. With regard the tests, I think we just need to install some extra packages on the machine.

rob-c:
@mesmith75 I'd argue having the command history for how you submitted the job is very important, I know a lot of users who archive jobs out of ganga due to the poor performance of the XML repo who use the command history as a powerful tool to know what's going on.
This shouldn't be a painful thing to test as per moving from 0.17 to 1.x but I think if there are steps we need to take to preserve this it should be done.

mesmith75:
test this

mesmith75:
Sounds a bit odd but fine I guess. Often we find that the ipython history is corrupted and has to be removed anyway.

More importantly this is incompatible with the existing CVMFS (and AFS) installations. Once this has been released the old versions will no longer work.

rob-c:
Ah yes the whole AFS multiple session issue.
This is only something hit by a small number of users as most users on lxplus still strictly run only 1 session at a time.

If you're going for a blank slate approach I would make this very clear to users in the release notes if the history is not going to be migrated as there is no way of getting the binary data out of the existing. Last time I forward migrated the history once on first launch in a one shot approach and that caused enough emails on the dast list and elsewhere about commands not being available from the history.

mesmith75:
A quick installation suggests that the old history is maintained ok.

egede (commented):
We still use the externals area on afs for eventually creating the release on /cvmfs. I guess that means that we have to install IPython 5 there for this to work?

mesmith75:
@egede Yes - that should be taken care of by the install script used for the AFS and CVMFS installs when the release is done. I don't think we can have two versions together though so as I say previous versions of Ganga wouldn't work which is unfortunate.

egede:
@mesmith75 There is version information in the install script, so I do think that we should be able to keep installation of past versions working as well.

However, it might be a better idea to kill the install script and move to pip instead. We could do that before we merge this?

mesmith75:
@egede Ah, I didn't realise we could have two versions of ipython installed in the externals.

I agree it would be best just to move to pip and a virtualenv.

mesmith75:
Now fixes #1235 too.

I think updating to absl-py doesn't require any changes from us.

mesmith75:
The installation on CVMFS seems to be working fine - can I merge this?

egede (approved):
Has the code for installing in /cvmfs been updated as well? Where does that code reside?

mesmith75:
@egede It has not. There is a script on the CVMFS machine but that never gets updated. I'll write a new one and get it to copy the version in the new release over itself so it will be updated.

PR #1230, opened 2018-02-27
Cache warnings

### LINKED ISSUE #1211: develop failing to cache logs
I think this was addressed a short while ago but I'm seeing the main thread getting spammed while I try to type a complex command out.
(on latest develop)

Did something change with regard to the cache size or the log caching to keep logging from spamming the user while they're busy?

mesmith75:
Fixes #1211 .

Instinctively I don't think we should cache `ERROR` but I don't feel strongly if anyone wants to cache them as well.


egede (approved):
That this change was requested makes me wonder if we have too many pointless WARNINGS in the code.

PR #1236, opened 2018-03-06
Control DiracProxy time

### LINKED ISSUE #1232: Add control of proxy lifetimes.
Currently there is no way to control the lifetime of a proxy created by ganga. Only the default 24hours is possible. Proxies with much longer lifetimes can be made by the underlying Dirac tools, so it would be very useful if ganga could expose this throughs he DiracProxy class.

mesmith75:
Fixes #1232

### DIFF
--- a/ganga/GangaDirac/Lib/Credentials/DiracProxy.py
+++ b/ganga/GangaDirac/Lib/Credentials/DiracProxy.py
@@ -25,7 +25,6 @@ class DiracProxyInfo(VomsProxyInfo):
     A wrapper around a DIRAC proxy file
     """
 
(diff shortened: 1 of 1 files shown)


egede:
test me

egede:
test this please

egede (approved):
I think many users in LHCb are still using the `gridProxy.create()` short cut. So I guess we either have to teach them to use the (cumbersome) `credential_store.create(DiracProxy(validTime='48:00'))` or that they should change the value of `[defaults_DiracProxy]validTime` in their `.gangarc`.

PR #1237, opened 2018-03-07
Fix status update when loading damaged jobs

mesmith75:
Fixes bug introduced in #1193 .

@rob-c I'm not sure what you were aiming for here - but I assume it was this?

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Registry/JobRegistry.py
+++ b/ganga/GangaCore/GPIDev/Lib/Registry/JobRegistry.py
@@ -137,7 +137,7 @@ class JobRegistry(Registry):
                             haveKilled = True
                     if not haveKilled:
                         logger.warning("Job status re-updated after potential force-close")
-                        v.updateStatus()
+                        v.updateMasterJobStatus()
                 else:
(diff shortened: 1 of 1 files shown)


rob-c:
@mesmith75 I have to ask what bug as I never hit problems in testing?

mesmith75:
@rob-c `updateStatus` needs an argument. What are you updating the status to?

Unfortunately I lost the traceback and wonky job registry that let me discover it.

ganga/GangaCore/GPIDev/Lib/Registry/JobRegistry.py
```diff
@@ -137,7 +137,7 @@ def _killJob(this_job):
                             haveKilled = True
                     if not haveKilled:
                         logger.warning("Job status re-updated after potential force-close")
-                        v.updateStatus()
```
rob-c, line 140: This was for updating master jobs which are in the wrong state with subjobs in a monitorable state. so I suppose that the updateMasterJobStatus should just be as compatible but I don't know what the difference between these 2 methods now is

rob-c (approved):

PR #1240, opened 2018-03-09
Bump submitting to submitted when recovering damaged jobs

mesmith75:
If the subjobs are `running` but the master is `submitting` the `updateMasterJobStatus` tries to update the master to `running` which is not an allowed transition. An exception is thrown and ganga exits before it starts.

This bumps the master to "submitted" before attempting to update the master job status to allow for the transition.


mesmith75:
Could someone please approve this. Users have noticed the bug on the DEV version.

egede (approved):

PR #1243, opened 2018-03-25
Return errors from Dirac script

### LINKED ISSUE #1239: Return errors from DIRAC submission script
We should try to return the DIRAC error from the submission script - at the moment it just complains that there isn't a dict which is not helpful.

mesmith75:
Fixes #1239 .

Edit: Change my mind


mesmith75:
In the Dirac script it checks for successful submission. If it is not successful then the error message gets returned in the dict for the subjob. Subsequently when the Dirac IDs are being applied to the jobs, if the dict for the subjob contains a string then the submission for that subjob is marked as failed and the error message printed. This way if only one or two subjobs fail they can be resubmitted individually.

mesmith75:
Could someone review this please - I am keen to get it tested.

alexanderrichards (approved):

PR #1247, opened 2018-04-03
Remove error handler from logging

mesmith75:
This removes the addition of an error handler to the loggers. It was causing some duplicates with errors:
```python
Ganga In [1]: a
ERROR    name 'a' is not defined
ERROR    name 'a' is not defined
```


drmarkwslater (approved):
Looks sensible to me 😄 This means we just have the error handler added on the `main_logger` which makes sense.

PR #1250, opened 2018-04-10
Fix object setting

mesmith75:
Remove proxies for objects to be set otherwise things go wrong.

### DIFF
--- a/ganga/GangaCore/GPIDev/Base/Objects.py
+++ b/ganga/GangaCore/GPIDev/Base/Objects.py
@@ -473,12 +473,6 @@ class Descriptor(object):
 
         _set_name = _getName(self)
 
-        from GangaCore.GPIDev.Base.Proxy import isProxy, stripProxy
-        if isProxy(val):
-            val = stripProxy(val)
-        if isProxy(obj):
-            obj = stripProxy(obj)
-
         if isinstance(val, str):
             if val:
                 from GangaCore.GPIDev.Base.Proxy import stripProxy, runtimeEvalString
@@ -527,7 +521,6 @@ class Descriptor(object):
 
         item = obj._schema[name]
 
-
         ## If the item has been defined as a sequence great, let's continue!
         if item['sequence']:
             # These objects are lists
@@ -561,6 +554,7 @@ class Descriptor(object):
 
         if isinstance(new_val, Node) and new_val._getParent() is not obj:
             new_val._setParent(obj)
+
         return new_val
 
     @staticmethod
@@ -573,6 +567,7 @@ class Descriptor(object):
             v (unknown): Object we want a new copy of
             extra_args (tuple): Contains the name of the attribute being copied and the object which owns the object being copied
         """
+
         name=extra_args[0]
         obj=extra_args[1]
         if isinstance(v, (list, tuple, GangaList)):
--- a/ganga/GangaCore/GPIDev/Base/Objects.py
+++ b/ganga/GangaCore/GPIDev/Base/Objects.py
@@ -387,6 +387,10 @@ class Descriptor(object):
         """
         item = obj._schema[name]
 
+        from GangaCore.GPIDev.Base.Proxy import isProxy, stripProxy
+        if isProxy(v):
+            v = stripProxy(v)
+
         if v is None:
             if item.hasProperty('category'):
                 assertion = item['optional'] and (item['category'] != 'internal')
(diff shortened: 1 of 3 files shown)


rob-c:
For a while we were excepting and generating a stack trace if a proxied object ended up in the GangaObject class as we decided to keep the object proxy managed/stripped in the Proxy class objects where it's more appropriate.
That being said I added a fix like this a long time ago which removed a lot of runtime problems but it masks higher level code problems in other classes which escape the proxy in some manner and do bad things.

I don't want to say whether this is the best way to fix the problem compared to tracing back the class which is mis-behaving (it will almost certainly fix the problems for the time being), but this will likely mask future bad behaviour is my warning.

(All of that being said the object is by definition an instance of the GangaObject so I think you need only strip the unknown user-defined `val`)

drmarkwslater:
@mesmith75 Do you have a reproducible error that this fixes? As I think @rob-c was saying, I would very much prefer to find the root cause as I was under the impression that all the proxy stuff was sorted at the Proxy Class stage to avoid these calls to `stripProxy` and `addProxy`...

ganga/GangaCore/GPIDev/Lib/GangaList/GangaList.py
```diff
@@ -241,9 +241,9 @@ def strip_proxy_list(self, obj_list, filter=False):
 
     def getCategory(self):
         """Returns a list of categories for the objects in the list. Returns [] for an empty list."""
-
+        from GangaCore.Utility.util import unique
```
drmarkwslater: Can this be put at the top of the file?

ganga/GangaCore/GPIDev/Base/Objects.py
```diff
@@ -473,6 +473,12 @@ def __set__(self, obj, val, root_obj=None):
...
 
+        from GangaCore.GPIDev.Base.Proxy import isProxy, stripProxy
+        if isProxy(val):
+            val = stripProxy(val)
+        if isProxy(obj):
+            obj = stripProxy(obj)
+
         if isinstance(val, str):
             if val:
                 from GangaCore.GPIDev.Base.Proxy import stripProxy, runtimeEvalString
```
drmarkwslater: This can be partially removed because of the above.

ganga/GangaCore/GPIDev/Lib/GangaList/GangaList.py
mesmith75: yes

mesmith75:
@rob-c I am pretty sure you are right, this is not the best solution. I found this from a problem a user was having with `addQuery` with an LHCb transform.

```python
Ganga In [2]: tf = LHCbTransform()

[10:36:05]
Ganga In [3]: myquery = BKQuery('/MC/2011/Beam3500GeV-2011-MagUp-Nu2-Pythia8/Sim08e/Digi13/Trig0x40760037/Reco14a/Stripping20r1NoPrescalingFlagged/12195001/ALLSTREAMS.DST')

Ganga In [4]: tf.addQuery(myquery)
INFO     Retrieving latest bookkeeping information for transform, please wait...
INFO     Constructing LHCbDataset
INFO     Checking for new and removed data for query 0, please wait...
ERROR    v: [10 Entries of type 'DiracFile']
ERROR    GangaException: Error: found Object: [10 Entries of type 'DiracFile'] of type: <class 'GangaCore.GPIDev.Base.Proxy.GangaList'> expected an object inheriting from Node!
```
The error arises at https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaLHCb/Lib/Tasks/LHCbTransform.py#L272

This is where a GangaList gets cloned into another GangaList. The issue seemed to be that in the `__set__` function which leads to the `cleanValue` and then down into the `cloneListOrObject` and `cloneObject` the GangaList isn't being shown as such but a proxy.

rob-c:
*ow* This looks like something I may have done for performance when moving around _huge_ lists of DiracFiles.
It looks like the output from the BKQuery is being used in a proxied format.
(All of the output from https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaLHCb/Lib/LHCbDataset/BKQuery.py makes use of addProxy).
If the datasets are stripped after being requested from BKQuery in the Transforms layer does this work as expected?

(This is why it would be nice to have the proxy layer manage all of proxy-ing output from public methods)

mesmith75:
@rob-c yes it does.

mesmith75:
@rob-c @drmarkwslater Is this ok to be merged as it is now?

mesmith75:
test this please

rob-c (approved):

mesmith75:
@rob-c  thanks!

PR #1251, opened 2018-04-13
For release 7.0.2

mesmith75:
Release notes.

I also disabled the load check by default so we can test it a little more.


egede (approved):
I guess we in general should avoid introducing code changes in these release branches.

mesmith75:
Yes, good point.

PR #1258, opened 2018-05-11
WIP: Compressed lhcb dataset

ArnauBrossa:
It includes the LHbDataset_custom class  which intends to substitute the LHCbDataset class.

It intends to subsutitute the current attribute files (GangaList of DiracFiles) with two attributes names (list of strings) and prefix (string).
this attributes contain the most general prefix between all the lfns and the remaining subfixes as a list of strings.
It's implementation is using python a property and getter called files2 that returns a GangaList() of DiracFiles() with the lfns without storing the GangaList() in the LHCbDataset().

The new property and setter and implemented in line 475.
other changes respect LHCbDataset are only in the schema definition of attributes and in __init__

To Do:
- [ ] Solve compatibility of setter and property with Proxy, currently it only work if it is initialized as:
stripProxy(LHCbDataset_custom())

- [ ] Solve compatibility of isinstance with the Proxy in the Setter. currently LHCbDataset().files2 = GangaList() returns a wrong type exception while LHCbDataset().files2 = stripProxy(GangaList()) works fine.

- [ ] Fix formatting on the DiracFiles returned by the files2 property, which does not macth with the DriacFiles() stored in LHCbDataset.()files

- [ ]  Update __init__ to be consistent with the new attributes


egede:
test please

egede (commented):
Should this not be implemented as a full replacement for LHCbDataset, rather than a parallel structure? In that case, you will also get all the dataset test cases running on the branch.

rob-c:
@Erni1619 If you're feeling brave you may want to consider overloading the methods which (de)serialize the data stored in the class.
You may be able to get around this if you intercept the setter for the files attribute in the main class and store the data in a more lightweight array internally rather than only in-memory.

I think you're heading toward what I would suggest is a much more performant version of this class that stores much less internal information and is basically a factory which returns IGangaFile instances on demand.
(You may want to cache the objects once they've been generated once as the construction of many DiracFile objects is often the biggest bottleneck in profiles I've taken).

@egede Is it worth dropping support for anything other than DiracFile in this class at some point?
The main sticking point in simplifying the code is that a _proper_ class factory has to support all IGangaFile objects which means (de)serializing the data becomes tricky.

mesmith75:
@egede I think it would be nice to still have a class that stores everything. For sure if you are just running over a massive dataset from BKQuery this is best but a lot of people run over private productions, output of other jobs etc where it might be useful to keep some extra info.

egede:
@rob-c Remind me of the reason for the `LHCbDataset` class. Will a `GaudiExec` job work if I just attach a normal list of `LocalFile` objects? I try to understand if the "extras" that the `LHCbDataset` provide in reality are only useful for `DiracFile` objects already.

rob-c:
Looking at the class the main functionality that is of use is:

 1. Generating the XML catalogue
 2. Generating an options file (really the same thing as 1.)
 3. Allowing a user to perform set based logic to determine which LFN they need to re-submit
 4. Get a list of PFN or LFN

I think due to 1. and 2. GaudiExec for instance will still need the LHCbDataset rather than the GangaDataset or a flat array of IGangaFile objects.
Certainly when running on a local batch system these requirements mean we can't just drop the class as it stands.


I think that replacing the `files` attribute with a 2 new attributes internally each only supporting `DiracFile` and `MassStorageFile` satisfies all of the above requirements.

Storing many DiracFile objects in a GangaList is unfortunately expensive so maybe a new version of the class which intercepts the get/set calls to the `files` attribute and silently stores the data in a more economical format (2 arrays of filenames and LFNs would be my suggestion) is more useful.

I don't think there is any requirement for the full Dirac/MassStorage-File functionality for file objects inside this class and keeping this adds some significant overheads.

I think a user might decide they want to do something with a DiracFile returned from within an LHCbDataset but I don't think it makes sense to do say `myLHCbDataset.files[1].remoteDir = someValue` as these LFN are not usually user managed.



Some very quick Pseudo-code similar to the actual PR. This should be compatible with users looking to use the `files` attribute but store much smaller objects on disk and allow for several methods to be made much simpler internally to the class.
```

class LHCbDataset:
    files: GangaArray(IGangaFiles)
    lfn_files:  GangaArray(strings)
    pfn_files: GangaArray(strings)

def __set__(value, object):
  ...
  if value == files:
    if object is DiracFile:
      self.lfn_files = object.lfn
    if object is MassStorageFile:
      self.pfn_files = object.full_path
  ...

def __get__(name):
  ...
  if name == files:
      newlist = GangaList()
      for file in self.lfn_files:
        newlist.append(DiracFile(lfn=file))
      for file in self.pfn_files:
        newlist.append(MassStorageFile(path=file))
    return newlist
  ...

```

egede:
@Erni1619 Let me know if you anticipate working further on the PR.

mesmith75:
Replaced by #1557

PR #1259, opened 2018-05-11
Added listFiles DiracCommand to list old lfns

jplews17:
Added diracCommand listFiles for use in cleaning up grid storage. It follows a similar logic to dirac-dms-user-lfns, taking a base directory and a minimum file age. It returns a list of LFNs for files in the base directory and subdirectories older than the given age.

This is to be used by another function, that I currently do not know where to put, which takes these LFNs and filters them using a wildcard argument and then deletes the selected files:

`def remove_lfns(basedir = "", days = 0, months = 0, years = 0, wildcard = "*", FORCE = False):
    import fnmatch
    from datetime import datetime, timedelta

    from GangaDirac.Lib.Utilities.DiracUtilities import execute

    vo = ""
    username = credential_store[DiracProxy()].username
    vo = credential_store[DiracProxy()].group

    if not vo:
        raise Exception("Could not determine VO.")

    defaultDir = False
    if not basedir:
        basedir = "/lhcb/user/{0}/{1}/".format(username[0], username)
        defaultDir = True

    totalDays = 0
    if years:
        totalDays += 365*years
    if months:
        totalDays += 30*months
    if days:
        totalDays += days
    td = timedelta(days = totalDays)

    if defaultDir and not td and wildcard == "*":
        if not FORCE:
            raise Exception("This will delete everything!!! Use argument 'FORCE=True' to force deletion")

    all_lfns = execute("listFiles('{0}', {1})".format(basedir, td.__repr__()))
    selected_lfns = fnmatch.filter(all_lfns, wildcard)

    if selected_lfns:
        print "Deleting following files:"
    else:
        print "No files to be deleted."

    for lfn in selected_lfns:
        print lfn

    execute("removeFile({0})".format(selected_lfns))

`

Advice on where to put this function would be much appreciated.


ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,62 @@ def checkSEStatus(se, access = 'Write'):
...
+
+    def isOlderThan(cTimeStruct, days):
+        timeDelta = timedelta( days = days )
+        maxCTime = datetime.utcnow() - timeDelta
+        if cTimeStruct < maxCTime:
+            return True
+        return False
+
+    baseDir = baseDir.rstrip('/')
+    print ("Will search for files in %s" % baseDir)
```
mesmith75: There shouldn't be `print` in DiracCommands.

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,62 @@ def checkSEStatus(se, access = 'Write'):
...
+            raise Exception("Error retrieving directory contents", "%s %s" % ( currentDir, res['Message'] ) )
+        elif currentDir in res['Value']['Failed']:
+            raise Exception("Error retrieving directory contents", "%s %s" % ( currentDir, res['Value']['Failed'][currentDir] ) )
+        else:
+            dirContents = res['Value']['Successful'][currentDir]
+            subdirs = dirContents['SubDirs']
+            files = dirContents['Files']
+            if not subdirs and not files:
+                emptyDirs.append( currentDir )
+                print ('%s: empty directory' % currentDir)
```
mesmith75: `print`

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,62 @@ def checkSEStatus(se, access = 'Write'):
...
+                for filename in sorted(files):
+                    fileOK = False
+                    if (not withMetaData) or isOlderThan(files[filename]['MetaData']['CreationDate'], days):
+                        fileOK = True
+		    if not fileOK:
+                        files.pop(filename)
+                allFiles += sorted(files)
+
+                if len( files ) or len( subdirs ):
+                    print ("%s: %d files, %d sub-directories" % ( currentDir, len(files), len(subdirs) ) )
```
mesmith75: `print`

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,62 @@ def checkSEStatus(se, access = 'Write'):
...
+    activeDirs = [baseDir]
+
+    allFiles = []
+    emptyDirs = []
+
+    while len(activeDirs) > 0:
+        currentDir = activeDirs.pop()	
+        res = fc.listDirectory(currentDir, withMetaData, timeout = 360)
+        if not res['OK']:
+            raise Exception("Error retrieving directory contents", "%s %s" % ( currentDir, res['Message'] ) )
```
mesmith75: It doesn't make sense to raise exceptions here. If something doesn't work it should be propagated to the `return`. I *think* `execute` should catch if you don't return a dict so simply returning a string with an error message would be fine.

egede:
@jplews17 A few comments on the suggested remove function above.

I wonder if it would be better that the user is simply given a list of DiracFile objects. They can then decide to remove them if that is what they want to do. The way to specify the age should be more pythonlike by accepting a datetime.timedelta object.

ganga/GangaDirac/Lib/Server/DiracCommands.py
jplews17: Ok, I've changed this now.

jplews17:
@egede The overall aim is to use this function to automate the cleanup of things like output sandboxes for old jobs. Of course the function could be used directly by the user too. removeFile takes an LFN or list of LFNs as an argument (though I presume the LFN can be obtained from a DiracFile object) so is it not better to use LFNs rather than DiracFiles?

jplews17:
@egede I've changed the argument of the listFiles command to take a datetime.timedelta object which is then compared to "now" in the loop. Is this the sort of thing you meant? I could alternatively calculate the datetime.datetime of the minimum age inside the remove_lfns function and pass that, but I'm also aware that I want the function to be as general as possible.

EDIT: I don't understand why the test failed, everything seems to work for me

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,50 @@ def checkSEStatus(se, access = 'Write'):
     '''
     result = dirac.checkSEAccess(se, access)
     return result
+
+@diracCommand
+def listFiles(baseDir, td):
+
```
egede, line 454: Docstring should be added

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,50 @@ def checkSEStatus(se, access = 'Write'):
...
+    from DIRAC.Resources.Catalog.FileCatalog import FileCatalog
+    fc = FileCatalog()
+
+    withMetaData = False
+    if td:
```
egede: Does that imply that you should have td=None in the function definition?

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +405,50 @@ def checkSEStatus(se, access = 'Write'):
...
+            return "Error retrieving directory contents", "%s %s" % ( currentDir, res['Value']['Failed'][currentDir] )
+        else:
+            dirContents = res['Value']['Successful'][currentDir]
+            subdirs = dirContents['SubDirs']
+            files = dirContents['Files']
+            if not subdirs and not files:
+                emptyDirs.append( currentDir )
+            else:
+                for subdir in sorted( subdirs, reverse=True):
+                    if (not withMetaData) or subdirs[subdir]['CreationDate'] < (datetime.utcnow() - td):
```
egede: Put time calculation outside while function

egede:
test please

egede:
@rob-c Can you help @jplews17 on how to get the suggested function above added as a global function that is exposed to the GPI.

rob-c:
@egede OK I'll take a look when I get a chance, Just a quick question from me, could this find lookup be performed using a query against the  Metadata of the files?
I think I prefer this way of doing it as we're slightly less prone to timeouts with huge directory trees but I'm just curious if the filecatalogue might be able to get slices of files by date internally?

egede:
@rob-c I don't know. But for sure it would be better to ask for files of a given age rather than asking for everything and then filter afterwards.

rob-c:
@egede I suppose just asking for folders modified after a certain date would be sane from within Ganga, however I wouldn't want to make a recursive request over a large directory tree as debugging any problems just due to timeouts could prove painful.

rob-c:
@jplews17 Could you investigate if directory metadata is modified correctly by DIRAC when a new file is created? (This could avoid walking through a directory tree when there is no need).
(Do directories have correct metadata from their contents? This might not necessarily be the case)
If you could see if there is a way of restructuring the query as a metadata query this could be useful.
I'll write a short wrapper to expose this to the GPI otherwise at some point this week and update the PR for you to have a look at.

mesmith75:
I've added a `listFiles` function to the GPI to use this.

mesmith75:
Superseded by #1465

PR #1262, opened 2018-05-12
WIP:Added an option in GaudiExec that automatically finds and sets the DB tags 

paguenther:
We (I and @ZhenziW) added the option 'autoDBtags' to GaudiExec which calls a Dirac command that retrieves the DB and CondDB tag from the bookkeeping. It is meant to be used for MC tags only. The new Dirac command function takes the production number from the LFN and asks the bookkeeping for information about this production. The information contains the DB tags for all production steps, however, only the tags of the last production step are passed via 'extraOpts' to DaVinci. 
Thanks @VanyaBelyaev for the idea to do that and @chrisburr for a lot of help!


ganga.egg-info/PKG-INFO
```diff
@@ -0,0 +1,45 @@
+Metadata-Version: 1.1
```
chrisburr: All of the files in `ganga.egg-info` shouldn't be committed, can you remove them?

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +404,26 @@ def checkSEStatus(se, access = 'Write'):
     '''
     result = dirac.checkSEAccess(se, access)
     return result
+
+@diracCommand
+def getDDBtags( lfn ):
```
chrisburr, line 409: You should check that the LFN corresponds to MC and raise an error if not. I think this be done using the return value of `getProductionInformations`.

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +404,26 @@ def checkSEStatus(se, access = 'Write'):
...
+    from LHCbDIRAC.BookkeepingSystem.Client.BookkeepingClient import BookkeepingClient
+    import types
+    bk = BookkeepingClient()
+    prod = long(lfn.split('/')[5]) # not sure if this works in all cases 
+    res = bk.getProductionInformations( prod )
+    ddb = ''
+    conddb = ''
+    if res['OK']: # there should probably also be an 'else' for cases where no information could be retrieved 
+        val = res['Value']
+	steps = val['Steps']
```
chrisburr, line 422: There should be 8 spaces in front of lines 422-426.

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -309,6 +310,10 @@ def constructExtraFiles(self, job):
             # Add the extra opts file to the job
             tinfo = tarfile.TarInfo(extra_opts_file)
             tinfo.mtime = time.time()
+	    if self.autoDBtags:
+		ddb, conddb = execute('getDDBtags("{0}")'.format(job.inputdata[0].lfn)) # take the tags only from the first file
+		self.extraOpts += '"' + 'DaVinci().DDDBtag = '   + "'" + ddb    + "'" + ';"'
```
chrisburr, line 316: What is the first `'"'` for here?

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +404,26 @@ def checkSEStatus(se, access = 'Write'):
...
+def getDDBtags( lfn ):
+    ''' returns the DDB and CONDDB tags of the last 
+	production step for a given LFN 
+    '''
+    from LHCbDIRAC.BookkeepingSystem.Client.BookkeepingClient import BookkeepingClient
+    import types
+    bk = BookkeepingClient()
+    prod = long(lfn.split('/')[5]) # not sure if this works in all cases 
+    res = bk.getProductionInformations( prod )
+    ddb = ''
```
chrisburr, line 418: `ddb`  should be changed to `dddb` as it stands for Detector Description DataBase. This also applies to the changes in `GaudiExec.py`.

ganga/GangaDirac/Lib/Server/DiracCommands.py
chrisburr, line 409: I think this function should be called either `getDBtags` or `getDBtagsforMC`.

rob-c:
@paguenther Looks good but are the ganga.egg files needed as part of the PR?

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -309,6 +310,10 @@ def constructExtraFiles(self, job):
             # Add the extra opts file to the job
             tinfo = tarfile.TarInfo(extra_opts_file)
             tinfo.mtime = time.time()
+	    if self.autoDBtags:
+		ddb, conddb = execute('getDDBtags("{0}")'.format(job.inputdata[0].lfn)) # take the tags only from the first file
```
chrisburr, line 315: I think this function call will need to be cached in some way as it is called for every subjob, but I'll defer to the ganga developers here for advice.

paguenther:
@rob-c ah damn, no I think they should not be there, sorry. I will remove them

chrisburr (commented):
Thanks @paguenther and @ZhenziW. I really like what you achieved with this project and look forward to using it!

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
rob-c: @chrisburr The general intention of GaudiExec is that all files are constructed and uploaded before submitting the job so this is done once and stored.
That being said yes it appears this will be called once per-subjob.
Moving this to the configure step and modifying the `self.extraOpts` there should be enough to ensure all subjobs inherit this change. I'd make sure this is correct with a test job but if this breaks let me know but you can always cache the output of this once in the `master_job.application`.

This _will_ crash if a user has forgotten to declare any inputfile(s) here though and it might be nice to catch this in a nicer way.

egede (commented):
Putting this information into `extraopts` is a poor idea. We should always avoid adding information automatically to fileds that the user modify. At the moment it is undefined if the `extraopts` lines added should be copied if a user is copying the job.

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +404,26 @@ def checkSEStatus(se, access = 'Write'):
...
+
+@diracCommand
+def getDDBtags( lfn ):
+    ''' returns the DDB and CONDDB tags of the last 
+	production step for a given LFN 
+    '''
+    from LHCbDIRAC.BookkeepingSystem.Client.BookkeepingClient import BookkeepingClient
+    import types
+    bk = BookkeepingClient()
+    prod = long(lfn.split('/')[5]) # not sure if this works in all cases
```
chrisburr, line 416: It might be a better idea to move this to outside the function and pass the production ID as an argument. You can then check that you get the same DB tags for all production IDs.

mesmith75:
@egede Yes I agree. It should be as for the metadata options. These are written into their own options files which is added to the command to be run.

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -405,3 +404,26 @@ def checkSEStatus(se, access = 'Write'):
...
+    res = bk.getProductionInformations( prod )
+    ddb = ''
+    conddb = ''
+    if res['OK']: # there should probably also be an 'else' for cases where no information could be retrieved 
+        val = res['Value']
+	steps = val['Steps']
+	last_step = steps[-1] # the tags are taken from the last step of production
+	ddb = last_step[4]
+	conddb = last_step[5]
+	return ddb, conddb
```
mesmith75, line 426: What exactly are you returning here? Are these dicts ?

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
paguenther: This should lead to "DaVinci().DDDBtag='simwhatever';" which is the format that is required by extraOpts if I am not mistaken.

ganga/GangaDirac/Lib/Server/DiracCommands.py
paguenther: should be strings

mesmith75: does that work?

chrisburr: [The dirac wrapper handles it](https://github.com/paguenther/ganga/blob/4a1938bdc0e0a023218ec4ddaf4b793d41c34aa8/ganga/GangaDirac/Lib/Server/DiracDefinition.py#L35-L41)

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
egede: This could fail in several ways so should probably be put into a try except that then throws an exception with a clear explanation. As @rob-c said, a missing dataset, but it could also be an error from Dirac in terms of no information for the given lfn. Also, this will fail if the dataset is not given in terms of `DiracFile` objects.

mesmith75:
Now in #1317 with some corrections.

Issue #1269, opened 2018-05-23
runing multicores job


sabahsalih:
Dear All,  How could I submit a multicore jobs using ganga and ARC please

Many Thanks, Sabah.

egede:
@sabahsalih Do you know what the underlying syntax is for asking ARC for this. From the Ganga side, we will just pass on an option. A secondary question is if there are any multicore resources available to run on.

sabahsalih:
Dear Ulrik,
         Thank you so much for your kind reply.

I think is this

 j.backend.xRSLextras= {"count":8}
  j.backend.xRSLextras= {"count":8, "countpernode":8}

It came from


https://github.com/ganga-devs/ganga/commit/52b367862ebc7af1a81878f52ee46264437073f8
https://github.com/ganga-devs/ganga/pull/1267

Yes the Grid farm have alot of multicore resources available to run on


   Many Thanks, Sabah.

From Sabah Salih
Particle Physics Group,
The School of Physics and Astronomy,
The University of Manchester,
Schuster Laboratory,
Brunswick Street,
Manchester M13 9PL.
Tel: +44 1612754171 or  x4171
E-mail: sabah.salih@manchester.ac.uk
________________________________

mesmith75:
@sabahsalih Would it be possible to confirm that commit works as intended?

egede:
So the ability will come through #1267. So I will just close this request. As #1267 has already been merged, you can just pull the default version of Ganga from git and you will get it - or if you are running from `/cvmfs` you can use the DEV version that already has it included.

sabahsalih:
Dear Mark,
         I wil run it I feed back this afternoon.

   Many Thanks, Sabah.

From Sabah Salih
Particle Physics Group,
The School of Physics and Astronomy,
The University of Manchester,
Schuster Laboratory,
Brunswick Street,
Manchester M13 9PL.
Tel: +44 1612754171 or  x4171
E-mail: sabah.salih@manchester.ac.uk
________________________________

sabahsalih:
Dear Mark,
         I submitted the job with

     j.backend.xRSLextras= {"count":8}
    j.backend.xRSLextras= {"count":8, "countpernode":8}



   When I go to ganga
I get the following:


jobs(208)
Ganga Out [2]:
 Job (
   comment = ,
   parallel_submit = True,
   outputdir = /afs/cern.ch/user/s/sabah/gangadir/workspace/sabah/LocalXML/208/output/,
[...]

PR #1272, opened 2018-05-24
Display dictionary keys and values

Sam-Tygier:
When displaying a ganga object, check if it is a mapping type and if
so display both the keys and the values.

For example in the ARC backend, the xRSLextras attribute is currently displayed as
xRSLextras = ['count', 'countpernode'],
this fixes it to
`xRSLextras = {'count':8, 'countpernode':8}`

See #1269


mesmith75:
@Sam-Tygier Thanks for this. SO the submission works ok?

mesmith75 (approved):

Issue #1282, opened 2018-06-06
DiracFiles in inputfiles should not be labelled as inputdata in Dirac


egede:
When an LFN is listed in the inputsandbox in the Dirac API, the file pointed to is copied to the PWD on the worker node when the job starts. This is waht happens if a `DiracFile` is used a an inputfile for the `Dirac` backend. However, in the line 

https://github.com/ganga-devs/ganga/blob/8ca6ee73653e32b272d4d5055187d1892f5558e3/ganga/GangaDirac/Lib/RTHandlers/DiracRTHUtils.py#L148

we set it as `inputdata` in the Dirac API as well. This seems wrong. It will cause the job just to be matched to sites where the LFN is present, but that seems wrong in this case (it is typically not a datafile). I suggest to delete the above line.

egede:
Suggested fix above is not correct. The line above actually correctly assigns the inputdata field. I have so far failed to find where DiracFiles in inputfiles are assigned as inputdata as well.

mesmith75:
Is this just an issue for GaudiExec ?

egede:
No, I just reproduced this myself with a standard executable application.

```
j = Job(backend=Dirac(), inputfiles = [DiracFile('LFN:/lhcb/test')])
```
is producing the script

```
resultdict = {}

# dirac job created by ganga
from DIRAC.Core.Base.Script import parseCommandLine
parseCommandLine()
from DIRAC.Interfaces.API.Dirac import Dirac
from DIRAC.Interfaces.API.Job import Job
dirac = Dirac()
sjNo='2'

j = Job()

# default commands added by ganga
j.setName('{Ganga_Executable_(2)}')
j.setExecutable('exe-script.py','','Ganga_Executable.log')
j.setInputSandbox(['/home/hep/egede/gangadir/workspace/egede/LocalXML/2/input/_input_sandbox_2.tgz', 'LFN:/lhcb/test', 'LFN:/lhcb/test', ])
j.setInputData(['/lhcb/test'])
...
```

Also note that `/lhcb/test` is repeated twice in the inputsandbox.

rob-c:
A workaround for the moment might be to try `j.backend.settings({'InputData':[]})` within ganga as this should overload the `setInputData` attribute before the job is submitted.
Granted this doesn't fix the double inputsandbox attribute which is a little odd. Could this have come from the recent changes to collect the jobs before submission?

mesmith75:
[...]

PR #1283, opened 2018-06-07
Initial attempt at fixing vanilla Dirac exe

rob-c:
This is a first attempt to address #1282.

I've not tested this with any other application other than Executable and so don't know if any other RTHandlers are impacted (I suspect not).

Hopefully this fixes everything as I was able to submit a correct looking JDL with this but haven't tested if the job runs/completes/etc.

### DIFF
--- a/ganga/GangaDirac/Lib/Backends/DiracBase.py
+++ b/ganga/GangaDirac/Lib/Backends/DiracBase.py
@@ -356,17 +356,10 @@ class DiracBase(IBackend):
         input_sandbox += self._addition_sandbox_content(subjobconfig)
 
         ## Add LFN to the inputfiles section of the file
-        input_sandbox_userFiles = []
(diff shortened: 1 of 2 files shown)


egede (approved):

rob-c:
Suppose I should submit some LHCb  test jobs with this to check it doesn't break things in new and exciting ways, if someone else tests this and it works feel free to merge from me :smile:

mesmith75:
test this please

mesmith75:
No, this has broken it. Now DiracFiles in inputfiles are not included at all.

mesmith75:
@rob-c I apologise, I was wrong. This does indeed work fine for executable jobs. There is an issue with GaudiExec not including DiracFiles which I think I have found the source of. Let's merge this and I'll fix that in a separate PR.

PR #1287, opened 2018-06-12
Set process name to Ganga

mesmith75:
Sets the name of the process to `ganga` in `top` and `ps`.

This should be implemented at the same time as the update to `iPython 5` and the `absl-py` as it requires the installation of the `libcap-dev` library and `python-pctrl`

### DIFF
--- a/bin/ganga
+++ b/bin/ganga
@@ -55,6 +55,9 @@ import GangaCore.Runtime
 GangaCore.Runtime._prog = None
 import sys
 
+import prctl
+prctl.set_name('ganga')
+
 def log(level, err):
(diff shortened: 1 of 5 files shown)


egede (commented):
Is it really a good idea to add a binary dependence to Ganga for the sake of changing the process ID name. This could break things for users in two ways when running from /cvmfs. Either they do not have the package, or if we provide the package, it might not be binary compatible with their OS.

mesmith75:
Yes I think you are right. Although for iPython 5 we need gcc anyway but I guess that is more typical.

PR #1290, opened 2018-06-26
Add master job's DiracFiles inputfiles to input sandbox

mesmith75:
This fixes a bug that DiracFiles in the `inputfiles` for GaudiExec jobs get ignored. It is because in the master prepare when the sandbox gets made it only gathers together the local ones in order to tar them up, so the remote ones get forgotten about.


mesmith75:
Once we get this merged I'll release a new version.

egede (approved):

rob-c:
This may be not needed due to #1305 . I'll try and go through an check at some point to avoid adding un-needed fixes to other handlers which should be in the core code.

PR #1294, opened 2018-07-03
Extras for CVMFS virtualenv

mesmith75:
Some extra things for installing on CVMFS.
The first is to check that the user has python 2.7.
The second adds the virtualenvs site-packages directory to the path to access the requirements.

### DIFF
--- a/bin/ganga
+++ b/bin/ganga
@@ -32,11 +32,16 @@ def standardSetup():
     """   
     import os.path
 
-    # insert the path to Ganga itself - this is required for the installation on CVMFS
+    # insert the path to Ganga itself - this is required for the installations without pip 
     exeDir = os.path.abspath(os.path.normpath(os.path.dirname(os.path.realpath(__file__))))
     gangaDir = os.path.join(os.path.dirname(exeDir), 'ganga' )
-
     sys.path.insert(0, gangaDir)
+
+    #On CVMFS we need to point to the site-packages directory as we don't start the virtualenv
+    if exeDir.startswith('/cvmfs/ganga.cern.ch', 0, 20):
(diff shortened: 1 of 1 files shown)


bin/ganga
```diff
@@ -21,17 +21,28 @@
...
 
-    # insert the path to Ganga itself - this is required for the installation on CVMFS
+    # insert the path to Ganga itself - this is required for the installations without pip 
     exeDir = os.path.abspath(os.path.normpath(os.path.dirname(os.path.realpath(__file__))))
     gangaDir = os.path.join(os.path.dirname(exeDir), 'ganga' )
-
     sys.path.insert(0, gangaDir)
 
+    #On CVMFS we need to point to the site-packages directory as we don't start the virtualenv
+    if exeDir.startswith('/cvmfs/ganga.cern.ch', 0, 20):
```
egede, line 41: Could the check of not using virtualenv be made a bit more general? If somebody copied the /cvmfs installation to somewhere else (stupid but they might), then it would stop working which seems a bit silly.

mesmith75: I see your point but I am not sure we can, unless we just add this always.

egede: Is https://stackoverflow.com/questions/1871549/determine-if-python-is-running-inside-virtualenv helpful?

mesmith75: But on CVMFS ganga won't be run in a virtualenv - nor will it be if you copy it. I can imagine users will install ganga via pip both with and without virtualenv so that won't help us.

egede: Fair point.

egede (approved):

PR #1296, opened 2018-07-04
Bash script for installation on CVMFS

mesmith75:
A copy of this should sit in the home area of the CVMFS publishing machine. There is also a simple python script that checks for a new release and then runs this - that script will also copy this from the new release back into the home area so it gets updated with each release if need be.

The installation is done from github as it can take several minutes for new releases to appear on pypi.


egede (approved):
Would it make sense to make the script email stdout and stderr at the end such that we get alerted if something didn't work correctly.

mesmith75:
@egede That is taken care of by the python script run by the cron job. I'll add it to this PR for bookkeeping.
```python
#!/usr/bin/python
import os
import requests
import commands

msg = ""

# put in a massive try..except block so I can get some feedback
try:
    # Request the list of releases
    resp = requests.get('https://api.github.com/repos/ganga-devs/ganga/releases')
    if resp.status_code == 200:
        # assume the first is the latest
        rel_ver = resp.json()[0]['name']

        # Now check if it's installed
        if rel_ver not in  os.listdir('/cvmfs/ganga.cern.ch/Ganga/install/'):
            # we need to install it
            msg = commands.getstatusoutput("/home/cvganga/ganga-cvmfs-install.sh %s" % rel_ver)[1]

    else:
        msg = "Problem getting the list of releases"

except Exception as err:
    msg = str(err)

# send any messages
if msg:
    os.system('echo "%s" | mail -s "Message from CVMFS Installer" "project-ganga-developers@cern.ch"' % msg)

```

egede (approved):
Great.

PR #1308, opened 2018-07-29
Ganga swan integration

apsknight:
Enable Ganga to share Jobs between multiple session such that one Ganga session can submit the Job and other session can monitor the same job and completes it.

Now there can be 2 sessions, Session-1 with Monitoring Enabled and Session-2 with Monitoring Disabled.

So, if user submits a Job in [2], the monitoring loop in [1] will pick the Job and update it. Now if a user wish to see the Job update in [2], he/she can run reloadJob(job_id) in [2] and the Job with id job_id will update from disk.

These changes can be enabled by creating an environment variable named `GANGA_SWAN_INTEGRATION`.

### DIFF
--- a/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -4,6 +4,7 @@ import time
 import copy
 from contextlib import contextmanager
 
+from GangaCore import GANGA_SWAN_INTEGRATION
 from GangaCore.Core.GangaThread import GangaThread
 from GangaCore.Core.GangaRepository import RegistryKeyError, RegistryLockError
 from GangaCore.Core.exceptions import CredentialRenewalError
@@ -523,6 +524,9 @@ class JobRegistry_Monitor(GangaThread):
 
         self._runningNow = False
 
+        if GANGA_SWAN_INTEGRATION:
+            self.newly_discovered_jobs = []
+
     def isEnabled( self, useRunning = True ):
         if useRunning:
             return self.enabled or self.__isInProgress() and not self.steps
@@ -541,6 +545,16 @@ class JobRegistry_Monitor(GangaThread):
         log.debug("Starting run method")
 
         while self.alive:
+            if GANGA_SWAN_INTEGRATION:
+                # Monitor Jobs from other sessions
+                new_jobs = stripProxy(self.registry_slice).objects.repository.update_index(True, True)
+                self.newly_discovered_jobs = list(set(self.newly_discovered_jobs) | set(new_jobs))
+                for i in self.newly_discovered_jobs:
+                    j = stripProxy(self.registry_slice(i))
+                    job_status = lazyLoadJobStatus(j)
+                    if job_status in ['new']:
+                        stripProxy(self.registry_slice).objects.repository.load([i])
+
             checkHeartBeat(JobRegistry_Monitor.global_count)
             log.debug("Monitoring Loop is alive")
             # synchronize the main loop since we can get disable requests
@@ -636,6 +650,18 @@ class JobRegistry_Monitor(GangaThread):
         self.__updateTimeStamp = time.time()
         self.__sleepCounter = config['base_poll_rate']
 
+    
+    def reloadJob(self, i):
+        """
+        Reload a Job from disk.
+        Parameters:
+        i: Job ID
+        Return:
+            True, if Job is successfully reloaded from disk.
+        """
+        stripProxy(self.registry_slice).objects.repository.load([i])
+        return True
+
     def runMonitoring(self, jobs=None, steps=1, timeout=300):
         """
         Enable/Run the monitoring loop and wait for the monitoring steps completion.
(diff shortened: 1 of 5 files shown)


apsknight:
@egede Sir, there are few changes I made in Ganga to make Job Scripts both Py2 and Py3 compatible. I haven't included those changes in this PR. Should I add these changes ?

These changes can be found here : https://github.com/apsknight/ganga/commit/a2ca466591af3af36f86789d4b8b9d284edefae5

ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -636,6 +650,18 @@ def __monStep(self):
...
+    
+    def reloadJob(self, i):
+        """
+        Reload a Job from disk.
+        Parameters:
+        i: Job ID
+        Return:
+            True, if Job is successfully reloaded from disk.
+        """
+        stripProxy(self.registry_slice).objects.repository.load([i])
```
rob-c, line 677: I think from code inspection `registry_slice` is a `RegistrySlice` object. Therefore the `objects` attribute is an oDict. Shouldn't you be calling the first entry on the oDict here?

Also this is then an object of type `Job` which doesn't have a repository attribute or have I missed what is being called here?

I think what you're after is something similar to:
`jobs.select(47,47)._impl.objects.values()[0]._registry.repository.load([47])`

apsknight: Yes, registry_slice is a RegistrySlice object but `objects` is `JobRegistry` object.
I'm accessing `load` from `GangaRepositoryXML` in similar way as `update_index` method is being called here:

https://github.com/apsknight/ganga/blob/95b6de6aff10947cb4a56ad1d9d2877ab08b39c3/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py#L550

Is there any better way to achieve this reloading of Job from disk. I tried modifying the getter here: 
https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/GPIDev/Base/Objects.py#L328 

but couldn't make it work.

rob-c: OK, the `RegistrySlice` object initially passed has a different form to the object than is constructed from `jobs.select` this is bad behavior and I suspect to fix that behavior requires some serious work in the XXXRegistry classes... (not for now)

I suspect this was an attempt to get the monitoring to pick up new jobs through a dynamically updating reference list but this seems to cause more problems probably due to the confusing different ways `RegistrySlice` can be constructed. This was likely done before the proxy layer was introduced and the new proxy objects which can be constructed at runtime are static objects which have an implicitly different form to the dynamic one.


Exposing the `reloadJob`method will almost certainly fail if called with something like `reloadJob(jobs.select(a,a))` as the object will in this instance be an oDict so it will either need refactoring or at least not exposing through the GPI as `RegistrySlice` objects don't use a very standardized interface. I'd at least suggest not exposing it for now unless this is needed, in which case some work will need to be done to consider the case of `reloadJob(jobs.select(1,1))` from the GPI.


Aside from clearing the contents of an object before re-loading it from disk there is no better fix for things getting out of sync. 


I did some digging and I've spotted at least 1 bug in the monitoring code.
running `runMonitoring(jobs.select(a,b))` will cause all future monitoring to not work correctly as the dynamic job list is incorrectly replaced by a static list of just the objects in the select... this is potentially annoying and might explain some problems I've seen in the past... I might suggest a short PR to fix this.

mesmith75: @rob-c @apsknight what is the status with this?

rob-c: @mesmith75 I think removing this method from the GPI, or adding some inspection for the slice being passed through should be enough, depending on what is needed. I'm tempted to say this _should_ go in, but it won't work as expected for people using `runMonitoring` given a bug in the monitoring code.

As for a bug in the monitoring: https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py#L699 appears to be incorrect as it will potentially replace the main jobresgistry object with a static oDict behind the scenes.

I would suggest a fix but which should just be to remove this line and pass through a custom slice to the methods in the monitoring. This bug exposes itself such that `runMonitoring` works fine but attempting to enable `enableMonitoring` after this will cause it to fault. As would running `runMonitoring(...)` with the core monitoring task enabled as this will replace the monitoring loop with only the jobs requested by the `runMonitoring` method which isn't the correct behavior.
Apologies I've not had the time to make a PR over this I'm a bit busy and only just spotted the bug in the logic after re-visiting it recently, so if someone else wants to step up it should be an easy fix.

I've no idea if the problem with re-loading subjobs remains but that I think is a separate issue to this PR.

apsknight:
All the changes in this PR are protected by an environment variable switch `GANGA_SWAN_INTEGRATION`. So without setting this environment variable, nothing will change in existing Ganga.

But personally, I also think there need to be some better way for reloading Job from disk instead of the current way. I tried to update getter here as suggested by @rob-c earlier :

https://github.com/ganga-devs/ganga/blob/81e0a80a51da26019c18f8a5d2870042789d4594/ganga/GangaCore/GPIDev/Base/Objects.py#L328

to load from disk
https://github.com/ganga-devs/ganga/blob/81e0a80a51da26019c18f8a5d2870042789d4594/ganga/GangaCore/GPIDev/Base/Objects.py#L364

before loading from existing data

https://github.com/ganga-devs/ganga/blob/81e0a80a51da26019c18f8a5d2870042789d4594/ganga/GangaCore/GPIDev/Base/Objects.py#L350

but couldn't make it work as it created many problems which I had no clue of, so I went with this sub-optimal method.

This method `reloadJob` also has a limitation which I described in Github Team discussions, but as of now I don't know how to implement the reloading of job directly from disk and update all existing data of Job in memory.

It'd be better if someone expert in this thing from Ganga Team give a try for implementing a method for reloading Job from disk.

mesmith75 (approved):

mesmith75:
Ok, let's merge this and leave the job reloading for another PR as it is likely to be a little more involved,

PR #1313, opened 2018-08-06
Small monitoring fix

rob-c:
This fixes a minor bug in `runMonitoring`.

When a user executed this method it had the ability to stall the main monitoring thread and then in cases of it being disabled it wouldn't correctly become enabled again after executing this method.

Additionally I've moved the callbacks to use static methods which call instances which makes the code a little easier to follow and from some quick testing improves the stability of the monitoring because it's not trying to track the global `self` state automatically through the monitoring.

### DIFF
--- a/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -470,7 +470,7 @@ def get_jobs_in_bunches(jobList_fromset, blocks_of_size=5, stripProxies=True):
 
 
 class JobRegistry_Monitor(GangaThread):
-
+    """Job monitoring service thread."""
     
     uPollRate = 1.
     minPollRate = 1.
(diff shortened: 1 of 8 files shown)


rob-c:
OK, the additional changes here are needed due to the tests in question being slightly broken in design. They don't rewrite the functional aspect of the tests so I don't see any real problems here.
This was interestingly only exposed in this PR which surprises me but hey it's working now :smile:

mesmith75:
@rob-c Surely the testing issues are more deepset? all the other PRs are failing the tests which they passed previously.

That said these modifications seem to make them more robust.

rob-c:
@mesmith75 best I can tell this was from the migration to the newer versions of the testing packages and how that is handling the different classmethods which are hooked.

With that in mind technically the tests were completely wrong before these changes. And unfortunately the bug was only present in these tests as only these tests attempt to cleanup objects only after all of the functions within a class have finished. I spent ages looking for a filesystem problem due to missing files before I realized it was a class initialization bug.

mesmith75:
ok great! This has been winding us up for several days - see the issues with the release notes PR.

rob-c:
Ah OK, well I claim I've fixed it here, feel free to comment or merge

mesmith75 (approved):

PR #1315, opened 2018-08-08
Fix monitoring fix

mesmith75:
Reverts some code that was broken by #1313


rob-c:
Ah oops, sorry for botching that

rob-c (approved):

PR #1317, opened 2018-08-09
Auto set db tags

mesmith75:
My improvements to #1262 .


mesmith75:
This now adds a new options file with the DB tags in it.

mesmith75:
test this please

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -142,6 +142,7 @@ class GaudiExec(IPrepareApp):
         'jobScriptArchive': GangaFileItem(defvalue=None, hidden=1, copyable=0, doc='This file stores the uploaded scripts which are generated fron this app to run on the WN'),
         'useGaudiRun':  SimpleItem(defvalue=True, doc='Should \'options\' be run as "python options.py data.py" rather than "gaudirun.py options.py data.py"'),
         'platform' :    SimpleItem(defvalue='x86_64-slc6-gcc62-opt', typelist=[str], doc='Platform the application was built for'),
+        'autoDBtags' :  SimpleItem(defvalue=False, doc='Automatically set database tags for MC'),
```
egede, line 145: I suggest to change default to `True`. Otherwise the feature, most beneficial for new users, will never be used.

egede (approved):

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
mesmith75: Perhaps - although this can change the result of the job so I would rather have the user make a conscious decision to use it.

mesmith75:
Thanks to @VanyaBelyaev for the initial idea and code, @paguenther and @ZhenziW for implementation.

PR #1321, opened 2018-08-13
Remove stomp and associated message service

### LINKED ISSUE #1320: Strip out Ganga spyware
The spyware code (using STOMP) should be stripped out of Ganga. I do not think we really use it any more, and to be GDPR compliant it would anyway need a complete revision. First step would be to set `UsageMonitoringMSG` to `False` by default.

mesmith75:
Fixes #1320


alexanderrichards (approved):

egede (approved):
It would be good to disable the monitoring in previous releases from /cvmfs as well by changing the INI files for those (so update the existing one to disable the spyware, and then create a new one that doens't mention it).

mesmith75:
Ah yes good point. I've removed it from the templates here.

I'll add it and set it to false in the `LHCb.ini` in the old CVMFS releases.

Speaking of which do we want to clean up some of the old CVMFS releases? Maybe only keep the last bugfix of each minor version?

PR #1326, opened 2018-08-15
Fix stripProxy to support namedtuples

### LINKED ISSUE #1325: Can't assign a job's dataset if a namedtuple is instantiated in a script
I have a rather bizarre problem. Take this script:

```python
from collections import namedtuple

UserTuple = namedtuple('UserTuple', ['foo', 'bar'])
ut = UserTuple('baz', 123)

j = Job()
j.inputdata = []
```

Then run with Ganga 7.1.1 as a script:

```
$ ganga the_script.py
```

One gets this traceback:

```
  File "/cvmfs/ganga.cern.ch/Ganga/install/LATEST/bin/ganga", line 84, in <module>
    GangaCore.Runtime._prog.run(GPI.__dict__)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/Runtime/bootstrap.py", line 1111, in run
    execfile(script, local_ns)
  File "tmp.py", line 7, in <module>
    j.inputdata = []
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py", line 1089, in _setattr
    GPIProxyObject.__setattr__(self, x, new_v)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py", line 691, in __set__
    final_val = ProxyDataDescriptor._process_set_value(raw_obj, _val, attr_name)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py", line 657, in _process_set_value
    new_val = stripper(val)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py", line 647, in <lambda>
    stripper = lambda v: stripComponentObject(v, cfilter, item)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py", line 328, in stripComponentObject
    vv = cfilter(v, item)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py", line 510, in string_dataset_shortcut
    if isinstance(stripProxy(i), ObjectMetaclass)
  File "/cvmfs/ganga.cern.ch/Ganga/install/7.1.1/lib/python2.7/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py", line 262, in stripProxy
    return type(obj)(stripProxy(_) for _ in obj)
TypeError: __new__() takes exactly 3 arguments (2 given)
```

There is _no_ error if I make either one of the following two changes:

1. Change the definition of `UserTuple` to only have one field, e.g. `UserTuple('UserTuple', ['foo'])`
2. Do not assign the instantiated `UserTuple` to a variable, i.e. delete (`ut =`)

Based on 2, I can also run the script if I `del ut` just before assigning to `j.inputdata`.

So, that's already quite strange, but then I also _don't_ see the error if I run each line in this script in a interactive Ganga session!

I'm completely stumped as to what might be causing this.

chrisburr:
Fixes #1325.

This hasn't been tested as I don't have time at the moment but I think it fixes this issue.

It's not clear to me why the `namedtuple` instance is being give to `stripProxy` but the issue is caused by `namedtuple` changing the `__init__` method of tuple to take multiple arguments instead of a single iterable object.

Unfortunately `namedtuple` doesn't use a base class so you have to guess the type based on it's characteristics, see [this StackOverflow answer](https://stackoverflow.com/a/2166841/2685230).

### DIFF
--- a/ganga/GangaCore/GPIDev/Base/Proxy.py
+++ b/ganga/GangaCore/GPIDev/Base/Proxy.py
@@ -290,6 +290,8 @@ def addProxy(obj):
                 return GPIProxyObjectFactory(obj)
     elif isclass(obj) and issubclass(obj, GangaObject):
         return getProxyClass(obj)
+    elif is_namedtuple_instance(obj):
+        return type(obj)(*[addProxy(_) for _ in obj])
     elif isinstance(obj, (list, tuple)):
         return type(obj)(addProxy(_) for _ in obj)
     elif isinstance(obj, dict):
--- a/ganga/GangaCore/GPIDev/Base/Proxy.py
+++ b/ganga/GangaCore/GPIDev/Base/Proxy.py
@@ -252,13 +252,22 @@ def getName(_obj):
     returnable = _getName(obj)
     return returnable
 
+def is_namedtuple_instance(obj):
+    if len(type(obj).__bases__) != 1 or not isinstance(obj, tuple):
+        return False
+    fields = getattr(obj, '_fields', None)
+    if not isinstance(fields, tuple):
+        return False
+    return all(isinstance(n, str) for n in fields)
 
 def stripProxy(obj):
     """Removes the proxy if there is one
     Args:
         obj (object): This may be an instance or a class
     """
-    if isinstance(obj, (list, tuple)):
+    if is_namedtuple_instance(obj):
+        return type(obj)(*[stripProxy(_) for _ in obj])
+    elif isinstance(obj, (list, tuple)):
         return type(obj)(stripProxy(_) for _ in obj)
     elif isinstance(obj, dict):
         return dict((k, stripProxy(v)) for k, v in obj.items())


rob-c:
@chrisburr Is it possible to add an equivalent check for the `addProxy` method so that objects that are constructed using this type within Ganga have their child objects appropriately protected

alexpearce:
Amazing work!

I'd love to know how the `namedtuple` ends up in this piece of Ganga code. Particularly because I only saw the error when assigning to `Job.inputdata`, which seemed particularly odd to me.

Still, thanks a lot for the fix!

chrisburr:
@rob-c Done!

chrisburr:
It might be worth adding a test [here](https://github.com/ganga-devs/ganga/blob/b024d67066f02bfa545f928b324de7d9842abf65/ganga/GangaCore/test/Unit/TestProxy.py#L8-L13) to ensure the proxy supports `namedtuple`. I haven't done it as I don't fully understand the schema property and I don't have a test environment set up to figure it out with.

This branch should be editable by members of `ganga-devs` if anyone wants to add it in this PR.

rob-c:
@alexpearce I suspect this problem is from the `ProxyDataDescriptor._process_set_value` level of the stack trace in #1325 
When assigning an object to an attribute of a `GangaObject` type class Ganga will immediately attempt to understand the object being assigned to a member variable in the context of the current namespace. This would allow it to use things such as types defined at runtime by the user.
This is really a fiddly area of the code as the objects have to be instantiated in the correct order.

The actual _bug_ here is that Ganga seems to be wasting CPU by parsing all possible objects in the namespace prior to the attribute assignment. (which in Ganga terms creates a new instance of the object upon assignment and is therefore closer to a deep copy which is needed for thread-safe internal behavior).
That being said I don't know another way of walking up/down an objects inheritance path to determine if the proxy layer has added/removed the appropriate proxies to any children to build up the Ganga public/private model within Python. I can semi-construct some solutions but these quickly become complex pieces of code.

@chrisburr Thanks for adding the `addProxy` code this means that should you have assigned this object to something within Ganga hopefully you will get something sensible back from the XML file if this gets saved. (I've not checked but that's the hope)

chrisburr:
@rob-c Thanks for the explanation. It's odd as it only happens when running as a script, is it something like an IPython feature that is only used for interactive sessions? I tried attaching pdb to investigate but I couldn't get it to work. (Likely my fault, I don't use it very often.)

mesmith75 (approved):

mesmith75:
@chrisburr Thanks!

PR #1334, opened 2018-08-31
New merger for GaudiExec

### LINKED ISSUE #1029: Make how Root is setup for mergers more configurable
At present, GangaLHCb uses
```
[ROOT]
path = /cvmfs/lhcb.cern.ch/lib/RootConfig/pro/x86_64-slc6-gcc49-opt
```

And then sets ROOTSYS to this, LD_LIBRARY_PATH to this+/lib and then runs this+/bin/hadd. This fails badly if LHCb users are just running ganga straight (i.e. outside of lb-run) as the compiler (and other things) are not added to the LD_LIBRARY_PATH. This could be completely fixed by doing:
```
lb-run ROOT hadd
```
Instead but there is no config option to completely change the `rootprefix` used in the `RootMerger:mergefiles` function. I would propose adding this as a config option and setting the default LHCb setting to the above.

### LINKED ISSUE #1333: ROOT not correctly setup
The ROOT that ganga is trying to use for merging is not usable the way ganga tries to use it, as it requires a newer `libstdc++` than available on the system. A LHCb-specific solution could be to give the possibility to configure the entire prefix and set it to `lb-run root ` for LHCb. I'm not sure how version selection can be implemented in this scheme. 

Version: 7.1.3
Tested platforms: `x86_64-slc6-gcc62-opt` and `x86_64-centos7-gcc62-opt`
Reproducer:
```python
import GangaCore
import commands
cmd = GangaCore.Utility.root.getrootprefix()[1] + 'hadd'
commands.getstatusoutput(cmd)

# /cvmfs/lhcb.cern.ch/lib/RootConfig/pro/x86_64-centos7-gcc62-opt/bin/hadd: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /cvmfs/lhcb.cern.ch/lib/RootConfig/pro/x86_64-centos7-gcc62-opt/bin/hadd)
# /cvmfs/lhcb.cern.ch/lib/RootConfig/pro/x86_64-centos7-gcc62-opt/bin/hadd: /lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by /cvmfs/lhcb.cern.ch/lib/RootConfig/pro/x86_64-centos7-gcc62-opt/lib/libGraf3d.so)
# ...
```

PS It might be useful to run this or something similar to see where libraries come from.
```sh
lb-run root bash -c 'ldd `which hadd`'
```

mesmith75:
Fixes #1333 . Fixes #1029 .

A new merger for the the LHCb GaudiExec application. This just uses the run script of the application to do the `hadd`.

### DIFF
--- a/ganga/GangaLHCb/Lib/Mergers/GaudiExecMerger.py
+++ b/ganga/GangaLHCb/Lib/Mergers/GaudiExecMerger.py
@@ -41,9 +41,9 @@ class GaudiExecMerger(IMerger):
 
     # will produce the specified files
     j = Job() 
-    j.outputsandbox = ['hist.root','trees.root']
+    j.outputfiles = ['hist.root','trees.root']
     j.splitter = SomeSplitter()
-    j.merger = gm
+    j.postprocessors = gm
     j.submit()
 
     The merge object will be used to merge the output of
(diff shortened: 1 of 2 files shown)


ganga/GangaLHCb/Lib/Mergers/GaudiExecMerger.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+    def mergefiles(self, masterjob, file_list, output_file):
+
+        #First grab the job object etc.
+        j = masterjob
+        sharedir = j.application.is_prepared.path()
+        tmp_dir = tempfile.gettempdir()
+        #Extract the run script
+        tar = tarfile.open(os.path.join(sharedir, "cmake-input-sandbox.tgz"), "r:gz")
```
egede, line 225: The temporary file and its directory should be cleaned up after use. This should be done in a way such that it happens if an exception is thrown.

egede (changes requested):
This is a very good idea.

egede (approved):

PR #1338, opened 2018-09-05
Update athena-utility.sh - use the 'here' asetup option ...

will-cern:
…for all project types

Was only enabled for AthAnalysisBase project before.


egede:
@will-cern Can you update branch to the latest version of develop. I think it is an ancient one as it is still attempting to run test cases against python 2.6 (and then fails already at setup).

PR #1339, opened 2018-09-05
use asetup 'here' option in all releases

will-cern:
ok last attempt, then I am bored!


egede:
From Will: To test:

```
asetup 20.20.8.8,here
cmt new_skeleton MyPackage
cmt find_packages
cmt compile
mkdir run; cd run
ganga athena --condor .... --extOutFile=PoolFileCatalog.xml MyPackage/MyPackageAlgJobOptions.py
```

egede:
@drmarkwslater If we had a docker image with the ATLAS software (via cvmfs), it should be easy to build a test that always runs for things like this.

egede:
@will-cern Thanks for the multiple iterations. The tests are indeed running correctly now.

will-cern:
There is a docker image that contains cvmfs so that setupATLAS and then 'asetup' can be run inside it: https://hub.docker.com/r/atlas/atlas_external_cvmfs/

mesmith75:
What is the status of this? Are we waiting for the ATLAS test to be set up?

mesmith75:
test this please

mesmith75 (approved):

Issue #55, opened 2015-11-27
Import missing in Grid.py for CREAM backend


drmarkwslater:
>> Good afternoon,
>>
>> I have tried using Ganga-6-1-6-hotfix1 from cvmfs:
>>
>> /cvmfs/ganga.cern.ch/Ganga/install/LATEST/bin/ganga
>>
>> to submit jobs to Manchester Analysis Facility but I get the error
>> below. Could you please have a look?
>>
>> Thank you,
>> Rustem
>>
>> ---------------------------------------------------------------------------------------------------------------------
>> *** Welcome to Ganga ***
>> Version: Ganga-6-1-6-hotfix1
>> Documentation and support: http://cern.ch/ganga
>> Type help() or help('index') for online help.
>>
>> This is free software (GPL), and you are welcome to redistribute it
>> under certain conditions; type license() for details.
>>
>> Ganga.Utility.Config               : INFO     reading config file
>> /afs/hep.man.ac.uk/u/rustem/.gangarc
>> Ganga.GPIDev.Lib.Job               : INFO     submitting job 3251
>> Ganga.GPIDev.Lib.Job               : INFO     job 3251 status changed
>> to "submitting"
>> Ganga.Lib.Executable               : INFO     Preparing Executable application.
>> Ganga.Lib.Executable               : INFO     Created shared
>> directory: conf-30cf5d3b-450d-4f1b-b234-d4ea6cf07192
>> Ganga.GPIDev.Adapters              : INFO     Sending file object
>> /bin/bash to shared directory
>> Ganga.GPIDev.Lib.Job               : INFO     Preparing subjobs
>> Ganga.Lib.Executable               : INFO     Submitting a prepared
>> application; taking any input files from
>> conf-30cf5d3b-450d-4f1b-b234-d4ea6cf07192
>> Ganga.Lib.LCG                      : WARNING  CREAM CE assigment from
>> AtlasCREAMRequirements failed.
>> Ganga.GPIDev.Lib.Job               : WARNING  ---------- error in
>> user/extension code ----------
>> Ganga.GPIDev.Lib.Job               : WARNING  Traceback (most recent call last):
>> Ganga.GPIDev.Lib.Job               : WARNING    File
>> "/cvmfs/ganga.cern.ch/Ganga/install/LATEST/python/Ganga/GPIDev/Lib/Job/Job.py",
>> line 1592, in submit
>> Ganga.GPIDev.Lib.Job               : WARNING      r =
[...]

PR #1343, opened 2018-09-13
subjobs progress column

mesmith75:
Add a subjobs progress column to the job registry. Just an idea.

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -112,11 +112,11 @@ class SubjobStatuses(GangaObject):
     _category = 'subjobstatuses'
     _name = 'SubjobStatuses'
 
-    def __init__(self, nSubjobs):
+    def __init__(self):
         super(SubjobStatuses, self).__init__()
-        self.totalSubjobs = nSubjobs
+        self.totalSubjobs = 0
         self.jobStatuses = {
-            'new': nSubjobs,
+            'new': 0,
             'submitting': 0,
             'submitted': 0,
             'running': 0,
@@ -176,7 +176,7 @@ class JobInfo(GangaObject):
         'monitor': ComponentItem('monitor', defvalue=None, load_default=0, comparable=0, optional=1, doc="job monitor instance"),
         'uuid': SimpleItem(defvalue='', protected=1, comparable=0, doc='globally unique job identifier'),
         'monitoring_links': SimpleItem(defvalue=[], typelist=[tuple], sequence=1, protected=1, copyable=0, doc="list of tuples of monitoring links"),
-        'subjob_statuses': SimpleItem(defvalue=SubjobStatuses(0))
+        'subjob_statuses': SimpleItem(defvalue=SubjobStatuses())
     })
 
     _category = 'jobinfos'
@@ -674,16 +674,18 @@ class Job(GangaObject):
 
         elif self.master:
             if not hasattr(self.master.info, 'subjob_statuses'):
-                self.master.info.subjob_statuses = SubjobStatuses(len(self.master.subjobs))
+                self.master.info.subjob_statuses = SubjobStatuses()
+                self.master.info.subjob_statuses.totalSubjobs = len(self.subjobs)
                 for sj in self.master.subjobs:
                     self.master.info.subjob_statuses.incrementStatus(sj.status)
         else:
            if not hasattr(self.info, 'subjob_statuses'):
                if not self.subjobs:
-                   self.subjob_statuses = SubjobStatuses(0)
-                   self.incrementStatus(final_status)
+                   self.info.subjob_statuses = SubjobStatuses()
+                   self.info.subjob_statuses.incrementStatus(final_status)
                else:
-                   self.subjob_statuses = SubjobStatuses(len(self.subjobs))
+                   self.info.subjob_statuses = SubjobStatuses()
+                   self.info.subjob_statuses.totalSubjobs = len(self.subjobs)
                    for sj in self.subjobs:
                        self.info.subjob_statuses.incrementStatus(sj.status)
 
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -103,68 +103,6 @@ class FakeError(GangaException):
     def __init__(self):
         super(FakeError, self).__init__()
 
-class SubjobStatuses(GangaObject):
-    """ This just stores the overall status of the subjobs. This rather relies on good bookkeeping instead of keeping tabs on the individual jobs."""
-    _schema = Schema(Version(0,1), {
-        'totalSubjobs' : SimpleItem(defvalue=0, typelist=[int], copyable=0),
-        'jobStatuses': SimpleItem(defvalue={}, typelist=[dict], copyable=0, doc="a dict that summarises the subjobs statuses")
-    })
-    _category = 'subjobstatuses'
-    _name = 'SubjobStatuses'
-
-    def __init__(self):
-        super(SubjobStatuses, self).__init__()
-        self.totalSubjobs = 0
-        self.jobStatuses = {
-            'new': 0,
-            'submitting': 0,
-            'submitted': 0,
-            'running': 0,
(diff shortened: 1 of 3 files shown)


mesmith75:
Now prevents loading the whole job. If a field is not in the index then it just returns `---`.

ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -729,13 +812,18 @@ def getSubJobStatuses(self):
 
 	return stats
 
+    def returnSubjobStatuses(self):
+        if hasattr(self.info, 'subjob_statuses'):
```
egede: I don't think that the `info` part is in the index file. So will this not trigger loading of the full job?

mesmith75: I don't think so. The change in `RegistrySlice.py` should prevent a job being loaded if the new column is not in the index so this function won't be called until the job is updated by the monitoring.

rob-c: I think I agree with Ulrik here, and upon inspection I realize I forgot to add a test to make sure the job summary is safe for lazy loading so this can be broken and pass the integration suite.

rob-c: @mesmith75 does this PR pass if you try to run #1346?

mesmith75: @rob-c Looks to be.

rob-c: Ah, OK, I think I see the code that checks for the lazy loading being broken. this looks like a quite cool extension :+1:

egede (approved):

mesmith75:
The one wrinkle with this is that jobs submitted with it cannot be loaded with an older version of ganga. It gives loading errors if you try;
```python
ERROR    #1 Error(s) Loading File: /afs/cern.ch/work/m/masmith/GangaTest/test/repos/repository/masmith/LocalXML/6.0/jobs/0xxx/55/data
ERROR    err: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
ERROR    #1 Error(s) Loading File: /afs/cern.ch/work/m/masmith/GangaTest/test/repos/repository/masmith/LocalXML/6.0/jobs/0xxx/55/data~
ERROR    err: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
ERROR    XML File failed to load for Job id: 55
ERROR    Actual Error was:
Repository 'jobs' object #55 is not accessible because of an InaccessibleObjectError: Repository 'jobs' object #55 is not accessible because of an PluginManagerError: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
ERROR    Loading: 55 into incomplete_objects to avoid loading it again in future
ERROR    Error Loading Job! '55'
ERROR    Caught Unknown Callback Exception
ERROR    Callback:
 Repository 'jobs' object #55 is not accessible because of an InaccessibleObjectError: Repository 'jobs' object #55 is not accessible because of an PluginManagerError: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")

[18:19:12]
Ganga In [1]: ERROR    Caught Unknown Callback Exception
ERROR    Callback:
 'IncompleteObject' object has no attribute 'status'
ERROR    Caught Unknown Callback Exception
ERROR    Callback:
 'IncompleteObject' object has no attribute 'status'
ERROR    Caught Unknown Callback Exception
ERROR    Callback:
 'IncompleteObject' object has no attribute 'status'
ERROR    Caught Unknown Callback Exception
ERROR    Callback:
 'IncompleteObject' object has no attribute 'status'
ERROR    Caught Unknown Callback Exception
ERROR    Callback:
```

I am not sure how much of an issue this is - hopefully people won't be going backwards but I guess it is a possibility.

rob-c:
> The one wrinkle with this is that jobs submitted with it cannot be loaded with an older version of ganga. It gives loading errors if you try;
> 
> ```python
> ERROR    #1 Error(s) Loading File: /afs/cern.ch/work/m/masmith/GangaTest/test/repos/repository/masmith/LocalXML/6.0/jobs/0xxx/55/data
> ERROR    err: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
> ERROR    #1 Error(s) Loading File: /afs/cern.ch/work/m/masmith/GangaTest/test/repos/repository/masmith/LocalXML/6.0/jobs/0xxx/55/data~
> ERROR    err: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
> ERROR    XML File failed to load for Job id: 55
> ERROR    Actual Error was:
> Repository 'jobs' object #55 is not accessible because of an InaccessibleObjectError: Repository 'jobs' object #55 is not accessible because of an PluginManagerError: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
> ERROR    Loading: 55 into incomplete_objects to avoid loading it again in future
> ERROR    Error Loading Job! '55'
> ERROR    Caught Unknown Callback Exception
> ERROR    Callback:
>  Repository 'jobs' object #55 is not accessible because of an InaccessibleObjectError: Repository 'jobs' object #55 is not accessible because of an PluginManagerError: PluginManagerError: (PluginManagerError(...), u"cannot find 'SubjobStatuses' in a category 'subjobstatuses', or elsewhere")
> 
> [18:19:12]
> Ganga In [1]: ERROR    Caught Unknown Callback Exception
> ERROR    Callback:
>  'IncompleteObject' object has no attribute 'status'
> ERROR    Caught Unknown Callback Exception
> ERROR    Callback:
>  'IncompleteObject' object has no attribute 'status'
> ERROR    Caught Unknown Callback Exception
> ERROR    Callback:
>  'IncompleteObject' object has no attribute 'status'
> ERROR    Caught Unknown Callback Exception
> ERROR    Callback:
>  'IncompleteObject' object has no attribute 'status'
> ERROR    Caught Unknown Callback Exception
> ERROR    Callback:
> ```
> 
> I am not sure how much of an issue this is - hopefully people won't be going backwards but I guess it is a possibility.

This is because of the schema being designed to not allow for data loss to be caused when downgrading to an older version of Ganga. I think the only way around this is to store the data in a dictionary or other flat object which somehow seams unappealing.

Looking at this all the information I think is available in the subjob-xml which is lazy loaded with the main job isn't it?
https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Core/GangaRepository/SubJobXMLList.py#L484

If you change the class to parse this info from the `SubJobXMLList` rather than store it and have to be a `GangaObject` then you also remove the risk in the data becoming out of sync between the 2 indexes.

egede:
I think it is fine that they can't be loaded with the older version. We have in the past bumped up the second version number when this happens.

mesmith75:
@rob-c You are right, that is a much more elegant solution.

ganga/GangaCore/GPIDev/Lib/Registry/RegistrySlice.py
```diff
@@ -479,6 +479,8 @@ def _display(self, interactive=0):
                         #pass
                         if item == "fqid":
                             vals.append(self._get_display_value(obj, item))
+                        elif item == 'subjob status':
```
rob-c, line 482: Is this exception still needed?

mesmith75: I think so - Would not the `self.subjobs` end up loading the job? With this check that won't happen for old jobs.

rob-c: Ah, yes, you're probably correct. It will load the master job before the subjob data is loaded.
Can't think of a nicer solution that just returning `----` in that case for unloaded jobs. Actually that's also a nice way by eye of seeing which master jobs have been loaded into memory.

PR #1349, opened 2018-09-24
Make Dirac sandbox unpacking optional

mesmith75:
From Dirac v6r20 and LHCbDirac v9r2p7


mesmith75:
There was a mistake in the DIRAC change I made so the default needs to be `False` for now otherwise the sandbox won't get downloaded at all. I would like to get this merged today so we can release it as a temporary fix.

egede (approved):

PR #1357, opened 2018-10-03
Remove credential decorator from splitter

mesmith75:
#1352  broke the splitter. I think the issue was something else.


egede:
In which way did it break? The decorator should be transparent if you have a proxy.

mesmith75:
```python
ERROR    Exception raised executing 'getLFNReplicas' in Thread 'None':
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/ganga/GangaCore/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 115, in __worker_thread                                                                
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/ganga/GangaCore/GPIDev/Credentials/__init__.py", line 26, in cred_wrapped_method                                                                                 
    cred_req = self.credential_requirements                                                                                                                                                                                                 
AttributeError: 'list' object has no attribute 'credential_requirements'     
```
I cannot reproduce the original issue - I am not sure how he managed it. Looking at SplitByFiles, the `split` function already has a `require_credential` decorator in place.

egede (approved):

PR #1359, opened 2018-10-18
Update Notifier.py

ghost:
Add in the notification mail from ganga also the job name.
It would be nice to add as well in the mail sent the user which has done that in such a way if the notifier sends mails to a MailingList of a shared analysis project one can know which one has completed the job. 
I am just not sure how to get the ```whoami``` inside ganga libraries.


mesmith75:
According to stack overflow the following will get you the user:
```python
import getpass
getpass.getuser()
```

egede (approved):

PR #1360, opened 2018-10-18
Update GridftpSandboxCache.py

mesmith75:


mesmith75:
So this now definitely works for ARC and has been tested.

The behaviour with `arcget` is rather different than previously so the required changes were a little more involved.

egede (approved):

PR #1361, opened 2018-10-22
Default leave Dirac output sandbox compressed

mesmith75:
Now set the default to leave DIRAC sandboxes compressed. Is that what we want (probably for LHCb but maybe not others) or should we make it configurable?

Also wrap the dirac command in a try/except in case an old version of DIRAC is used which doesn't have this option.


mesmith75:
test this please

egede (commented):
I suggest to make it configurable, leave the default as uncompressed and then in the LHCb.ini file set it to compressed.

mesmith75:
It turns out this doesn't work with DIRAC as it is. If you set `unpack` to `False` the tar file ends up in a temp directory and the return value says `None` so you don't know its name.

I'll raise it with them but it might be best to jump to not downloading at all as in #1145 .

PR #1366, opened 2018-10-24
Update ganga-cvmfs-install.sh

mesmith75:


egede (approved):
Is this a result of setuptools becoming deprecated and that we should move to something else?

mesmith75:
@egede I don't think it is deprecated, it is just the version with python installed in LCG is ancient (v12 compared to the current v40)

egede:
@mesmith75 OK

PR #1367, opened 2018-10-30
 Compare to `None` using identity `is` operator 

janisozaur:
This is a trivial change that replaces `==` operator with `is` operator, following PEP 8 guideline:

> Comparisons to singletons like None should always be done with is or is not, never the equality operators.

https://legacy.python.org/dev/peps/pep-0008/#programming-recommendations


egede (approved):
This looks innocent enough as part of a general code clean-up and can be merged. Out of curiosity, was it triggered by anything failing?

janisozaur:
> was it triggered by anything failing?

No, just spotted it by chance.

mesmith75 (approved):

PR #1369, opened 2018-11-01
improve Dirac finalisation

mesmith75:
More improvements to Dirac finalisation.

Puts each block of jobs to finalise into a thread to prevent clogging up the terminal.

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -611,7 +611,10 @@ class Job(GangaObject):
 
         if new_status in ['completed', 'failed', 'killed']:
             if len(self.postprocessors) > 0:
-                logger.info("Running postprocessor for Job %s" % self.getFQID('.'))
+                if self.master:
+                    logger.debug("Running postprocessor for Job %s" % self.getFQID('.'))
+                else:
+                    logger.info("Running postprocessor for Job %s" % self.getFQID('.'))
                 passed = self.postprocessors.execute(self, new_status)
                 if passed is not True:
                     new_status = 'failed'
@@ -664,7 +667,7 @@ class Job(GangaObject):
                 if outputfileClass in backend_output_postprocess[backendClass]:
 
                     if not self.subjobs:
-                        logger.info("Job %s Setting Location of %s: %s" % (self.getFQID('.'), getName(outputfile), outputfile.namePattern))
+                        logger.debug("Job %s Setting Location of %s: %s" % (self.getFQID('.'), getName(outputfile), outputfile.namePattern))
                         try:
                             outputfile.setLocation()
                         except Exception as err:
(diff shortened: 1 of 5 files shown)


ganga/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -1160,91 +1160,99 @@ def finalise_jobs(allJobs, downloadSandbox = True):
 
         #First grab all the info from Dirac
         inputDict = {}
-        #I have to reduce the no. of subjobs per process to prevent timeouts
+        #I have to reduce the no. of subjobs per process to prevent DIRAC timeouts
         nPerProcess = int(math.floor(configDirac['maxSubjobsPerProcess']/2))
```
egede: Why the /2 here. Is that because it is the same number that control the number of jobs for submission. If yes, should it not be two different parameters instead?

ganga/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -1160,91 +1160,99 @@ def finalise_jobs(allJobs, downloadSandbox = True):
...
-                                postprocesslocationsfile.flush()
-
-                    logger.debug("Written: %s" % open(lfn_store, 'r').readlines())
-
-                #Set the status of the subjob
-                sj.updateStatus(statusmapping[statusList['Value'][sj.backend.id]['Status']])
+    @staticmethod
+    def finalise_jobs_thread_func(jobSlice, downloadSandbox = True):
+        """
+        Finalise the jobs given. This downloads the output sandboxes, gets the final Dirac stati, completion times etc.
```
egede: typo "stati" -> "statistics"

egede (commented):
Looks really good overall.

ganga/GangaDirac/Lib/Backends/DiracBase.py
mesmith75: I wanted to use the same parameter as a more general limit to the number of things we can do in one go. The problem is the finalisation takes rather longer than the submission so I had to divide by two to avoid timeouts. I'll make it a separate configurable.

ganga/GangaCore/Utility/logging/__init__.py
```diff
@@ -600,3 +600,18 @@ def log_unknown_exception():
 
     tb_logger.debug('Bare except clause triggered {0}:{1}'.format(caller.filename, caller.lineno))
     tb_logger.debug('Exception caught:', exc_info=True)
+
+def supress_logging(func):
```
egede: IS this function problematic in the multithreaded environment of Ganga? It could cause other messages not to come out, and if the method is entered twice, the `previousloglevel` variable might get the wrong value.

mesmith75: I don't think this does what I want. Really I was trying to suppress the messages about the postprocessors running but they occur elsewhere.

I think I'll get rid of it.

ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -664,15 +667,15 @@ def postprocessoutput(self, outputfiles, outputdir):
                 if outputfileClass in backend_output_postprocess[backendClass]:
 
                     if not self.subjobs:
-                        logger.info("Job %s Setting Location of %s: %s" % (self.getFQID('.'), getName(outputfile), outputfile.namePattern))
```
egede, line 667: To me it looks like you get DEBUG for the master and INFO for the subjob with this change. Don't you want it the other wayr around? Other places seems to have similar logic.

mesmith75: `not self.subjobs` implies the job doesn't have subjobs, so it is itself a subjob. I agree it is confusing. I'll change it to `if self.master`

mesmith75: Actually I think it is fine as is - from my tests it works as intended.

egede (approved):

PR #1370, opened 2018-11-01
Update OfflineGangaDiracSplitter.py

mesmith75:
Potential fix for exception thrown on worker thread causing main thread to hang.

Would be better if there was some way of registering a queued task and checking if it's completed, but addressing that looks beyond the scope of this.


mesmith75:
thanks @rob-c !

egede (commented):
I can see how this avoids the infinite loop. A message is emitted when we catch the exception and we then continue just making an empty match. I am unsure about if the job can still submit correctly, i.e. is there just some match missing but if others are there it doesn't matter. If that is the case, the message should be changed to a WARNING, rather than an ERROR. If the submission will always fail in this case, how will than then happen?

egede (approved):

mesmith75:
@egede You are right. The offending SE gets removed from the mapping if the returned list is empty (at L286) so the splitting could still be successful. If the lack of the SE means the problem is intractable that will be caught later on and the submission will fail.

I have therefore changed that to a warning.

PR #1371, opened 2018-11-02
Check no of locations of GaudiExec files

mesmith75:
Check if the file already has replicas and do nothing if it has.


egede (commented):
Is this the correct place to implement this. The method it is inside is having a name that seems to indicate that it will replicate to a raondom SE, but now we get code that might not result in a replication. So should this check be moved to the called instead, or is the name if the method misleading?

mesmith75:
ah yes, good point

egede (approved):

PR #1372, opened 2018-11-02
Set KRB5CCNAME in Dirac env

### LINKED ISSUE #1368: permission errors when downloading to eos
When downloading output files of grid jobs to eos via the eos fuse mount on lxplus I get errors like this:

```
GangaDiracError: {'Successful': {}, 'Failed': {'/lhcb/user/t/ttekampe/2018_10/233615/233615802/Bu_D0pipipi_Kenu_MC_12185008_2016_Sim09c_MagDown.root': 'Operation not permitted ( 1 : Could not copy srm://lhcbsdrm.
t1.grid.kiae.ru:8443/srm/managerv2?SFN=/t1.grid.kiae.ru/data/lhcb/lhcbuser/lhcb/user/t/ttekampe/2018_10/233615/233615802/Bu_D0pipipi_Kenu_MC_12185008_2016_Sim09c_MagDown.root to file:///eos/lhcb/user/t/ttekampe/K
1sim_noStrip/134/10/Bu_D0pipipi_Kenu_MC_12185008_2016_Sim09c_MagDown.root, [1] Could not open destination: errno reported by local system call Operation not permitted)'}}
```

What I do exactly is putting `DiracFile.Get(...)` into the ganga queues. That way multiple files are downloaded at the same time. I reported the issue to the eos support and they might have found the source of the problem.

Quoting  Andreas Joachim Peters:

> So what I think is, that GANGA switches authentication tokens to talk to the GRID, which result in a permission denied error on the FUSE mount. You probably use a grid certificate to fetch files and to do so the KRB5CCNAME variable might get invalidated and when this happens in an uncoordinated way by parallel agents you end up revoking the authentication token to talk to EOS. I see that the initial requests are mapped to your kerberos name while later they are mapped to a GRID pool account which has no permission to work in your directory. It might be that in the past this has worked because there was a nother credential binding mechansim used on the FUSE mount.

So a possible workaround, which seems to work for some people is to just download one file at a time, but maybe there is also a way to fix this in ganga?

mesmith75:
Fixes #1368

### DIFF
--- a/ganga/GangaDirac/Lib/Utilities/DiracUtilities.py
+++ b/ganga/GangaDirac/Lib/Utilities/DiracUtilities.py
@@ -226,7 +226,8 @@ def execute(command,
 
     if cred_req is not None:
         env['X509_USER_PROXY'] = credential_store[cred_req].location
-        env['KRB5CCNAME'] = os.getenv('KRB5CCNAME')
+        if os.getenv('KRB5CCNAME'):
+            env['KRB5CCNAME'] = os.getenv('KRB5CCNAME')
 
     if cwd is None:
         # We can in all likelyhood be in a temp folder on a shared (SLOW) filesystem
--- a/ganga/GangaDirac/Lib/Utilities/DiracUtilities.py
+++ b/ganga/GangaDirac/Lib/Utilities/DiracUtilities.py
@@ -226,6 +226,7 @@ def execute(command,
 
     if cred_req is not None:
         env['X509_USER_PROXY'] = credential_store[cred_req].location
+        env['KRB5CCNAME'] = os.getenv('KRB5CCNAME')
 
     if cwd is None:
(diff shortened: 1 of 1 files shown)


egede (changes requested):
Test if environment variable is defined firt. Otherwise this could throw an exception.

mesmith75:
I don't think it does return an exception - you just get `None`.
```
 os.getenv(varname[, value])

    Return the value of the environment variable varname if it exists, or value if it doesn’t. value defaults to None.
```

egede:
OK, but you will then still create the  KRB5CCNAME environment variable in the Dirac process with an empty value. This could give rise to an odd error (kerberos expects to find a file name as a string in this variable). It would be better not to set it.

egede (approved):

PR #1374, opened 2018-11-07
Check for failure in Condor status check

### LINKED ISSUE #1354: Condor monitoring error
We sometimes get an error when monitoring Condor jobs:
```python
[20:16:02]
Ganga In [3]: ERROR    Exception raised executing 'updateMonitoringInformation' in Thread 'Backend Monitor':
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/ganga/GangaCore/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 115, in __worker_thread
  File "/afs/cern.ch/user/m/masmith/cmtuser/GANGA/GANGA_HEAD/install/ganga/ganga/GangaCore/Lib/Condor/Condor.py", line 583, in updateMonitoringInformation
    allDict[id]["status"] = Condor.statusDict[status]
KeyError: 'performing'
```

mesmith75:
Fixes #1354 .

Sometimes there is an authentication error in the query - it is intermittent (not a sign of something fatal) but it puts something odd in the status field. This just checks that it has not returned failure.

### DIFF
--- a/ganga/GangaCore/Lib/Condor/Condor.py
+++ b/ganga/GangaCore/Lib/Condor/Condor.py
@@ -575,7 +575,8 @@ class Condor(IBackend):
             tmpList = infoString.split()
             id, host, status, cputime = ("", "", "", "")
             if 3 == len(tmpList):
-                id, status, cputime = tmpList
+                if not 'Failure' in tmpList[0]:
+                    id, status, cputime = tmpList
             if 4 == len(tmpList):
                 id, host, status, cputime = tmpList
             if id:


ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -575,7 +575,8 @@ def updateMonitoringInformation(jobs):
             tmpList = infoString.split()
             id, host, status, cputime = ("", "", "", "")
             if 3 == len(tmpList):
-                id, status, cputime = tmpList
+                if not 'Failure' in tmpList[0]:
+                    id, status, cputime = tmpList
             if 4 == len(tmpList):
                 id, host, status, cputime = tmpList
```
egede, line 581: Should you not make a similar type of check here. If it can go wrong with a length of 3, I guess is can with a length of 4 as well. Would it be better to just have an `if not 'Failure' in tmpList` check earlier?

mesmith75: Yes ok.

This is due to a specific message giving something like `['AUTHENTICATION:Failed', 'performing', 'handshake']` so I think it is only an issue for length 3 but I will make it general.

egede (approved):

ganga/GangaCore/Lib/Condor/Condor.py
mesmith75: On second thoughts it is a specific error message to catch. Other ones are `failured`, `failed` etc but with different lengths, so I'll keep as is. If there are more problems we can revisit.

PR #1376, opened 2018-11-14
combine standard output and standard error for all ND280 validation jobs

JacekHoleczek:


egede (approved):
Looks fine. Just remember that both stdout and stderr are buffered in memory. When writing them to a single file, they will thus not always be synchronised (causing confusion about where in the programme the error came from.

JacekHoleczek:
Thanks. Actually it seems to be working quite fine (i.e. the output is quite in sync) and that's why I decided to backport it from my production ganga 6.0.44 to the current "develop" version.

PR #1377, opened 2018-11-14
Subjob inputdata references master job

mesmith75:
This only works with `SplitByFiles`. When subjobs are created their `inputdata` now points to the `inputdata` of the master job. Should save a bit of space.


ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -15,6 +15,7 @@
 from GangaDirac.Lib.Backends.DiracUtils import get_result
 from GangaCore.GPIDev.Lib.GangaList.GangaList import GangaList, makeGangaListByRef
 from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
+import GangaCore.GPI as GPI
```
egede: Looks odd. Why do we want to use the GPI? Isn't that only for the end user?

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -55,7 +57,7 @@ class LHCbDataset(GangaDataset):
                       'getLFNs', 'getFileNames', 'getFullFileNames',
                       'difference', 'isSubset', 'isSuperset', 'intersection',
                       'symmetricDifference', 'union', 'bkMetadata',
-                      'isEmpty', 'hasPFNs', 'getPFNs']  # ,'pop']
+                      'isEmpty', 'hasPFNs', 'getPFNs', 'setReference', 'createIndex']  # ,'pop']
```
egede: Should these new methods be exposed to the end user?

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -110,20 +112,37 @@ def isFileTest(_file):
...
+
     def __getitem__(self, i):
         '''Proivdes scripting (e.g. ds[2] returns the 3rd file) '''
         #this_file = self.files[i]
         # print type(this_file)
         # return this_file
         # return this_file
         # return this_file
+        if self.ref:
+            return GPI.jobs(self.ref).inputdata[self.files][i]
```
egede: So is this really the correct way to do this?

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
```diff
@@ -110,20 +112,37 @@ def isFileTest(_file):
...
         logger.debug("Dataset Created")
 
+    def setReference(self, jobNo, indices):
+        '''A function to set up the references of a subjob's dataset to the master job'''
+        self.ref = jobNo
+        self.files = indices
+
+    def createIndex(self):
+        '''Populate a list of the dataset's LFNs for some quick sorting'''
+        self.lfnList = [df.lfn for df in self.files]
```
egede, line 179: Will this create a very large list in memory that we then never garbage collect? Is it a problem?

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: I agree it is not ideal. But `LHCbDataset` should exist as a standalone object so this was the only way I could think of to access a job on the registry.

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: Ah no, that was for my testing. I'll remove that.

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: I couldn't think of a better way...but there probably is one.

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: Yes. I should get the splitter to clear this after splitting.

ganga/GangaLHCb/Lib/LHCbDataset/LHCbDataset.py
mesmith75: Found a better way.

egede (approved):

PR #1382, opened 2018-11-30
Use old dirac-install script

alexanderrichards:
New dirac-install script requires non-standard python package six to be already installed. Either use old version of script like this PR does or we will need to to a local install to the docker container until this is fixed upstream.


mesmith75:
@alexanderrichards Trying from the v6r19 release branch.

mesmith75 (approved):

PR #1384, opened 2018-12-03
Set ganga_jobid when running GaudiExec

mesmith75:
Sets an environment variable with the subjob id when running GaudiExec applications.


egede (commented):
Why is this an option in the schema. Can't you just always do it?

mesmith75:
That was my first effort. But now it just happens always.

egede (approved):
Sorry for misunderstanding.

PR #1387, opened 2018-12-10
update LHCb config template

mesmith75:


egede (commented):
I thought this was no longer used?

mesmith75:
Ah, good point. Perhaps this is better...

egede (approved):
Much better!

PR #1391, opened 2018-12-14
Improve site SE mapping

mesmith75:
Uses the function to give the mapping for every SE. Much quicker (20s -> 6s, most of which is the connection overhead) than doing it one by one and most large datasets will likely use most SEs anyway.


mesmith75:
test this please

egede (approved):
Faster and neater!

PR #1394, opened 2018-12-25
Start a DIRAC process

mesmith75:
Resurrect having a running process for DIRAC commands.

To do:

- [x]  Check it is running
- [x] Close itself after some amount of time
- [x] Deal with returns that are not dictionaries (i.e. deprecation messages) 
- [x] Sort path to script
- [x] Security
- [x] Port range
- [x] Thread safe


mesmith75:
test this please

mesmith75:
So there are some improvements:
The port to use is just picked at random by starting a socket, binding it to 0 and then finding out that port no.

There is a random string created that is given to the server via a env variable in the subprocess. Commands need to have this string in them to be executed, otherwise the connection is closed by the server.

egede:
> There is a random string created that is given to the server via a env variable in the subprocess. Commands need to have this string in them to be executed, otherwise the connection is closed by the server.

Environment variables for a process can be viewed by everybody so it is no better than using an option on the command line.

mesmith75:
Oops.

How about just piping it directly?

mesmith75:
test this please

ganga/GangaDirac/BOOT.py
```diff
@@ -42,16 +49,81 @@ def diracAPI(cmd, timeout=60, cred_req=None):
...
 #\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/\/#
 
+running_dirac_process = False
+dirac_process = None
+dirac_process_ids = None
+def startDiracProcess():
+    '''
+    Start a subprocess that runs the DIRAC commands
+    '''
+    HOST = '127.0.0.1'  #Connect to localhost
```
egede: Would it be better to use the name "localhost" instead? In that case you are not forcing IPv4.

ganga/GangaDirac/BOOT.py
```diff
@@ -42,16 +49,81 @@ def diracAPI(cmd, timeout=60, cred_req=None):
...
+
+    #Pass the port no as an argument to the popen
+    serverpath = os.path.join(os.path.dirname(inspect.getsourcefile(runClient)), 'DiracProcess.py')
+    popen_cmd = [serverpath, str(PORT)]
+    dirac_process = subprocess.Popen(popen_cmd, env = getDiracEnv(), stdin=subprocess.PIPE)
+    global running_dirac_process
+    running_dirac_process = (dirac_process.pid, PORT)
+
+    #Now set a random string to make sure only commands from this sessions are executed
+    rand_hash = uuid.uuid1()
```
egede: Better to use uuid4?

ganga/GangaDirac/Lib/Server/DiracProcess.py
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env python
+import sys
+import os
+import errno
+import socket
+import traceback
+HOST = '127.0.0.1'  # Standard loopback interface address (localhost)
```
egede: Same localhost comment

ganga/GangaDirac/Lib/Utilities/DiracUtilities.py
```diff
@@ -237,21 +227,63 @@ def execute(command,
...
+    returnable = ''
+    if not new_subprocess:
+        with Dirac_Exec_Lock:
+            # First check if a Dirac process is running
+            from GangaDirac.BOOT import running_dirac_process
+            if not running_dirac_process:
+                startDiracProcess()
+            #Set up a socket to connect to the process
+            from GangaDirac.BOOT import dirac_process_ids
+            HOST = '127.0.0.1'  # The server's hostname or IP address
```
egede: and here

egede (approved):

PR #1398, opened 2019-01-11
Start dirac process with python

mesmith75:
Can't start a bash script due to file permissions so start with python.

### DIFF
--- a/ganga/GangaDirac/BOOT.py
+++ b/ganga/GangaDirac/BOOT.py
@@ -70,7 +70,7 @@ def startDiracProcess():
     s.close()
     #Pass the port no as an argument to the popen
     serverpath = os.path.join(os.path.dirname(inspect.getsourcefile(runClient)), 'DiracProcess.py')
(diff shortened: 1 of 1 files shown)


ganga/GangaDirac/BOOT.py
```diff
@@ -70,7 +70,7 @@ def startDiracProcess():
     s.close()
     #Pass the port no as an argument to the popen
     serverpath = os.path.join(os.path.dirname(inspect.getsourcefile(runClient)), 'DiracProcess.py')
-    popen_cmd = [serverpath, str(PORT)]
+    popen_cmd = ['python',serverpath, str(PORT)]
```
egede, line 73: Could it be undefined what version of python this will give you? Does it matter?

mesmith75: This will give whatever is in the path so it is undefined. I don't think it matters - there is nothing that looks to be version specific in the script.

egede (approved):

PR #1408, opened 2019-01-16
SLURM improvements, /tmp/tmp* management fixes

JacekHoleczek:


ganga/GangaCore/__init__.py
```diff
@@ -29,7 +29,7 @@ def getLCGRootPath():
 # ------------------------------------------------
 # store Ganga version based on new git tag for this file
 _gangaVersion = '7.1.9'
-_development = True
+_development = False
```
mesmith75, line 32: Please don't change this.

JacekHoleczek:
Sorry, I did not change it myself. I was modifying original files from the ganga-7.1.9 release and that's how they came.
Can you fix it or do you want me to do something about it?

ganga/GangaCore/__init__.py
JacekHoleczek: Sorry, I did not change it myself. I was modifying original files from the ganga-7.1.9 release and that's how they came.
Can you fix it or do you want me to do something about it?

JacekHoleczek:
Let me try to fix it myself with a new pull request ... https://github.com/ganga-devs/ganga/pull/1409

egede:
Great to put in a pull request. Just remember to branch from the `develop` branch in your fork and make sure that you have merged in the latest changes from there before you create the pull request.

JacekHoleczek:
I work with stable released versions which I try to get working. I modify them in-place. The easiest way for me would be to send you a standard patch file. Doing a pull request is additional work and as you can see rather error-prone.

egede:
Hi,

Sending a patch file is OK as well, but it might take a bit longer to 
implement a change s we then haev to include it ourselves in a pull request.

Ulrik.

PR #1409, opened 2019-01-16
SLURM improvements, /tmp/tmp* management fixes

JacekHoleczek:


mesmith75 (approved):
Thanks for these updates!

Issue #1401, opened 2019-01-12
Problem with Executable (exe always copied into sandbox)


JacekHoleczek:
I've got a strange problem (ganga-7.1.9).
My simple script uses:
```
t = CoreTask()
trf = CoreTransform()
trf.application = Executable()
```

[According to the documentation](https://ganga.readthedocs.io/en/stable/GPI/classes.html#Executable), if I say:
```
trf.application.exe = File('/bin/sh')
```
then ganga will make a "copy" of the executable.

But if I say:
```
trf.application.exe = '/bin/sh'
```
then ganga should simply use the given path (without copying the executable).

However, what I get is that in BOTH cases, ganga copies "sh" into the "sandbox".

You can easily test this behaviour, if you take [the "test_exit.py" script from one of my recent issues](https://github.com/ganga-devs/ganga/issues/1400) and play with the "trf.application.exe =" line inside (you may also need to change "Slurm" into your own backend, of course).

mesmith75:
The documentation is not quite correct. From the prepare stage:
```
        The application wil have a Shared Directory object created for it. 
        If the application's 'exe' attribute references a File() object or
        is a string equivalent to the absolute path of a file, the file 
        will be copied into the Shared Directory.
```
Ganga checks if it can resolve the string to a file. If so, it gets copied. This makes sense to me as you cannot be sure the file is available on the WN (particularly with the grid).

Therefore I will update the docstring and close this.

JacekHoleczek:
So, I have met some cases in which exactly the opposite is the case. The executables exist ONLY on batch nodes and NOT on the head node where Ganga runs. Two situations are implemented:
1. the "/full/path/to/exe" file does NOT exist on the head note at all (but it exists on batch nodes),
2. the "/full/path/to/exe" file on the head node is a "dummy" shell script which simply informs the user that the real executable is allowed to be executed on batch nodes only.

So, I really think there should be an easy way in Ganga to specify a batch-node-only-path without copying the executable into the "sandbox" (otherwise one needs to write "dummy" wrapper / proxy scripts which encapsulate the real call to the real "exe").

Well, I could also imagine situations in which "batch nodes" run another operating system than the "head node" where Ganga runs ... so no executable should be copied in sandbox.

[...]

PR #1412, opened 2019-01-18
Never copy executable string

mesmith75:


JacekHoleczek:
Please see: https://github.com/ganga-devs/ganga/issues/1401#issuecomment-455644446

JacekHoleczek:
This looks almost great.
Just one note.
You write "If the string can be resolved to an existing file, ganga will copy the file to the sandbox to be shipped to the WN." but I think in case of a "string", it should never be copied to the sandbox, even if the file exists on the node where Ganga runs (in order to make sure that false "dummy" executables are not copied from the head node do WN). Only explicit "File('string')" should be copied.
That was actually the original idea, wasn't it: "When you want to execute a command on the worker node you should specify it as a string. Ganga will call the command with its full path on the worker node:"

Please again see (I added there another two cases when copying into sandbox needs to be prohibited): https://github.com/ganga-devs/ganga/issues/1401#issuecomment-455644446

JacekHoleczek:
Could I have a question here, please.
In the [update docstring](https://github.com/ganga-devs/ganga/pull/1412/commits/1f2055ede427d1dfe1a51873280f56e859b50608) patch, I find: "In this case, a wrapper script which calls this same command is created and placed into the Shared Directory."
But I cannot locate the source code which does it in the other patch.
I just want to ask you the following ... when you create a wrapper script, please make sure that you use a ganga specific name for it ... like `'./_ganga_'+original_exe_name` (or `'./_sandbox_'+original_exe_name`).
I think we should not create a script with the same name as the original executable name ... just as a precaution for cases when the user prepends  the `os.environ["PATH"]` with "." in the "preexecute" string (and then somewhere in some script uses the `original_exe_name` without specifying the full path, for example, in order to "locate" its configuration files, data or modules).

mesmith75:
test this please

JacekHoleczek:
I'm afraid I would need some help ... how to get it into a new ganga distribution installed with "pip" (in a "virtualenv").

JacekHoleczek:
I looked at the "changed file" ("Executable.py") and it seems to me that you need to "reverse" conditions in lines starting with 126 (well, "str" takes precedence and the "source" may not exist where the ganga executable runs):
```
            if isinstance(self.exe, str):
                logger.debug("exe is a string so no copying")
            elif not os.path.exists(source):
                logger.debug("Error copying exe: %s to input workspace" % str(source))
            else:
```

See: https://github.com/ganga-devs/ganga/pull/1421

egede (commented):
Seems to have an overlap with #1421 . Maybe this PR can just be retired?

JacekHoleczek:
Yes, I took this patch, tried it and found that I needed to patch yet another file in order to get it working (and a patch for both files is #1421  which got "mixed" yesterday with my new SLURM related patch).

There is one point which I did not cover. As written in one of my previous comments here.
I just want to ask you the following ... if you need to create a wrapper script, please make sure that you use a ganga specific name for it ... like `'./_ganga_'+original_exe_name` (or `'./_sandbox_'+original_exe_name`).

BTW. A corresponding patch for the old Ganga-6-0-4[432] is here: https://github.com/ganga-devs/ganga/issues/1401#issuecomment-462399166

PR #1421, opened 2019-02-11
Never copy executable string and SLURM improvements

JacekHoleczek:
One needs to patch two files.

I still always get the "Created shared directory: conf-..." message, but the directory is empty (I tried both, "relative" and "absolute" executable paths as strings). I don't know how to prevent the creation of this directory, when it is not needed.

BTW. A corresponding patch for the old Ganga-6-0-4[432] is here: https://github.com/ganga-devs/ganga/issues/1401#issuecomment-462399166


JacekHoleczek:
I wanted to create a separate new "pull request" for the "SLURM improvements", but my today's patches to `ganga/GangaCore/__init__.py` got added to my old (unrelated) pending patches here. You can "split" it into two separate "pull requests" if you like.

egede (approved):

PR #1427, opened 2019-03-13
Fix job removal

### LINKED ISSUE #1426: DiracFile with inputfiles deleted too early
If the following is executed

```
j.prepare()
j.submit()

:
Wait some days
:
j2 = j.copy()
j2.submit()
j.remove()
```
the job `j2` will fail as the `j.remove` deleted the uploaded sandbox. The logic for deleting uploaded sandboxes should be the same as for deleting them in the shared area.

mesmith75:
Fixes #1426 .

For some reason every subjob was decrementing the shareref counter when a job gets remove so the sharedir was getting deleted.


egede (approved):
Easy if that is indeed the full fix.

PR #1434, opened 2019-03-18
Make location of file for used versions configurable

### LINKED ISSUE #1403: ${HOME}/.cache/Ganga/.used_versions
It seems that ganga-7.1.9 unconditionally creates a "${HOME}/.cache/Ganga/.used_versions" file.
The old ganga-6.0.44 was creating the ".used_versions" file in the "gangadir" (I always set a ganga version specific "gangadir").
Could you, please, restore the old behaviour (so that ganga does not pollute "${HOME}/.cache").

mesmith75:
Fixes #1403


egede (commented):
What is the motivation for this change? Does it really matter for anybody where this file is? Adding flexibility could in principle just make it harder to support users.

mesmith75:
@egede I agree entirely, but it was requested in #1403 . If we agree we don't like the suggestion we can forget this and close the issue,

JacekHoleczek:
Well, I like the idea of setting the ".used_versions" file location.
I've got a small question. If I define "gangadir = /my/own/ganga/dir", can I then somehow define the ".used_versions" file location using something like "used_versions_path = ${gangadir}"?

mesmith75:
no

JacekHoleczek:
O.K. Thanks.
It seems I cannot set / modify environment variables from inside of ".gangarc" but I guess I can simply "export MY_GANGA_DIR=/my/own/ganga/dir" (before calling ganga) and define "gangadir = ${MY_GANGA_DIR}" and also "used_versions_path = ${MY_GANGA_DIR}" which should also work fine.

egede (approved):
Doesn't harm I guess, so fine to go ahead with.

PR #1436, opened 2019-03-19
Update DiracCommands.py

mesmith75:


egede (approved):
Nice that you picked up on this one from MatterMost.

PR #1439, opened 2019-04-01
Download uploaded Dirac sandboxes

mesmith75:
If the initial sandbox download fails this check if there is a sandbox uploaded to grid storage and downloads it.

A few questions:
- Do we want to download it automatically? It may block up the monitoring if jobs take a long time to finalise due to downloading large sandboxes.
- The sandbox is still compressed - do we want to decompress it as well?  I think we do.
- Do we want to give a warning about oversize sandboxes?
- Do we want to store the sandbox LFN?

### DIFF
--- a/ganga/GangaDirac/Lib/Server/DiracCommands.py
+++ b/ganga/GangaDirac/Lib/Server/DiracCommands.py
@@ -157,6 +157,16 @@ def getOutputSandbox(id, outputDir=os.getcwd(), unpack=True, oversized=True, noJ
             os.system('mv -f %s/* %s/. ; rm -rf %s' % (tmpdir, outputDir, tmpdir))
         
         os.system('for file in $(ls %s/*_Ganga_*.log); do ln -s ${file} %s/stdout; break; done' % (outputDir, outputDir))
+    #So the download failed. Maybe the sandbox was oversized and stored on the grid. Check in the job parameters and download it
+    else:
+        parameters = dirac.getJobParameters(id)
+        if parameters is not None and parameters.get('OK', False):
+            parameters = parameters['Value'][id]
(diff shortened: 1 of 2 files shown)


egede:
>    Do we want to download it automatically? It may block up the monitoring if jobs take a long time to finalise due to downloading large sandboxes.

Yes, I think it should be automatic. 

>    The sandbox is still compressed - do we want to decompress it as well? I think we do.

Yes.

>    Do we want to give a warning about oversize sandboxes?

Yes, I think that is a nice idea.

>    Do we want to store the sandbox LFN?

No, I would actually suggest to remove the LFN after we have downloaded it.

ganga/GangaDirac/Lib/Server/DiracCommands.py
```diff
@@ -157,7 +157,14 @@ def getOutputSandbox(id, outputDir=os.getcwd(), unpack=True, oversized=True, noJ
...
         os.system('for file in $(ls %s/*_Ganga_*.log); do ln -s ${file} %s/stdout; break; done' % (outputDir, outputDir))
-
+    #So the download failed. Maybe the sandbox was oversized and stored on the grid. Check in the job parameters and download it
+    else:
+        parameters = dirac.getJobParameters(id)
+        if parameters is not None and parameters.get('OK', False):
+            parameters = parameters['Value'][id]
+            if 'OutputSandboxLFN' in parameters:
+                sandbox_lfn = parameters['OutputSandboxLFN']
+                result = dirac.getFile(sandbox_lfn, destDir=outputDir)
```
egede: So emit warning here and then if download worked, remove the LFN.

egede (approved):
Looks good now.

PR #1445, opened 2019-04-09
Account for subfiles in auto file removal

mesmith75:
Fixes #1442

PR #1447, opened 2019-04-09
Improve DiracFile removals

mesmith75:
Dirac's `removeFile` can take a list of LFNs so it is quicker to do that if we are removing multiple files. Therefore I have added the massRemove function to DiracFile.

Once this and #1445 have been merged I'll put the massRemove into the auto output remove in the job to be as quick as possible.


ganga/GangaDirac/Lib/Files/DiracFile.py
```diff
@@ -403,6 +403,49 @@ def remove(self):
...
+                logger.warning(msg)
+                return False
+            #Check that the number of removed files is the same as the number requested
+            if result.get('Successful', False):
+                _successes.extend(result['Successful'].keys())
+            if result.get('Failed', False):
+                _failures.extend(result['Failed'].keys())
+            i = i+200
+
+        def clearFileInfo(f):
```
egede, line 440: You are changing information of objects that might sit inside a completed job. Will the job get correctly marked as dirty and the change persisted?

mesmith75: I'm not sure how one would mark the job. The change of the DiracFile object should be persisted though.

egede: If it works, I am not too worried.

egede (approved):

mesmith75:
I don't think I like this. I will try to do it better.

PR #1456, opened 2019-04-12
fix hostname()

### LINKED ISSUE #1452: ACTUALCE missing in all "__jobstatus__" files (ganga 7.1.12)
It is possible that the problem originates in the `ganga/GangaCore/Lib/Batch/Batch.py` file, in the line:
```
        '###INLINEHOSTNAMEFUNCTION###' : inspect.getsource(Utility.util.hostname),
```
It does copy the "hostname()" source code into the `__jobscript__` but, I think that one vital line (present in the `ganga/GangaCore/Utility/util.py` file) is not copied (and then it is missing):
```
hostname._hostname_cache = ''
```

It seems to me that the "hostname()" function would internally need to check if the `hostname._hostname_cache` exists and, if not, it should create it automatically (then the additional external line which defines it would not be needed).

JacekHoleczek:
hopefully fixes https://github.com/ganga-devs/ganga/issues/1452

### DIFF
--- a/ganga/GangaCore/Utility/util.py
+++ b/ganga/GangaCore/Utility/util.py
@@ -150,7 +150,7 @@ def hostname():
     """ Try to get the hostname in the most possible reliable way as described in the Python LibRef."""
 
     # cache the result to prevent lockups in gethostbyaddr calls with queues
-    if hostname._hostname_cache == '':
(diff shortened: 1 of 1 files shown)


alexanderrichards (approved):
This looks sensible to me

PR #1457, opened 2019-04-15
Move logging flush to pre-prompt hook

mesmith75:
For some reason IPython reprints the input prompt for every key stroke. This means the loggers are getting flushed in the middle of typing commands - very annoying.

Here I have moved the flushing to the post execute hook (it could equally well be in the pre-prompt one). The problem is that a cell now needs to be executed for the logging to be printed - just hitting return on an empty cell does nothing. I am not sure how to change that behaviour.

### DIFF
--- a/ganga/GangaCore/Runtime/bootstrap.py
+++ b/ganga/GangaCore/Runtime/bootstrap.py
@@ -1218,8 +1218,6 @@ under certain conditions; type license() for details.
         import IPython
 
         # Based on examples/Embedding/embed_class_long.py from the IPython source tree
-        from IPython.core.displayhook import DisplayHook
-
         # First we set up the prompt
         from traitlets.config.loader import Config
         
@@ -1236,7 +1234,7 @@ under certain conditions; type license() for details.
         from IPython.terminal.embed import InteractiveShellEmbed
 
         ipshell = InteractiveShellEmbed(config=cfg, banner1=banner, exit_msg=exit_msg)
-        ipshell.events.register("post_execute", ganga_prompt)
+        ipshell.events.register("pre_execute", ganga_prompt)
 
         # Add our custom error handler to ignore stack traces for GangaExceptions
         ipshell.set_custom_exc((Exception,), error_handler)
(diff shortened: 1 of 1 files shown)


alexanderrichards:
Quoting from IPython [docs](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.hooks.html#IPython.core.hooks.pre_prompt_hook):

> IPython.core.hooks.pre_prompt_hook(self):
>
>     Run before displaying the next prompt
>     Use this e.g. to display output from asynchronous operations
>     (in order to not mess up text entry)

Possibly, modifying the following line:

https://github.com/ganga-devs/ganga/blob/03a3032ebc9857b8482072c74d352edcb712bc5f/ganga/GangaCore/Runtime/bootstrap.py#L1239

along the lines of:
```diff
-        ipshell.events.register("post_execute", ganga_prompt)
+        ipshell.set_hook("pre_prompt_hook", ganga_prompt)
```

---
*P.s.* Shows how long since I tinkered with the boostrap, I assumed that this was the way we already did it but of course as you pointed out it all changed with the IPython 5 updates. :wink:

ganga/GangaCore/Runtime/bootstrap.py
```diff
@@ -1220,6 +1218,7 @@ def launch_IPython(local_ns, error_handler, ganga_prompt):
         import IPython
 
         # Based on examples/Embedding/embed_class_long.py from the IPython source tree
+        from IPython.core.displayhook import DisplayHook
```
alexanderrichards: Looking at the diff of this patch, unless I missed it, DisplayHook is never used so does it need to be imported?

mesmith75: Ah yes - this was my trying to alter the display hook.

mesmith75:
I tried the display hook but it gives the same behaviour, i.e. not printing when `return` is pressed - only if the cell gets run.

alexanderrichards:
> I tried the display hook but it gives the same behaviour

Do you mean pre_prompt_hook?

mesmith75:
yes

alexanderrichards:
hmmm... not sure then.

alexanderrichards (approved):

PR #1462, opened 2019-05-10
Add LbDevTools to requirements

mesmith75:


egede (commented):
Is it possible to make Ganga complain in a helpful way if a user installs via pip (but without the LHCb dependency) and then subsequently is enabling the LHCb extension).

egede (dismissed):

mesmith75:
It seems we have to source directly `LbEnv` to run Dirac now.

Also now `ROOTSYS` is not initially set in the LHCb environment. This makes no different as previously it was just empty so I got rid of the root environment store. Maybe we should just set this to be empty by default in case it is needed elsewhere?

egede (dismissed):

egede (dismissed):

egede:
So this seems more or less ready. Should we merge it and let it go to the DEV version? How do we manage the requirement that LHCb projects are checked out again? Can we check if they are wrong and then abort the submission with a helpful message? Will this still work for old releases or are we creating a problem there?

mesmith75:
So I don't think it is easy to check how the project was checked out.
This is a bit of an issue - Ganga versions will directly relate to the LHCb software giving people a reason not to use the latest.

egede:
Last time that we had an issue like this, we created a whole new set of application classes (moved from `DaVinci` etc, to `GaudiExec`. Do we really need to do that again?

egede (dismissed):

mesmith75:
So I think this is right now. It checks the Makefile for LbEnv so it should be transparent to the users and not require checking out the application again.

Perhaps we should add a deprecation warning?

egede (approved):

PR #1471, opened 2019-05-28
Adds Memory Profiler and CPU Profiler

ishanrai05:


ganga/GangaCore/Utility/Monitoring.py
```diff
@@ -0,0 +1,54 @@
...
+        datafn = mpath+func.__name__ +timestr+".profile"
+        prof = cProfile.Profile()
+        retval = prof.runcall(func, *args, **kwargs)
+        prof.dump_stats(datafn)
+        return retval
+    return wrapper
+
+def cpu_profiler(cls, profile_cpu=c['Profile_CPU']):
+    if profile_cpu:
+        for key, value in list(vars(cls).items()):
```
alexanderrichards: `items()` presumably already returns a list so the explicit list construction seems irrelevant. Also since these are temporary intermediary lists simply used in the iteration perhaps a generator would be better.

i.e. does this not simplify to:
```python
for key, value in vars(cls).iteritems():
```

ganga/GangaCore/Utility/Monitoring.py
```diff
@@ -0,0 +1,54 @@
...
+    else:
+        pass
+    return cls
+
+def mem_profiler(cls, profile_memory=c['Profile_Memory']):
+    if profile_memory:
+        cpath = makedir(os.path.join(path,'memory_profiles/'))
+        file_name = cpath+cls.__name__+timestr+'.log'
+        fp = open(file_name, 'w+')
+        for key, value in list(vars(cls).items()):
```
alexanderrichards: ditto

alexanderrichards (approved):

ganga/GangaCore/Utility/Monitoring.py
```diff
@@ -0,0 +1,54 @@
+from GangaCore.Utility.Config import getConfig
```
mesmith75: Is this a new file? Perhaps a better name would be `Profiling`? `Monitoring` makes me think of other things, like the job monitoring.

ganga/GangaCore/Utility/Monitoring.py
```diff
@@ -0,0 +1,54 @@
...
+
+import os
+from memory_profiler import profile
```
mesmith75: I don't think `memory_profiler` is standard. Is the purpose of this branch to go into a release? If so `memory_profiler` needs adding to the requirements.

egede: Yes. To install memory_profiler from pip also requires that the python-devel package is installed on any centos/fedora system. If not already in the centos7 docker image, it will need to be added to  ganga/ganga/GangaCore/test/Dockerfile for the testing to work.

ganga/GangaCore/GPIDev/Base/Objects.py
```diff
@@ -25,6 +25,8 @@
 
 from GangaCore.Utility.Plugin import allPlugins
 
+from GangaCore.Utility.Monitoring import cpu_profiler, mem_profiler
```
egede: Where is the cpu_profiler package from. I did not find it with pip.

ganga/GangaCore/Utility/Monitoring.py
```diff
@@ -0,0 +1,54 @@
...
+    if profile_memory:
+        cpath = makedir(os.path.join(path,'memory_profiles/'))
+        file_name = cpath+cls.__name__+timestr+'.log'
+        fp = open(file_name, 'w+')
+        for key, value in list(vars(cls).items()):
+            if callable(value):
+                setattr(cls, key, profile(value, stream=fp, precision=6))
+    else:
+        pass
+    return cls
```
egede: Add the missing newline.

ganga/GangaCore/GPIDev/Base/Objects.py
```diff
@@ -687,6 +689,8 @@ def __init__(cls, name, bases, this_dict):
             this_schema.createDefaultConfig()
 
 
+@cpu_profiler
+@mem_profiler
```
egede, line 694: Is there a performance price to pay for adding these two, even if the cpu and memory profiler is turned off?

ganga/GangaCore/GPIDev/Base/Objects.py
ishanrai05: @egede  cProfile comes pre-installed with python. [docs](https://docs.python.org/2/library/profile.html)

ganga/GangaCore/Utility/Monitoring.py
ishanrai05: Yes. Thank you for this.

ganga/GangaCore/Utility/Monitoring.py
ishanrai05: Yes, `Profiling` would be a better name.

ganga/GangaCore/GPIDev/Base/Objects.py
ishanrai05: There is no performance lag on my machine. However, we should test this in real world world scenarios just to be sure.

ganga/GangaCore/Utility/Monitoring.py
ishanrai05: I have updated the code to meet PEP 8 requirements.

ganga/GangaCore/GPIDev/Base/Objects.py
alexanderrichards: No shouldn't really be as these are only run once at module import time and are essentially a no-op when deactivated

ganga/GangaCore/Utility/Monitoring.py
ishanrai05: I added  `memory_profiler` to requirements.txt

egede (approved):
Looks good now.

mesmith75:
Shall we merge this?

egede:
> Shall we merge this?
OK for me. @ishanrai05 what do you say?

ishanrai05:
I added a decorator to count the number of calls made to GangaObject. We can merge this  if you think it is okay.

egede (approved):
As agreed.

PR #1473, opened 2019-06-03
Fix splitter copy

### LINKED ISSUE #1472: Undefined variable in job class
While doing the 2 to  3 conversion...
https://github.com/ganga-devs/ganga/blob/1547fa1d73b1e29d09ee4a7708f411e7a7e518f2/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L2182

here variable `do_not_copy` appears undefined unless I'm missing something?

mesmith75:
Fixes #1472

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -26,7 +26,6 @@ from GangaCore.GPIDev.Lib.Job.MetadataDict import MetadataDict
 from GangaCore.GPIDev.Schema import ComponentItem, FileItem, GangaFileItem, Schema, SimpleItem, Version
 from GangaCore.Utility.Config import ConfigError, getConfig
 from GangaCore.Utility.logging import getLogger, log_user_exception
-from GangaCore.GPIDev.Base.Objects import do_not_copy
 
 from .JobTime import JobTime
 from GangaCore.Lib.Localhost import Localhost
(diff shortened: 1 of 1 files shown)


alexanderrichards (dismissed):
This will undoubtedly fix the problem however I wonder if, given this hasn't tripped anyone up, the function that uses this import is even needed/used at all?

egede:
@mesmith75 The code that this is fixing is just 6 months old, see https://github.com/ganga-devs/ganga/commit/d6422fa79d5255d93f1678876f8b75419cf8a134 . Is this part of some project that was never finished? Is the real solution just to delete the method that contains the missing import?

alexanderrichards:
looks like it came in as part of #1392

egede:
So we got a speed-up from adding code that is never called ;-)

mesmith75:
As I say this is a derivation of https://github.com/ganga-devs/ganga/blob/f1448c8849a3e1393a52222bd2caae53bf773cc1/ganga/GangaCore/GPIDev/Base/Objects.py#L806 .

Essentially there were a whole load of checks that were taking a lot of time so I got rid of them. The part that is not called would also not be called in the original although I didn't check. I guess we can do away with it.

mesmith75:
I guess the implication is that the things that get copied to a new subjob when splitting are not Nodes. As I don't really understand what a Node means here I cannot tell you if that is reasonable.

In any case the code is removed.

alexanderrichards (approved):

PR #1478, opened 2019-06-11
Nice error when trying to download master job sandbox

mesmith75:


egede (approved):
Should have been there a long time ago. Thanks.

PR #1480, opened 2019-06-13
Adds proxy for copy on write

ishanrai05:


ganga/GangaCore/GPIDev/Base/Proxy.py
```diff
@@ -1205,6 +1205,165 @@ def _getattribute(self, name):
...
+    The proxy is a transparent wrapper that intercepts all access to the
+    underlying object. When first created, it creates a shallow copy
+    of the underlying object and behaves like a shadow reference. When
+    any attribute of the underlying object is modified, the COWProxy either
+    creates a complete deep copy (it deep copies the entire object it was
+    tracking) or creates a partial deep copy (it deep copies only the
+    attribute that was touched). This behavior is triggered by an appropriate
+    flag.
+    """
+    __slots__ = ["_obj", "__weakref__", "__slots__", "_is_copied",
```
alexanderrichards: You don't need to define `"__slots__"` within `__slots__`

ganga/GangaCore/GPIDev/Base/Proxy.py
```diff
@@ -1205,6 +1205,165 @@ def _getattribute(self, name):
...
+                 "_enable_partial_copy", "_attr_map"]
+
+    _is_copied = False
+    _attr_map = {}
+
+    def __init__(self, obj, _partial_copy=True):
+        object.__setattr__(self, "_obj", obj)
+        object.__setattr__(self, "_enable_partial_copy", _partial_copy)
```
alexanderrichards: as long as the attribute names `_obj` and `_enable_partial_copy` are defined within `__slots__` (as they are) you should be able to assign these using the standard:
```python
self._obj = obj
self._enable_partial_copy = _partial_copy
```
rather than this unbounded (static) call to object base class's `__setattr__`.

ganga/GangaCore/GPIDev/Base/Proxy.py
```diff
@@ -1205,6 +1205,165 @@ def _getattribute(self, name):
...
+            cls._class_proxy_cache = cache = {}
+        try:
+            theclass = cache[obj.__class__]
+        except KeyError:
+            cache[obj.__class__] = theclass = cls._create_class_proxy(obj.__class__)
+        ins = object.__new__(theclass)
+        theclass.__init__(ins, obj, *args, **kwargs)
+        return ins
+
+
```
alexanderrichards: My general feeling is that this is quite dense, especially with all the static calls to the object base class. I have thrown together a simple first approach. What are the deficiencies of the following?
```python
class COWProxy(object):
    __slots__ = ('_obj', '_partial', '_partial_dict', '_copied')

    def __init__(self, obj, partial=False):
        self._obj = obj
        self._partial = partial
        self._partial_dict = {}
        self._copied = False

    def __getattribute__(self, attr):
        if self._copied:
            return getattr(self._obj, attr)
        return self._partial_dict.setdefault(attr, COWProxy(getattr(self._obj, attr)))

    def __setattr__(self, attr, value):
        if attr in COWProxy.__slots__:
            return object.__setattr__(self, attr, value)
        if self._copied:
            return setattr(self._obj, attr, value)
        if not self._partial:
            self._obj = deepcopy(self._obj)
            self._copied = True
            return setattr(self._obj, attr, value)
        self._partial_dict[attr] = value
```

ganga/GangaCore/GPIDev/Base/Proxy.py
alexanderrichards: I see it's because you have overridden the `object.__setattr__` and you don't want it triggered on these instance variables

ganga/GangaCore/GPIDev/Base/CoW.py
```diff
@@ -0,0 +1,504 @@
...
+        duck typing of python to make this all transparent.
+    """
+
+    # __slots__ = ['__weakref__','__cow_add_cb_pointer','_my_flyweight_cb_func','_flyweight_cache']
+
+    # Using weakref to allow garbage collection
+    # dict[<var_type>][__hash__]
+    _flyweight_cache = {}
+
+    def __init__(self, *args, **kwargs):
```
mesmith75, line 97: Could you also define some comparison operators as well? It might be useful when this gets integrated into python3

ganga/GangaCore/GPIDev/Base/CoW.py
```diff
@@ -0,0 +1,504 @@
...
+                    if self[key] is old:
+                        self[key] = new
+
+            # Iterate over set
+            elif issubclass(type(self), set):
+                if old in self:
+                    self.remove(old)
+                    self.add(new)
+
+        except Exception as e:
```
mesmith75, line 269: I'm not sure I like this treatment of the exception. Can we catch a particular one and deal with it properly?

mesmith75:
Is this quicker than the current copying? When I did the splitter copy for jobs I ended up taking out a whole load of code that just slowed it down - it looks like this is quite dense so I wonder if it is still quick.

mesmith75 (changes requested):

ganga/GangaCore/GPIDev/Base/CoW.py
ishanrai05: yes. on it.

ganga/GangaCore/GPIDev/Base/CoW.py
ishanrai05: yes, I am working on it. also while switching to python3 we could use `OrderedDict` instead of `dict` in ProxyDict, we won't have to sort the dictionary for getting hash keys then.

alexanderrichards: If you feel that the `OrderedDict` approach is better and should be what we are using for python 3 then once this code is merged into the python 3 branch please make the change and open a PR against the python 3 branch.

ishanrai05: Yes.. Sure @alexanderrichards

egede (approved):

egede (commented):
Do you need the specific version number here? If not, I would take it away. These locked in version often cause trouble in the future.

ishanrai05:
@egede The Jenkins tests are not running. 
I have removed the version number.

alexanderrichards:
I just killed the jenkins jobs as they were indeed stuck. I'll let the backlog clear to determine if it was just for this PR they were stuck or if there is a more general problem. It could be that the docker image store has filled up again.

ishanrai05:
I can't see why jenkins tests are failing. The site is unreachable.

egede:
[fail.txt](https://github.com/ganga-devs/ganga/files/3791402/fail.txt)
I see a large number of failures in the automatic testing. Please see the file attached here. 

You should be able to run the testing yourself using the Dockerfile https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/test/Dockerfile 

Just build the image and run it. It should be fairly self explaining what it does.

mesmith75:
What is the plan for this?

PR #1481, opened 2019-06-13
Tests fix

mesmith75:
The latest LbScripts uses gcc8 which is throwing a fit when compiling psutil.

Although that is a worry as we shouldn't be relying on compilers...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,7 @@ setup(name='ganga',
           'absl-py>=0.1.2',
           'google-api-python-client>=1.1',
       ],
-      extra_require={'profiler' : ['memory_profiler']}
+      extra_require={'profiler' : ['memory_profiler']},
       classifiers=[
           'License :: OSI Approved :: GNU General Public License v2 (GPLv2)',
           'Programming Language :: Python :: 2.7',
--- a/setup.py
+++ b/setup.py
@@ -90,6 +90,7 @@ setup(name='ganga',
           'absl-py>=0.1.2',
           'google-api-python-client>=1.1',
       ],
+      extra_require={'profiler' : ['memory_profiler']}
       classifiers=[
           'License :: OSI Approved :: GNU General Public License v2 (GPLv2)',
           'Programming Language :: Python :: 2.7',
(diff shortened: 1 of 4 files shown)


egede (commented):
The dependency is created by the `memory_profiler` dependency added by @ishanrai05 . As we do not want this to be run by all users, we can maybe think about implementing this in a different way (like the optional dependency we have for LHCb now).

The actual problem is that the wrong gcc is used for building. If I keep the gcc8 configuration and do 
`source /cvmfs/lhcb.cern.ch/lib/lcg/releases/LCG_95/gcc/8.2.0/x86_64-centos7/setup.sh`
it works for me. Obviously that is not a great solution either.

egede:
Maybe to solve the problem short-term and get the testing running again, just remove memory_profile from the dependencies. That can then be fixed by solving #1483.

mesmith75:
I think the importing of the modules needs to be moved as well.

egede:
@ishanrai05 Can you move the import of `memory_profile` such that `ganga` can run without `memory_profile` installed if it is not enabled in the `.gangarc` file. The best will be to do it on this PR such that it confirms that the testing is working again.

ishanrai05:
Ref : #1483

setup.py
```diff
@@ -90,6 +90,9 @@ def run(self):
           'absl-py>=0.1.2',
           'google-api-python-client>=1.1',
       ],
+      extra_require=[
+          'memory_profiler',
+      ],
```
mesmith75: This needs to be a dict with the name of the option and then a list of the requirements.
i.e.
```python
extra_require={'profiler' : ['memory_profiler']}
```

mesmith75:
You also need to move the places the module is imported until after you have checked if you are running the profiling.

setup.py
```diff
@@ -90,6 +90,7 @@ def run(self):
           'absl-py>=0.1.2',
           'google-api-python-client>=1.1',
       ],
+      extra_require={'profiler' : ['memory_profiler']},
```
alexanderrichards, line 93: Just question to @mesmith75 didn't we already have an extra_require for LHCb? What happened to that?

alexanderrichards (approved):

setup.py
mesmith75: Its in another PR - #1462

alexanderrichards: thanks

PR #1484, opened 2019-06-14
Set python version in ganga

mesmith75:
Helps if there are multiple available versions.

### DIFF
--- a/bin/ganga
+++ b/bin/ganga
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 
 ################################################################################
 # Ganga Project. http://cern.ch/ganga


alexanderrichards (commented):
we will have to change this for Python 3

mesmith75:
Yes. It's just that on slc6 the first python in the PATH is 2.6.

alexanderrichards:
> Yes. It's just that on slc6 the first python in the PATH is 2.6.

Fair enough. Although has this not caused issues before?

alexanderrichards (approved):

egede (approved):
Given the move to python3 very soon, I actually think it is a good idea to be explicit here.

PR #1487, opened 2019-06-27
Some LHCb python3 things

mesmith75:


alexanderrichards:
Just let me know when your ready for it to be merged

ganga/GangaCore/Utility/util.py
```diff
@@ -53,6 +53,7 @@ def unique(s):
 
         # Sort to bring duplicate elements together and weed out the
         # duplcates in on sinle pass
+    t = []
```
alexanderrichards, line 56: How did that not cause an error before ?!

egede: @alexanderrichards You mean why the `del t` did not fail before? I guess the `sorted(s)` has never failed? In any case, why is the `del t` required? As an aside, I find the `try ... else` construct quite contrived.

alexanderrichards: yes indeed, I didn't read the whole file, merely saw the diff that the `t = [ ]` was added. What's more worrying is that for Mark to have added this change implies that the sorted has now started failing and he is now seeing an error. I wonder why sorted is now failing?

We are nudging ever closer... :wink:

alexanderrichards: P.s. I agree about the `del t` and the whole `try... else` construct. This I think is very ancient code.

mesmith75: That is a good question....

I noticed this when testing the GaudiExec submission.

mesmith75: @egede @alexanderrichards 
So its catching the `TypeException` now:
```python
'<' not supported between instances of 'File' and 'File'
```
I am seeing this a lot all over the place for different kind of objects, including the monitoring.

Edit: Add the `__gt__` and `__lt__` methods for `Job` .

alexanderrichards: Ok this is because in python 3 you have to implement the rich comparison operators rather than just CMP. Do you have a list of the object that our complains about then I can implement them

mesmith75: Here is the monitoring issue:
```python
ERROR    Monitoring Error: '<' not supported between instances of 'function' and 'function'
```
This is from `IBackend`: https://github.com/ganga-devs/ganga/blob/d4fa1cce3cf5d4d8d8a3fdc990cfdf1d82b41ef4/ganga/GangaCore/GPIDev/Adapters/IBackend.py#L501

alexanderrichards: This one looks a bit more worrying as it's for a `function` rather than a class.

mesmith75:
I think we should merge this now.
It successfully submits and monitors a standard GaudiExec job.
I haven't quite got the old CMT applications working - they are a bit more involved due to the options parsing but I can do that more slowly in another PR.

alexanderrichards:
Great, Thanks for the help with the LHCb plugin mark. I'll merge it into the python3 branch now.

PR #1488, opened 2019-07-05
More LHCb python3

mesmith75:
For some reason putting strings into options and inputfiles no longer works.

### DIFF
--- a/ganga/GangaCore/Runtime/gangadoc.py
+++ b/ganga/GangaCore/Runtime/gangadoc.py
@@ -185,7 +185,7 @@ def mygetmodule(object):
             return builtin
 
 
-def doc2(thing, title='Python Library Documentation: %s', forceload=0):
+def doc2(thing, title='Python Library Documentation: %s', forceload=0, output=None):
     """Display text documentation, given an object or a path to an object."""
     try:
         object, name = pydoc.resolve(thing, forceload)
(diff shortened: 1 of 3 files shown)


ganga/GangaCore/Runtime/gangadoc.py
```diff
@@ -185,7 +185,7 @@ def mygetmodule(object):
             return builtin
 
 
-def doc2(thing, title='Python Library Documentation: %s', forceload=0):
+def doc2(thing, title='Python Library Documentation: %s', forceload=0, output=None):
```
alexanderrichards, line 188: @mesmith75 Am I missing something? I can't see this argument used in the function anywhere

alexanderrichards (changes requested):

ganga/GangaCore/Runtime/gangadoc.py
mesmith75: It is because the method this is replacing now has the output option. I think we always want to output with the `pager` so I ignore it, but the keyword argument still needs to be there.

alexanderrichards: ok thanks.

alexanderrichards (approved):

alexanderrichards:
@mesmith75 is this ready for me to merge?

mesmith75:
Yes

alexanderrichards:
thanks @mesmith75

PR #1490, opened 2019-07-11
Virtualization ui

inzamam-iqbal:
Virtualization works for both Docker and Singularity in LocalBackend

example usage:
```python
j = Job()
j.application.exe = File("pathToSomeExecutablePythonFile")
j.virtualization = Docker(imageUrl = "python:2.7-slim")
j.submit()
```
In the above case if user doesn't have Docker (or if user doesn't have permission to run docker) it will check if UDocker is available, if not it will download uDocker and will run the application inside the container. You will get the results and errors as usual in the output directory. 

simillarly to use Singularity use
```python
j.virtualization = Singularity(imageUrl = "shub://PathToImage")
```


ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -219,6 +219,7 @@ class Job(GangaObject):
                                      'subjobs': ComponentItem('jobs', defvalue=GangaList(), sequence=1, protected=1, load_default=0, copyable=0, comparable=0, optional=1, proxy_get="_subjobs_proxy", doc='list of subjobs (if splitting)', summary_print='_subjobs_summary_print'),
                                      'master': ComponentItem('jobs', getter="_getMasterJob", transient=1, protected=1, load_default=0, defvalue=None, optional=1, copyable=0, comparable=0, doc='master job', visitable=0),
                                      'postprocessors': ComponentItem('postprocessor', defvalue=MultiPostProcessor(), doc='list of postprocessors to run after job has finished'),
+                                      'virtualization': ComponentItem('virtualization', defvalue=None, load_default=0, optional=1, doc='optional virtualization to be used'),
```
alexanderrichards: Are there any adverse effects of this addition on existing jobs in a repository? If for example I had some jobs before upgrading to this Ganga version will they still load?

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -184,10 +186,16 @@ def preparejob(self, jobconfig, master_input_sandbox):
...
         fileutils = File( inspect.getsourcefile(GangaCore.Utility.files), subdir=PYTHON_DIR )
+        virtualizationutils = File( inspect.getsourcefile(GangaCore.Utility.Virtualization), subdir=PYTHON_DIR )
 
         sharedfiles = jobconfig.getSharedFiles()
 
-        subjob_input_sandbox = job.createPackedInputSandbox(jobconfig.getSandboxFiles() + [ fileutils ] )
+        subjob_input_sandbox = job.createPackedInputSandbox(jobconfig.getSandboxFiles() + [ fileutils, virtualizationutils] )
+        
+        import tempfile
+        workdir = tempfile.mkdtemp(dir=config['location'])
```
alexanderrichards: Why did these two lines move? They are not directly used by your changes right?

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -231,7 +236,11 @@ def preparejob(self, jobconfig, master_input_sandbox):
         script = script.replace('###ENVIRONMENT###', repr(environment))
         script = script.replace('###WORKDIR###', repr(workdir))
         script = script.replace('###INPUT_DIR###', repr(job.getStringInputDir()))
-
+        script = script.replace('###VIRTUALIZATION###', repr(virtualization.__class__.__name__))
+        script = script.replace('###VIRTUALIZATIONIMAGE###', repr(virtualization.imageUrl))
+        script = script.replace('###VIRTUALIZATIONMODE###', repr(virtualization.mode) if hasattr(virtualization, 'mode') else repr(None)  )
+        print("app:")
+        print(repr(appscriptpath))
```
alexanderrichards: Please remove prints from Ganga running code. If these are generally useful then please use the logging system otherwise remove them

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -231,7 +236,11 @@ def preparejob(self, jobconfig, master_input_sandbox):
         script = script.replace('###ENVIRONMENT###', repr(environment))
         script = script.replace('###WORKDIR###', repr(workdir))
         script = script.replace('###INPUT_DIR###', repr(job.getStringInputDir()))
-
+        script = script.replace('###VIRTUALIZATION###', repr(virtualization.__class__.__name__))
+        script = script.replace('###VIRTUALIZATIONIMAGE###', repr(virtualization.imageUrl))
+        script = script.replace('###VIRTUALIZATIONMODE###', repr(virtualization.mode) if hasattr(virtualization, 'mode') else repr(None)  )
```
alexanderrichards: If I understand the code correctly then the default value for the `Job.virtualization` attribute is None. If a user is not using virtualisation and just submitting a normal job then the `Job.virtualization` attribute will have this default value. The above lines will then break with an `AttributeError` right as `None` clearly doesn't have these attributes?

ganga/GangaCore/Lib/Virtualization/Singularity.py
```diff
@@ -0,0 +1,27 @@
...
+logger = GangaCore.Utility.logging.getLogger()
+
+class Singularity(IVirtualization):
+
+    """
+    Handles all the config related to Singularity
+    """
+    _name = 'Singularity'
+    _schema = Schema(Version(1, 0), {
+        'imageUrl': SimpleItem(defvalue="", doc='Link to the container image')
```
alexanderrichards: This attribute seems ubiquitous across virtualisation clases so should probably exist in the `IVirtualization` base class?

ganga/GangaCore/Lib/Virtualization/Docker.py
```diff
@@ -0,0 +1,26 @@
...
+class Docker(IVirtualization):
+
+    """
+    Handles all the config related to Docker
+    """
+
+    _name = 'Docker'
+    _schema = Schema(Version(1, 0), {
+        'imageUrl': SimpleItem(defvalue="", doc='Link to the container image'),
+        'mode': SimpleItem(defvalue="P1", doc='Mode of container execution'
```
alexanderrichards: This mode attribute only exists in the Docker class however `LocalHost` has to have some knowledge about if this attribute is present or not with a conditional at the following line:
https://github.com/ganga-devs/ganga/pull/1490/files#diff-39811eca09c833c4f431c0bddee31ca9R241

This sort of two way knowledge doesn't scale well and makes debugging harder as it requires ever greater conditional checking. I would suggest having an abstract method defined in the `IVirtualization` class and implemented in the concrete derived classes that can add whatever is needed to the script. That way the `LocalHost` backend doesn't care what the virtualisation is, it merely requests any script additions.

alexanderrichards (changes requested):

ganga/GangaCore/Lib/Localhost/Localhost.py
inzamam-iqbal: fixed

ganga/GangaCore/Lib/Localhost/Localhost.py
inzamam-iqbal: I removed them

ganga/GangaCore/Lib/Localhost/Localhost.py
inzamam-iqbal: Fixed

ganga/GangaCore/Lib/Virtualization/Singularity.py
inzamam-iqbal: moved to IVirtualization class

ganga/GangaCore/Lib/Virtualization/Docker.py
inzamam-iqbal: created getters in IVirtualization class and I overide them in Docker and SIngularity class

inzamam-iqbal:
There seems to be some issue with unit test I will fix them update the PR.

inzamam-iqbal:
fixed bug in unit test

ganga/GangaCore/Lib/Localhost/Localhost.py
alexanderrichards: I don't think I was perhaps clear. The getters in the `IVirtualization` class haven't changed the fact that `LocalHost.py` either has to have knowledge of the virtualisation attributes type or have a certain level of if/elsery to catch all cases. What I was suggesting would see these lines here reduced to the following:

```python
if job.virtualization is not None:
    script = job.virtualization.modify_script(script)
```
Where you can call the actual method what you like.

---
To achieve this I was suggesting the base class will have an abstract method like so:
```python
class IVirtualization:
...
def modify_script(self, script):
        return script.replace('###VIRTUALIZATION###', repr(self.__class__.__name__))\
                     .replace('###VIRTUALIZATIONIMAGE###', repr(self.imageUrl))
...
```

And the derived classes like `Docker` for example would override this hence using polymorphism instead of if/else/else/else ... for however many derived classes we end up making:

```python
class Docker(IVirtualization):
...

def modify_script(self, script):
    return super(Docker, self).modify_script(script)\
                              .replace('###VIRTUALIZATIONMODE###', repr(self.mode))
...
```

This way there is no implicit knowledge of the type of the virtualization within `LocalHost.py`, no if /elsery  and everything is self contained, modular and scalable.

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
```diff
@@ -0,0 +1,28 @@
...
+        'imageUrl': SimpleItem(defvalue="", doc='Link to the container image')
+    })
+    _category = 'virtualization'
+    _name = 'IVirtualization'
+    _hidden = 1
+
+    def getImageUrl(self):
+        return ""
+
+    def getMode(self):
```
alexanderrichards: I think that these should be changed as described in https://github.com/ganga-devs/ganga/pull/1490#discussion_r303339614.

In any case these getters simply return public attributes so are redundant.

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -231,6 +236,9 @@ def preparejob(self, jobconfig, master_input_sandbox):
         script = script.replace('###ENVIRONMENT###', repr(environment))
         script = script.replace('###WORKDIR###', repr(workdir))
         script = script.replace('###INPUT_DIR###', repr(job.getStringInputDir()))
+        script = script.replace('###VIRTUALIZATION###', repr(virtualization.__class__.__name__))
+        script = script.replace('###VIRTUALIZATIONIMAGE###', repr(virtualization.getImageUrl()) if virtualization else repr(None))
+        script = script.replace('###VIRTUALIZATIONMODE###', repr(virtualization.getMode()) if virtualization else repr(None))
```
alexanderrichards: see https://github.com/ganga-devs/ganga/pull/1490#discussion_r303339614

ganga/GangaCore/Lib/Virtualization/Singularity.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, imageUrl):
+        super(Singularity, self).__init__()
+        
+        if isinstance(imageUrl, str):
+            self.imageUrl = imageUrl
+        else:
+            logger.error("Unkown type: %s . is not a valid format for image Url" % type(imageUrl))
+
+    def getImageUrl(self):
+        return self.imageUrl
```
alexanderrichards: add new line at end of file please

alexanderrichards (changes requested):

inzamam-iqbal:
@alexanderrichards please review it now

ganga/GangaCore/Lib/Virtualization/Singularity.py
inzamam-iqbal: done

ganga/GangaCore/Lib/Localhost/Localhost.py
inzamam-iqbal: done

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
inzamam-iqbal: removed them

ganga/GangaCore/Lib/Localhost/Localhost.py
inzamam-iqbal: done

mesmith75:
A few more docstrings might be nice.

ganga/GangaCore/Lib/Virtualization/Docker.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, imageUrl, mode):
+        super(Docker, self).__init__(imageUrl)
+        self.imageUrl = imageUrl
+        self.mode = mode
+
+    def modify_script(self, script):
+        script = super(Docker, self).modify_script(script)
+        script = script.replace('###VIRTUALIZATION###', repr("Docker"))
+        script = script.replace('###VIRTUALIZATIONMODE###', repr(self.mode))
+        return script
```
alexanderrichards: new line at end of file

ganga/GangaCore/Lib/Virtualization/Singularity.py
```diff
@@ -0,0 +1,23 @@
...
+    _schema = IVirtualization._schema.inherit_copy()
+
+    def __init__(self, imageUrl):
+        super(Singularity, self).__init__(imageUrl)
+
+    def modify_script(self, script):
+        script = super(Singularity, self).modify_script(script)
+        script = script.replace('###VIRTUALIZATION###', repr("Singularity"))
+        script = script.replace('###VIRTUALIZATION###', repr(None))
+        return script
```
alexanderrichards: new line at end of file

alexanderrichards (changes requested):
Looks good now, just minor new line and end of file to fix. P.s. do you have any idea the answer to https://github.com/ganga-devs/ganga/pull/1490#discussion_r302611496

ganga/GangaCore/Lib/Virtualization/Singularity.py
inzamam-iqbal: done

ganga/GangaCore/Lib/Virtualization/Docker.py
inzamam-iqbal: done

ganga/GangaCore/GPIDev/Lib/Job/Job.py
inzamam-iqbal: yes, I checked it by first installing ganga from develop branch then I created few jobs. then I installed ganga from this branch and the old jobs are still compatible.

inzamam-iqbal:
> A few more docstrings might be nice.

inzamam-iqbal:
@mesmith75 added more docstrings

ganga/GangaCore/GPIDev/Lib/Job/Job.py
alexanderrichards: ok excellent, thanks.

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -85,6 +88,27 @@ errorfile=open('stderr','w')
...
+    if (checkDocker()):
+        print("Using Docker")
+        appscriptpath = ['docker', 'run', '-v', workdir+":"+"/work_dir", virtualization_image] + appscriptpath        
+    else:
+        print("Docker not availabe or user doesn't have the permission to run docker demon")
+        if not (checkUDocker()):
+            installUdocker()
+        print("Using UDocker")
+        os.environ["PROOT_NO_SECCOMP"] = "1"
+        appscriptpath = ['/home/inzamam/udocker', '--quiet', 'run', '--nosysdirs', '--volume', workdir+":"+"/work_dir", virtualization_image] + appscriptpath
```
alexanderrichards: This needs to be pointed at wherever udocker is installed rather than hard coded to your home dir which wont exist if I (or anyone else) run it.

alexanderrichards (changes requested):

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
inzamam-iqbal: oh sorry it's a mistake :)

alexanderrichards (dismissed):

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -184,10 +186,13 @@ def preparejob(self, jobconfig, master_input_sandbox):
         import inspect
 
         fileutils = File( inspect.getsourcefile(GangaCore.Utility.files), subdir=PYTHON_DIR )
+        virtualizationutils = File( inspect.getsourcefile(GangaCore.Utility.Virtualization), subdir=PYTHON_DIR )
 
         sharedfiles = jobconfig.getSharedFiles()
 
-        subjob_input_sandbox = job.createPackedInputSandbox(jobconfig.getSandboxFiles() + [ fileutils ] )
+        subjob_input_sandbox = job.createPackedInputSandbox(jobconfig.getSandboxFiles() + [ fileutils, virtualizationutils] )
```
alexanderrichards: Basically all ok now I think. However we could consider whether we want to be packaging up the virtualisation utility tools with every job or only if a virtualisation is requested. i.e. maybe this line could go inside the 
```python
if virtualization:
    here
```

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -85,6 +88,27 @@ errorfile=open('stderr','w')
 sys.stdout=open('./__syslog__','w')
 sys.stderr=sys.stdout
 
+from Virtualization import checkDocker, checkUDocker, checkSingularity, installUdocker
```
mesmith75: I think we need to be careful here. Is it such a good idea to be importing things from ganga? This will cause issues with different python versions.

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -85,6 +88,27 @@ errorfile=open('stderr','w')
...
+
+if (virtualization == "Docker"):
+    appscriptpath[0] = "/work_dir/"+appscriptpath[0]
+    if (checkDocker()):
+        print("Using Docker")
```
mesmith75, line 97: Same kind of thought with the printing. If the default python version on your system is 2 then this will fail.

mesmith75 (changes requested):

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
inzamam-iqbal: I'm sending the virtualization.py module to the backend. so it is not calling a gangaclass it's calling a class within the same folder

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
inzamam-iqbal: Sorry I haven't used any versions before 2.7. So can you please explain why it won't work?

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
alexanderrichards: I think @mesmith75 point was about versions. e.g. if we switch the ganga code to Python 3 will this still work

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
alexanderrichards: I think this one is ok @mesmith75 Python 2.7+ supports the print function without the use of `__future__`. I think we already decided we don't support versions <2.7

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
inzamam-iqbal: Virtualization module will only be used in backend. So I think we don't need to upgrade the virtualization module to python 3. We can maintain it as python 2.7.  But it will be bit odd to have only one class in 2.7 within the ganga code base.

ganga/GangaCore/Lib/Localhost/Localhost.py
inzamam-iqbal: Done had to make minor other changes to make this possible.

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
egede: Let us discuss this issue at the meeting today. While we have agreed that the main part of Ganga very soon will transfer to python3, then the backend template scripts might need to be kept such that they are both python2 and python3 compatible.

alexanderrichards: I had a brief look at your `Virtualization` utility class and it looks python 2/3 compatibly already so it may well be that this is a non-issue. Could you just try running the 2to3 tool over this module to check.

```bash
pip install 2to3
2to3 Virtualization.py
```

inzamam-iqbal:
@alexanderrichards yes the Virtualization.py is python3 compatiable. As you suggested I ran 2to3 and checked it

alexanderrichards (dismissed):

ganga/GangaCore/GPIDev/Lib/Job/Job.py
egede: Remove single whitespace to align with lines above

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -85,6 +88,28 @@ errorfile=open('stderr','w')
...
+        if not (checkUDocker()):
+            installUdocker()
+        print("Using UDocker")
+        fullenvironment["PROOT_NO_SECCOMP"]="1"
+        appscriptpath = [os.path.expanduser("~") + '/udocker', '--quiet', 'run', '--nosysdirs', '--volume', workdir+":"+"/work_dir", virtualization_image] + appscriptpath
+elif (virtualization == "Singularity"):
+    print("Using singularity")
+    appscriptpath[0] = "/work_dir/"+appscriptpath[0]
+    appscriptpath = ['singularity', '-q', 'exec', '--bind', workdir+":"+"/work_dir", "--no-home" ,virtualization_image] + appscriptpath   
+else:
```
egede: I think we should just print nothing if virtualization is not requested. We do not want to pollute stdout for users that never care.

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Contains the interface for all virtualization classes, all virtualization classes should inherit from this object.
+    """
+
+    def __init__(self, imageUrl):
+        super(IVirtualization, self).__init__()
+        self.imageUrl = imageUrl
+
+    _schema = Schema(Version(1, 0), {
+        'imageUrl': SimpleItem(defvalue="", doc='Link to the container image')
```
egede: See general comment. Maybe we should at least just change name to "image", rather than imageUrl. Then it will be easier later to extend the list of accepted types without having to migrate anything for users.

egede (dismissed):
This looks pretty good to me. I am a little bit worried about if we limit ourselves by having the image as a URL in the Ivirtualization interface. At the moment where we only discuss docker images, this is fine. But if we later extend to use singularity files, then they could be supplied as a GangaFile object instead. Simply changing the name of the schema object will at least avoid having a misleading name in the future.

egede (dismissed):
Looks good now. There are some failing tests that I have problems with getting hold of the output from (a firewall issue at Imperial). I have asked @alexanderrichards to have a look.

egede:
test please

PR #1491, opened 2019-07-17
More python3

mesmith75:
More bits and pieces.


mesmith75:
Can we merge this?

ganga/GangaCore/Runtime/bootstrap.py
```diff
@@ -722,7 +722,7 @@ def _versionsort(s, p=re.compile(r'^(\d+)-(\d+)-*(\d*)')):
                     return int(m.group(1)), int(m.group(2)), int(m.group(3))
             if s == 'SVN':
                 return 'SVN'
-            return None
+            return 0, 0, 0
```
mesmith75, line 725: In case you are wondering - you can sort a list of `None` now so something has to be returned.

PR #1492, opened 2019-07-19
More Python3

mesmith75:
More bits and pieces


alexanderrichards:
thanks @mesmith75 should I merge it in now?

PR #1493, opened 2019-07-22
More python3

mesmith75:


ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -881,15 +881,19 @@ def createPackedInputSandbox(self, files, master=False):
...
         name = '_input_sandbox_' + self.getFQID('_') + '%s.tgz'
-
         if master:
             if self.master is not None:
                 name = '_input_sandbox_' + self.master.getFQID('_') + '%s.tgz'
             name = name % "_master"
         else:
             name = name % ""
 
-        files = [f for f in files if hasattr(f, 'name') and not f.name.startswith('.nfs')]
```
mesmith75, line 892: I'm not sure why this line had to be un-condensed but files was empty if now when making the master sandbox. I think it has something to do with `files` being a `map`.

mesmith75:
Now the XML summary merging works - the LHCb scripts are all python 2 so I had to copy them into ganga and convert. I couldn't think of a better solution.

mesmith75:
This can be merged now

alexanderrichards (approved):

alexanderrichards:
thanks @mesmith75

PR #1494, opened 2019-08-03
Virtualization ui

inzamam-iqbal:
Virtualization works for both Docker and Singularity in LocalBackend

example usage:
```python
j = Job()
j.application.exe = File("pathToSomeExecutablePythonFile")
j.virtualization = Docker(imageUrl = "python:2.7-slim")
j.submit()
```
In the above case if user doesn't have Docker (or if user doesn't have permission to run docker) it will check if UDocker is available, if not it will download uDocker and will run the application inside the container. You will get the results and errors as usual in the output directory. 

simillarly to use Singularity use
```python
j.virtualization = Singularity(image = "shub://PathToImage")
```


egede:
So for Singularity, I would like that the following should work as well
```
j.virtualization = Singularity(image = LocalFile('/some/path/mycontainer.sif'))
```
The image object should be allowed to be any type of `GangaFile` object. To get this object into the working directory of the job, it should be possible to link into what is already done for `inputfiles` of a Ganga job.

inzamam-iqbal:
@egede I have updated the PR to support Singularity images as file just as you mentioned above

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Contains the interface for all virtualization classes, all virtualization classes should inherit from this object.
+    """
+
+    def __init__(self, image):
+        super(IVirtualization, self).__init__()
+        self.image = image
+
+    _schema = Schema(Version(1, 0), {
+        'image': SimpleItem(defvalue="", typelist=[str, File], doc='Link to the container image')
```
egede: I was unclear in my specification here. The object that we want to accept is a `GangaFile` object. See https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/GPIDev/Adapters/IGangaFile.py. This will allow us to attach files from the Local file system but also a file that sits on a Google Drive as an example.

alexanderrichards: Was this addressed yet? I don't see the changes on this branch in Github. I think that other than this the branch looks ready to merge.

inzamam-iqbal: @alexanderrichards sorry, Still not updated. I'm looking into it. But I'm having trouble implementing it.

mesmith75:
Is this for release?

alexanderrichards:
Think so. but will need updating from develop

egede:
While testing, I came across some issues here.

- [x] The virtualization property of the job object should be copyable.
- [x] By default udocker is installed in `~/udocker`. On remote nodes this is unlikely to work.
- [x] Error message if udocker installation fails is unhelpful
- [x] The worker node script is not dealing correctly with the situation where `j.application.exe="cat"` or similar (so when it is a command and not a script.

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -85,6 +88,26 @@ errorfile=open('stderr','w')
...
+    appscriptpath[0] = "/work_dir/"+appscriptpath[0]
+    if (checkDocker()):
+        print("Using Docker")
+        appscriptpath = ['docker', 'run', '-v', workdir+":"+"/work_dir", virtualization_image] + appscriptpath        
+    else:
+        print("Docker not availabe or user doesn't have the permission to run docker demon")
+        if not (checkUDocker()):
+            installUdocker()
+        print("Using UDocker")
+        fullenvironment["PROOT_NO_SECCOMP"]="1"
```
egede: The `fullenvironment` is just a dictionary. This will not change anything for the `udocker` script.

ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
```diff
@@ -85,6 +88,26 @@ errorfile=open('stderr','w')
...
+        appscriptpath = [os.path.expanduser("~") + '/udocker', '--quiet', 'run', '--nosysdirs', '--volume', workdir+":"+"/work_dir", virtualization_image] + appscriptpath
+elif (virtualization == "Singularity"):
+    print("Using singularity")
+    appscriptpath[0] = "/work_dir/"+appscriptpath[0]
+    appscriptpath = ['singularity', '-q', 'exec', '--bind', workdir+":"+"/work_dir", "--no-home" ,virtualization_image] + appscriptpath   
+
```
egede: After the script has been running inside the container, it will run again outside the container. Problem with logic here.

egede:
The implementation here is very specific for running on the `Local` backend with the `Executable` application. I think we need to somehow get the virtualization part into a wrapper instead which can then wrap whatever worker node script that we have.

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -183,11 +186,21 @@ def preparejob(self, jobconfig, master_input_sandbox):
...
+        virtualization = job.virtualization
 
+        utilFiles= []
+        fileutils = File( inspect.getsourcefile(GangaCore.Utility.files), subdir=PYTHON_DIR )
+        utilFiles.append(fileutils)
+        if virtualization:
+            virtualizationutils = File( inspect.getsourcefile(GangaCore.Utility.Virtualization), subdir=PYTHON_DIR )
+            utilFiles.append(virtualizationutils)
+            if type(virtualization.image) is File:
+                utilFiles.append(virtualization.image)
```
egede: We should not allow an image to be a `File` object. It should always be a `GangaFile` object.

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -183,11 +186,21 @@ def preparejob(self, jobconfig, master_input_sandbox):
...
+            if type(virtualization.image) is LocalFile:
+                utilFiles.append(File(virtualization.image.location()[0]))
```
egede: This will only work for `LocalFile` and not for any of the other `GangaFile` style objects. We should use the general machinery as done for `j.inputfiles`.

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
egede: I implemented this today.

ganga/GangaCore/Lib/Localhost/Localhost.py
egede: At the moment left it as the user responsibility to add file to the inputfiles.

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Contains the interface for all virtualization classes, all virtualization classes should inherit from this object.
+    """
+
+    def __init__(self, image):
+        super(IVirtualization, self).__init__()
+        self.image = image
+
+    _schema = Schema(Version(1, 0), {
+        'image': SimpleItem(defvalue="", typelist=[str], doc='Link to the container image')
```
alexanderrichards, line 20: Why did we remove the `GangaFile` from the typelist?

ganga/GangaCore/Lib/Localhost/Localhost.py
```diff
@@ -5,6 +5,7 @@
 import GangaCore.Utility.util
 
 from GangaCore.GPIDev.Lib.File import FileBuffer
+from GangaCore.GPIDev.Lib.File.LocalFile import LocalFile
```
alexanderrichards: Unused import

ganga/GangaCore/Lib/Virtualization/Singularity.py
```diff
@@ -0,0 +1,49 @@
...
+    j.inputfiles = j.inputfiles + [imagefile]
+    """
+    _name = 'Singularity'
+    _schema = IVirtualization._schema.inherit_copy()
+    _schema.datadict['image'] = SimpleItem(defvalue="",
+                                           typelist=[str,'GangaCore.GPIDev.Adapters.IGangaFile.IGangaFile'],
+                                           doc='Link to the container image')
+
+    def __init__(self, image):
+        super().__init__(image)
```
alexanderrichards: superfluous default constructor

ganga/GangaCore/Lib/Virtualization/__init__.py
```diff
@@ -0,0 +1,3 @@
+from __future__ import absolute_import
```
alexanderrichards: Now that we are fully Python3 I don't think we need `__future__`  imports

ganga/GangaCore/Utility/Virtualization.py
```diff
@@ -0,0 +1,64 @@
...
+            return True
+    return False
+
+
+def installUdocker():
+    """Download and install UDocker
+
+        Return value: True (If Success) or False"""
+
+    fname = os.path.join(os.path.expanduser("~"),"udocker")
```
alexanderrichards: As @egede points out, this may not work on workernode. I assume we change it to
```suggestion
    fname = os.path.join(os.getcwd(), "udocker")
```

alexanderrichards (changes requested):

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
alexanderrichards: I understood this when looking at the Singularity vs Docker concrete implementations. We can't give a file to Docker but we can have a standalone singularity image.

ganga/GangaCore/Utility/Virtualization.py
egede: I think we might just need to let the installation location be dependent on the backend. For the Local backend it makes sense to install to a shared location to facilitate re-use. It should probably be a configurable though. An even bigger deal in fact is the location of the images that are downloaded. By default they go into `~/.udocker`

alexanderrichards: ah sorry your right I though I was only looking at Dirac (non local) code.

egede:
Tested this with a singularity image stored as a `DiracFile` which worked on the first go. This opens up for a reasonable implementation on the `Dirac` backend as well.

ganga/GangaCore/Lib/Localhost/Localhost.py
egede: Removed import

ganga/GangaCore/Lib/Virtualization/Singularity.py
egede: Removed

ganga/GangaCore/Utility/Virtualization.py
egede: Locations of both the script and the images is now configurable.

egede:
@alexanderrichards I believe that all issues are now addressed. Please have a look for another review.

alexanderrichards (approved):

PR #1498, opened 2019-08-28
Travis tests

mesmith75:


egede:
Interesting to see the number of small issues that the Travis testing is turning up.

alexanderrichards:
thats fixed the easier two... just the execute ones to go

mesmith75:
I have fixed the datetime printing. Apparently it is a bit of an issue: https://stackoverflow.com/questions/24805105/unpickling-python2-datetime-under-python3 

I am not sure if this breaks other things. I guess testing will tell. Perhaps we just want to scrap the tests with datetime as it uncovers a more general python issue?

mesmith75:
Apparently pickling/unpickling across versions is problematic - particularly with regards datetime which is causing a failing test:
https://bugs.python.org/issue22005
I guess this points to something rather deeper - we shouldn't be using pickle. That is perhaps a change we need to make at a later date due to the complexity.

In the mean-time I'll try and work around it, maybe with time.

ganga/GangaCore/test/Unit/Utility/TestUtilitiesExecutePrint.py
```diff
@@ -40,10 +40,10 @@ def test_execute_cwd():
 def test_execute_output():
     ''' This tests the various levels when an import can occur and that complex objects can be returned via stdout '''
 
-    # Test printout of pickle dump interpreted correctly
-    d = execute('import pickle, datetime\nprint(pickle.dumps(datetime.datetime(2013,12,12)))', shell=False)
-    assert hasattr(d, 'month')
-    assert d.month == 12
+    # Test printout of pickle dump interpreted correctly - we have to use time here due to issues with datetime and pickling between 2 and 3
```
mesmith75: The travis tests only use python3, so the subprocesses run with python 3. This therefore returns a string with a byte representation inside, i.e.:
```python
stdout = "b'\x80\x03ctime\nstruct_time\nq\x00(M\xe3\x07K\nK\x07K\x0bK\x0fK:K\x00M\x18\x01K\x01tq\x01}q\x02(X\x07\x00\x00\x00tm_zoneq\x03X\x03\x00\x00\x00BSTq\x04X\t\x00\x00\x00tm_gmtoffq\x05M\x10\x0eu\x86q\x06Rq\x07.'"
```
This then messes up the `pickle.loads()` part, especially as the encoding of a time or datetime object is not "utf-8".

alexanderrichards: for the datetime I think you can change the encoding to "latin1"

alexanderrichards:
This latest commit should now fix the pickling datetime issue.

mesmith75:
Can we merge this?

alexanderrichards (approved):

PR #1499, opened 2019-09-12
Rename Gaudi Job script to have the same prefix for all users

chrisburr:
DIRAC sets the application status based on the `basename(executable_fn).split('_')[0]`. Currently for Gaudi jobs this is the user's name but this makes filtering in the DIRAC web app inconvenient (https://github.com/DIRACGrid/DIRAC/issues/4267). 

Would it be a problem to start with the application name instead?


mesmith75:
Sounds reasonable. Can we hold this for a week or two? We are aiming for the python3 branch to be merged shortly.

chrisburr:
Yeah I think it can wait 👍

mesmith75 (approved):

mesmith75:
Let's merge this now.

PR #1501, opened 2019-09-26
Update Dockerfile for core testing

egede:


egede:
@mesmith75 @alexanderrichards Can I get you to review and merge this into the python3 branch. While there are still many tests failing in Jenkins, at least it runs inside python3 now.

alexanderrichards (approved):

PR #1506, opened 2019-10-22
Add deprecation warning for old Ganga

mesmith75:
Add a deprecation warning to the last Ganga 7 release.


egede (changes requested):
Rather than leaving this a crap in the code, can you just enter it in without the if statement, make the release and then remove it again as part of the next release.

mesmith75:
That makes more sense

alexanderrichards (approved):

egede (approved):

Issue #1524, opened 2019-11-12
Undefined if wrapper scripts are running in python2 or python3 environment


egede:
For Localhost the wrapper script is started as
```
process = subprocess.Popen(["python", scriptpath, 'subprocess'])
```
which means that it in a default virtualenv installation will be executed as python3. As the wrapper scripts (including the injected code) are not converted to python3, this will fail.

PR #1525, opened 2019-11-12
Make the Local wrapper script execute explicitly in python2

egede:


egede:
Solves #1524. In the longer term a better solution will be required.

alexanderrichards (approved):

PR #1527, opened 2019-11-13
Move testing to Github Actions

alexanderrichards:
Testing GitHub Actions


egede:
Do you understand why you have to make changes to the syntax of the `mock` packages? These tests are running fine in Python 3.6 on the Jenkins tests.

alexanderrichards:
> Do you understand why you have to make changes to the syntax of the `mock` packages? These tests are running fine in Python 3.6 on the Jenkins tests.

Yes it's because we are installing from the requirements file which then drags in the setup.py. Since we have the setup.py we don't really also need requirements. I looked at what gets installed in requirements and it's just for testing. This is better done using a conditional dependency list within setup.py. It includes some things that aren't needed in python 3, chief among them being mock (i.e. the python 2 backport of mock which will also work on python 3) but this is already now part of python under unittest.mock. Therfore when we remove the dependency we must also ensure there is no python 2 compatible mock code left

alexanderrichards:
essentially I'm trying to streamline the process which (lets face it) is quite complicated

alexanderrichards:
> Do you understand why you have to make changes to the syntax of the `mock` packages? These tests are running fine in Python 3.6 on the Jenkins tests.

I should add that most of these import switches were done with the python 3 migration. We either missed these or they were added afterwards

alexanderrichards:
This is now matching the jenkins box for full integration tests on Core. I'll now try adding Dirac and LHCb

.github/workflows/ci_push_testing.yml
```diff
@@ -0,0 +1,98 @@
...
+        echo $ROBOT_KEY > ~/.globus/userkey.pem
+    - name: Install DIRAC UI
+      run: |
+        yum install -y wget
+        mkdir ~/dirac_ui
+        cd ~/dirac_ui
+        wget -np -O dirac-install https://raw.githubusercontent.com/DIRACGrid/DIRAC/integration/Core/scripts/dirac-install.py
+        # add --no-check-certificate to wget if needed
+        chmod u+x dirac-install
+        ./dirac-install -r v6r22p26 -i 27 -g v14r1
```
egede: The code that installs the latest series 6 release seems to have been undone here.

egede (commented):
Great to see the progress here. Remember to add in the GangaGUI tests as well.

egede:
Amazing that the system gives you full control to install things like the DiracUI as well. In principle the LHCb container should work as well - I am a bit worried how much content that we will pull from an uncached /cvmfs though.

.github/workflows/ci_push_testing.yml
alexanderrichards: yup my fault I forgot that

alexanderrichards:
> Amazing that the system gives you full control to install things like the DiracUI as well. In principle the LHCb container should work as well - I am a bit worried how much content that we will pull from an uncached /cvmfs though.

yes i agree, do we pull a lot from cvmfs for the testing?

alexanderrichards:
hit a stop now as the robot cert I have has expired

egede:
Secrets were updated for Ganga Robot.

alexanderrichards:
> Secrets were updated for Ganga Robot.

thanks

egede:
I played a bit with the LHCb tests. If `/cvmfs` is not mounted from the host, then the container will try to use `parrot` to create a mount point in userspace. From looking in the `LHCb` documentation, I see 
"in-container mount of /cvmfs requires '--security-opt seccomp=unconfined' option to docker run". Lookin at https://cctools.readthedocs.io/en/latest/parrot/#a-note-on-docker gives the same message.

alexanderrichards:
> I played a bit with the LHCb tests. If `/cvmfs` is not mounted from the host, then the container will try to use `parrot` to create a mount point in userspace. From looking in the `LHCb` documentation, I see
> "in-container mount of /cvmfs requires '--security-opt seccomp=unconfined' option to docker run". Lookin at https://cctools.readthedocs.io/en/latest/parrot/#a-note-on-docker gives the same message.

That actually looks like more trouble (also I don't know what options in github actions to use) than just setting up cvmfs in the outer ubuntu container and mounting it within the lhcb continer.

egede:
> That actually looks like more trouble (also I don't know what options in github actions to use) than just setting up cvmfs in the outer ubuntu container and mounting it within the lhcb continer.

If that is possible then that would of course be fine as well. But you do not think that you will get the same problem there?

alexanderrichards:
possibly but actually looking again im not sure if I can do some actions before starting the container. If so then I imagine that setting it up on the latest ubuntu should be doable

alexanderrichards:
actually just found this on the docs for github actions
```yaml
jobs:
  my_job:
    container:
      image: node:10.16-jessie
      env:
        NODE_ENV: development
      ports:
        - 80
      volumes:
        - my_docker_volume:/volume_mount
      options: --cpus 1
```

looks like we could us the options flag?

alexanderrichards:
also looks like I missed in the existing jenkins file we run the `docker run` command with the `--privileged`  option. Maybe this is sufficient?

alexanderrichards:
@egede Right, I have successfully installed CVMFS in a plain CentOS7 container using fuse. You can see from the latest "Fuse Test" log that I can list the `/cvmfs/lhcb.cern.ch` dir. Can we use this instead of the lhcb image? Do you know what extra it provides?

egede:
@alexanderrichards  Looks like you have a success !!

egede:
If you just do a 
```bash
source /cvmfs/lhcb.cern.ch/lib/LbEnv
```
I think the LHCb environment should be there.

alexanderrichards:
> If you just do a
> 
> ```shell
> source /cvmfs/lhcb.cern.ch/lib/LbEnv
> ```
> 
> I think the LHCb environment should be there.

I'll add that to the `Fuse Test` test to see if it works

alexanderrichards:
That looks good! Now I'll apply that to the LHCb tests

alexanderrichards:
🎉 SUCCESS!!! 🎉  (Now for the LHCb integration tests...)

```
Update: LHCb unit tests run BUT it looks like pytest is skipping all the tests.
I'll have to check that is the same jenkins box

...

ok seems that most of the LHCb tests on jenkins also actually get skipped,
including ALL unittests. I think only 5 actually run. Is this known about.
```

mesmith75:
Are they marked external?

alexanderrichards:
> Are they marked external?

I don't know, certainly I don't think so in the logs. Where is said marking occurring? Would you expect only 5 tests to be running?

---
In any case this is a separate issue since we are able to reproduce the jenkins results here in github. Whether these results are right or not is another matter 😉

mesmith75:
Looking at the files most of the LHCb tests are marked external. Whether that is a good idea I don't know as we don't have the nightly testing any more.

alexanderrichards:
> Looking at the files most of the LHCb tests are marked external. Whether that is a good idea I don't know as we don't have the nightly testing any more.

I have no understanding of what marking them external means? Only run in nightlies?

mesmith75:
It used to be that. You run them with `pytest --run-externals`. Without that they just get skipped.

.github/workflows/ci_push_testing.yml
```diff
@@ -0,0 +1,223 @@
...
+        DIRAC_DEPRECATED_FAIL: True
+      run: python3 -m pytest --cov-report term-missing --cov ganga/GangaDirac ganga/GangaDirac/test/GPI
+
+  lhcb-unit:
+    name: GangaLHCb Unit
+    needs: lint
+    runs-on: ubuntu-latest
+    container: 
+      image: centos:7
+      options: --privileged
```
egede, line 169: Is this still required - or was it only added during the testing and then not removed again when different approach was taken?

.github/workflows/ci_push_testing.yml
```diff
@@ -0,0 +1,223 @@
...
+        yum install -y cvmfs cvmfs-config-default
+        cvmfs_config setup
+        echo "CVMFS_REPOSITORIES=lhcb.cern.ch" > /etc/cvmfs/default.local
+        echo "CVMFS_HTTP_PROXY=DIRECT" >> /etc/cvmfs/default.local
+        mkdir -p /cvmfs/lhcb.cern.ch
+        mount -t cvmfs lhcb.cern.ch /cvmfs/lhcb.cern.ch
+    - name: Test CVM-FS
+      run: ls /cvmfs/lhcb.cern.ch
+    - name: Load LHCb environment
+      run: source /cvmfs/lhcb.cern.ch/lib/LbEnv
```
egede: Guess that this can just be deleted now.

.github/workflows/ci_push_testing.yml
alexanderrichards: Yes it is still required, without it the fuse mount of cvmfs doesn't work. I note also we use it in the current Jenkins tests too.i assume for a similar reason

.github/workflows/ci_push_testing.yml
alexanderrichards: Yes indeed. I will do that today as well as adding the LHCb integration tests.

.github/workflows/ci_push_testing.yml
```diff
@@ -0,0 +1,239 @@
...
+      run: |
+        . ~/venv/bin/activate
+        python3 -m pip install --upgrade pip setuptools wheel
+        python3 -m pip install -e .[LHCb]
+        python3 -m pip install coverage pytest pytest-cov pytest-pylint pytest-mock
+    - name: Test with pytest
+      env:
+        GANGA_CONFIG_PATH: GangaLHCb/LHCb.ini
+      run: |
+        source /cvmfs/lhcb.cern.ch/lib/LbEnv && source LbLogin.sh
```
egede: I think you can get rid of the `LbLogin.sh` - even if it should not do any damage.
```bash
source /cvmfs/lhcb.cern.ch/lib/LbEnv
```

.github/workflows/ci_push_testing.yml
```diff
@@ -0,0 +1,239 @@
...
+        . ~/venv/bin/activate
+        lhcb-proxy-init
```
egede: I have no idea why the `lhcb-proxy-init` fails. 
```bash
(ganga) [centos7] ~/programming/ganga % which lhcb-proxy-init
/cvmfs/lhcb.cern.ch/lib/var/lib/LbEnv/796/stable/x86_64-centos7/bin/lhcb-proxy-init
```
can you take a look at the `PATH`?

.github/workflows/ci_push_testing.yml
egede: And why do I not see the stdout from the LbEnv command
```bash
********************************************************************************
*                               ---- LbEnv ----                                *
********************************************************************************
 --- CMAKE_PREFIX_PATH is set to:
    /cvmfs/lhcb.cern.ch/lib/lhcb
    /cvmfs/lhcb.cern.ch/lib/lcg/releases
    /cvmfs/lhcb.cern.ch/lib/lcg/app/releases
    /cvmfs/lhcb.cern.ch/lib/lcg/external
    /cvmfs/lhcb.cern.ch/lib/contrib
    /cvmfs/lhcb.cern.ch/lib/var/lib/LbEnv/796/stable/x86_64-centos7/lib/python2.7/site-packages/LbDevTools/data/cmake
--------------------------------------------------------------------------------
```

alexanderrichards: indeed, it looks like it's not sourcing it properly.

.github/workflows/ci_push_testing.yml
alexanderrichards: will do

alexanderrichards:
Ok LHCb integration tests are running but looks like the GPI isn't being populated
```python
>       from GangaCore.GPI import GaudiExec, Job, LocalFile, DiracFile
E       ImportError: cannot import name 'GaudiExec' from 'GangaCore.GPI'
```

@mark have you seen this before with LHCb testing?

---
Also, one thing I haven't yet replicated from the current [testrunner](https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaLHCb/test/testrunner) script is:

```bash
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/cvmfs/sft.cern.ch/lcg/releases/LCG_97python3/Python/3.7.6/x86_64-centos7-gcc9-opt/lib
```

Do you remember why this was necessary I assume for our external dependencies?

mesmith75:
@alexanderrichards I haven't seen that before. It may be related to the GANGA_CONFIG_PATH? Do you need to set the GANGA_CONFIG_FILE as well?

For the LD_LIBRARY_PATH variable I think that can be dropped - it was for setting up the virtualenv on the CVMFS installation.

alexanderrichards:
> It may be related to the GANGA_CONFIG_PATH

I lifted this directly from the current testing so it should be the same.

> Do you need to set the GANGA_CONFIG_FILE as well

I don't think we set this currently do we? anyway, worth a try

.github/workflows/ci_push_testing.yml
alexanderrichards: seems to be sourcing properly but not showing the output

.github/workflows/ci_push_testing.yml
alexanderrichards: `PATH` is correctly updated but seems it doesn't matter. Solution was to give absolute path to `lhcb-proxy-init`. I assume this is to do with the fact it's on cvmfs.

alexanderrichards:
🎉 SUCCESS 🎉 
Looks good now. I'll check to make sure we see the same number of passes/fails in Jenkins but from memory as I've been going along I think they are now in agreement

egede:
This is great work. Not having to maintain our own test box will save us time in the future. I also notice that it is more than twice as fast - I think this is mainly from that it builds the containers much faster.

egede (dismissed):

alexanderrichards:
> This is great work. Not having to maintain our own test box will save us time in the future. I also notice that it is more than twice as fast - I think this is mainly from that it builds the containers much faster.

We can also get rid of the travis tests as well now since these one do it all.

egede:
Should `mock` be removed from `requirements.txt` to avoid problems in the future with not using the right one?

alexanderrichards:
> Should `mock` be removed from `requirements.txt` to avoid problems in the future with not using the right one?

actually I'm not using the requirements at all now. Just using the setup.py. Hence I moved all the relevant dependencies into a [dev] option. We can leave the requirements there for legacy reasons (so as not to break someone else's workflow) but removing mock seems sensible

egede (dismissed):
This is fantastic work. This was the only area where I thought that GitLab was better than GitHub.

alexanderrichards:
Github actions seem more stable than jenkins too. Even though the last change was to the github actions yaml file, jenkins unexpectedly failed even though nothing should have changed for it. I had to log into the jenkins box and rerun the tests after which they passed with no additional changes

---
This is now ready so unless there are and other changes anyone wants I'll merge it on Monday

egede (approved):

PR #1528, opened 2019-11-18
Add platform to executable and generic Dirac

mesmith75:
This needs some input from Dirac before we can merge.


ganga/GangaCore/Lib/Executable/Executable.py
```diff
@@ -55,6 +55,7 @@ class Executable(IPrepareApp):
         'exe': SimpleItem(preparable=1, defvalue='echo', typelist=[str, File], comparable=1, doc='A path (string) or a File object specifying an executable.'),
         'args': SimpleItem(defvalue=["Hello World"], typelist=[str, File, int], sequence=1, strict_sequence=0, doc="List of arguments for the executable. Arguments may be strings, numerics or File objects."),
         'env': SimpleItem(defvalue={}, typelist=[str], doc='Dictionary of environment variables that will be replaced in the running environment.'),
+        'platform': SimpleItem(defvalue='ANY', typelist=[str], doc='Platform the application was built for'),
```
egede: Change the docstring to
- Give an example of the syntax required.
- Say "platform where the job should be executed. 
The last change is to avoid confusion around Virtualization. I might require a centos7 platform to get Singularity to work correctly, but my executable might be built for some Debian virtual image.

egede (dismissed):

egede (dismissed):

mesmith75:
So apparently we cannot do this in DIRAC for reasons that are beyond me. Therefore I shall close it.

mesmith75:
I have added the ExeDiracRTHandler to LHCb so executable jobs use LHCbDirac when submitting, rather than generic Dirac. This allows us to set the platform and is apparently more correct.

alexanderrichards (approved):

PR #1530, opened 2019-11-18
Fix xml processing

mesmith75:
Added the LHCb XML processing code to ganga and converted to python3


egede (approved):
Does this mean that the code should be deleted from some other project?

mesmith75:
No I don't think so. I just took this from LHCb/Kernel which is where the applications would look for it.

alexpearce:
I'm working on the Python 3 port of LHCb in [LHCb!2214](https://gitlab.cern.ch/lhcb/LHCb/merge_requests/2214); it would be great to have a patch for the `XMLSummarySvc` there!

(I've seen problems with it locally, but haven't had the time to patch it yet.)

egede:
@alexpearce Looking at the files changed for the PR should give you what you want!

PR #1531, opened 2019-11-18
Fix files on the batch

mesmith75:
Prevents batch jobs trying to use ganga code when copying output

### DIFF
--- a/ganga/GangaCore/Lib/Batch/BatchScriptTemplate.py.template
+++ b/ganga/GangaCore/Lib/Batch/BatchScriptTemplate.py.template
@@ -124,7 +124,7 @@ except:
     filefilter = None
 
 from files import multi_glob, recursive_copy
-    
+
 createOutputSandbox(outputpatterns,filefilter,sharedoutputpath)
 
 def printError(message):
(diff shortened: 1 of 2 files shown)


egede (changes requested):
We have the capability to ship code along with the job and place it in a subdirectory. Would that be better here rahter than making the template longer.

mesmith75:
So that is what we do currently. The problem is the code we ship is just a copy of https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Utility/files.py so it fails when trying to import the ganga modules.
As we only use two functions I figured this would be the simplest solution

egede:
It seems like the only import is the logger for a single debug statement. Can't we just delete that one? Anyway this is not a big deal either way. Your proposed solution is of course fine as well, but duplication of code should be avoided where possible.

egede (dismissed):

mesmith75:
Yes, that is better.

mesmith75:
Could someone approve this please? It is messing up my analysis jobs at Imperial.

egede (approved):
Sorry, I forgot that I did not approve it first time round.

mesmith75:
thanks!

PR #1535, opened 2019-11-21
Fix splitter with banned sites

mesmith75:


ganga/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
```diff
@@ -257,7 +257,7 @@ def calculateSiteSEMapping(file_replicas, uniqueSE, CE_to_SE_mapping, SE_to_CE_m
         SE_dict[lfn] = sitez
 
     # Remove the banned sites (CE) from the mappings
-    for iSE in CE_to_SE_mapping.keys():
+    for iSE in list(CE_to_SE_mapping.keys()):
```
alexanderrichards, line 260: Why is this necessary? `.keys()` should return an iterable generator.

Could also do the following:
```suggestion
    for iSE in CE_to_SE_mapping:
```

alexanderrichards (changes requested):

ganga/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
mesmith75: But you need a copy of the list if you start to remove items. Other wise you end up with
```python
RuntimeError: dictionary changed size during iteration
```

egede (approved):

ganga/GangaDirac/Lib/Splitters/OfflineGangaDiracSplitter.py
alexanderrichards: Ah sure, I didn't read the following code, just this line.

alexanderrichards (approved):

PR #1539, opened 2019-11-26
Update of Virtualization

egede:
- Dirac and Batch backend support
- Authentication support
- Refactoring of wrapper scripts


alexanderrichards (dismissed):

ganga/GangaCore/GPIDev/Adapters/IVirtualization.py
```diff
@@ -17,7 +18,12 @@ def __init__(self, image):
         self.image = image
 
     _schema = Schema(Version(1, 0), {
-        'image': SimpleItem(defvalue="", typelist=[str], doc='Link to the container image')
+        'image': SimpleItem(defvalue="", typelist=[str], doc='Link to the container image'),
+        'username': SimpleItem(defvalue="", typelist=[str], doc='Deploy token username'),
+        'password': SimpleItem(defvalue="", typelist=[str], doc='Deploy token password'),
```
mesmith75: Is it ok to keep a password like this?

egede: Fair question. It is what is called a deploy token which means that it is only valid for pulling the image and nothing else.

egede:
@alexanderrichards @mesmith75 I have discovered a problem with the vritualization implementation for Local and Batch backends. If I do
```
j = Job()
j.application.exe='/foo/bar/myscript'
j.virtualization = Singularity(image = 'docker://fedora')
```
then the job will fail because `/foo/bar` is not mounted inside the container. Have not discovered this so far as I had done all testing with commands like `echo` and `cat`. Now this problems can be solved in two ways

1. We can mount the directory containing the script
2. We can copy the script into the worker node

Which of the models should we adopt? The first one will expose more of the system inside the container (without the knowledge of the user) while the second one is quite a big change (if `/foo/bar` is a large program this could cause trouble). My suggestion would be to go for option 1. for now, but then maybe consider if we should change behaviour here in the future to make local and grid running more identical.

mesmith75:
I think I prefer option 2. I don't think it is always the case that the local file system is available on batch systems.

mesmith75 (approved):

PR #1543, opened 2019-12-02
Release notes 8.0.3

mesmith75:


egede (dismissed):

mesmith75:
I want to do a bit more testing of the LbEnv things before releasing this.

mesmith75:
So the LbEnv code doesn't work in C shell. The complaints look related to LbScripts/LbEnv so I am not sure if it is from our side or theirs.

I assume most people use bash so perhaps we don't care. If so shall we release anyway?

alexanderrichards (approved):

PR #1544, opened 2019-12-03
Set shell subprocess stdin to NULL

### LINKED ISSUE #1504: Ganga process suspended for LHCb users
In the past few weeks, LHCb users have observed that the Ganga process gets suspended. It can then be restarted with foregrounding the process. Below is an analysis by Chris Burr

```
Hi Ulrik,

I've had a look and it think it's related to leaking file handles to subprocesses which then close the stdin pipe and disconnect the process.

For me it only happens on launch if I have a proxy and seems to happen when the background call to lhcb-proxy-info exits. I've also seen it happen when calling gridProxy.renew():

Ganga In [2]: gridProxy.renew()
Generating proxy...

Your certificate will expire in 26 days. Please renew it!


[1]+  Stopped                 ganga
bash-4.2$ fg
ganga
Enter Certificate password:

I can reproduce independently of lhcb-proxy-info in various ways such as this:

import subprocess, os, signal
for i in range(1000):
     process = subprocess.Popen(['/bin/sh', '-c', 'ls > /dev/null 2>&1'], shell=True)
     os.kill(process.pid, signal.SIGKILL)

This reproducer isn't reliable but if you run enough loops of that it stops eventually. Replacing ls with voms-proxy-info makes it much more reliable but I haven't figured out why.

I can fix launching ganga by modifying the lhcb-proxy-init wrapper to prevent the stdin pipe from being inherited by the child process with this patch:

diff --git a/tmp/cburr/bin/lhcb-proxy-info b/tmp/cburr/bin-fixed/lhcb-proxy-info
index eb579c1..c503c47 100755
--- a/tmp/cburr/bin/lhcb-proxy-info
+++ b/tmp/cburr/bin-fixed/lhcb-proxy-info
@@ -18,7 +18,8 @@ while i < len(sys.argv):
     arguments = arguments + " " +sys.argv[i]
   i = i + 1

-s = subprocess.Popen(arguments, stdin=sys.stdin, shell=True, env=dict(os.environ))
-s.communicate()
+with open(os.devnull, 'w') as devnull:
+    s = subprocess.Popen(arguments, stdin=devnull, shell=True, env=dict(os.environ))
+    s.communicate()

I think the "correct" fix is that subprocesses in background threads should never inherit stdin.

Cheers,

Chris
```

### LINKED ISSUE #1505: Make stdin=devnull for all subprocess calls in background threads.
To avoid trouble with stdin getting closed by a subprocess,, we should have stdin=devnull for any subprocess started from a background thread. See #1504 for some details.

mesmith75:
Fixes #1504 Fixes #1505 .

This just fixes the most obvious uses of Popen where stdin is clearly not required. I guess there are more tricky cases that need tackling as well but I think this already helps. Maybe @egede you can try this branch and see if it works for you? It seems to fix the issue for me.

### DIFF
--- a/ganga/GangaCore/GPIDev/Adapters/IGangaFile.py
+++ b/ganga/GangaCore/GPIDev/Adapters/IGangaFile.py
@@ -175,7 +175,7 @@ class IGangaFile(GangaObject):
         mystderr = ''
 
         try:
-            child = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            child = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.DEVNULL)
             (mystdout, mystderr) = child.communicate()
             exitcode = child.returncode
         finally:
--- a/ganga/GangaCore/Lib/Checkers/CustomChecker.py
+++ b/ganga/GangaCore/Lib/Checkers/CustomChecker.py
@@ -7,7 +7,6 @@ from GangaCore.GPIDev.Adapters.IPostProcessor import PostProcessException
 from GangaCore.GPIDev.Adapters.IChecker import IChecker
 from GangaCore.GPIDev.Schema import FileItem
 from GangaCore.Utility.logging import getLogger
-import subprocess
 import copy
 import os
 import re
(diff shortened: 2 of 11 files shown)


egede (approved):
I did not manage to get it to fail.

PR #1550, opened 2019-12-05
Update fill_gpi.py

mesmith75:


egede (approved):
When was the import statement changed?

mesmith75:
The previous PR. For some reason I changed the import and didn't think about when it gets used. I am confident this will get built by buildthedocs which I wanted so the virtualisation is properly documented.

PR #1555, opened 2019-12-11
Fix file outputs

mesmith75:


alexanderrichards (approved):

egede (commented):
Why did this cause trouble? It was meant to clean up `### XYZ ###` stuff in the wrapper scripts that were not edited. I fear that we now end up with lines like

```
var = ### XYZ ###
```
which will result in a syntax error.

mesmith75:
This replaced all of these things with `None` so there was no difference between them when the substitution was made. Also in the job script the dict that contained the elements to write the script to transfer the files only had keys of `None`, so ultimately there was only one entry.

I do agree though, these things really need tidying up.

egede:
I do not understand because all the replacements were made in the lines above. If you look in the old version

https://github.com/ganga-devs/ganga/blob/061d92ea63bc406ab5c82f772167040b26ca866c/ganga/GangaCore/Lib/Localhost/Localhost.py#L250

you see that all the replacements were made in the lines 228 to 244.

In the same way for Batch.

https://github.com/ganga-devs/ganga/blob/061d92ea63bc406ab5c82f772167040b26ca866c/ganga/GangaCore/Lib/Batch/Batch.py#L482

The substitution happens in lines 479 to 480 (in a different way!) and then the cleanup in 482. I do not see the problem.

mesmith75:
Here is an excerpt from a `__jobscript__` for a local job.
The script for moving the output files is as follows:
```python
import os
## NB parseCommandLine first then import Dirac!!
import datetime
import gzip
import sys
from contextlib import closing
parseCommandLine()
dirac=Dirac()
errmsg=''
file_name='None'
file_label=file_name

compressed = None
if compressed:
    file_name += '.gz'
    f_in = open(file_label, 'rb')
    f_out = gzip.open(file_name, 'wb')
    f_out.writelines(f_in)
    f_out.close()
    f_in.close()

lfn=os.path.join('None', file_name)
wildcard='None'
storage_elements=None

with closing(open('None','ab')) as locationsfile:
    for se in storage_elements:
        sys.stdout.write('\\nUploading file: \"%s\" as \"%s\" at \"%s\"\\n' % (file_name, lfn, se))
        try:
            result = dirac.addFile(lfn, file_name, se)
        except Exception as x:
            sys.stdout.write('Exception running dirac.addFile command: %s' % str(x))
            break
        if result.get('OK',False) and lfn in result.get('Value',{'Successful':{}})['Successful']:
            guid = dirac.getMetadata(lfn)['Value']['Successful'][lfn]['GUID']
            locationsfile.write("DiracFile:::%s&&%s->%s:::['%s']:::%s\\n" % (wildcard, file_label, lfn, se, guid))
            locationsfile.flush()
            locationsfile.close()
            break
        errmsg+="(%s, %s), " % (se, result)
    else:
        locationsfile.write("DiracFile:::%s&&%s->None:::File '%s' could not be uploaded to any SE (%s):::NotAvailable\\n" % (wildcard, file_label, file_name, errmsg))
        sys.stdout.write('Could not upload file %s' % file_name)
```
Everything in that which is `None` would normally be marked `###LFNNAME###` etc. This in itselft causes issues.

The job script then performs a substitution from a dict containing all the necessary attributes (file name, postprocess file, storage elements etc). However now that dict only has key values `None`:
```python
    replace_dict = {
            'None' : file_name,
            'None' : lfn_base,
            'None' : str(compress),
            'None' : wildcard,
            'None' : str(storage_elements),
            'None' : postprocesslocations.name
            }
```
Which of course results in a dict `{'None':postprocesslocations.name}`
This does not work.

mesmith75:
Here is a jobscript from a failing job:
[__jobscript__.txt](https://github.com/ganga-devs/ganga/files/3953012/__jobscript__.txt)

mesmith75:
Here is a jobscript from a working job:
[__jobscript__.txt](https://github.com/ganga-devs/ganga/files/3953018/__jobscript__.txt)

egede:
I see, this is the stuff from the code that gets injected for the files. All understood.

PR #1556, opened 2019-12-17
Another LHCb exec fix

mesmith75:


mesmith75:
This changes the submission of `Executable` applications from the LHCb plugin to Dirac. It now used the LHCbDirac `LHCbJob` instead of the vanilla Dirac `Job`. For this to work the platform must be specified. Therefore I have added a config for the default application platform the LHCb plugin, which is now also used by GaudiExec.

ganga/GangaLHCb/__init__.py
```diff
@@ -54,6 +54,7 @@
                           splitInputDataBySize and splitInputData')
     defaultLHCbDirac = 'prod'
     configLHCb.addOption('LHCbDiracVersion', defaultLHCbDirac, 'set LHCbDirac version')
+    configLHCb.addOption('defaultPlatform', 'x86_64-centos7-gcc8-opt', 'The default platform for applications to use')
```
egede: This value is bound to be outdated pretty soon. Is there some way we can set it to be the latest value?

mesmith75: This now takes the CMTCONFIG from the standard login by default.

alexanderrichards (approved):

Issue #1496, opened 2019-08-07
Excessive CPU usage of Ganga when idle


egede:
So I am getting some understanding of this. I managed to get the profiling to work for all threads and I looked at a simple Ganga session that does absolutely nothing. That takes up 20% of CPU time, by just keeping the threads alive. Look at tables below.

First just CPU time used.

```
Tue Aug  6 09:04:44 2019    pstats

         3025011165 function calls (3045501047 primitive calls) in 18608.766 seconds

   Ordered by: internal time
   List reduced from 2552 to 20 due to restriction <20>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
111672296/111672382 5302.618    0.000 5302.619    0.000 time:0(sleep)
15955981/15956009 3455.435    0.000 10966.966    0.001 threading.py:309(_Condition.wait)
203432054/203432055 1564.325    0.000 1564.325    0.000 __builtin__:0(<method 'acquire' of 'thread.lock' objects>)
13893115/13893158  623.255    0.000 10083.479    0.001 Queue.py:150(PriorityQueue.get)
178002954  485.361    0.000  485.361    0.000 time:0(time)
101874671  463.914    0.000  463.914    0.000 __builtin__:0(min)
 24576283  446.015    0.000  731.661    0.000 threading.py:147(_RLock.acquire)
 20650850  429.269    0.000 1753.586    0.000 GangaThread.py:10(GangaThread.decorated)
35396110/54437560  394.032    0.000  916.908    0.000 lexer.py:616(Python3TracebackLexer.get_tokens_unprocessed)
 24576282  357.426    0.000  523.322    0.000 threading.py:187(_RLock.release)
366362488  283.677    0.000  283.677    0.000 __builtin__:0(<method 'match' of '_sre.SRE_Pattern' objects>)
 67826741  202.180    0.000  202.180    0.000 threading.py:64(_RLock._note)
  4249734  186.170    0.000 1303.766    0.000 html.py:738(HtmlFormatter._format_lines)
```

and now who is calling the first ones in the table above

```
pstats% callers
   Ordered by: internal time

Function                                                                          was called by...
                                                                                      ncalls  tottime  cumtime
time:0(sleep)                                                                     <- 2109382/2109379  115.992  115.992  Registry.py:254(RegistryFlusher.run)
                                                                                   932748/932747   50.326   50.326  SessionLock.py:114(SessionLockRefresher.run)
[...]

PR #1558, opened 2020-01-01
cpu consumption even when threads are idle

ishanrai05:
This should probably decrease the CPU consumption when threads are idle. The timeout time can be adjusted in .gangarc file.

Ref: #1496

### DIFF
--- a/ganga/GangaCore/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
+++ b/ganga/GangaCore/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
@@ -12,6 +12,9 @@ from GangaCore.GPIDev.Base.Proxy import getName
 
 from collections import namedtuple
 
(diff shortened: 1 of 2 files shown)


ganga/GangaCore/Core/GangaThread/WorkerThreads/WorkerThreadPool.py
```diff
@@ -12,6 +12,12 @@
 
 from collections import namedtuple
 
+try:
+    timeout = getConfig('Queues')['ShutDownTimeout']
```
egede: I do not see any reason for the `try ... except` code. When you have defined the new configuration object, you know that it will exist. So I siggest just to have the single line `timeout = getConfig('Queues')['ShutDownTimeout']`.

egede (approved):

mesmith75:
Is this for merging?

ishanrai05:
@mesmith75 Yes.

PR #1564, opened 2020-02-04
Add GPI function runfile

### LINKED ISSUE #1523: Add execfile to the GPI
Unfortunately `execfile` is no longer there in python 3. We should try and add an equivalent to the GPI as it was useful.

hedonhermdev:
Fixes #1523 
-Add a function runfile which resembles the function of execfile() in python2.


hedonhermdev:
I added the function in GangaCore.Runtime.GPIFunctions. Please let me know if there's a better place for the function to be added.

ganga/GangaCore/Runtime/GPIFunctions.py
```diff
@@ -93,3 +95,9 @@ def convert_merger_to_postprocessor(j):
         mp.process_objects.append(stripProxy(j).merger)
         stripProxy(j).postprocessors = mp
 
+def runfile(path_to_file):
+    """ A wrapper for the runpy.run_path() function. """
```
egede: I suggest adding an example of usage here as well. Maybe just the simple usage `runpy('myfile.py')` but then also add an example of what can be in `myfile.py`. The important thing to illustrate is how it behaves with the namespace of the `Ganga` session. So can I have 

```
j = Job()
j.submit()
```
in the file? And comment on if the object `j` in this case will be visible at the main prompt after running.

ganga/GangaCore/Runtime/GPIFunctions.py
```diff
@@ -93,3 +95,9 @@ def convert_merger_to_postprocessor(j):
...
+    if not type(path_to_file) == str:
```
egede: Avoid direct type comparisons. Always use `if not isinstance(path_to_file, str):` instead as somebody might in principle have a class that inherits from `str`.

ganga/GangaCore/Runtime/GPIFunctions.py
```diff
@@ -93,3 +95,21 @@ def convert_merger_to_postprocessor(j):
...
+    Usage:
+    Ganga In[]: runfile('myfile.py')
+    Note: If you want to run a file which includes something like:
+
+    j = Job()
+    j.submit()
+
+    then you have to run the function with init_globals=globals() in the Ganga interpreter like this
+
+     Ganga In[]: runfile('myfile.py', init_globals=globals())
```
egede: As we are exporting this to the Ganga API, I would make the `init_globals=globals()` the default. I actually think it should not even be an option. If a user wants anything else, it is not really a part of the Ganga API, and they can just import runpy themselves.

egede (commented):
Please add a test case that demonstrates that this is working. It should be placed in a new file in `GangaCore/test/GPI`. You can run the test yourself with `pytest mytest.py`. It will also run as part of the unit testing for every push made to this pull request.

ganga/GangaCore/Runtime/GPIFunctions.py
hedonhermdev: Makes sense, but I had trouble making it the default as setting init_globals=globals() in the function sets it to the globals of that file rather than the namespace of the GPI interpreter. Any ideas?

egede: I am not quite sure what will work here. Can you try something like

```
    from GangaCore.GPI import *
    return runpy.run_path(path_to_file, init_globals=globals())
```
where the idea is that we populate the local namespace with what has been exposed to the API of Ganga (called the GPI).

hedonhermdev:
How can I spawn a Ganga Interpreter in a GangaUnitTest?

ganga/GangaCore/Runtime/GPIFunctions.py
egede: Did you try this?

egede:
> How can I spawn a Ganga Interpreter in a GangaUnitTest?

See https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/test/GPI/TestArgSplitter.py but I do not think that you actually want this?

ganga/GangaCore/Runtime/GPIFunctions.py
```diff
@@ -93,3 +96,22 @@ def convert_merger_to_postprocessor(j):
...
+    Usage:
+    Ganga In[]: runfile('myfile.py')
+    Note: If you want to run a file which includes something like:
+
+    j = Job()
+    j.submit()
+
+    then you have to run the function with init_globals=globals() in the Ganga interpreter like this
+
+     Ganga In[]: runfile('myfile.py', init_globals=globals());
```
egede: The function no longer takes the `init_globals` argument, so this line should be fixed.

ganga/GangaCore/test/GPI/TestRunfile.py
```diff
@@ -0,0 +1,11 @@
...
+from GangaCore.testlib.GangaUnitTest import GangaUnitTest, start_ganga, stop_ganga
+
+class TestRunfile(GangaUnitTest):
+    def test(self):
+        from GangaCore.Runtime.GPIFunctions import runfile
+        with open('test_file.py', 'w') as f:
+            f.write('j = Job()\nj.submit()')
+            f.close()
+        runfile('test_file.py')
+        os.remove('test_file.py')
```
egede: Add newline at end of file

egede (changes requested):
Tests is fine. See other small comments. Nearly ready.

egede (approved):

hedonhermdev:
Why is the CI failing? I cannot see the details for the Jenkins CI.

egede:
> Why is the CI failing? I cannot see the details for the Jenkins CI.

Not quite sure. It runs again now as I made a few more tiny changes. We have had some intermittent problems lately with the CI failing during the build stage of the docker container (that is used for the testing).

egede:
> > Why is the CI failing? I cannot see the details for the Jenkins CI.
> 
> Not quite sure. It runs again now as I made a few more tiny changes. We have had some intermittent problems lately with the CI failing during the build stage of the docker container (that is used for the testing).

And it worked now.

egede:
Please delete this branch and start on a new one if you make other contributions. Many thanks for the work.

PR #1569, opened 2020-02-10
Update Dirac version in Dockerfile for testing

egede:
It seems like the current Dirac version can no longer be installed. Try to update.


egede:
OK, this seems to work. Please merge such that it can be integrated into other PRs with tests that fail at the moment.

mesmith75 (approved):

PR #1577, opened 2020-02-12
Add database support to ganga

ishanrai05:


egede:
Can we close this?

ishanrai05:
@egede  yes, i think this can be closed now.

PR #1578, opened 2020-02-12
Revive GangaDirac

dilawari2008:
On executing
export GANGA_CONFIG_PATH=GangaDirac/Dirac.ini
the followign error is shown

ERROR    problems with loading Plugin /home/vagish/ganga/src/ganga/ganga/GangaDirac
ERROR    Reason: GangaValueError: DIRAC Proxy `group` is not set. Set this in ~/.gangarc in `[defaults_DiracProxy]/group`

seemed to be an issue.

Changes have been made to fix it.

In GangaLSST, the default was mentioned as
[defaults_DiracProxy]
group = lsst_user
but so was not in GangaDirac.


dilawari2008:
Sorry for the typo.
Description Updated.

egede (commented):
This is not a bug. Any user of DIRAC will have to explicitly set this value, so there can be no default.

Issue #1566, opened 2020-02-05
Revive GangaTutorial package


egede:
The GangaTutorial package is not working in the latest release.

```
export GangaTutorial/Tutorial.ini`
(ganga) (base) [desktop] ~/programming/ganga % bin/ganga

*** Welcome to Ganga ***
Version: 8.2.0 - DEV
Documentation and support: http://cern.ch/ganga
Type help() or help('index') for online help.

This is free software (GPL), and you are welcome to redistribute it
under certain conditions; type license() for details.


INFO     reading config file /home/egede/.gangarc
INFO     reading config file /home/egede/programming/ganga/src/ganga/ganga/GangaTutorial/Tutorial.ini
ERROR    problems with loading Plugin /home/egede/programming/ganga/src/ganga/ganga/GangaTutorial
ERROR    Reason: No module named 'Lib'
PluginError: Failed to load plugin: /home/egede/programming/ganga/src/ganga/ganga/GangaTutorial. Ganga will now shutdown to prevent job corruption.
(consider --debug option for more information)
INFO     Stopping Job processing before shutting down Repositories
INFO     Shutting Down Ganga Repositories
INFO     Registry Shutdown
```
Looking at https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaGaudi/PACKAGE.py and https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaGaudi/__init__.py should make it clear what changes are required. Inside package there might be further changes required to deal with python3.

advait-l:
I would like to work on this issue.

egede:
You are very welcome.

advait-l:
I have set up the code but I'm unable to replicate the issue with the GangaTutorial package.
How do I do that?

egede:
Typo in instructions. It should say
```
[desktop] ~/programming/ganga % export GANGA_CONFIG_PATH=GangaTutorial/Tutorial.ini
[desktop] ~/programming/ganga % bin/ganga
```

egede:
Sorry, updated the instructions.

Ulrik.

advait-l:
Thanks, I was actually unable to replicate the issue.

Could you please elaborate upon what needs to be done to fix this issue?

egede:
If I ask `Ganga` to load the tutorial package on startup with the line
```
export GANGA_CONFIG_PATH=GangaTutorial/Tutorial.ini
```
and then start `Ganga`, I get the error
```
[...]

PR #1579, opened 2020-02-12
Revive GangaTutorial package #1566

dilawari2008:
Doesn't throw error anymore


ganga/GangaTutorial/__init__.py
```diff
@@ -2,8 +2,8 @@
 
 def loadPlugins( config = {} ):
 
-    import Lib.PrimeFactorizer
-    import Lib.PrimeFactorizerSplitter
-    import Lib.PrimeTableDataset
+    import GangaTutorial.Lib.PrimeFactorizer
+    import GangaTutorial.Lib.PrimeFactorizerSplitter
```
egede, line 6: I think this would better be
```
    import .Lib.PrimeFactorizer
```
and the same for all the other lines.

egede (changes requested):
This is a good start. However, the plugin is still not working. The code below should work and return a valid result.
```
j = Job()
j.application=PrimeFactorizer(number=20)
j.inputdata = PrimeTableDataset(table_id_lower=1, table_id_upper=1)
j.submit()
```
There are (at least) a further two problems. There is an error on submission:
```
Ganga In [9]: j.submit()                                                                                 
INFO     submitting job 178
INFO     job 178 status changed to "submitting"
INFO     Preparing subjobs
ERROR    Exception raised executing '_prepare_sj' in Thread 'None':
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/Core/GangaThread/WorkerThreads/WorkerThreadPool.py", line 124, in __worker_thread
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/GPIDev/Lib/Job/Job.py", line 1241, in _prepare_sj
    if app.is_prepared in [None, False]:
AttributeError: 'PrimeFactorizer' object has no attribute 'is_prepared'
```
This problem will just require a minor modification to the schema of the object I believe, like adding a line equivalent to https://github.com/ganga-devs/ganga/blob/e5b5da6f449f6c22cf6b7029896468152e738f5c/ganga/GangaCore/Lib/Executable/Executable.py#L59

A second problem is that the location `http://cern.ch/diane/download/primes` where the prime number table is picked up from is no longer valid. A different location should be found.

ganga/GangaTutorial/__init__.py
dilawari2008: @egede 
the line
import .Lib.PrimeFactorizer
is showing the following error
Screenshot from 2020-02-13 13-50-02

egede: Sorry, look at https://github.com/ganga-devs/ganga/blob/e5b5da6f449f6c22cf6b7029896468152e738f5c/ganga/GangaDirac/__init__.py#L111 for the syntax.

egede:
Stale. Will be fixed by #1615 instead.

PR #1585, opened 2020-02-21
Fix missing python3 upgrade

mesmith75:
One slipped through.

### DIFF
--- a/ganga/ganga/__init__.py
+++ b/ganga/ganga/__init__.py
@@ -43,7 +43,7 @@ try:
 except Exception as x:
     logger.critical('Ganga system plugins could not be loaded due to the following reason: %s', x)
     logger.exception(x)
-    raise(GangaException(x), None, sys.exc_info()[2])
+    raise GangaException.with_traceback(sys.exc_info()[2])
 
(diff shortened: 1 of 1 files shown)


ganga/ganga/__init__.py
```diff
@@ -43,7 +43,7 @@ def ganga_license():
 except Exception as x:
     logger.critical('Ganga system plugins could not be loaded due to the following reason: %s', x)
     logger.exception(x)
-    raise GangaException(x), None, sys.exc_info()[2]
+    raise(GangaException(x), None, sys.exc_info()[2])
```
alexanderrichards: This is probably more Pythonicly written in Python3 as:
```suggestion
    raise GangaException.with_traceback(sys.exc_info()[2])
```

alexanderrichards (changes requested):
One slipping through is not so bad but surprised given it's location it was missed. I assume you saw this while answering my question about importing little ganga

alexanderrichards (approved):

egede (approved):
Thanks. Was just about to file a bug report on this.

PR #1604, opened 2020-03-04
Revive ganga tutorial

DumbMachine:
Aims to solve [1566](https://github.com/ganga-devs/ganga/issues/1566). I took reference from https://github.com/ganga-devs/ganga/blob/e5b5da6f449f6c22cf6b7029896468152e738f5c/ganga/GangaDirac/__init__.py#L111.


egede:
The problem is a bit deeper than just getting the imports working. Take a look at the comment https://github.com/ganga-devs/ganga/pull/1579#pullrequestreview-357911652 . That PR seems to have got stale, but the comment still applies.

DumbMachine:
On running the job, after including the suggested line: 
```python
'is_prepared': SimpleItem(defvalue=None, strict_sequence=0, visitable=1, copyable=1, hidden=0, typelist=[None, ShareDir], protected=0, comparable=1, doc='Location of shared resources. Presence of this attribute implies the application has been prepared.')
```
I observed a error:
```python
ERROR    JobError: Error: GangaAttributeError: Object 'PrimeFactorizer' does not have attribute: 'prepare'
```
I think these means that `PrimeFactorizer` needs a `prepare()` function. I tried dummy functions and something similar to :
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaCore/Lib/Executable/Executable.py#L79

But I wasn't really sure about what was to be implemented in `prepare()` for this function. Could you provide some information here?

>A second problem is that the location http://cern.ch/diane/download/primes where the prime number table is picked up from is no longer valid. A different location should be found.

I think the table is similar to [this?](https://en.wikipedia.org/wiki/List_of_prime_numbers). We could upload a sample to this repo or as a raw file, so as to replace 
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaTutorial/Lib/primes/primes.py#L66

egede:
> But I wasn't really sure about what was to be implemented in prepare() for this function. Could you provide some information here?

A prepare function can pack up configuration files or similar such that it only needs to be done once, even if the application is used multiple times. One option would be to download the table of primes in this step and then save it as a `shared` file. But we can also just leave that to happen at runtime and let the prepare step be a `noop`.

egede:
>>   A second problem is that the location http://cern.ch/diane/download/primes where the prime number table is picked up from is no longer valid. A different location should be found.

> I think the table is similar to this?. We could upload a sample to this repo or as a raw file, so as to replace

Yes, that seems reasonable.

egede:
Solved by #1615

PR #1608, opened 2020-03-05
Update Jenkinsfile

### LINKED ISSUE #1563: Jenkins failures
```
#4 docker-image://docker.io/docker/dockerfile:1.0-experimental@sha256:cbd64...

#4 CACHED

failed to solve with frontend dockerfile.v0: failed to solve with frontend gateway.v0: frontend grpc server closed unexpectedly

script returned exit code 1
```
This is the cause of the recurring failures in Jenkins.

egede:
Test if introducing small wait makes the build of the DockerDirac container fail less often.


egede:
So at least this test worked. As the problem is intermittent, I think that we should just merge it and then see if the testing becomes more reliable.

mesmith75 (approved):

PR #1609, opened 2020-03-08
Extend @add_config Functionality

varunbankar:
Modified @add_config functionality such that now user can externally define config sections which can be used during Tests.

**Problem:** @add_config decorator was unable to create custom config sections which can be used while performing the tests

### **Findings:**

- `pytest` uses `gpi` fixture to create a ganga.GPI namespace with few modified configurations to perform the tests. 

- @add_decorator : adds `_config_values` attribute to the function passed to it and stores the passed configuration values in it.

**Example:** 
```
@add_config([('Configuration', 'user', 'test')])
def test_add_config(gpi):
    assert gpi.config['Configuration'].user == 'test'
```
When executed will add `_config_values` to the function `test_add_config(gpi)` like:
```
print(test_add_config._config_values)
# Output: {('Configuration', 'user'): 'test'}
```

So now when `pytest` is run, the `gpi` fixture takes the values from `_config_values`, converts them into list of 3 tuples. This list is then passed to `start_ganga` function as `extra_opts`

These `extra_opts` are then iterated and each 3 tuple is passed to a function `setUserValue`.

Function `setUserValue`: takes in 3 tuple (config_name, option_name, value) and checks if config_name (config section) exists in config, If it:
- exists, and option_name also exists then modifies the option_name.
- exists but option_name does NOT exist, then if (.is_open=True) then new option is added to the section
- exists but option_name does NOT exist, then if (.is_open=False) then nothing is changed
- rest all remaining config are stored in a dictionary which is not used later to makeConfig
- `.is_open` attribute decides if options are allowed to be added to that specific section.

After this is done, Ganga is bootstrapped and the test is ran in the testing ganga.GPI namespace. After the test is finished all the configuration is changed back to normal hence none of the changed configuration remains.

### **Modification**

Function `setUserValue` to `setUserValueForTest`: takes in 3 tuple (config_name, option_name, value) and checks if config_name (config section) exists in config, If it:
- exists, and option_name also exists then modifies the option_name.
- exists but option_name does NOT exist, then if (.is_open=True) then new option is added to the section
- exists but option_name does NOT exist, then if (.is_open=False) then nothing is changed
- does NOT exist then created the section using `makeConfigForTest` and adds the option.

Function `makeConfig` to `makeConfigForTest`: takes in all the same arguments as `makeConfig` but just the bootstrap restriction is removed as config whose sections are to be created for testing purposes can be done after bootstrapping too.

**Results**
`pytest` were created to confirm the functionality.
Screenshot 2020-03-09 at 4 06 53 AM

After the `pytest`, persistence of modified configuration was tested in normal ganga GPI
Screenshot 2020-03-09 at 4 08 13 AM
and the configuration did NOT exists.

### **Location**
**@add_config**
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaCore/testlib/decorators.py#L4
**gpi**
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaCore/testlib/fixtures.py#L34
**start_ganga**
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaCore/testlib/GangaUnitTest.py#L62
**setUserValue**
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaCore/Utility/Config/Config.py#L930
**makeConfig**
https://github.com/ganga-devs/ganga/blob/f79c9597443512b67ab451ce3e536c4751094b45/ganga/GangaCore/Utility/Config/Config.py#L180

### DIFF
--- a/ganga/GangaCore/Utility/Config/__init__.py
+++ b/ganga/GangaCore/Utility/Config/__init__.py
@@ -1,5 +1,5 @@
 
-from .Config import getConfig, makeConfig, ConfigError, setSessionValuesFromFiles, allConfigs, setConfigOption, expandConfigPath, config_scope, setSessionValue, setUserValue
+from .Config import getConfig, makeConfig, ConfigError, setSessionValuesFromFiles, allConfigs, setConfigOption, expandConfigPath, config_scope, setSessionValue, setUserValue, setUserValueForTest
 import os.path
 
 ## from Config import getConfigDict
--- a/ganga/GangaCore/test/GPI/TestAddConfigForTest.py
+++ b/ganga/GangaCore/test/GPI/TestAddConfigForTest.py
@@ -1,0 +1,26 @@
+from GangaCore.testlib.decorators import add_config
+from GangaCore.Utility.Config import getConfig
+
+# Case 1: Config Section exists & given option also exists -> Outcome expected: Should change the value of the option.
+@add_config([('Configuration', 'user', 'test')])
+def test_add_config_1(gpi):
+    c = getConfig('Configuration')
+    assert c['user'] == 'test'
+
+# Case 2: Config Section exists BUT (.is_open=False) & given option does NOT exists -> Outcome expected: Should not add option to the config.
+@add_config([('Configuration', 'TestOption', 'TestValue')])
+def test_add_config_2(gpi):
+    try:
+        c = getConfig('Configuration')
+        c['TestOption'] == 'TestValue'
+        assert False, "Should throw an Error"
+    except:
+        assert True
+    
+# Case 3: Config Section does NOT exists -> Outcome expected: Should create a config section with (.is_open=True) & add option to it.
+@add_config([('TestSection', 'TestOption', 'TestValue')])
+def test_add_config_3(gpi):
+    c = getConfig('TestSection')
+    assert c['TestOption'] == 'TestValue'
+
+#TODO Add test case with Config Section (.is_open=True) & option does NOT exist.
--- a/ganga/GangaCore/test/GPI/TestAddConfigForTest.py
+++ b/ganga/GangaCore/test/GPI/TestAddConfigForTest.py
@@ -1,5 +1,5 @@
 from GangaCore.testlib.decorators import add_config
-from GangaCore.Utility.Config import getConfig
+from GangaCore.Utility.Config import getConfig, ConfigError
 
 # Case 1: Config Section exists & given option also exists -> Outcome expected: Should change the value of the option.
 @add_config([('Configuration', 'user', 'test')])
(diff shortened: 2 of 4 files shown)


ganga/GangaCore/test/GPI/TestAddConfigForTest.py
```diff
@@ -0,0 +1,26 @@
...
+
+# Case 2: Config Section exists BUT (.is_open=False) & given option does NOT exists -> Outcome expected: Should not add option to the config.
+@add_config([('Configuration', 'TestOption', 'TestValue')])
+def test_add_config_2(gpi):
+    try:
+        c = getConfig('Configuration')
+        c['TestOption'] == 'TestValue'
+        assert False, "Should throw an Error"
+    except:
+        assert True
```
egede: You should test for the specific exception that `c['TestOption']` should raise. I suggest to use [assertRaises](http://docs.python.org/library/unittest.html#unittest.TestCase.assertRaises).

ganga/GangaCore/Utility/Config/__init__.py
```diff
@@ -1,5 +1,5 @@
 
-from .Config import getConfig, makeConfig, ConfigError, setSessionValuesFromFiles, allConfigs, setConfigOption, expandConfigPath, config_scope, setSessionValue, setUserValue
+from .Config import getConfig, makeConfig, ConfigError, setSessionValuesFromFiles, allConfigs, setConfigOption, expandConfigPath, config_scope, setSessionValue, setUserValue, setUserValueForTest
```
egede, line 2: Many of these functions are not used in the code below. Why import them here?

egede (changes requested):
Nice work. Just a few minor edits requested.

ganga/GangaCore/Utility/Config/__init__.py
varunbankar: I have just appended `setUserValueForTest` to the list, for `from GangaCore.Utility.Config import setUserValueForTest`

ganga/GangaCore/test/GPI/TestAddConfigForTest.py
varunbankar: I have modified the code for the specific exception, but I am unsure of how to use `assertRaises` here as it needs to inherit `GangaUnitTest`, where as the @add_config decorator needs `gpi` fixture to be used.

egede: That is fine as well.

egede (approved):

PR #1610, opened 2020-03-09
Fix error message semantics

admorris:
was: "Unable to finalise job after {job ID} retries due to error:\n{error message}"
now: "Unable to finalise job {job ID} after {num retries} retries due to error:\n{error message}"


mesmith75 (approved):

mesmith75:
Thanks!

PR #1613, opened 2020-03-13
Update Merger.py

rishabh2566:
For CustomMerge, a mergefiles() is used instead of merge()


egede (approved):
Thanks for updating and fixing. I changed a few other places where documentation was not accurate.

PR #1620, opened 2020-03-24
Allow for LocalFile in task chain input data

mesmith75:
Allows for LHCbTransform to chain input data with LocalFile in case of batch running.


egede (commented):
Why is this limited to spcific datafiles. Should it not be able to work with all GangaFile objects? MassStorageFile would be another sensible file type to use here.

ganga/GangaLHCb/Lib/Tasks/LHCbTransform.py
```diff
@@ -197,11 +199,30 @@ def createChainUnit(self, parent_units, use_copy_output=True):
                             for pat in excl_pat_list:
                                 if re.search(pat, f.lfn) and "LFN:" + f.lfn in flist:
                                     flist.remove("LFN:" + f.lfn)
+                    elif isType(f, LocalFile):
```
egede: Should istype be isinstance instead to allow for inheritance?

mesmith75: yes

mesmith75:
I think you are right, we should allow for MassStorageFile as well. But I guess we still want to keep the inputdata to a single type?

egede:
> I think you are right, we should allow for MassStorageFile as well. But I guess we still want to keep the inputdata to a single type?

Is this not part of a larger somewhat unresolved issue in Ganga where we allow inputdata to be a mix of different types of `GangaFile` objects. Often things will fail if that is the case. I do not think this is the place where we want to enforce that restriction (as it implies to a broader area).

ganga/GangaLHCb/Lib/Tasks/LHCbTransform.py
mesmith75: Actually no - these files are wrapped up in proxies so `isinstance` doesn't work on them

mesmith75:
So this now takes LocalFile and MassStorageFile.

ganga/GangaLHCb/Lib/Tasks/LHCbTransform.py
```diff
@@ -197,11 +199,30 @@ def createChainUnit(self, parent_units, use_copy_output=True):
                             for pat in excl_pat_list:
                                 if re.search(pat, f.lfn) and "LFN:" + f.lfn in flist:
                                     flist.remove("LFN:" + f.lfn)
+                    elif isType(f, LocalFile) or isType(f, MassStorageFile):
```
egede, line 203: MassStorageFile is not imported.

egede (changes requested):

egede (approved):

PR #1622, opened 2020-03-27
LHCbCompressedDatasetFix

mesmith75:
`isEmpty` was backwards.

### DIFF
--- a/ganga/GangaLHCb/Lib/LHCbDataset/LHCbCompressedDataset.py
+++ b/ganga/GangaLHCb/Lib/LHCbDataset/LHCbCompressedDataset.py
@@ -306,10 +306,7 @@ class LHCbCompressedDataset(GangaDataset):
 
     def isEmpty(self):
         '''Does this contain files'''
-        if len(self) > 0:
-            return False
-        else:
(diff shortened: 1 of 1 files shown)


ganga/GangaLHCb/Lib/LHCbDataset/LHCbCompressedDataset.py
```diff
@@ -307,9 +307,9 @@ def getRunNumbers(self):
     def isEmpty(self):
         '''Does this contain files'''
         if len(self) > 0:
-            return True
-        else:
             return False
+        else:
+            return True
```
egede: What about just
```
return not len(self) > 0
```

egede (dismissed):
Fine. But can we not simplify it a bit as in the suggestion?

ganga/GangaLHCb/Lib/LHCbDataset/LHCbCompressedDataset.py
mesmith75: Yes that is better

egede (approved):

PR #1624, opened 2020-04-01
fix deprication warning

ishanrai05:
Camel cases functions in threading and importing ABCs from collections are deprecated and will be soon removed in a future version of python

```bash
PendingDeprecationWarning: isAlive() is deprecated, use is_alive() instead
```
```bash
DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```


egede (changes requested):
Please do not make changes in the GangaAtlas directory. Otherwise OK.

egede (approved):

PR #1625, opened 2020-04-01
GoogleFile fix

DumbMachine:
Aims to fix GoogleFile partially broken #1623
TODO:
- [x] fix auth()
- [x] fix put()
- [x] fix restore()
- [x] fix remove()
- [x] implement get()
- [x] fix internalCopyTo()
- [ ] update tests

Issue #1617, opened 2020-03-15
Create a method for user to upload output of `report` functionality.


egede:
The `report()` function has been updated but we can no longer use the CERN based server. There is a need to create a new one. See #1587.

egede:
@RajatSingh95 Can you please add a link to the specification for the server backend.

RajatSingh95:
@egede https://aws.amazon.com/ec2/instance-types/
In the given link, you can have a look at **T2** instance. This instance is sufficient for the purpose.
Currently, I have hosted the service for accepting requests from `report()` function on **ec2 T2 instance**.

egede:
> @egede https://aws.amazon.com/ec2/instance-types/
> In the given link, you can have a look at T2 instance. This instance is sufficient for the purpose.

But you then setup some server on it?

RajatSingh95:
Yes. After creating the instance, you have to host a web server(in my case, it was django server) on it..

egede:
I am no expert here, so is there a way to more or less automate this (so having requested a T2, can I simply execute a script to setup, configure and start the server)?

RajatSingh95:
Yes. We can automate the whole process (from **requesting a T2 instance** to **starting the server**).
I will update you once I am ready with the script which can automate the whole process.

RajatSingh95:
@egede I have created a repository where I have pushed all the codes and instructions which need to be done to automate the whole process that we have discussed earlier. Below is the repo link:
https://github.com/RajatSingh95/ec2-automation

Please have a look at it and let me know if you encounter any problem.

RajatSingh95:
Currently, The setup of the Django webserver through the above automation script is the clone of the repo link given below:
https://github.com/RajatSingh95/report-server

If there are some changes to be made to this repo before deploying a permanent server, we can discuss it.

egede:
Many thanks.

DumbMachine:
@egede Once the [`GoogleFile`](https://github.com/ganga-devs/ganga/pull/1625) PR is merged I think we can use that for submitting reports. With the use of `report(job)` function, the required files (which are available [here](https://github.com/ganga-devs/ganga/blob/f9ca0c5b1c007ecdbc71d4af96566dc430c78a96/ganga/GangaCore/Utility/feedback_report.py#L149), I assume) will be uploaded by adding the corresponding file to a `GoogleFile` object.
[...]

PR #1629, opened 2020-04-06
report feedback using any GangaFile type

DumbMachine:
Aims to fix. #1617
This method works by replacing the old permanent server by google drive. The feedback report is uploaded to the user's account.
Can be merged after #1625 

Progress:
- [x] Support for `filetype` argument
- [x] Return the `File` Object


ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -23,15 +23,9 @@ def deny_modification(name, x):
         pass
 _initconfigFeed()
 
-def send_request(url,data,files):
-    response = requests.post(url, data=data, files=files)
-    return response
 def report(job=None):
```
egede: I would like to see this changed such that the user can get report in different locations (LocalFile, DiracFile, GoogleFile) in addition to the default upload mechanism of a GoogleFile. So maybe the following signatures.
```
report()
report(job=jobs[47])
report(filetype=LocalFile)
```
The docstring should be updated to reflect this.

Finally the report function should return the file object created such that I can do
```
myreport = report()
```

egede (changes requested):

ganga/GangaCore/Utility/feedback_report.py
DumbMachine: Do you want it like: 
1. if the user uses `GoogleFile`, the report is uploaded to his google drive.
2. if the user uses `LocalFile`, the report is saved as a LocalFile on the user's disk
.... ?

egede: Yes, that is right. We have other file types as well that it in principle should work with. I feel that our abstract interface for IGangaFile is a bit weak though in how to implement this well. That can be the source of another issue though.

DumbMachine:
The function now supports all the `File` types. Someone with proper credentials should test the `DiracFile` and other similar file types.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -72,57 +71,47 @@ def get_content_type(filename):
...
+                from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
+                from GangaCore.GPIDev.Lib.File.LocalFile import LocalFile
+                from GangaCore.GPIDev.Lib.File.LCGSEFile import LCGSEFile
+                from GangaCore.GPIDev.Lib.File.GoogleFile import GoogleFile
+                from GangaCore.GPIDev.Lib.File.MassStorageFile import MassStorageFile
+                
+                filetype = stripProxy(filetype)
+                assert(issubclass(filetype, IGangaFile))
+                print(filename, localdir)
+                if issubclass(filetype, LocalFile):
```
DumbMachine: `LocalFIle` doesn't really need te execution of `put` as we point it to the exact location of the file. 
Another question, I get the following error when using the `put` function of LocalFile.
image
Though the [`put`](https://github.com/ganga-devs/ganga/blob/d057b9e9f7fce0713190b4597c7e7ef8580608e3/ganga/GangaCore/GPIDev/Lib/File/LocalFile.py#L245) method seems to be implemented.

edit: It seems that LocalFile (without the `proxy`) doesn't give the error. The function that `put` method serves is not really used here, so I think `if` condition is justified.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -72,57 +71,47 @@ def get_content_type(filename):
...
+        try:
+                from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
+                from GangaCore.GPIDev.Lib.File.LocalFile import LocalFile
+                from GangaCore.GPIDev.Lib.File.LCGSEFile import LCGSEFile
+                from GangaCore.GPIDev.Lib.File.GoogleFile import GoogleFile
+                from GangaCore.GPIDev.Lib.File.MassStorageFile import MassStorageFile
+                
+                filetype = stripProxy(filetype)
+                assert(issubclass(filetype, IGangaFile))
+                print(filename, localdir)
```
egede: We should only use the logger to print anything.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -23,15 +25,12 @@ def deny_modification(name, x):
         pass
 _initconfigFeed()
 
-def send_request(url,data,files):
-    response = requests.post(url, data=data, files=files)
-    return response
-def report(job=None):
+def report(job=None, filetype=None):
```
egede: I think we should pick a default value for the `filetype`. `GoogleFile` is probably a good default as it allows for the easy sharing.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -72,57 +71,47 @@ def get_content_type(filename):
...
-            files = {'file': path}
-
-            return send_post(server, files)
-
-        return upload_file
+    def upload(filetype, filename, localdir):
+        try:
+                from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
+                from GangaCore.GPIDev.Lib.File.LocalFile import LocalFile
+                from GangaCore.GPIDev.Lib.File.LCGSEFile import LCGSEFile
```
egede: I do not think that we need these imports. Will need the GoogleFile one for the default filetype.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -72,57 +71,47 @@ def get_content_type(filename):
...
+                from GangaCore.GPIDev.Lib.File.GoogleFile import GoogleFile
+                from GangaCore.GPIDev.Lib.File.MassStorageFile import MassStorageFile
```
egede: No need for this

ganga/GangaCore/Utility/feedback_report.py
egede: Just posted #1633 to fix that problem. We can then avoid the if statement.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -72,57 +71,47 @@ def get_content_type(filename):
...
+        if issubclass(filetype, GoogleFile):
+            logger.info(
+                f'Your error report was uploaded as {feedback.downloadURL}.')  
+            logger.info(
+                'You may include this URL and the following summary information in your bug report or in the support email to the Ganga developers.')
+        else:
+            logger.info(
+                f'Your error report was saved as {filetype.location}.')
+            logger.info(
+                'You may include this file and the following summary information in your bug report or in the support email to the Ganga developers.')
```
egede: Avoid this if/else again. Even for LocalFile, there should be a URL (file:// ....).

egede (changes requested):

ganga/GangaCore/Utility/feedback_report.py
DumbMachine: Posted #1634 to fix this problem. I'll make the changes after its merger.

ganga/GangaCore/Utility/feedback_report.py
DumbMachine: This was for debugging, will be removed in the next commit

ganga/GangaCore/Utility/feedback_report.py
DumbMachine: All `Files` have an `accessUrl` while `GoogleFile` has a `downloadUrl` attribute. This means we would require an `if` condition to cater for this.
Edit: We could rename `downloadUrl` to `accessUrl` for `GoogleFile`, this will make for a uniform interface.

egede: Yes, please rename in `GoogleFile`.

DumbMachine:
Necessary updates have been made, please check the support of `DiracFile` and others.

DumbMachine:
I believe https://github.com/ganga-devs/ganga/blob/fc4b88a99ac1c826cb133f6cd225c7eda184bfd6/.gitignore#L5
should be 
` ganga.egg-info`. Since they are always included in the root directory of the project and not `python` folder.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -72,57 +68,33 @@ def get_content_type(filename):
...
+                from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
+                
+                filetype = stripProxy(filetype)
+                assert(issubclass(filetype, IGangaFile))
 
-    def send_post(url, files):
-        logger.debug("Sending Post to %s ,  containing %s" % (url, files))
+                feedback = filetype(filename)
+                feedback.localDir = localdir 
+                feedback.put()
```
egede: Please stick to 4 whitespace (not tabs) for indentation.

ganga/GangaCore/Utility/feedback_report.py
```diff
@@ -23,15 +25,9 @@ def deny_modification(name, x):
         pass
 _initconfigFeed()
 
-def send_request(url,data,files):
-    response = requests.post(url, data=data, files=files)
-    return response
-def report(job=None):
+def report(job=None, filetype=GoogleFile):
     """ Upload error reports (snapshot of configuration,job parameters, input/output files, command history etc.). Job argument is optional. """
```
egede: Docstring should be updated. It should explain that report can be provided as any GangaFile, that GoogleFile is the default and give an example where LocalFile is used.

egede (changes requested):
If I call `report` without a job (or with a job in the new status), I get traceback
```
Ganga In [5]: report()                                                                                   
UnboundLocalError                         Traceback (most recent call last)
~/programming/ganga/src/ganga/ganga/GangaCore/GPI/__init__.py in <module>
----> 1 report()

~/programming/ganga/src/ganga/ganga/GangaCore/Utility/feedback_report.py in report(job, filetype)
    677 
    678         resultArchive, tempDir = report_inner(
--> 679             job, isJob, isTask)
    680 
    681         report_bytes = os.path.getsize(resultArchive)

~/programming/ganga/src/ganga/ganga/GangaCore/Utility/feedback_report.py in report_inner(job, isJob, isTask)
    600         # Copy thread stack trace file
    601         try:
--> 602             thread_trace_source_path = os.path.join(getConfig('Configuration')['gangadir'], thread_trace_file_name)
    603             thread_trace_target_path = os.path.join(fullLogDirName, thread_trace_file_name)
    604             shutil.copyfile(thread_trace_source_path, thread_trace_target_path)

UnboundLocalError: local variable 'getConfig' referenced before assignment
```
This is caused by the unfortunate placement of the import at 
https://github.com/ganga-devs/ganga/blob/495b74d1747a4ce6de6d761047d0d61748bd2277/ganga/GangaCore/Utility/feedback_report.py#L443

I believe that PEP8 recommends that all imports are at the beginning of the file.

Do we have any test cases for the report. With the `LocalFile` support it should be easy to write a few that do not depend on external services.

egede (approved):

PR #1634, opened 2020-04-09
exposed put and get fns to GPI

### LINKED ISSUE #1633: Export put and get methods for LocalFile to the GPI

DumbMachine:
Aims to fix : #1633

Progress:
- [x] expose `get`
- [x] expose `put`


egede (approved):

DumbMachine:
Note: If we do not merge this change, #1629 will still function. Since we call the `put` method on the `Proxy` stripped version of `LocalFile`.

PR #1642, opened 2020-04-20
Change where python3 is sourced from

egede:


egede:
OK, this change seems to work. It changes the LHCb testing to use the same python3 installation as is used for the actual LHCb installation in /cvmfs. The reason that the testing stopped working was that the LHCb login made changes to the virtualenv.

ganga/GangaLHCb/test/testrunner
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 source /cvmfs/lhcb.cern.ch/lib/LbEnv && source LbLogin.sh
 export GANGA_CONFIG_PATH=GangaLHCb/LHCb.ini
-virtualenv -p python3 venv
-rm -f venv/bin/python
+export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/cvmfs/sft.cern.ch/lcg/releases/LCG_92python3/Python/3.6.3/x86_64-slc6-gcc62-opt/lib
```
mesmith75: Might be worth using a more recent release and with centos7

mesmith75 (changes requested):

mesmith75:
The remaining issue is that the environment guessing doesn't work with python3. `os.environ` now returns an `os._environ` class object rather than just a dict so trying to `eval` the string into a dict doesn't work.

egede:
If you can fix it, please go ahead.

mesmith75 (approved):

mesmith75:
@egede I think that is ok

egede:
Thanks for fixing this.

PR #1645, opened 2020-04-21
Update TestReport.py

### LINKED ISSUE #1643: Update testing of report()
The test for the `report()` caused an infinite hang. Most likely because it tried to open a web page for Google authentication on the test machine. The test should be fixed in a way such that the core functionality is tested but that this doesn't happen. Maybe just write to a `LocalFile`? The file in question is `ganga/GangaCore/test/GPI/TestReport.py`. At the moment the test is disabled.

varunbankar:
Aims to fix #1643 
Fixed the test to use LocalFile instead of GoogleFile.


egede (approved):
Perfect. Many thanks.

PR #1651, opened 2020-04-22
Update Dirac dockerfile

### LINKED ISSUE #1650: Use DIRAC_DEPRECATED_FAIL=True in your tests
In order to avoid future problems like this one https://github.com/ganga-devs/ganga/pull/1640
could you please run your DIRAC tests with the environment variable `DIRAC_DEPRECATED_FAIL=True` ? See https://dirac.readthedocs.io/en/latest/AdministratorGuide/ServerInstallations/environment_variable_configuration.html

This will result in an exception being thrown rather than a warning message

mesmith75:
Fixes #1650 .

In another PR I'll make this a configurable for experts.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,11 +13,14 @@ pipeline {
     stage('Build Package Images') {
       parallel {
         stage('GangaDirac') {
+          environment {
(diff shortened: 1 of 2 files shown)


egede (commented):
For this to be efficient, we need to look at updating the DIRAC version that we test against. At the moment this is just hard coded as 
```
RUN ./dirac-install -r v6r22p15 -i 27 -g v14r1
```
which I suspect is highly out of data.

mesmith75:
Ah yes - unfortunately it looks as if Dirac don't create releases on their github so we cannot find the latest automatically.

egede:
@mesmith75 Why was this one closed without getting merged?

mesmith75:
I didn't like it. #1652 replaces it

PR #1653, opened 2020-04-23
GangaND280 updated to work with python3

NickHastings:
I updated the code under GangaND280 to work with python3. Hopefully this can make it into one of the upcoming releases.


ganga/GangaCore/Lib/Batch/Batch.py
```diff
@@ -151,6 +151,7 @@ def submit(self, jobconfig, master_input_sandbox):
             jobnameopt = False
 
         if self.extraopts:
+            global re
```
egede: This (and the same one just below) looks really odd. There is an `import re` at the top of this file, so the regular expression module should be available without any further magic.

egede (changes requested):
Many thanks for making this update.

Please take the change to `Batch.py` out of this PR. It is unrelated, so just make another PR (or file as a bug in the issues). I say this mainly as I do not understand the reason for the change there so it might require further discussion.

As a side comment, it would probably be useful if the may `print` statements in the plugin were changed  to `logger` messages instead.

ganga/GangaCore/Lib/Batch/Batch.py
NickHastings: Actually, you'll see in the other places where re is used, the previous line has an additional "import re" within. So it looks like someone else encountered the same problem using other methods in Batch.py.  A bit of reading indicated to me that the correct fix is to add "global re", but I confess that I didn't take the time to understand why.

NickHastings:
> Many thanks for making this update.

No problem. I saw that the issues related the poor performance problems in 7.x were closed stating a fix  so thought it time to move to 8.x.

> Please take the change to `Batch.py` out of this PR. It is unrelated, so just make another PR (or file as a bug in the issues). I say this mainly as I do not understand the reason for the change there so it might require further discussion.

Ok, will do. However please note that it is related since the ND280 code that I was testing after making the changes for python3 would not work without that commit.

> As a side comment, it would probably be useful if the may `print` statements in the plugin were changed to `logger` messages instead.

I can look into it.

Cheers,

Nick.

egede (approved):
Many thanks for this. It will go into the next release. Let us know if there is any testing that you would like to happen for this package. If yes, just create an issue and we can talk further about it there.

PR #1655, opened 2020-04-24
Regex global

### LINKED ISSUE #1654: Use "global re" instead of importing inside methods.
While testing the python3 updates in the GangaND280 (see https://github.com/ganga-devs/ganga/pull/1653#issue-408228505 ) I found the Batch.submit() and Batch.resbumit() methods did not see the re module. The following error was issued:

```
"UnboundLocalError: local variable 're' referenced before assignment".
```
In two other places where re is used, the previous line has an additional "import re". So it looks like someone else encountered the same problem in other parts of Batch.py. A bit of reading indicated to me that the correct fix is to add "global re", but I confess that I didn't take the time to understand why.

So, in summary, this seems to fix a problem that I don't really understand. Anyone able to shed some light?

I've made a branch with my fix implemented and made a pull request https://github.com/ganga-devs/ganga/pull/1655#issue-408286503

NickHastings:
See https://github.com/ganga-devs/ganga/issues/1654#issue-605981860

### DIFF
--- a/ganga/GangaCore/Lib/Batch/Batch.py
+++ b/ganga/GangaCore/Lib/Batch/Batch.py
@@ -290,7 +290,7 @@ class Batch(IBackend):
         if rc == 0:
             with open(soutfile) as sout_file:
                 sout = sout_file.read()
-            import re
+            global re
             m = re.compile(
                 self.config['submit_res_pattern'], re.M).search(sout)
             if m is None:
@@ -331,7 +331,7 @@ class Batch(IBackend):
         if rc == 0:
             return True
         else:
-            import re
+            global re
             m = re.compile(self.config['kill_res_pattern'], re.M).search(sout)
             logger.warning('while killing job %s: %s', self.getJobObject().getFQID('.'), sout)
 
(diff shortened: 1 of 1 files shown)


egede:
Hi @NickHastings 
Something odd is for sure going on here. With the change implemented here, both the Jenkins and the Travis CI tests give errors of the type below. 

```
---------------------------- Captured stderr call -----------------------------
[0;31mCRITICAL Ganga system plugins could not be loaded due to the following reason: name 're' is used prior to global declaration (Batch.py, line 293)[0;0m
[0;31mERROR    name 're' is used prior to global declaration (Batch.py, line 293)
```

NickHastings:
@egede 
Hmm, maybe python version specific? I've been testing this with with python 3.6.10.

NickHastings:
Ok, my bad. I didn't test it properly after changing the in method "import re"s to "global re"s. Latest commit 8c9ddb62566401b4cf1ea14425468154dbd6ddb2 seems to work for me and it looks like Travis is happy with it, but seems to be still pending on Jenkins.

egede (approved):

Issue #1663, opened 2020-05-07
Project GangaGUI (Start Point)


varunbankar:
# Ganga GUI Project

This issue will act as the start point for the Ganga GUI project and will contain the plans and progress of the project. All the future issues will also be mentioned here and this will be a singular base for understanding the project pathway.

## About Project

This project aims to implement a Graphical User Interface (GUI) directly into Ganga, which could run alongside the existing text-based command line in Python. The implementation of the GUI will add an additional interface for the user to submit, monitor, and manage the job. Moreover, this project is more than just an implementation of a GUI, the creation of Web APIs will add modularity to the existing software and give users an ability to integrate the tool in other applications, to pull jobs data and perform actions on it, locally as well as remotely. This project has 7 major components along with 1 experimental component. The components are:

* Integration of a web server into Ganga.
* Creation of Web APIs for every possible interaction with the frontend.
* Creation of a clean, minimal looking, mobile responsive frontend UI.
* Connection of the frontend to the backend using AJAX requests.
* Improving the server security and implementation of authentication methods.
* Enhancing the reliability and responsiveness of the GUI and the Web APIs.
* Creation of unit tests and documentation of the features implemented.
* (Experimental) Integration of a CLI directly into the frontend UI.

You can check out the proposal here: [Ganga Graphical User Interface Proposal.pdf](https://github.com/ganga-devs/ganga/files/5143354/Ganga.Graphical.User.Interface.Proposal.pdf)

### List of PRs
* #1675 - **[Merged]**
* #1683 - **[Merged]**
* #1685 - **[Merged]** 
* #1700 - **[Merged]** 
* #1702 - **[Merged]** 
* #1703 - **[Merged]** 
* #1704 - **[Merged]** 
* #1705 - **[Merged]** 
* #1706 - **[Merged]** 
* #1707 - **[Merged]** 
* #1716 - **[Merged]** 
* #1717 - **[Merged]** 
* #1721 - **[Merged]** 
* #1728 - **[Merged]** - Most important of all

### Summary of PRs

**#1675** 
- [x] Create `GangaGUI` top level package
- [x] Understand the CI testing
- [x] Create `start_gui()` & `stop_gui()` functions
- [x] Set up testing framework

**#1683**
- [x] Add separate database support to GangaGUI
- [x] Create token-based authentication
- [x] Create tests for database and token authentication

**#1685**
- [x] Provides a fix for a small bug in `stop_gui` function

[...]

PR #1675, opened 2020-05-19
GangaGUI - Initial Server Set Up using GangaThread

varunbankar:
### Summary

__Project Structure__
We are going be using the Flask web framework for serving GUI to the user. In this Pull Request, I have created a `GangaGUI` a top-level package containing the Flask application. The entire Flask application resides inside the directory `GangaGUI/gui` and only it's configuration file `GangaGUI/config.py` and the start file `GangaGUI/start.py` resides outside of the `GangaGUI/gui` directory. All the tests related to the package are present under `GangaGUI/test` directory. 

__How To Start The Server__
After cloning the repository and downloading the updated requirements. The user can start the Flask server in 2 following ways:

* While starting ganga command line interface use the `--webgui` flag. 
* By setting `RUNTIME_PATH = GangaGUI` in the `.gangarc` and then after starting the ganga command-line interface use `start_gui()` function to start the Flask server and `stop_gui()` to stop the Flask server.

After starting the server, open the browser and navigate to `127.0.0.1:5000/`. In either case, when exiting ganga, a check will be performed if the Flask server GangaThread is alive or not, if alive the Flask server is stopped ultimately stopping the GangaThread.

__Understanding The Code - Review In The Following Order__ 

__`GangaGUI/gui`__
* This is where the Flask server is set up. The `__init__.py` creates a Flask application which is then used all over the other modules. 
* The `routes.py` is rather an important file containing what should be done when the user tries to access the specific route. For example after seeing the file, you may understand that when a user tries to access the `https://127.0.0.1/` routes, it executes the dashboard function and returns the output to the user's browser. 

__`config.py`__
* Contains all the config configuration of the Flask app which is used by `GangaGUI/gui/__init__.py`

__`start.py`__
* Contain code to start the Flask server on a GangaThread so it can access the ganga namespace while the command line interface is running without stopping the monitoring.
* A `GUIServerThread` is defined which is a typeof `GangaThread`. It need the `self.host` and `self.port` in order to start the Flask server on that specific `host` and `port`.
* It has a `run` method that logs the `host` and `port` to the `stdout` and starts the Flask server.
* It has a `shutdown` method, which makes the request to the Flask app on `https://host:port/shutdown` route to stop the Flask app.
* The `start_gui` function creates a GUIServerThread and starts it and the `stop_gui` function stop the already existing GUIServerThread. 
* Additional take note than the `start_gui()` function can take additional arguments of `host` and `port`, but they have a default value if they aren't provided.

__`GangaCore/Core/InternalServices/ShutdownManager.py`__
* Added a check if the webserver is running then call the `stop_gui()` function to stop the webserver for the smooth shutdown of ganga.

__`GangaCore/Runtime/bootstrap.py`__
* Modified `--webgui` flag to run `start_gui()` function.

__`BOOT.py`__
* Exports the `start_gui()` and `stop_gui` to the GPI.

__`GangaGUI/test`__
* Contains the tests for the Flask application.
* The test uses `GangaUnitTest` and during `setUp` it creates an app test client (which is built into the Flask` which tests the application without starting the server.

__Would need help in setting up the Jenkin CI for testing.__

__What's Next?__
In the next following week, I plan to
- [x] Creation of `secret_key` which is returned to the user after starting the server.
- [x] Implement token-based authentication for accessing the APIs. 
- [ ] Creation of job APIs. - Moved to #1683 


ganga/GangaGUI/config.py
```diff
@@ -0,0 +1,7 @@
+"""Contains the configuration settings for the GUI Flask App"""
+
+import os
+
+
+class Config(object):
+    SECRET_KEY = os.environ.get("SECRET_KEY") or "f3aa26t8b537abf6ee6305eefea0a10a"
```
egede, line 7: I guess this is just something temporary ...

ganga/GangaGUI/gui/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+
+# Disable development server warning (as it is not applicable to our GUI Flask App)
+cli = sys.modules['flask.cli']
+cli.show_server_banner = lambda *x: None
+
+# Flask App and get configuration from config.py file
+app = Flask(__name__)
+app.config.from_object(Config)
+
+from GangaGUI.gui import routes
```
egede, line 13: Put all imports at the start (unless there is an issue with it).

ganga/GangaGUI/start.py
```diff
@@ -0,0 +1,46 @@
...
+        except Exception as err:
+            logger.error("Failed to start GUI Server: {}".format(err))
+
+    def shutdown(self):
+        res = requests.post("http://localhost:{port}/shutdown".format(port=self.port))
+        logger.info("{} - {}".format(res.status_code, res.text))
+
+
+def start_gui(host: str = "localhost", port: int = 5000):
+    """Start GUI Flask App on a GangaThread"""
```
egede, line 31: As these functions will be exported to the user, it is important that the doc lines explain well what is going on. For now it is fine though.

egede (approved):

ganga/GangaGUI/config.py
varunbankar: Yes, but the user will always need to specify a secret key in order to run Flask. We can maybe generate it in `.gangarc`, and then take it from there in the future?

ganga/GangaGUI/gui/__init__.py
varunbankar: In Flask, due to the circular import issue, we have to import routes are the bottom. (It has become a kind of a convention)

ganga/GangaGUI/start.py
varunbankar: Okay, I will add the usage section also in the docstring along with more details on how it runs.

PR #1681, opened 2020-05-23
[WIP]Respoistory json

DumbMachine:
Work in progress for adding the following:
- `RepositoryJson`: Replacement of the current `xml` based repository

TODO:
- [x] function to dump json
- [x] function to load job from json
- [] working `repositoryjson`
- [x] tests

Pardon the spelling error in the branch name 😅


egede:
We agreed last week that we wanted to have a feature branch for the database work. So that means that we do not want to merge into develop now. Do we just make this `repositoryJson` the feature branch such that future developments will be branched off that pne and then merged into it via pull requests?

Tests are failing now, which is no surprise. I think. We might want to modify the testing such that we will get useful feedback from it.

ganga/GangaCore/Core/GangaRepository/GangaRepositoryJSON.py
```diff
@@ -0,0 +1,4 @@
+"""
+Currently a work in progress
+
+"""
```
egede, line 4: Remember a newline at the end of all files. You might be able to set your editor to add one automatically.

ganga/GangaCore/Core/GangaRepository/JStreamer.py
```diff
@@ -0,0 +1,196 @@
...
+
+from GangaCore.Utility.Plugin import PluginManagerError, allPlugins
+
+from GangaCore.GPIDev.Base.Objects import GangaObject, ObjectMetaclass
+from GangaCore.GPIDev.Schema import Schema, Version
+from GangaCore.GPIDev.Lib.GangaList.GangaList import makeGangaList
+
+# from GangaCore.Core.GangaRepository import SchemaVersionError
+
+import copy
```
egede: Move to the other import statements?

ganga/GangaCore/Core/GangaRepository/JStreamer.py
```diff
@@ -0,0 +1,196 @@
...
+        self.job_json = JsonDumper.object_to_json(starting_name, starting_node)
+        
+
+    @staticmethod
+    def object_to_json(name, node):
+        """Will give the attribute information of the provided `node` object as a python dict
+        """
+        node_info = {
+            "name": node._schema.name,
+            "version": node._schema.version.minor,
```
egede, line 90: Let us discuss if we should take the opportunity to get rid of the schema version attribute. We have in reality never used it.

ganga/GangaCore/Core/GangaRepository/JStreamer.py
```diff
@@ -0,0 +1,196 @@
...
+        for attr_name, attr_object in node._schema.allItems():
+            value = getattr(node, attr_name)
+            # if attr_name == "time":
+                # node_info[attr_name] = JsonDumper.handleDatetime(attr_name, getattr(node, attr_name))
+            if isType(value, (list, tuple, GangaList)):
+                node_info[attr_name] = list(value)        
+            elif isinstance(value, GangaObject):
+                node_info[attr_name] = JsonDumper.object_to_json(attr_name, getattr(node, attr_name))
+
+            # ForReview : We could use pickle to save the datetime
```
egede: We have (tried to) avoid pickle in the past as there is no promise that an object pickled with version X of python can be read by version Y.

ganga/GangaCore/Core/GangaRepository/JStreamer.py
```diff
@@ -0,0 +1,196 @@
...
+            else:
+                node_info[attr_name] = getattr(node, attr_name)
+        return node_info
+
+
+    def dump_json(self):
+        """Dump the json file
+        """
+        if self.dump_location is None:
+            print(self.job_json)
```
egede: Get in the habit of using logger rather than print.

ganga/GangaCore/Core/GangaRepository/JStreamer.py
```diff
@@ -0,0 +1,196 @@
...
+        #     if not obj._schema.version.isCompatible(version):
+        #         part_attr['currversion'] = '%s.%s' % (
+        #             obj._schema.version.major, obj._schema.version.minor)
+        #         self.errors.append(SchemaVersionError(
+        #             'Incompatible schema of %(name)s, repository is %(version)s currently in use is %(currversion)s' % part_attr))
+        #         obj = EmptyGangaObject()
+        #         self.ignore_count = 1
+        #     else:
+        #         # Initialize and cache a c class instance to use as a classs factory
+        #         print("I have no ideam when this case will be used")
```
egede: Can just be ignored given discussion on versions above.

egede (commented):
Consider implementing some useful tests that can be used to demonstrate that the things that have been implemented so far actually works.

DumbMachine:
> We agreed last week that we wanted to have a feature branch for the database work. So that means that we do not want to merge into develop now. Do we just make this `repositoryJson` the feature branch such that future developments will be branched off that pne and then merged into it via pull requests?
.

Yes this branch will be the base for future work related to the database as having the json repository is  required for implementing database support.

ganga/GangaCore/Core/GangaRepository/JStreamer.py
DumbMachine: Currently, I do not see any part of the new JStreamer requiring it. If it serves no purpose then I believe no harm lies in removing it.

DumbMachine: Currently, no part of `JStreamer` requires the use of `version` attr. If it serves no purpose then I believe no harm lies in removing it. What was the reason behind introducing it?

egede: It was introduced such that it would be possible to do more advanced upgrading of old jobs when we introduced a new feature. So you could take two old attributes and then maybe merge the information from them into a single new attribute. In reality we have never used such an ability, and in most cases it would just cause confusion.

DumbMachine:
I believe, once the tests pass and review is completed. I will branch off this feature branch to further work on the `GangaRepositoryJson`. The changes introduced in with the `GangaRepositoryJson.py` file will be > +1000 LOC and thus it would be convenient for that change to have its own feature branch, whose base is this current PR

egede (commented):
OK, so tests are passing now. As this will BE the feature branch, I will close this pull request for now. Please create new branches off this one and also merge in changes from the develop branch at a regular rate to make sure that there is minimal divergence.

PR #1683, opened 2020-06-01
GangaGUI - Token Based Authentication

varunbankar:
### Summary

__Background__
This PR aims at adding token-based authentication which will be used by the web APIs during the GangaGUI project.

__What is Token Based Authentication?__
The creation of web APIs for Ganga will enable access to its data over the network. The access to such sensitive data cannot be given to all and hence to limit the access to specific users, token-based authentication is implemented. 

__How Does It Work?__
Whenever a user will start the GUI server either by using `--webgui` flag or `start_gui()` in the command-line interface, he will be returned a tuple containing `(host, port, user, password)` data. This information will be also displayed to the user when they start the server. This is sensitive information and must not be shared publicly.

When the `start_gui()` function is invoked, under the hood, it generates a default user `GangaGUIAdmin`, a 7 character `password` and a unique `public_id`. All of this information is stored in an SQLite database named `gui_db.sqlite` and a tuple containing `(host, port, user, password)` is returned.

__Note:__ `start_gui()` can take custom `host`, `port` and `password` values, if the user wishes to set it themselves. 

Now in order to access the protected resource, the user will need to first generate a JWT token using the `/token` route using the authentication details that were generated above. The token contains information about the `public_id` and the `expiration_date` of the generating user. The `expiration_date` by default has be set to `5 days`.

__Example:__

For this example, we will take the `host`=`127.0.0.1` and `port`=`5000`.

* `User A` invokes `start_gui()` function in command-line interface.
*  `User A` is returned with `('127.0.0.1', 5000, 'GangaGUIAdmin', 'H7UN99L')` - password is randomly generated each time `start_gui()` function is called.
* `User A` wants to access the API at `http://127.0.0.1:5000/resource` but it is protected.
*  `User A` generated a token by making a `POST` request to `http://127.0.0.0.1:5000/token` with the data `{"user": "GangaGUIAdmin", "password": "H7UN99L"}`
* `User A` received a response like `{"token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJwdWJsaWNfaWQiOiI3Y2MxYjdjOS0wYzJmLTRjZGUtOTkyZC04MzJiYWRmM2ZlYjAiLCJleHAiOjE1OTEyNzQ1Nzd9.hU44_42ThLCGgWqIGC4zAktE2EZvKK8lhFVFH_r2MAo"}`
* `User A` uses the `token` in each request header that he makes to the protected route `http://127.0.0.1:5000/resource`. The header's name is `X-Access-Token`.
* `User A` in now granted access to the protected API and is happy.

---

### Understanding The Code

__`GangaGUI/gui/__init__.py`__:
* Added support for Flask-SQLAlchemy (ORM for Flask).
* Created a db instance of SQLAlchemy.

__`GangaGUI/config.py`__:
* Added configuration for SQLAlchemy - location of the SQLite database.

__`GangaGUI/gui/models.py`__:
* Created a Users model - which will, in turn, create a table in SQLite database with the columns: `id, public_id, user, password_hash, role`.
* Passwords are not stored directly but their hashed values are stored - additional security layer.
* Added a method to the `User` class to generate PyJWT token using the `public_id` information and expiration date of `5` days from creation.

__`GangaGUI/gui/routes.py`__:
* Created a `/token` route to the authenticated user and accordingly send an error message or token.
* Create a `@token_requred` which will be later used by the API to authenticate the token and identify the user.

__`GangaGUI/start.py`__:
* Updated the docstring to be more elaborate.
* Added random `public_id` and `password` generation at each startup and stored in a separate database.
* `start_gui()` now returns a tuple containing (host, port, user, password) information.
*  User can set his own password.
* Added database creation for development server startup.

__`GangaGUI/test/`__:
* Created test to test database creation and modifications.
* Created test to test token authentication system.

Other than this, updated few docstrings, added comments, and added new dependencies to `setup.py`.

---

### What's Next
- [ ] Creation of `/job` endpoint
- [ ] Creation of `/jobs` endpoint
- [ ] Creation of `/config` endpoint
- [ ] Creation of `/templates` endpoint

---
__Footnotes:__
I apologize for the delay with this PR, it should not have taken this long. Next week I will try to make up for the delay.


egede (commented):
This looks really good. It would be nice if the message to the user includes the specific URL in the printout such that the user can simply cut-and-paste, maybe like
```
INFO: You can now access the GUI at http://127.0.0.1:5000/resource
```
It would be even better if the password could be encoded into the URL that is printed such that the user only has to cut-and-paste a single thing. We might also need to help users with setting up port forwarding via ssh. So maybe print something like.
```
INFO: If on a remote system you may need to set up port forwarding to reach the web server. 
This can be done with "ssh -D 5000 <remote-ip>" from a terminal.

File: ganga/GangaGUI/gui/models.py [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+
+    def verify_password(self, password: str) -> bool:
+        return check_password_hash(self.password_hash, password)
+
+    def generate_auth_token(self, expires_in_days: int = 5) -> str:
+        return jwt.encode({"public_id": self.public_id, "exp": datetime.datetime.utcnow() + datetime.timedelta(days=expires_in_days)},
+                          app.config["SECRET_KEY"], algorithm="HS256")
+
+    def __repr__(self):
+        return "User {}: {} (Public ID: {}, Role: {})".format(self.id, self.user, self.public_id, self.role)
```
egede: Please add newlines at the end of all files.

ganga/GangaGUI/gui/routes.py
```diff
@@ -1,14 +1,81 @@
...
+from GangaGUI.gui.models import User
 
 
 @app.route("/")
 def dashboard():
     return "Hello GangaGUI"
 
 
-@app.route('/shutdown', methods=['POST'])
+# Generate token for API authentication - token validity 5 days
```
egede, line 13: What happens if the user runs the Ganga session and GUI for more than 5 days? Do they need to stop and start the GUI again?

egede (changes requested):
This all looks good. Just a few questions and comments to look at before this can be merged.

varunbankar:
> This looks really good. It would be nice if the message to the user includes the specific URL in the printout such that the user can simply cut-and-paste, maybe like
> 
> ```
> INFO: You can now access the GUI at http://127.0.0.1:5000/resource
> ```
> 
> It would be even better if the password could be encoded into the URL that is printed such that the user only has to cut-and-paste a single thing. We might also need to help users with setting up port forwarding via ssh. So maybe print something like.
> 
> ```
> INFO: If on a remote system you may need to set up port forwarding to reach the web server. 
> This can be done with "ssh -D 5000 <remote-ip>" from a terminal.
> ```

Thank you! 

Yes, I will add these lines for the logger to print at the start of the server.

When I will start working on the frontend, that time I will try to incorporate the user and password also in the URL.

ganga/GangaGUI/gui/routes.py
varunbankar: Nope, simply renew the token. If the user is just using the GUI, then no worries I'll add one popup just re-enter the password and token will be renewed.

ganga/GangaGUI/gui/models.py
varunbankar: Okay, will do

egede (approved):

PR #1684, opened 2020-06-01
Prompt user to set necessary config variables

alexanderrichards:


ganga/GangaDirac/__init__.py
```diff
@@ -115,12 +115,34 @@ def loadPlugins(config=None):
...
+    from GangaCore.GPIDev.Lib.Config.Config import getConfig#, config_file_as_text
+    from GangaCore.Runtime.bootstrap import GangaProgram
+    regenerate = False
+
     dirac_conf = getConfig('DIRAC')
     if not dirac_conf['DiracEnvSource'] and not dirac_conf['DiracEnvJSON']:
-        return
+        source = input("Please enter the location of your Dirac UI bashrc file.")
+        if not source:
+            logger.warning("No location specified, Dirac plugin will likely not work!")
```
egede: I think this message should explicitly say how this should be set in the `.gangarc` file. Also, would it not be better to make the message a `FATAL` to make ganga abort?

ganga/GangaDirac/__init__.py
```diff
@@ -115,12 +115,34 @@ def loadPlugins(config=None):
...
+            return
+        dirac_conf.setGangarcValue("DiracEnvSource", source)
+        regenerate = True
+#            config_file_as_text(False)
+
+    dirac_proxy = getConfig("defaults_DiracProxy")["group"]
+    if not dirac_proxy["group"] or dirac_proxy["group"] == "None":
+        group = input("Which group would you like to use for you proxy (e.g. <VO>_user)")
```
egede: Again, is it not better to get the user to write a correct `.gangarc` file and then abort. Maybe we can collate everything that is wrong and then write a single message (to avoid that user will have to start Ganga 4 times and each time be told that something new is missing.

ganga/GangaDirac/__init__.py
egede: I found that I also had to set
```
proxyInitCmd = dirac-proxy-init -g gridpp_user -M
userVO = gridpp
```
in the `[DIRAC]` section.

egede (changes requested):

ganga/GangaDirac/__init__.py
alexanderrichards: is this because you already had the LHCb package loaded as I don't think I have needed these. I have still to run tests on this branch but I'll check these as well

ganga/GangaDirac/__init__.py
alexanderrichards: The idea here is that it prompts the user and then writes the gangarc options for them. This code shouldn't need a ganga restart.

egede: I see. That is fine as well.

egede (approved):

egede (dismissed):

egede:
@alexanderrichards More work required or can it be merged?

alexanderrichards:
> @alexanderrichards More work required or can it be merged?

Actually I think I want to test it first. Also need to change the warnings to fatal. Just been working on the other PR. I'll do this today and then we can merge

alexanderrichards:
@egede I have tested this now and I can successfully submit a DIRAC hello world job from a completely clean generated config file with the only modification being loading the GangaDirac plugin with:
```ini
[Configuration]
RUNTIMEPATH = GangaDirac
```

The two necessary config options are prompted for and the config file is updated all without any restart so you can just go ahead and submit jobs immediately.

This all works for me and I only needed to prompt for the following config options in addition to the one above:

```ini
[DIRAC]
DiracEnvSource = 

[defaults_DiracProxy]
group = 
```

-----
@egede can you check if this works for you or if you really do need:

```ini
[DIRAC]
proxyInitCmd =
userVO =
```

As I said before if you do need them then it could possibly be either that they are needed for something more specific than my simple Hello World job test or perhaps they are needed if you also load the GangaLHCb or GangaGaudi plugins?

egede:
Hi Alex. You are right in that I do not need the two extra variable.s I think I just added things in the wrong order and then got confused.

egede (approved):

PR #1685, opened 2020-06-02
Update 'ShutdownManager.py' To Conditionally Run 'stop_gui()'

varunbankar:
This provides a fix by adding a conditional check to only run `stop_gui()` function is the GUIServerThread is running when shutting down Ganga.


egede (dismissed):

egede:
Please just resolve conflict. You can then just merge yourself afterwards.

egede (approved):

PR #1686, opened 2020-06-06
Disk checker for Dirac

### LINKED ISSUE #1682: Stop downloading when localdisk disk is full
Hi,
we regularly have people downloading over night the output of their zillion jobs, rapidly filling their localdisk. The outcome is that we see in our DIRAC monitoring huge pics of download failures. Would it be possible for you to stop the downloading loop if the localdisk is full ?

mesmith75:
Fixes #1682 

Add a decorator to check the available disk space before running the function.

I have wrapped everything in `DiracBase.py` that does downloading. If the check fails then an exception is raised and the job ends up in the failed state.

The message is hopefully clear enough so that users know to use `j.backend.reset()` to push the job back to `submitted` and then finalise again. Would it be preferable to leave it in completing instead?

Do we want this decorator in GangaCore for general usage?

Also this doesn't work if the `gangadir` is on EOS but users shouldn't be using that anyway.

### DIFF
--- a/ganga/GangaCore/Core/exceptions.py
+++ b/ganga/GangaCore/Core/exceptions.py
@@ -20,6 +20,10 @@ class GangaFileError(GangaException):
     This is intended to be thrown as an IGangaFile Error during Runtime
     """
 
+class GangaDiskSpaceError(GangaException):
+    """
+    Specific for issues with running out of disk space
+    """
 
 class PluginError(GangaException):
     """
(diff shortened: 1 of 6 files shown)


egede (commented):
I think it is a good idea to put this decorator in the Core. Only point is that the j.backend.reset comment is only valid for Dirac. It would be nice to see a test that demonstrates the use (through a mock of the space check).

egede:
Another question here. This exception will typically be raised in a worker thread. What happens then? I guess the thread will stop. If a new one is started then the code will just keep failing. If not, then all the worker threads will just die.

mesmith75:
I have added some tests. I am not sure it is the best way but I couldn't come up with anything better. I think mocking decorators is not so straightforward.

egede (approved):

PR #1691, opened 2020-06-12
Read comment from index cache

### LINKED ISSUE #1687: jobs.comment not lazy loading
From the time it takes to print it looks as if `j.comment` is loading the whole job rather than just reading the index.

mesmith75:
Fixes #1687 

+ fixes some testing warnings


egede (commented):
Will this trigger that all old jobs are read into memory in order to create this (mostly empty) field in the index cache?

mesmith75:
No I don't think so. The display index already exists so the repository shows up fine. It only loads the job the first time when you do j.comment and then saves it.

The second time it reads it direct from the cache.

alexanderrichards (approved):

PR #1699, opened 2020-06-21
Ganga repository json

DumbMachine:
Will introduce full support for a new repository type `json` which will act as the base/starting point for the Database Support.

- [ ] **Ready for review**

Progress:
- [X] Converted all the XML persistency to json based persistency 
- [X] Converted all the pickle persistency to json based persistency 
- [X] Run simple Jobs
- [X] Pass tests for simple Jobs 
- [X] Run Jobs with subjobs
- [ ] Pass tests for  Jobs with subjobs
     - [X] Some tests pass
     - [ ] Some do not as they require a change in files: IBackend.py which will change the behavior for the current XML backend


DumbMachine:
Since the code from this PR is just an intermediary for database support, I think we should close this PR and start tracking the PR which will introduce database support. We now have all the things required to store information inside of databases so I will create another PR for that which will track the progress of integration and testing of database-related functions.

PR #1700, opened 2020-06-22
GangaGUI - Job API

varunbankar:
### Summary

**Background**
In order to keep updating the job monitoring data without refreshing the whole web page, `Job` API endpoints are required. The front end makes requests to these APIs and can get information about the job, create a job, perform a certain action on the job, delete the job. These API endpoints can also be consumed by other applications with proper authentication.

**Work Overview (5 Endpoints | 37 Tests | 91% Test Coverage)**
Following endpoints are created for the `Job` API:
* `/api/job/<job_id>` - `GET` - Retrieves the general details about the job which is seen when `jobs` command is used in the command-line interface.
* `/api/job/<job_id>/<attribute>` - `GET` - Retrieves the details about the specific attribute of the job (can be any attribute of the Job object).
* `/api/job/create` - `POST` - Need to provide `template_id` value in the request body. Using the `template_id` create a job in the `new` state.
* `api/job/<job_id>/<action>` - `PUT` - Performs certain action on the job. Any method that the Job object has can be called using this endpoint. Just replace the placeholder `<action>` with the method name. Ex. To call the `submit` method on the job (with ID 13), the endpoint will be `/api/job/13/submit`.
    * If the method takes an argument then it can be passed in the body of the HTTP request in the field with same name as the method and the value of the field as the value to be passed in the method as an argument. Ex. `force_status` method on the job (ID: 13) takes an argument `failed`, so the request will be made to the endpoint `/api/job/13/force_status` with `force_status="killed"` in the request body. 
        * Please note that the value of the argument in the request body, if it is a `string` then it must be enclosed between `""` (double quotes), if it is is an `integer` then it shouldn't be enclosed between `""`, similarly even the boolean value `true` and `false` shouldn't be enclosed between `""` (double quotes).
    * With this endpoint, the user can also change the attributes of the job. Ex. User can change the `do_auto_resubmit` value to `False` for job (ID 13) by making a `PUT` request to `/api/job/13/do_auto_resubmit` with `do_auto_resubmit=false` in the request body.
* `/api/job/<job_id>` - `DELETE` - Deletes the job.

---

### How to Use the API

**Step 1: Start the GUI Server**

ganga_start_webserver

ganga_generate_token

**Step 3a: `/api/job/<job_id>` - `GET` Endpoint**

ganga_get_job_info

**Step 3b: `/api/job/<job_id>/<attribute>` - `GET` Endpoint**

ganga_job_attribute_info

**Step 3c: `/api/job/create` - `POST` Endpoint**

ganga_create_job

**Step 3d: `/api/job/<job_id>/<action>` - `PUT` Endpoint**

ganga_job_action

**Step 3e: `/api/job/<job_id>` - `DELETE` Endpoint**

ganga_job_delete

---


ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,365 @@ def decorated(*args, **kwargs):
...
+
+    # Imports
+    from GangaCore.GPI import jobs
+
+    # Job existence check
+    if job_id not in list(jobs.ids()):
+        return jsonify({"success": False, "message": "Job with ID {} does not exist".format(job_id)}), 400
+
+    supported_actions = ["do_auto_resubmit", "name", "parallel_submit", "copy", "kill", "force_status", "resubmit",
+                         "runPostProcessors", "submit"]
```
egede: This seems to replicate information that is already in the Job class, see https://github.com/ganga-devs/ganga/blob/92c6ac046b8b323d5428434289ee430c7dd2375a/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L233. Can't we just take it from there to avoid having to maintain the same information in two different places?

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,365 @@ def decorated(*args, **kwargs):
...
+    from GangaCore.GPI import jobs
+
+    # Job existence check
+    if job_id not in list(jobs.ids()):
+        return jsonify({"success": False, "message": "Job with ID {} does not exist".format(job_id)}), 400
+
+    supported_attributes = ["application", "backend", "do_auto_resubmit", "fqid", "id", "info", "inputdir", "inputdata",
+                            "inputfiles",
+                            "master", "name", "outputdir", "outputdata",
+                            "outputfiles", "parallel_submit", "splitter", "status", "subjobs", "time"]
```
egede: Is this list of information available elsewhere?

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,365 @@ def decorated(*args, **kwargs):
...
+        if value is None:
+            return jsonify({"success": False,
+                            "message": "Please provide do_auto_submit value in the request body."}), 400
+
+        # Value validator
+        if value in ["True", "False", "true", "false"]:
+            value = False if value == "False" or value == "false" else True
+        else:
+            return jsonify({"success": False,
+                            "message": "Please provide do_auto_submit value as True or False."}), 400
```
egede: This type checking should be factored out to make the code more readable.

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,365 @@ def decorated(*args, **kwargs):
...
+        if value is None:
+            return jsonify({"success": False,
+                            "message": "Please provide name value in the request body."}), 400
+
+        # Change the attribute value of the Job
+        try:
+            j = jobs(job_id)
+            setattr(j, action, str(value))
+        except Exception as err:
+            return jsonify({"success": False, "message": str(err)}), 400
```
egede: Can we get all the attributes from the schema of the `GangaObject`, in this case the `Job`. See https://github.com/ganga-devs/ganga/blob/92c6ac046b8b323d5428434289ee430c7dd2375a/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L201 . If there is a `SimpleType` in the schema, it seems to me that we could just auto generate the code for it here.

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,365 @@ def decorated(*args, **kwargs):
...
+    job_data["id"] = str(j.id)
+    job_data["fqid"] = str(j.fqid)
+    job_data["status"] = str(j.status)
+    job_data["name"] = str(j.name)
+    job_data["subjobs"] = len(j.subjobs)
+    job_data["application"] = str(type(j.application))
+    job_data["backend"] = str(type(j.backend))
+    job_data["backend.actualCE"] = str(j.backend.actualCE)
+    job_data["comment"] = str(j.comment)
+    job_data["subjob_statuses"] = str(j.returnSubjobStatuses())
```
egede: Maybe again some way to get this information automatically.

ganga/GangaGUI/gui/routes.py
egede: Or maybe we should not do it at all. Python in general is duck typed and type checking can become very heavy. In this specific case, the only thing you care about is probably if the argument can evaluate to True or False (but it could be an int).

egede (commented):
All my comments are of the same type. I am worried about that there is duplication of information here (like the attributes of a Job) here and elsewhere. While this is quick to implement and will work for now, it will lead to trouble later when changes are made elsewhere and the GUI then doesn't automatically follow.

ganga/GangaGUI/gui/routes.py
varunbankar: Actually the list is just to identify what attribute info is the user requesting. But essentially we are taking the data from the actual attribute implementation of the Job only so if any changes are made to the internal working of the attribute in the GangaCore, they will be reflected in the GUI. 

But if a case comes that a new attribute is added to the Job class, then it will be needed to add that in this list for it to be accessible to the GUI.

ganga/GangaGUI/gui/routes.py
varunbankar: The supported_actions list is just for validating that the action the user wants to perform is actually allowed or not. 

Eg: Suppose a user tries to change the name of the job (with ID 13), then they will make a `PUT` request at the endpoint `/job/13/name`. The `name` part of the endpoint is checked if it exists in the supported_actions list so that it is allowed. Which in this case of is allowed. But suppose, if the user uses the endpoint `/job/13/newname`, here the `newname` action is not allowed, and the user is returned with the HTTP code 400.

ganga/GangaGUI/gui/routes.py
varunbankar: I can refactor it, as the value that comes from the client-side is basically of type `str`, and in order to set the value of `do_auto_submit` it needs to be of type `bool`. This is just the type conversion.

ganga/GangaGUI/gui/routes.py
varunbankar: This is just creating the response object to be returned. As each the values to be sent back as the response in JSON format needs to be `str`. It's just making sure that the values are `str` and the server doesn't give an error.

DumbMachine: Since you already know the information you are going to show: `id, fqid, status, name, subjobs, application, backend, backend.actualCE, comments, subjob_statuses]`, I believe a good strat would be to loop over the displayed attr and use `getattr` to fetch them..
```python
for attr in [id, fqid, status, name, subjobs, application, backend, backend.actualCE, comments, subjob_statuses]:
  value = getattr(job, attr)
  job_data[attr] = value
```

DumbMachine: Same can be done with the multiple `functions` called in `/action`. 
```rust
getattr(job, actioname)() # equivalent to job.actioname()
```

varunbankar: This is really helpful, thanks a lot @DumbMachine

varunbankar:
@egede I have updated various endpoints to be generic and almost all of the error handling is left to Ganga. This has significantly shortened my code! 

Please have a look (especially at PUT method) and then I can update the rest of the endpoints and restructure my documentation, comments, and tests.

A brief explanation regarding the PUT method:

So whatever `action` the user requests, it checks if it exists in the `_exportmethods` of `GPIDev Job Class` then if require any kind of argument to be passed, it checks if the request body has that argument, if it does not it just tries to call the method. During all of this if any error is caused it is caught and returned to the user.

Eg:

A user wants to `force_status` of Job with id `13`, they will just call `/job/13/force_status` provide the `force_status: failed` field in the request body. The method will be automatically identified and the body data will be extracted and passed as an argument while calling the method.

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,190 @@ def decorated(*args, **kwargs):
...
+# Single Job Attribute Info API - GET Method
+@app.route("/job/<int:job_id>/<attribute>", methods=["GET"])
 @token_required
-def resource(current_user):
-    return jsonify({"success": True, "message": "Token is valid"})
+def job_attribute_endpoint(current_user, job_id: int, attribute: str):
+    """
+    Given the job id and attribute, return the attribute information in the string format via JSON.
+
+    Supported attributes: application, backend, do_auto_resubmit, fqid, id, info, inputdir, inputfile, master, name, outputdir, outputfiles, parallel_submit, splitter, status, subjobs, time
```
egede: Delete line or change to describe how it works now.

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,190 @@ def decorated(*args, **kwargs):
...
+
+
+# Perform certain action on the Job - PUT Method
+@app.route("/job/<int:job_id>/<action>", methods=["PUT"])
+# @token_required
+def job_action_endpoint(job_id: int, action: str):
+    """
+    Given the job_id and action in the endpoint, perform the action on the job.
+
+    Supported actions: do_auto_resubmit, name, parallel_submit, copy, kill, force_status, resubmit, runPostProcessors, submit
```
egede: As above

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,190 @@ def decorated(*args, **kwargs):
...
+
+    Important Info:
+    1) do_auto_resubmit: User needs to provide 'do_auto_resubmit' field and it's value in the request body.
+    2) name: User needs to provide 'name' field and it's value in the request body.
+    3) parallel_submit: User needs to provide 'parallel_submit' field and it's value in the request body.
+    4) force_status: User needs to provide 'force_status' field and it's value in the request body.
```
egede: Delete

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,190 @@ def decorated(*args, **kwargs):
...
+    if action not in Job._exportmethods:
+        return jsonify({"success": False,
+                        "message": "Job action {} is not currently supported or does not exist".format(action)}), 400
+
+    request_data = request.form.to_dict()
+
+    try:
+        j = jobs(job_id)
+        if action in request_data.keys():
+            getattr(j, action)(request_data[action])
```
egede: How does this work if there is an optional argument? Like the `resubmit` or `peek` methods?

egede (commented):
This looks good. Just a single question about how optional arguments are treated.

varunbankar:
> This looks good. Just a single question about how optional arguments are treated.

I will explain with an example. I will also have the request body data pretty printed. So you would be able to connect it with code.

Screenshot 2020-07-03 at 1 12 07 AM

 I created a new job which sleeps for 120 seconds.

Screenshot 2020-07-03 at 1 13 53 AM

Now I will try to submit it using the API by calling `http://localhost:5000/job/59/submit` and we know that `submit` method does not accept any arguments.

Screenshot 2020-07-03 at 1 15 12 AM

Here in this request I have deliberately added little data in the request body. So after sending the request, the `request_data` variable is:

Screenshot 2020-07-03 at 1 15 45 AM

and the status of the job is:

Screenshot 2020-07-03 at 1 16 09 AM

Now I will try to `force_status` of the job to `failed`. For this, I will make a `PUT` request to `http://localhost:5000/job/59/force_status`. We know that `force_status` method needs one argument. So I will add a field in the request body with the same name as the action name (which is `force_status`) and set its value as `"failed"`. Refer the following:

Screenshot 2020-07-03 at 1 16 39 AM

After making this request, the `request_data` variable is printed as:

Screenshot 2020-07-03 at 1 17 21 AM

The `if action in request_data.keys()` code just looks for the value that is provided by the user in the request body to be passed while calling the method.

and the status of the job now is:

Screenshot 2020-07-03 at 1 17 38 AM

varunbankar:
@egede I have made the necessary changes to the Job, Subjobs, Jobs, Templates APIs. If the code is satisfactory, I will rewrite few of the tests and update the documentation over the weekend whilst also working on the View routes.

ganga/GangaGUI/gui/routes.py
```diff
@@ -1,22 +1,34 @@
...
 from GangaGUI.gui.models import User
 
 
+# ******************** View Routes ******************** #
+
+# Dashboard route
 @app.route("/")
 def dashboard():
-    return "Hello GangaGUI"
+    return "GangaGUI"
```
alexanderrichards, line 15: Is this a placeholder, shouldn't it be returning the main dashboard?

alexanderrichards (commented):
Looks good to me, just one question about the root route "/"

ganga/GangaGUI/gui/routes.py
varunbankar: Yes it is actually placeholder, I am working on the view route as of now

egede (dismissed):

egede (dismissed):

egede (approved):

PR #1701, opened 2020-06-22
Don't download inputdata for files with accessURL

### LINKED ISSUE #1698: LHCbDataset downloads data for local jobs
When running a simple DaVinci job locally with `LHCbDataset` for the inputdata it downloads the file. It doesn't seem to be an issue for `LHCbCompressedDataset` though.

mesmith75:
Fixes #1698 .

This checks if a file from the inputdata has an accessURL . If it does then the inputdata won't be downloaded.


ganga/GangaCore/GPIDev/Lib/File/OutputFileManager.py
```diff
@@ -230,18 +230,34 @@ def doIHaveInputFiles(job):
...
     if job.inputdata:
         if job.inputdata and isType(job.inputdata, GangaDataset):
-            inputfiles_list += job.inputdata.files
+            for _f in job.inputdata:
+                try:
+                    _url = _f.accessURL()
+                    if _url:
+                        continue
+                    else:
+                        inputfiles_list.append(_f)
```
egede:
```python
if not _f.accessURL():
    inputfiles_list.append(_f)
```
seems shorter and clearer

ganga/GangaCore/GPIDev/Lib/File/OutputFileManager.py
```diff
@@ -230,18 +230,34 @@ def doIHaveInputFiles(job):
...
     elif job.master is not None:
         if job.master.inputdata and isType(job.master.inputdata, GangaDataset):
-            inputfiles_list += job.master.inputdata.files
+            for _f in job.master.inputdata:
+                try:
+                    _url = _f.accessURL()
+                    if _url:
+                        continue
+                    else:
+                        inputfiles_list.append(_f)
```
egede: as above

egede (changes requested):
I am a bit confused about what problem you are solving here. Is it about downloading to the worker node or about downloading after the job has finished?

mesmith75:
This solves #1698 . When running local or interactive jobs ganga is downloading remote inputdata to the WN. It doesn't make sense because the `data.py` file we create uses the remote access so we are filling up the tmp directory with massive data files for no reason.

egede (approved):

PR #1704, opened 2020-06-22
GangaGUI - Config API

varunbankar:
### Summary

**Background**
With the `Config` API, users can get a list of config sections and their options as well as effective value in JSON format.

**Work Overview (1 Endpoints | 1 Tests)**
Following endpoints are created for the `Jobs` API:
* `/api/config` - `GET` - Get a list of config sections, their docstring, their options, effective value and option's docstring.

---

### How to Use the API

ganga_start_webserver

ganga_generate_token

**Step 3: `/api/config` - `GET` Endpoint**

ganga_config

---


egede (dismissed):
Eventually I guess there should be a PUT method for the configuration as well. We have to make it clear for the user if they make a change that is persistent (so saved to the `.gangarc`) or just temporary (so valid for the session). Anyway that is another step and this can be merged.

varunbankar:
> Eventually I guess there should be a PUT method for the configuration as well. We have to make it clear for the user if they make a change that is persistent (so saved to the `.gangarc`) or just temporary (so valid for the session). Anyway that is another step and this can be merged.

Yes, it is on my TODO list. After the frontend page for the config is done. I am planning to implement it.

egede (dismissed):

egede (approved):

PR #1706, opened 2020-06-22
GangaGUI - Credential Store API

varunbankar:
### Summary

**Background**
`Credential Store` API can be used to get information of the stored credentials. It can also be used to renew expired credentials.

**Work Overview (2 Endpoints | 2 Tests)**
Following endpoints are created for the `Jobs` API:
* `/api/credential_store` - `GET` - Get a list of all the stored credentials in the credential store.
* `/api/credential_store/renew` - `PUT` - Renew expired credentials in the credential store.

---

### How to Use the API

**Step 1: Start the GUI Server**

ganga_start_webserver

ganga_generate_token

**Step 3a: `/api/credential_store` - `GET` Endpoint**

ganga_cr

**Step 3b: `/api/credential_store/renew` - `PUT` Endpoint**

ganga_cr_renew

---


egede (dismissed):

ganga/GangaGUI/gui/routes.py
```diff
@@ -64,13 +83,57 @@ def decorated(*args, **kwargs):
...
+    credential_info_list = []
+    try:
+        for c in credential_store:
+            credential_info = {}
+            credential_info["location"] = str(c.location)
+            credential_info["time_left"] = str(c.time_left())
+            credential_info["expiry_time"] = str(c.expiry_time())
+            credential_info["is_valid"] = str(c.is_valid())
+            credential_info["exists"] = str(c.exists())
+            credential_info_list.append(credential_info)
```
egede, line 534: Could this be built in some dynamic way (to avoid the hard coding of the method names and attributes)?

varunbankar: I can shorten this, but this is the monitoring information that is fixed and would be need for the front end. I could have made another route more generic use, in a sense users could just ask what they want like in case of `job` API but I looked into `ICredentialInfo` it didn't have a `schema` or `exportmethods`.

egede (approved):

PR #1710, opened 2020-06-26
Cleanup Jenkinsfile

mesmith75:
Do we still want the `codecov.yml`? We have a webhook for it but I don't know if we have it integrated?


alexanderrichards (dismissed):

alexanderrichards:
> Do we still want the `codecov.yml`? We have a webhook for it but I don't know if we have it integrated?

I don't think so. We run code coverage as part of the current testing. It's probably not in as nice a graphical format as using codecov but to me it makes sense to centralise everything on github

alexanderrichards (approved):

PR #1713, opened 2020-06-29
Fill in DiracEnvSource for LHCb

mesmith75:
Put a nonsense value in LHCb.ini so LHCb users are not prompted for it.


egede (approved):
Thanks for following up on this. I had a worry when @alexanderrichards checked in the code to query for this, but never got around to check.

PR #1715, opened 2020-07-03
Add remote inputfile check to GaudiExec

### LINKED ISSUE #1714: Check remote inputfiles exist
Currently users can enter nonsense as a DiracFile in `inputfiles`, even putting empty ones. We should check for a replica and raise an exception if there isn't one.

mesmith75:
Fixes #1714


alexanderrichards (dismissed):

mesmith75:
I have made some bugfixes and improvements.

The checks are now in `master_prepare` so they happen before any other uploading of the scripts and application tarballs happens.

mesmith75:
I've also cleaned up a random release notes file

egede (approved):

PR #1718, opened 2020-07-06
Fix compressed dataset length

mesmith75:
I would like to merge this and release a new version straight away as this bug is pretty bad.

The old method was just returning a quantity that is not persisted so loading the object from disk would always return length 0.


egede (approved):

niklasnolte:
hi, does this patch fix the generator behaviour aswell (aka `list(ds)`)?

mesmith75:
eurgh, this is proper amateur hour stuff. Thanks for pointing it out, I'll fix it in another PR

PR #1724, opened 2020-07-13
Changes to adapt to date/time format of condor 8.9.3

ductng:
Tested on
$CondorVersion: 8.9.3 Sep 16 2019 BuildID: 481344 PackageID: 8.9.3-1 $
$CondorPlatform: x86_64_RedHat7 $


ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -718,12 +718,12 @@ def getStateTime(self, status):
             splitLine = l.split()
             if checkstr == splitLine[0]:
                 year = datetime.datetime.now().year
-                if datetime.datetime.strptime(str(datetime.datetime.now().year)+'/'+splitLine[2]+' '+splitLine[3], "%Y/%m/%d %H:%M:%S") > datetime.datetime.now():
+                if datetime.datetime.strptime(splitLine[2]+' '+splitLine[3], "%Y-%m-%d %H:%M:%S") > datetime.datetime.now():
```
egede: This change should be made such that the old format (without the year as far as I can see) is accepted as well.

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -718,12 +718,12 @@ def getStateTime(self, status):
...
                     year = year - 1
-                timestr = str(year)+'/'+splitLine[2]+' '+splitLine[3]
+                timestr = splitLine[2]+' '+splitLine[3]
```
egede: As above. Might be useful to change the old notation to have a "-" and not a "/". That will mean you just have to check if there is a year in the string or not.

egede (changes requested):

ganga/GangaCore/Lib/Condor/Condor.py
ductng: Right, I was not aware that we were still using an old version. I have written a small parser now that checks the date format the first time the date is parsed and stores it in _condorDateFormat. It counts the number of date elements and checks the separation character
Whenever the date from the log file is requested, the date reformatted according to _condorDateFormat.

ganga/GangaCore/Lib/Condor/Condor.py
ductng: See my reply above, the changes are in a new commit

ganga/GangaCore/Lib/Condor/Condor.py
egede: HTCondor is not used just at CERN so we should guard against this type of changes.

egede (approved):

ductng:
The CI has failed and both failures are related to DIRAC. I am not familiar with it and I think my changes are not related to DIRAC. Can someone have a look and let me know, if there is anything I can do from my side?

egede:
> The CI has failed and both failures are related to DIRAC. I am not familiar with it and I think my changes are not related to DIRAC. Can someone have a look and let me know, if there is anything I can do from my side?

This is a brand new testing system, and I think it is caused by that the testing is using some secrets (a grid certificate) that is not allowed to be exposed to outside users. I think the solution is for us to clone your branch and then merge that one. Will let you know how it goes.

egede:
Issues with testing was a red herring. The real problem was that GitHub is suffering from some bandwith issues causing random failures. All good now.

alexanderrichards (approved):

egede:
@alexanderrichards Can you have a look at what goes wrong with the testing here. I am truly mystified. I am sure I saw it work yesterday, but now it fails again.

egede:
> @alexanderrichards Can you have a look at what goes wrong with the testing here. I am truly mystified. I am sure I saw it work yesterday, but now it fails again.

OK, so I did:
* I checked out this PR
* created a new branch from it
* pushed that branch to GitHub
* created a PR from it #1730

The effect of this was that the original PR (this one) started running the tests again and passed. I have no idea if all four of the actions above are required to trigger this ...

PR #1735, opened 2020-07-21
Dstreamer - Tools for reading/saving GangaObject into MongoDatabase

DumbMachine:
Introduces `DStreamer.py`, does the same things as `VStreamer.py` but uses `json` format instead of `XML` and `database` instead of `fs`.


egede (approved):
This all looks good.

PR #1737, opened 2020-07-22
Blank repository addition

DumbMachine:
Blank repository files have been added and conditional imports in all the files that required removal of locks.

### DIFF
--- a/ganga/GangaCore/Core/GangaRepository/DatabaseRegistry.py
+++ b/ganga/GangaCore/Core/GangaRepository/DatabaseRegistry.py
@@ -1,0 +1,699 @@
+
+
+import time
+
+from GangaCore.Utility.Config import getConfig
+from GangaCore.Utility.logging import getLogger
+from GangaCore.GPIDev.Schema import Schema, Version
+from GangaCore.GPIDev.Base.Objects import GangaObject
+from GangaCore.GPIDev.Base.Proxy import getName, isType
+from GangaCore.GPIDev.Lib.GangaList.GangaList import GangaList
+from GangaCore.Core.GangaThread.GangaThread import GangaThread
+from GangaCore.Core.exceptions import (
+    GangaException, InaccessibleObjectError,RepositoryError
+    )
+
+logger = getLogger()
+
+
+class RegistryError(GangaException):
+
+    def __init__(self, what=''):
+        super(RegistryError, self).__init__(self, what)
+        self.what = what
+
+    def __str__(self):
+        return "RegistryError: %s" % self.what
+
+
+class RegistryAccessError(RegistryError):
+
+    """ This error is raised if the request is valid in principle, 
+        but the Registry cannot be accessed at the moment."""
+
+    def __init__(self, what=''):
+        super(RegistryAccessError, self).__init__(what)
+
+    def __str__(self):
+        return "RegistryAccessError: %s" % self.what
+
+
+class RegistryLockError(RegistryError):
+
+    """ This error is raised if the request is valid in principle,
+        but the object is locked by another Ganga session"""
+
+    def __init__(self, what=''):
+        super(RegistryLockError, self).__init__(what)
+
+    def __str__(self):
+        return "RegistryLockError: %s" % self.what
+
+
+class ObjectNotInRegistryError(RegistryError):
+
+    """ This error is raised if an object has been associated to this registry,
+        but is not actually in the registry. This most probably indicates an internal Ganga error."""
+
+    def __init__(self, what=''):
+        super(ObjectNotInRegistryError, self).__init__(what)
+
+    def __str__(self):
+        return "ObjectNotInRegistryError: %s" % self.what
+
+
+class RegistryKeyError(RegistryError, KeyError):
+
+    """ This error is raised if the given id is not found in the registry """
+
+    def __init__(self, what=''):
+        super(RegistryKeyError, self).__init__(what)
+
+    def __str__(self):
+        return "RegistryKeyError: %s" % self.what
+
+
+class RegistryIndexError(RegistryError, IndexError):
+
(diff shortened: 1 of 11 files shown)


ganga/GangaCore/GPIDev/Adapters/IBackend.py
```diff
@@ -4,30 +4,31 @@
...
+from GangaCore.Core.exceptions import (
+    GangaException, GangaKeyError, IncompleteJobSubmissionError
+    )
 
 logger = GangaCore.Utility.logging.getLogger()
 
+if config["repositorytype"] != "Database":
+    from GangaCore.Core.GangaRepository.SubJobXMLList import SubJobXMLList
+else:
+    from GangaCore.Core.GangaRepository.SubJobJSONList import SubJobJsonList as SubJobXMLList
```
egede, line 29: This looks like it could cause confusion to me. You effectively overload the name `SubJobXMLList`. Somebody reading the code further down could be very confuse.d Would it not be better to use a single common name like just `SubJobList`. 

```python
if config["repositorytype"] != "Database":
    from GangaCore.Core.GangaRepository.SubJobXMLList import SubJobXMLList as SubJobList
else:
    from GangaCore.Core.GangaRepository.SubJobJSONList import SubJobJsonList as SubJobList
```

ganga/GangaCore/GPIDev/Adapters/IBackend.py
```diff
@@ -4,30 +4,31 @@
...
+from GangaCore.Core.GangaThread.WorkerThreads import getQueues
+from GangaCore.GPIDev.Base.Proxy import getName, isType, stripProxy
+from GangaCore.GPIDev.Credentials import credential_store, needed_credentials
+from GangaCore.Core.exceptions import (
+    GangaException, GangaKeyError, IncompleteJobSubmissionError
+    )
 
 logger = GangaCore.Utility.logging.getLogger()
 
+if config["repositorytype"] != "Database":
```
egede: I thinl we should have a positive identification here. So more a case like structure
```python
if config["repositorytype"] == "A":
    # do something
else if config["repositorytype"] == "B":
    # do something else
else:
   # log with a FATAL message that causes Ganga to quit.

File: ganga/GangaCore/GPIDev/Lib/Job/Job.py [hunk 3]
```diff
@@ -35,6 +34,11 @@
 config = GangaCore.Utility.Config.getConfig('Configuration')
 
 
+if config["repositorytype"] != "Database":
+    from GangaCore.Core.GangaRepository.SubJobXMLList import SubJobXMLList
+else:
+    from GangaCore.Core.GangaRepository.SubJobJSONList import SubJobJsonList as SubJobXMLList
+
```
egede: As above. If this behaviour is to appear in many places, we might need to factor it out.

ganga/GangaCore/__init__.py
```diff
@@ -961,3 +961,32 @@ def filefilter(fn):
...
+# Database
+db_config = makeConfig("DatabaseConfigurations",
+                       "Selection of database for ganga")
+db_config.addOption("database", "MONGODB",
+                    "others have not been implemented yet")
+db_config.addOption(
+    "containerName", "ganga_mongomon", "the identifier used to tag the docker container"
+)
+db_config.addOption("username", "default", "username")
+db_config.addOption("password", "default", "password")
```
egede, line 975: What is the role of this username and password? Is it such that another user on the same machine will not be able to communicate with the server? We do not want users to edit these themselves, so what is the best approach here?

egede:
Take a look at the failing test cases/. Some of them look directly related to your code. Others might no longer be relevant. If that is the case, then remove or disable them.

ganga/GangaCore/__init__.py
DumbMachine: >  Is it such that another user on the same machine will not be able to communicate with the server?

Yes.

> We do not want users to edit these themselves

Yes, we would not want that. That is why we will create a password different from the default Mongo password. In any case, the user explicitly wants to access the data, it can be done after reading the appropriate password from the config.

ganga/GangaCore/GPIDev/Adapters/IBackend.py
DumbMachine: The reason for name overloading is that I occasionally use the `XML` backend to compare the changes I made in the database `backend `. The overloading allows me to run any backend without having to change all occurrences backend-specific functions  .

ganga/GangaCore/GPIDev/Adapters/IBackend.py
DumbMachine: Agreed. Have made the change.

ganga/GangaCore/Core/GangaRepository/DatabaseRegistry.py
```diff
@@ -0,0 +1,699 @@
+
```
DumbMachine, line 1: This file is the same as `Registry.py`, minor changes to avoid having to acquire and release locks on jobs.

ganga/GangaCore/__init__.py
egede: OK, I see. So you will create a random password. So once created this password will stay the same the next time the server is started? I assume that it is only for the server connection and that the information itself in the database is not encrypted (so if password is lost, there is no loss of information).

ganga/GangaCore/GPIDev/Adapters/IBackend.py
egede: So you mean that this is kind of a debugging feature that will eventually be removed?

egede:
I still see a catastrophic failure of the tests. It seems like any test that actually try to start Ganga is failing (so the Unit tests are OK).

egede (approved):

PR #1742, opened 2020-07-23
add Dirac multicore options and fix disk checker

### LINKED ISSUE #72: Add multicore options to backend attribute
For all backends, the ability to request multicore processing should be added.

We discussed at the meeting if this should be one or two attributes to allow to distinguish between _as many cores as possible_ and _give me X cores_. As a request for _as many as possible_ might anyway not give access to a full worker node, then I think it is best left as a single attribute.
## Attribute

The attribute should be called _ncore_. The default value is 1. A value of -1 means as many as possible, a positive value means that you request this number of cores. Any other number (0 or below -1) results in an error at submission time.
## Backend implementation

Each backend will have to transmit this request onwards as part of the submission. If the backend does not support it, any ncore number different from 1 should result in a failed submission.
## Worker node implementation

On the worker node, an environment variable should be set that tells the number of cores accessible to the application. This should in the wrapper script be set in an environment variable `GANGA_NCORE`. For specific application objects in Ganga (i.e. Gaudi), we can just start up the application with the correct number of cores requested.

mesmith75:
Fixes #72 

I think I managed to find all the Dirac handlers. I didn't update LSST and SNO+ as I don't think they have been updated for python3.


alexanderrichards (dismissed):

egede (commented):
@alexanderrichards Does the (older) Dirac client and server at Imperial accept these options (it might not do anything with them but they should not cause an error).

alexanderrichards:
> @alexanderrichards Does the (older) Dirac client and server at Imperial accept these options (it might not do anything with them but they should not cause an error).

I don't know off hand. I'll find out

mesmith75:
What is the plan for this?

ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
```diff
@@ -27,5 +27,7 @@ j.setParametricInputData(###PARAMETRIC_INPUTDATA###)
 
 # submit the job to dirac
 j.setPlatform( '###PLATFORM###' )
+j.setNumberOfProcessors(minNumberOfProcessors=###MIN_PROCESSORS###, maxNumberOfProcessors=###MAX_PROCESSORS###)
+
```
egede: Can we just wrap this in a `try-except` clause and ignore it if the client raises an `AttributeError`

egede (changes requested):

ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
mesmith75: Ah yes, good idea.

mesmith75: This catches the `AttributeError` for very old Dirac versions, and a `TypeError` as the keyword arguments changed. In both cases it proceeds silently.

alexanderrichards:
> I don't know off hand. I'll find out

Sorry for the delay, I checked and Imperial are running at least v6r22p26 and that version has this method so should be all good. However a try: except: block couldn't hurt

alexanderrichards (approved):

alexanderrichards:
> Sorry for the delay, I checked and Imperial are running at least v6r22p26 and that version has this method so should be all good. However a try: except: block couldn't hurt

Ok scratch that, while they do have the method at v6r22p26 it doesn't take min/max as args, instead simply a number so you will need to try/catch

mesmith75:
Yes, that is the TypeError

alexanderrichards (dismissed):

egede (dismissed):
It would be good to update the docstring as well to include an example, and mention that setting min above 1 might limit places that job can run.

egede:
@mesmith75 There is a test failure. Not clear to me why from a quick look at it ...

mesmith75:
I think I have fixed the tests. They were failing because the disk checker now logs an error when something goes wrong.

The disk checker was not working on non-AFS file systems. Looking at the old code and output of `df` I'm not sure this function ever worked but the error was hidden in the debug. Updating to a regex should do it.

Also fixed a deprecation warning in a Condor regex whilst at it.

Edit: spoke too soon. I'm not sure how these tests ever passed. They seem unrelated to the original content of this PR.

egede (approved):

PR #1744, opened 2020-07-25
Migrating tool

DumbMachine:
Migration tool for XML users.
Proposed way:
```bash
$ ganga --migrate
```
Currently working method:
```bash
$ ganga migrater.py
```

### DIFF
--- a/ganga/GangaCore/Core/GangaRepository/migrate.py
+++ b/ganga/GangaCore/Core/GangaRepository/migrate.py
@@ -1,0 +1,162 @@
+# import ganga
+
+
+import os
+import time
+import pickle
+import pymongo
+
+from GangaCore.Core.GangaRepository.VStreamer import from_file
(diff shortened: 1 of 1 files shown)


egede:
A few initial questions about how this works
- Is the old XML files left in place (so that user can confirm that the new database works)?
- If an error happens during the conversion (running out of storage space probably the most likely), what happens?
- Related the the above, can we make an estimate of required disk space before the conversion happens?

DumbMachine:
>Is the old XML files left in place (so that user can confirm that the new database works)?

Yes, once the user confirms the database is working they can delete the root `repository` folder.

>If an error happens during the conversion (running out of storage space probably the most likely), what happens?

I will test this and update the comment here.

> Related the the above, can we make an estimate of required disk space before the conversion happens?

It requires 1/2 to 3/4 of the on-disk data.

egede (approved):

PR #1749, opened 2020-07-27
Fix quota check

### LINKED ISSUE #1746: checkDiskQuota possibly not working
To reproduce, start `ganga --debug` and type

```
from GangaCore.Runtime.Repository_runtime import checkDiskQuota
checkDiskQuota()
```

which shows

```
DEBUG    Error checking disk partition: a bytes-like object is required, not 'str' (Repository_runtime.py:93)
```

mesmith75:
Fixes #1746 
Another python3 issue!

### DIFF
--- a/ganga/GangaCore/Runtime/Repository_runtime.py
+++ b/ganga/GangaCore/Runtime/Repository_runtime.py
@@ -90,7 +90,7 @@ def checkDiskQuota():
         except GangaException as err:
             raise
         except Exception as err:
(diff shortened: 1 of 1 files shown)


egede (dismissed):

mesmith75:
I forgot to clean up the diagnostic print out. In any case I now changed the debug to an error so we are aware if something breaks but it won't be fatal.

alexanderrichards (approved):

PR #1752, opened 2020-08-06
Fix setting application platforms for LHCb

chrisburr:
In #1528 the [template for the DIRAC submission script](https://github.com/ganga-devs/ganga/commit/d65331e786b7c9073e7fe745f11e8ccbf95c587e#diff-99c765b1eb788172b420a7a55e58e926) was changed but `lhcbdiracAPI_script_template` wasn't updated to match.

This is the reason user jobs haven't been matched at various sites including `LCG.CSCS.ch`. Setting the platform in the JDL to `x86_64-centos7-gcc9-opt` (instead of `nehalem-centos7`) means it looks for worker nodes with an OS called `centos7-gcc9-opt`. Jobs are still matched to sites where singularity is useable as it bypasses the OS check.

It's probably worth adding some sanity checks here like `assert "j.setPlatform( \'###PLATFORM###\' )" in DiracScript` so this would have been caught sooner.

cc @VladimirRomanovsky

### DIFF
--- a/ganga/GangaLHCb/Lib/RTHandlers/RTHUtils.py
+++ b/ganga/GangaLHCb/Lib/RTHandlers/RTHUtils.py
@@ -47,7 +47,7 @@ j.setAncestorDepth(###ANCESTOR_DEPTH###)
 
 
     DiracScript = DiracScript.replace('\'###EXE_LOG_FILE###\'', '\'###EXE_LOG_FILE###\', systemConfig=\'###PLATFORM###\'')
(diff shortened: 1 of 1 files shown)


chrisburr:
It looks like the failures are due to the secrets not being available to this PR. I've tested this change locally and it works as expected.

egede:
> It looks like the failures are due to the secrets not being available to this PR. I've tested this change locally and it works as expected.

Yes, that is correct. I do not think that there is too much we can do about this. If we somehow changed it such that the secrets were available to PRs coming from forks, then it would be trivial to make a PR that simply published them. Would be nice if there was an easy way in GitHub to "pre-approve" and let the testing go ahead.

PR #1755, opened 2020-08-10
Revert "Hide the compressed filesets in LHCbCompressedDataset"

mesmith75:
Reverts ganga-devs/ganga#1754

It was rubbish and broke the space saving - the repository writes the whole file list.


egede (approved):
Oh well. Another try.

mesmith75:
I'm not quite sure of the best way forward. If I overwrite the `__getattribute__` for `files` then the compression gets lost when it is written to disk (the full file list get saved). Really the schema needs adjusting to get rid of files altogether and save the compressed part under a new name.  However I think that would break existing jobs in the repository.

The alternative is to do that in a new object and deprecate this one?

PR #1756, opened 2020-08-15
Container Handlers

DumbMachine:
Will allow the users to choose from the following services to act as the backend for the new `database` repository.
- [X] Native `mongo` Database
- [X] `Docker` Container
- [x] `uDocker` Container
- [x] `Singularity` Container


egede (commented):
Should the temp.py file be removed?

.database.rc
```diff
@@ -167,6 +167,31 @@ repositorytype = Database
...
+
+#  password
+#password = default
+
+#  port
+#port = default
+
+#  username
+#username = default
+
```
egede, line 194: Make file only have these lines and make it a file that is not hidden.

ganga/GangaCore/Core/GangaRepository/DStreamer.py
```diff
@@ -243,96 +219,10 @@ def handle_gangalist(glist):
...
-        self.simpleAttribute(attr_name, value, sequence)
-
-    @staticmethod
-    def componentAttribute(self, attr_name, value, ignore_subs):
-        """
-        """
-        if isType(value, (list, tuple, GangaList)):
-            ret_val = list(value)
-        else:
-            ret_val = JsonDumper.object_to_json(
```
egede, line 280: Could we replace the name `JsonDumper` with `JsonRepresentation`

ganga/GangaCore/Core/GangaRepository/GangaRepositoryCentral.py
```diff
@@ -21,12 +21,17 @@
...
 
 from GangaCore.Core.GangaRepository.DStreamer import (
-    EmptyGangaObject, object_to_database, object_from_database,
-    index_to_database, index_from_database,
+    EmptyGangaObject,
+    object_to_database,
+    object_from_database,
+    index_to_database,
+    index_from_database,
 )
```
egede, line 35: Make sure to make clear if this is not fully implemented.

ganga/GangaCore/Core/GangaRepository/GangaRepositoryDatabase.py
```diff
@@ -130,92 +147,63 @@ def __init__(self, registry):
...
 
-    # TODO: Add options to add custom option information for the database
 
-    def start_mongomon(self, options=None, backend="docker"):
+    def start_database(self):
         """Start the mongodb with the prefered back_end
         """
+        PORT = self.database_config["port"]
+        HOST = self.database_config["host"]
+        connection_string = f"mongodb://{HOST}:{PORT}/"
```
egede, line 182: We need to add some lightweight authentication for the connection.

ganga/GangaCore/Utility/Virtualization.py
```diff
@@ -8,6 +8,21 @@
 import urllib
 import os
 
+def checkNative():
```
egede, line 11: Move this to a separate file as it is misleading to have inside virtualization.py

ganga/GangaCore/__init__.py
```diff
@@ -966,20 +966,23 @@ def filefilter(fn):
...
 db_config.addOption(
-    "containerName", "ganga_mongomon", "the identifier used to tag the docker container"
+    "containerName", "gangaDB", "the identifier used to tag the docker container"
 )
 db_config.addOption("username", "default", "username")
 db_config.addOption("password", "default", "password")
-db_config.addOption("host", "default", "host")
-db_config.addOption("port", "default", "port")
+db_config.addOption("host", "localhost", "host")
+db_config.addOption("port", 27017, "port")
```
egede: Modify this to be a range (maybe a min and a max,

ganga/GangaCore/__init__.py
```diff
@@ -966,20 +966,23 @@ def filefilter(fn):
...
                        "Selection of database for ganga")
-db_config.addOption("database", "MONGODB",
-                    "others have not been implemented yet")
+db_config.addOption("controller", "docker",
+                    "Database Controller [Native, Docker, uDocker, Singularity]")
+db_config.addOption("baseImage", "mongo",
+                    "Docker Image for the database")
 db_config.addOption(
-    "containerName", "ganga_mongomon", "the identifier used to tag the docker container"
+    "containerName", "gangaDB", "the identifier used to tag the docker container"
```
egede: Make this name unique for a users Gangadir.

egede (changes requested):
The main change requested is to do with what happens when we have two users on the same system, or indeed the same user running two different Ganga sessions with different Gangadirs.

- The database inside the container should have a password for its access (that can be written into the Gangadir)
- The container should have a unique name (that again can be written into the gangadir)
- The port number should be taken from a range and a "free" one should be found dynamically.

PR #1758, opened 2020-08-18
Db testing

DumbMachine:
# Not ready for review


egede:
Note that there is a syntax error in the GitHub action file, see https://github.com/ganga-devs/ganga/pull/1758/checks

egede:
> Note that there is a syntax error in the GitHub action file, see https://github.com/ganga-devs/ganga/pull/1758/checks

 Error

every step must define a `uses` or `run` key

DumbMachine:
The history here is just a big mistake, have made the changes better in `db-tests`.

PR #1760, opened 2020-08-20
Fix Condor

mesmith75:
#1724 introduced bugs to Condor - I am not sure it worked.


egede (approved):
This underlies that tit would be really good to get some test cases that use the different backends.

PR #1762, opened 2020-08-23
Add ganga version to DIRAC JDL for easier debugging

chrisburr:
@VladimirRomanovsky suggested that it would be useful to be able know what Ganga version was used to submit jobs to DIRAC. This implements @fstagni's suggestion for how to include it in the JDL.

### DIFF
--- a/ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
+++ b/ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
@@ -26,7 +26,7 @@ j.setParametricInputData(###PARAMETRIC_INPUTDATA###)
 ###DIRAC_OPTS###
 
 # submit the job to dirac
-j.setPlatform( '###PLATFORM###' )
+j.setPlatform('###PLATFORM###')
 
 try:
 	j.setNumberOfProcessors(minNumberOfProcessors=###MIN_PROCESSORS###, maxNumberOfProcessors=###MAX_PROCESSORS###)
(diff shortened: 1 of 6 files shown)


mesmith75:
I'm not really sure I see the point of this

egede:
@marksmith @alexanderrichards  How do we trigger the testing to allow it to read "secrets" when a PR comes in from the outside? I thought it was sufficient to commit a noop to it, but it seems not. What is the easiest way?

alexanderrichards:
> @marksmith @alexanderrichards How do we trigger the testing to allow it to read "secrets" when a PR comes in from the outside? I thought it was sufficient to commit a noop to it, but it seems not. What is the easiest way?

It's not easy as for security reasons github doesn't expose secrets to forked repos. The easiest way is to create a new branch and have @chrisburr target that and simply merge it then the tests will run on the new branch

chrisburr:
> It's not easy as for security reasons github doesn't expose secrets to forked repos.

In case you're not aware, [there were some changes a few weeks ago](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks) that should make it possible to safely run the CI for PRs like this.

> I'm not really sure I see the point of this

This PR was triggered by the changes in #1752. After a fixed version of Ganga was deployed new jobs were still being submitted that had invalid platforms specified in the JDL. It was probably from the outdated version but it was impossible to know for sure. It's not going to be useful that often but it's sufficiently simple to implement that I think it's worth it.

egede (approved):
This seems sensible to me as a level of debug information.

PR #1774, opened 2020-09-11
Delete index information along with job information

DumbMachine:
gangaDB had a small bug, wherein the index would not be cleared when the job is deleted.
This removes that bug.
And removed `generate_container_rc` method from bootstrap.py


egede (approved):
All 👌

Issue #1778, opened 2020-09-17
Platform not being set correctly at submission for LHCb


CasaisA:
The following error is prompt 

```
INFO     Preparing subjobs
INFO     Submitting subjobs 0 to 2
ERROR    Job submission failed for job 61.1 : Job submission error ( 1503 : OS compatibility info not found)
```

when sending a `BenderModule`  job for the platform `x86_64+avx2+fma-centos7-gcc9-opt`. Apparently this is due to the following line in the generated `dirac-script-0.py`file:

```
j.setPlatform('x86_64+avx2+fma-centos7-gcc9-opt')
```

which should be changed to

```
j.setDIRACPlatform()
```
for LHCb jobs.

PR #1779, opened 2020-09-17
fix dirac script template

mesmith75:
A bug was introduced in #1762, which causes #1778 . This fixes that.

### DIFF
--- a/ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
+++ b/ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
@@ -26,7 +26,7 @@ j.setParametricInputData(###PARAMETRIC_INPUTDATA###)
 ###DIRAC_OPTS###
 
 # submit the job to dirac
-j.setPlatform('###PLATFORM###')
+j.setPlatform( '###PLATFORM###' )
 
 try:
 	j.setNumberOfProcessors(minNumberOfProcessors=###MIN_PROCESSORS###, maxNumberOfProcessors=###MAX_PROCESSORS###)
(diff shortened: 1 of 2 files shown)


chrisburr:
Urg, sorry about that.

It looks like it's worth adding a check in:

https://github.com/ganga-devs/ganga/blob/1d6130f7f0a592f12a022d8a95e1c48fe2e3ba0b/ganga/GangaLHCb/Lib/RTHandlers/RTHUtils.py#L50

to make sure that the substitutions actually find a match.

ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
```diff
@@ -26,7 +26,7 @@ j.setParametricInputData(###PARAMETRIC_INPUTDATA###)
 ###DIRAC_OPTS###
 
 # submit the job to dirac
-j.setPlatform('###PLATFORM###')
+j.setPlatform( '###PLATFORM###' )
```
alexanderrichards, line 29: Is this necessary as it violates PEP8?

alexanderrichards (approved):

ganga/GangaDirac/Lib/RTHandlers/DiracRTHScript.py.template
mesmith75: We can change the substitution if you prefer.
However I know this fixes the problem - I cannot guarantee I will find all the runtime handlers to change the substitution so I think this would be best.

alexanderrichards: no stick with what works 😉

PR #1598, opened 2020-02-29
Make backend script uniform for Batch and Local backend

ishanrai05:
Ref: #1592 #1588

PR #1785, opened 2020-09-29
iteritems() -> items() (python3) in  InteractiveScriptTemplate.py.template

admorris:


mesmith75:
I am not sure this is the fix we want. The other scripts are explicitly python2 (`#!/usr/bin/env python2`) - do we want to do that instead to bring this into line with the others?

I guess ultimately we need to make the scripts python 2 and 3 compatible but perhaps that is for #1598

egede (dismissed):

alexanderrichards:
I thought we were keeping the run scripts python 2 compatible (although I forget now the reason for this). This change however doesn't break Python 2 compatibility merely makes it less efficient. I don't think this is necessary therefore but if we are maintaining Python 2 for the scripts then I think being explicit `#!/usr/bin/env python2` should be done.

egede:
I think the optimal is if we can keep these scripts agnostic to python 2/3. The scripts are executed on remote nodes where we do not really have control over the python environment. So we can be in both the situation that we run on a plain centos7 system with only python2 available and on a system where python2 by default is no longer available (like Fedora 32).

mesmith75:
Ok, let's merge this then as it is a minimal change that works for both.

We should probably put some priority into sorting the other scripts.

mesmith75:
Does anyone remember how to get external branches to pass the tests?

alexanderrichards:
> Does anyone remember how to get external branches to pass the tests?

you need to merge it into a new local branch and then have that one target develop

mesmith75:
Ah, OK, I've replicated this in #1789 . I'll close this then.

PR #1790, opened 2020-10-10
removed psutil

### LINKED ISSUE #1787: Remove psutil from the database
Is it possible to remove `psutil` from the database code. This requires some compiling - we would prefer not to have any compiled dependencies. Indeed this doesn't work on lxplus now.

DumbMachine:
Aims to fix #1787.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,3 @@ coverage
 pygments
 pymongo
 docker
-psutil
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -8,5 +8,4 @@ unittest2
 coverage
 pygments
 pymongo
-docker
-psutil
\ No newline at end of file
+docker
```
mesmith75, line 11: Do we really want docker in the requirements?

mesmith75 (approved):

mesmith75:
Great thanks

requirements.txt
DumbMachine: Yes we do. That's how we control the docker instances.

PR #1791, opened 2020-10-10
Update setup.py

mesmith75:


DumbMachine:
> Singularity: We use a custom image sif file that is hosted here. Users can manually download the sif file and store it in the gangadir folder or let ganga do it for them. In the case of the latter, the file is downloaded and stored in the gangadir folder mentioned in the gangarc file.

The docs do tell users to download the `mongo.sif` file voluntarily. I used `gdown` to allow download if the user forget to.
Should we make it a requirement or instead add a warning that the user has not downloaded the file and does not have `gdown` installed?

DumbMachine:
`gdown` was used as the `sif` file was previously hosted on a google drive. Since that is not the case and a simple 
```bash
wget -O $DOWNLOAD_LOC https://cernbox.cern.ch/index.php/s/fv37NbLf24FpoWY/download
```
would service the purpose. I will open a new pr to remove dependency of `gdown` and instead make use of `subproc` call to `wget`

mesmith75:
Ok great. I'll close this then.

PR #1793, opened 2020-10-12
Add gdown dependency

egede:
Missing dependency added.

### DIFF
--- a/ganga/GangaCore/Core/GangaRepository/container_controllers.py
+++ b/ganga/GangaCore/Core/GangaRepository/container_controllers.py
@@ -4,6 +4,7 @@ import os
 import time
 import docker
 import subprocess
+import gdown
 
 from GangaCore.Utility import logging
 from GangaCore.Utility.Config import getConfig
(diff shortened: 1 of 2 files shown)


DumbMachine:
@egede do we want to add a `gdown` as a dependency or [use a simple `wget` call to download the files?](https://github.com/ganga-devs/ganga/pull/1791#issuecomment-706552599)

egede:
I first removed gdown and just used urlretrieve. But then realised it gave the nice progress bar. I prefer to keep it. Also prefer it to external unix commands.

DumbMachine:
Agreed, after testing the [removeGdown](https://github.com/ganga-devs/ganga/tree/removeGdown) branch that works using `wget` I would say creating a dangling process for a download is not the right choice.

DumbMachine (approved):

mesmith75:
@egede Does this also need to update the `requirements.txt`?

egede:
> @egede Does this also need to update the requirements.txt?
No, I believe the `-e` option in `requirements.txt` means that everything from `setup.py` is included.

PR #1794, opened 2020-10-12
added hash checking for mongo file

DumbMachine:
Added a check for the hash of `mongo.sif` before using it.
In case a hash match fails or another error occurs, we force quit ganga.

### DIFF
--- a/ganga/GangaCore/Core/GangaRepository/container_controllers.py
+++ b/ganga/GangaCore/Core/GangaRepository/container_controllers.py
@@ -77,7 +77,6 @@ def get_proc_environ(pid):
     return " ".join(output)
 
 
-
 def checkNative():
     """
     Check if the mongo database is instanlled locally
@@ -147,6 +146,7 @@ def get_database_config(gangadir):
 
     return config
 
+
 def mongod_exists(controller, cname=None):
     """
     Check of `mongod` process is running on the system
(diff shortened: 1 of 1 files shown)


ganga/GangaCore/Core/GangaRepository/GangaRepositoryDatabase.py
```diff
@@ -208,12 +208,16 @@ def start_database(self):
...
         self.connection = client[self.db_name]
 
         self.container_controller = controller_map[self.database_config["controller"]]
-        self.container_controller(
+        start_flag, error = self.container_controller(
             database_config=self.database_config, action="start", gangadir=self.gangadir)
+        if start_flag is False and error is not None:
+            # quit ganga process as the first registry could not start the database container
+            import sys
+            sys.exit()
```
DumbMachine: Is there a better way to quit ganga? Error during startup would imply that the first registry faced this error and thus I assume simply quitting everything would be fine.

egede: IIRC a
 ```
logger.fatal
```
will cause a shutdown after the message.

File: ganga/GangaCore/Core/GangaRepository/container_controllers.py [hunk 2]
```diff
@@ -173,10 +173,36 @@ def mongod_exists(controller, cname=None):
...
+    Args:
+        path (str): Location of the sif file
+    """
+    import hashlib
+
+    DEF_HASH = "a3833a7eb669f21f27de130495e2e1ba"
+
+    m = hashlib.md5()
+    hashfile = open(path, "rb")
+    while True:
```
mesmith75, line 193: I'm not sure I understand this bit - is it really the best way to read a file?

ganga/GangaCore/Core/GangaRepository/container_controllers.py
```diff
@@ -255,6 +281,15 @@ def singularity_handler(database_config, action, gangadir):
     if action not in ["start", "quit"]:
         raise NotImplementedError(f"Illegal Opertion on container")
 
+    has_correct_hash = check_sif_file_hash(sif_file)
+    if not has_correct_hash:
+        import sys
+        logger.fatal((
+            "sif_file does not match the expected hash." +
+            " Delete it to automatically download the correct file."
```
egede: Tell the location of the faulty file in this message. Otherwise the user will hunt around to delete a file with no idea about where it is.

ganga/GangaCore/Core/GangaRepository/container_controllers.py
egede: The idea ofthe code here (from StackOverflow) is to avoid having the full file in memory. OK for me.

egede (changes requested):
Please change the text in the FATAL message to be more explaining. As a different PR, I think we should move the URL for the container and the checksum to a configurable option. In that way it is easier to change in the future.

ganga/GangaCore/Core/GangaRepository/container_controllers.py
mesmith75: This looks resolved to me, so I'll merge and have a test.

mesmith75 (approved):

egede (approved):

PR #1804, opened 2020-10-20
--nojournal arg added

DumbMachine:
Branch to check of the journaling would speed up mongo


egede (changes requested):
This would make for a permanent change (which is not what we want). Can I suggest to make this part of the configuration system such that the journal file is ON by default, but can be turned OFF if requested (for testing or for living dangerously).

DumbMachine:
Yes, that would be great. I will add it to config, should we show a warning to the user when `journaling` is off?

egede:
> should we show a warning to the user when journaling is off?

Yes, that is a good idea.

PR #1812, opened 2020-11-18
Add mapping for DIRAC's "Completing" status

chrisburr:
When certifying the next release of LHCbDIRAC (v10r1) I noticed some jobs were being "randomly" marked as failed despite them appearing to be successful from DIRAC's point of view.

This issue turns out to be that DIRAC v7r1 has a new "Completing" status that should be thought of as "running" from ganga's perspective. I'm currently testing this by modifying my local `.gangarc` and it seems to have fixed the problem. I haven't checked if this change is needed elsewhere.


mesmith75 (approved):

mesmith75:
Thanks

PR #1814, opened 2020-11-20
Update GoogleFile to use private Oauth client

### LINKED ISSUE #1732: Update GoogleFile instructions in UserGuide to add something about first use.

### LINKED ISSUE #1808: Potential security issue with GoogleFile
To be able to write to your Google drive, Ganga has a client ID where the `client-secret` is exposed in the source. This follows the instructions on https://developers.google.com/identity/protocols/oauth2/?authuser=3#installed. Nevertheless this use pattern has been flagged as a security issue, so while investigating, the `client-secret` has been revoked. As a temporary measure, writing to Google drive will no longer work. When clarity is obtained, write access will be allowed again (but will require users to re-authenticate).

egede:
GoogFile has been updated such that you can use your own Google Oauth client ID. This is more secure and also prevents potential rate limits from other users. Closes #1808 .


egede:
Closes #1732

alexanderrichards (approved):

PR #1818, opened 2020-11-22
Remove expensive debug statement

### LINKED ISSUE #1638: Monitoring fails when jobs are created quickly
The following code 

```
for i in range(1000):
     print(i)
     j = Job()
```
fails sometimes (but not always) with an exception raised from the monitoring. Looks like a race condition between iterating over a dictionary in one thread (the monitoring one) and creating entries to it (in the main thread). The quick solution might be to remove the debug statement (which I think is actually really expensive to evaluate).

```
Exception in thread GANGA_Update_Thread_JobRegistry_Monitor:
  File "/usr/lib64/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py", line 584, in run
    log.debug("%s" % str(self.registry_slice))
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/GPIDev/Lib/Registry/RegistrySlice.py", line 473, in _display
    vals.append(str(obj._index_cache[display_str]))
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/GPIDev/Base/Objects.py", line 955, in _index_cache
    if self._fullyLoadedFromDisk():
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/GPIDev/Base/Objects.py", line 981, in _fullyLoadedFromDisk
    return self._getRegistry().has_loaded(self)
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/Core/GangaRepository/Registry.py", line 707, in has_loaded
    index = self.find(obj)
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/Core/GangaRepository/Registry.py", line 402, in find
    return next(id_ for id_, o in self._objects.items() if o is obj)
  File "/home/egede/programming/ganga/src/ganga/ganga/GangaCore/Core/GangaRepository/Registry.py", line 402, in <genexpr>
    return next(id_ for id_, o in self._objects.items() if o is obj)
RuntimeError: dictionary changed size during iteration
```
Thanks to @ast0815 for reporting in #1635.

egede:
This change removes an incredibly expensive debug message that at the same time could cause a race condition. In every loop of the monitoring it would take every job and search through all jobs currently loaded to find the identical one!

### DIFF
--- a/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
+++ b/ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
@@ -580,11 +580,7 @@ class JobRegistry_Monitor(GangaThread):
             # synchronize the main loop since we can get disable requests
             with self.__mainLoopCond:
 
-                log.debug("steps: %s" % str(self.steps))
-                log.debug("%s" % str(self.registry_slice))
-
-                log.debug("Monitoring loop __mainLoopCond")
-                log.debug('Monitoring loop lock acquired. Running loop')
+                log.debug('Monitoring loop lock acquired. Running loop: Steps %s' % str(self.steps))
                 # we are blocked here while the loop is disabled
                 while not self.enabled and self.steps <= 0:
                     log.debug("Not enabled")


egede:
Closes #1638

alexanderrichards (approved):

ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
```diff
@@ -580,11 +580,7 @@ def run(self):
             # synchronize the main loop since we can get disable requests
             with self.__mainLoopCond:
 
-                log.debug("steps: %s" % str(self.steps))
-                log.debug("%s" % str(self.registry_slice))
-
-                log.debug("Monitoring loop __mainLoopCond")
-                log.debug('Monitoring loop lock acquired. Running loop')
+                log.debug('Monitoring loop lock acquired. Running loop: Steps %s' % str(self.steps))
```
alexanderrichards, line 583: I can see that this is expensive for normal operation but if you would like to keep it only for the debug mode without impacting the normal operation you could consider changing this to the following.
```suggestion
                if log.isEnabledFor(logging.DEBUG):
                    log.debug('Monitoring loop lock acquired. Running loop: Steps %s' % str(self.steps))
```

alexanderrichards (commented):
Just a suggestion

ganga/GangaCore/Core/MonitoringComponent/Local_GangaMC_Service.py
egede: I considered this already but rejected for two reasons:

1. The statement is causing an occasional race condition where the list is looped on while it is added to. This is causing the monitoring thread to stop (and might actually be the cause of other trouble).
2. It is not even obvious that the debug statement was useful.

alexanderrichards: fair enough

mesmith75 (approved):

PR #1821, opened 2020-11-23
Make sys.path behave as iPython

### LINKED ISSUE #1811: CWD missing from path in ganga session
Currently one cannot import modules in the cwd into ganga.

egede:
For scripts, sys.path will behave in exactly the same way as ipython does now.

- `ganga /for/bar.py` will have `/foo` in `sys.path` when script executes.
- `runfile('/for/bar.py') from Ganga prompt will add `/foo` to `sys.path` for executing scirpt.
- simply starting Ganga does not add PWD to to `sys.path`

### DIFF
--- a/ganga/GangaCore/Runtime/GPIFunctions.py
+++ b/ganga/GangaCore/Runtime/GPIFunctions.py
@@ -1,3 +1,5 @@
+import sys
+import os.path
 import runpy
 from GangaCore.GPI import *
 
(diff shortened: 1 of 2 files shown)


egede:
Closes #1811

alexanderrichards (dismissed):

mesmith75 (approved):

mesmith75:
That's fixed it

PR #1817, opened 2020-11-22
Clean up Docker files

egede:

PR #1824, opened 2020-12-02
Update Dockerfile for Imperial DIRAC 7 instance

alexanderrichards:


egede (commented):
We heave a duplication with #1817 now as well where I put in the change to v7 (as i thought it was the cause of the failing tests).

alexanderrichards:
> We heave a duplication with #1817 now as well where I put in the change to v7 (as i thought it was the cause of the failing tests).

ah ok well I'll close this one then

alexanderrichards:
Out of curiosity, why do we parse the dirac ui version from `https://raw.githubusercontent.com/DIRACGrid/DIRAC/integration/release.notes` rather than hard coding it? Doesn't this make the tests less robust in that we could repeat the same tests on the same branch but get different results based on what DIRAC release is doing?

PR #1830, opened 2020-12-16
improve getOutputData

mesmith75:
Adds a check that all the files are available initially, and then that they are all successfully downloaded

### DIFF
--- a/ganga/GangaDirac/Lib/Backends/DiracUtils.py
+++ b/ganga/GangaDirac/Lib/Backends/DiracUtils.py
@@ -205,11 +205,21 @@ def getAccessURLs(lfns, defaultSE = '', protocol = '', credential_requirements=N
 
 exportToGPI('getAccessURLs', getAccessURLs, 'Functions')
 
-def getReplicas(lfns, credential_requirements=None):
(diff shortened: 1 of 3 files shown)


mesmith75:
Also added a `getReplicas` function to the GPI

ganga/GangaDirac/Lib/Backends/DiracUtils.py
```diff
@@ -204,3 +204,16 @@ def getAccessURLs(lfns, defaultSE = '', protocol = '', credential_requirements=N
     return myURLs
 
 exportToGPI('getAccessURLs', getAccessURLs, 'Functions')
+
+def getReplicas(lfns, credential_requirements=None):
+    """
+    Return a dict of files and their replicas.
+    lfns can be a string or a list
+    """
+    reps = execute('getReplicas(%s)' % str(lfns), cred_req=credential_requirements)
```
egede: This global function expects that `str(lfns)` will return a list of strings. Should it also be able to accept a dataset object?

mesmith75: Good idea, sorted.

egede (approved):

Issue #1828, opened 2020-12-08
"failed" is not enough to classify a job failure state


VukanJ:
A job that is failed can be failed because the backend is not up-to-date, the executable finished with errors, or some other problem. If the job is completed but the data was removed from the grid because it is outdated, the job is not even failed, it is "completed" but really should be failed if getOutputData has not been successfully called by the user. Consider the following scenarios:

**Scenario 1:**
1. User submits a job (DiracFile) with subjobs and 20 files per subjob; some jobs fail and need to be re-splitted, to have a change of succeeding
2. Now the original job must be preserved in its current failed state under all circumstances (no backend reset or resubmit allowed) so that we do not select the same data twice
3. If the user now: resets the backend of the original job, or resubmits some the original subjobs and they complete, we partially select the same data twice -> Physics analysis show-stopper

**Scenario 2:**
1. User submits a job (DiracFile) with subjobs and 20 files per subjob; some jobs fail and need to be re-splitted, to have a change of succeeding
2. Now some of the original job output files are removed from the grid because of overfull quota, or the user just does not make the download in time (or deletes the files himself to be pedantic)
3. Now, some "original" jobs are still marked as completed, but the user needs to resubmit them, right? So we force the job to failed.... and then it is indistinguishable from other failed jobs, that should stay failed.  -> Physics analysis show-stopper

In both cases the user is more or less required to do very detailed manual bookkeeping, which should be gangas job.
What I want to say is that there does not really seem to be a security measure in place in ganga, which ensures, that there is continuity between jobs. The user is not really prevented from losing data or selecting it twice. I think one reason for this problem, is that the "failed" state is just not enough to characterize a failure state. A job or subjob should ideally have more states like: "frozen" if the user does not want to change the job state ever again, or "unavailable" if the data turns out to be missing on the grid if getOutputData fails, which hardly can be called "completed", right?

mesmith75:
I am not sure what you are after. Bookkeeping of data is not ganga's responsibility.

VukanJ:
[...]

PR #1834, opened 2021-01-06
Add frozen states

mesmith75:
Finished the other part of #1828 .

Adds `completed_frozen` and `failed_frozen` jobs status, with `j.freeze()` and `j.unfreeze()` methods. You can freeze a master job or individual subjobs.

If a subjob is frozen it cannot be resubmitted (the status transition is not allowed). If a master job is frozen none of the subjobs can be resubmitted, even if they are not frozen (the master status transition is not allowed).

### DIFF
--- a/ganga/GangaGUI/api/routes.py
+++ b/ganga/GangaGUI/api/routes.py
@@ -528,8 +528,12 @@ def jobs_statistics():
 
     try:
         # Store statistics
-        for stat in ["new", "running", "completed", "failed", "killed", "completed_frozen", "failed_frozen"]:
+        for stat in ["new", "running", "completed", "failed", "killed"]:
             statistics[stat] = len(jobs.select(status=stat))
+            if stat == "completed":
+                statistics[stat] = len(jobs.select(status="completed_frozen"))
+            elif stat == "failed":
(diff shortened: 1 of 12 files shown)


egede (commented):
Is there a status list in the GUI somewhere that should be modified as well?

mesmith75:
Ah yes, good point. I had forgotten about the GUI. I think that will need some modification

ganga/GangaGUI/api/routes.py
```diff
@@ -530,6 +530,10 @@ def jobs_statistics():
         # Store statistics
         for stat in ["new", "running", "completed", "failed", "killed"]:
             statistics[stat] = len(jobs.select(status=stat))
+            if stat == "completed":
+                statistics[stat] = len(jobs.select(status="completed_frozen"))
+            elif stat == "failed":
+                statistics[stat] = len(jobs.select(status="failed_frozen"))
```
egede: Should these not be added to the numbers for completed and failed (rather than replacing them).

mesmith75: yes

egede (approved):

PR #1835, opened 2021-01-26
Improve job remove files

mesmith75:
Adds a `Dirac` `removeLFNs` help function. Also uses the Dirac `removeOutputData` when removing jobs for better speed. The `Dirac` function passes a list of LFNs to the Dirac thread, rather than removing each file one at a time.

There is a function in the Dirac API that takes a list of LFNs but it probably won't be much quicker than this.

### DIFF
--- a/ganga/GangaDirac/Lib/Backends/DiracBase.py
+++ b/ganga/GangaDirac/Lib/Backends/DiracBase.py
@@ -751,12 +751,15 @@ class DiracBase(IBackend):
 
         return True
 
-    def removeOutputData(self):
+    def removeOutputData(self, fileType = 'DiracFile'):
         """
         Remove all the LFNs associated with this job.
         """
         # Note when the API can accept a list for removeFile I will change
         # this.
+        if not fileType=='DiracFile':
+            return False
+
         j = self.getJobObject()
         logger.info("Removing all DiracFile output for job %s" % j.id)
         lfnsToRemove = []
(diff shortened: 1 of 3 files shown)


alexanderrichards (dismissed):

ganga/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -751,12 +751,15 @@ def getOutputSandbox(self, outputDir=None, unpack=True):
...
-    def removeOutputData(self):
+    def removeOutputData(self, fileType = 'DiracFile'):
         """
         Remove all the LFNs associated with this job.
         """
         # Note when the API can accept a list for removeFile I will change
         # this.
+        if not fileType=='DiracFile':
+            return False
+
```
egede, line 762: Should this somehow check if inheriting from `DiracFile` instead to allow for somebody to write a different type of `DiracFile`.

mesmith75: It is a good idea but I am not sure how it can be practically done as we only know the name of the file type - I am not sure that will tell you about inheritance.

egede: You are right. It is never a good idea to pass around strings for processing but that is not a failure of this piece of code.

egede (approved):

PR #1839, opened 2021-02-16
Generic substitution of illegal substrings in job names for Batch backends

### LINKED ISSUE #1838: Make substitution of illegal characters in batch job names generic
I am trying to submit Tasks (or `LHCbTask` in my case) using the Sun Grid Engine `SGE()` as the backend, but the colon in the job names are stopping the jobs being submitted. Any advice on what I should do here, other than maybe edit the source code on my end? Many thanks in advance.

https://github.com/ganga-devs/ganga/blob/f784982461cc8830160630e540a815902ab38d3f/ganga/GangaCore/GPIDev/Lib/Tasks/IUnit.py#L201

image

ryuwd:
Fixes #1838

It seems that a similar fix was implemented for `PBS()`, so this seemed like a good solution

### DIFF
--- a/ganga/GangaCore/Lib/Batch/Batch.py
+++ b/ganga/GangaCore/Lib/Batch/Batch.py
@@ -169,13 +169,18 @@ class Batch(IBackend):
             queue_option = queue_option + " " + self.extraopts
 
         if jobnameopt and job.name != '':
-            # PBS doesn't like names with spaces
             tmp_name = job.name
-            if isType(self, PBS):
-                tmp_name = tmp_name.replace(" ", "_")
-            # SGE doesn't like names with colons
-            if isType(self, SGE):
-                tmp_name = tmp_name.replace(":", "-")
+
+            # if jobnamesubstitution is set and not empty, then transform the job
+            # name to conform to requirements
+            if 'jobnamesubstitution' in self.config:
+                job_name_sub_cfg = self.config['jobnamesubstitution']
+                if len(job_name_sub_cfg) == 2:
+                    tmp_name = re.sub(*job_name_sub_cfg, tmp_name)
+                elif not len(job_name_sub_cfg) == 0:
+                    # list is not empty, and not of length 2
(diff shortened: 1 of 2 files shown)


mesmith75:
Excellent. Could you add in the whitespace replacement as well? It has been annoying me for ages.

ryuwd:
I was discussing with @egede in #1838 about using regular expressions to deal with illegal characters in job names. What we discussed there would definitely be able to deal with whitespace too since it would use regex.

If you like I could implement that in this PR, or save for a separate one and implement the whitespace replacement too for `SGE`.

mesmith75:
Whatever is easiest. It could probably be included here

ryuwd:
New `jobnamesubstitution` configurables added for `PBS`, `Slurm`, `SGE`, `LSF`.

`SGE`: whitespace and colons replaced with underscores.
`PBS`: whitespace replaced with underscores.
`Slurm`, `LSF` job names are not touched.

The regex could be configured more rigorously to replace characters not matching patterns like `[a-zA-Z0-9_-]`

egede (approved):

egede:
I tried out Slurm and it seems like just about any character was legal in the name.

ganga/GangaCore/Lib/Batch/Batch.py
```diff
@@ -169,10 +182,7 @@ def submit(self, jobconfig, master_input_sandbox):
             queue_option = queue_option + " " + self.extraopts
 
         if jobnameopt and job.name != '':
-            # PBS doesn't like names with spaces
-            tmp_name = job.name
-            if isType(self, PBS):
-                tmp_name = tmp_name.replace(" ", "_")
+            tmp_name = self._getLegalJobName()
```
mesmith75, line 185: this doesn't look right. See #1853

ryuwd: Ahh, thanks for finding this. Silly mistake on my part.

PR #1841, opened 2021-02-17
Make sure Popen used in MassStorageFile wrapper script returns stdout and stderr as string

ryuwd:
When using the `SGE()` backend with `MassStorageFile`, my Gaudi job runs successfully but fails inside the jobScript once the process is finished:
image

Setting `universal_newlines=True` in the `Popen` call makes sure that the stdout and stderr returned by Popen are returned as the expected `str` type.

In Python 3.7 onwards, `text=True` also achieves the desired result, however for backwards compatibility I've used `universal_newlines` instead which does the same thing in `>=3.7` and earlier versions.


egede (approved):

egede:
Many thanks for this.

mesmith75:
~~Does this work with interactive backends?~~ Ah I see this is just for MassStorageFile.

PR #1843, opened 2021-02-17
Ignore already downloaded files

### LINKED ISSUE #1837: ganga 8.4.5: getOutputData() throws error when there is nothing to do
The error is thrown when job(X).subjobs(Y).backend.getOutputData() is run twice. In both versions (8.4.4 and 8.4.5), the initial download works and the file is retrieved, but in ganga 8.4.5 a GangaDiracError is raised when the command is run a second time, whereas in 8.4.4 the function effectively does nothing (I guess the latter behaviour is the expected one). The error is
```
ERROR    GangaDiracError: Not all files downloaded successfully and ignoreMissing=false! Check your gangadir for missing files! len lfns: [<my lfn>], len success: []
```

mesmith75:
Fixes #1837

### DIFF
--- a/ganga/GangaDirac/Lib/Backends/DiracBase.py
+++ b/ganga/GangaDirac/Lib/Backends/DiracBase.py
@@ -817,7 +817,7 @@ class DiracBase(IBackend):
                         os.mkdir(output_dir)
                 dirac_file.localDir = output_dir
             if os.path.exists(os.path.join(dirac_file.localDir, os.path.basename(dirac_file.lfn))) and not force:
-                return
+                return dirac_file.lfn
             try:
                 dirac_file.get()
                 return dirac_file.lfn


egede (changes requested):
So code at the moment returns the lfn if the download go ahead. It also returns the lfn if the dowload has already happened. However, if the download fails with a `GangaDiracError` or `GangaFileError` it emits a warning and then returns `None`. Would it not be better to re-raise the exception again straight after the warning?

mesmith75:
So currently it downloads the rest of the files and then raises an exception at the end as the number of downloaded files doesn't match the number of LFNs so an exception does ultimately get raised.

alexanderrichards (approved):

egede (approved):

PR #1845, opened 2021-02-17
allow Batch extraopts to be changed at resubmission

ryuwd:
Try to allow `extraopts` to be changed at resubmit, which is helpful for resubmitting subjobs, etc to Batch systems with modified runtime or memory resource parameters.

I have marked this as draft as I'm not sure how `Schema` works. Am I right to increment the minor version?

### DIFF
--- a/ganga/GangaCore/Lib/Batch/Batch.py
+++ b/ganga/GangaCore/Lib/Batch/Batch.py
@@ -90,8 +90,8 @@ class Batch(IBackend):
     Kill job if it has exceeded the deadline (i.e. for your presentation)
     backend.extraopts = '-t 07:14:12:59' #Killed if not finished by 14 July before 1 pm
     """
-    _schema = Schema(Version(1, 0), {'queue': SimpleItem(defvalue='', doc='queue name as defomed in your local Batch installation'),
-                                     'extraopts': SimpleItem(defvalue='', doc='extra options for Batch. See help(Batch) for more details'),
+    _schema = Schema(Version(1, 1), {'queue': SimpleItem(defvalue='', doc='queue name as defomed in your local Batch installation'),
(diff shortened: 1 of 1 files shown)


ryuwd:
I tried this out but `Batch` resubmit method doesn't take a `backend` kwarg, so resubmitting doesn't work.

egede:
The Schema options are in effect never used, so they can actually just be ignored (some clean-up required).

ryuwd:
Adding `changable_at_resubmit=1` is enough to let me modify the `extraopts` attribute on the existing backend and let me `resubmit()` without any problems.

(Before I was trying to pass a new backend to `resubmit`, which doesn't seem to be supported for most backends anyway.)

mesmith75 (approved):

ganga/GangaCore/Lib/Batch/Batch.py
```diff
@@ -91,7 +91,7 @@ class Batch(IBackend):
     backend.extraopts = '-t 07:14:12:59' #Killed if not finished by 14 July before 1 pm
     """
     _schema = Schema(Version(1, 0), {'queue': SimpleItem(defvalue='', doc='queue name as defomed in your local Batch installation'),
-                                     'extraopts': SimpleItem(defvalue='', doc='extra options for Batch. See help(Batch) for more details'),
+                                     'extraopts': SimpleItem(defvalue='', changable_at_resubmit=1, doc='extra options for Batch. See help(Batch) for more details'),
```
mesmith75, line 94: Could you check the spelling?

mesmith75: Ah, its right.

ryuwd:
Thanks @mesmith75!

PR #1849, opened 2021-02-18
WIP: Move credential decorators for splitters

### LINKED ISSUE #1848: LHCbTask doesn't work with non-Dirac backend
In this file, as a workaround I can get LHCbTask to run on e.g. `Batch` backends, by setting `j.splitter = None`:
https://github.com/ganga-devs/ganga/blob/736cff033327a9753c44d5c71b14d15e600c5935/ganga/GangaLHCb/Lib/Tasks/LHCbUnit.py#L54

Are there any alternative splitters that can be used instead of `SplitByFiles` (which depends on `Dirac`) that work with `MassStorageFile` ? 

Many thanks!

P.S. for context, I am running private LHCb MC productions / tests, so in this case I have no need to access or use Dirac storage.

mesmith75:
Fixes #1848 

Moves the `requrie_credential` decorators to the creation of the splitter generators. Thus if `SplitByFiles` is not doing anything with Dirac it won't ask for a credential.

### DIFF
--- a/ganga/GangaDirac/Lib/Splitters/GangaSplitterUtils.py
+++ b/ganga/GangaDirac/Lib/Splitters/GangaSplitterUtils.py
@@ -3,6 +3,7 @@ from GangaDirac.Lib.Utilities.DiracUtilities import execute
 from GangaDirac.Lib.Backends.DiracUtils import result_ok
 from GangaCore.Utility.Config import getConfig
 from GangaCore.Utility.logging import getLogger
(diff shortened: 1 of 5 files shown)


egede (approved):
Very sensible change.

mesmith75:
Turns out that is broken. I'll fix it.

alexanderrichards (approved):

mesmith75:
This didn't work at all - try again.

PR #1859, opened 2021-02-24
Correctly pass Executable environment to Dirac

### LINKED ISSUE #1850: Should Executable's env var not be type dict?
Am I missing something here or should this not be type dict?

https://github.com/ganga-devs/ganga/blob/1e2877a8b6254cc3218bd2dd0b1a663532a93663/ganga/GangaCore/Lib/Executable/Executable.py#L57

alexanderrichards:
This PR fixes https://github.com/ganga-devs/ganga/issues/1850

### DIFF
--- a/ganga/GangaCore/Lib/Executable/Executable.py
+++ b/ganga/GangaCore/Lib/Executable/Executable.py
@@ -54,7 +54,7 @@ class Executable(IPrepareApp):
     _schema = Schema(Version(2, 0), {
         'exe': SimpleItem(preparable=1, defvalue='echo', typelist=[str, File], comparable=1, doc='A path (string) or a File object specifying an executable.'),
         'args': SimpleItem(defvalue=["Hello World"], typelist=[str, File, int], sequence=1, strict_sequence=0, doc="List of arguments for the executable. Arguments may be strings, numerics or File objects."),
-        'env': SimpleItem(defvalue={}, typelist=[str], doc='Dictionary of environment variables that will be replaced in the running environment.'),
+        'env': SimpleItem(defvalue={}, typelist=[dict], doc='Dictionary of environment variables that will be replaced in the running environment.'),
         'platform': SimpleItem(defvalue='ANY', typelist=[str], doc='Platform where the job will be executed, for example "x86_64-centos7-gcc8-opt"'),
         'is_prepared': SimpleItem(defvalue=None, strict_sequence=0, visitable=1, copyable=1, hidden=0, typelist=[None, ShareDir], protected=0, comparable=1, doc='Location of shared resources. Presence of this attribute implies the application has been prepared.'),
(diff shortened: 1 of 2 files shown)


egede (dismissed):

mesmith75 (dismissed):

ganga/GangaCore/Lib/Executable/Executable.py
```diff
@@ -54,7 +54,7 @@ class Executable(IPrepareApp):
     _schema = Schema(Version(2, 0), {
         'exe': SimpleItem(preparable=1, defvalue='echo', typelist=[str, File], comparable=1, doc='A path (string) or a File object specifying an executable.'),
         'args': SimpleItem(defvalue=["Hello World"], typelist=[str, File, int], sequence=1, strict_sequence=0, doc="List of arguments for the executable. Arguments may be strings, numerics or File objects."),
-        'env': SimpleItem(defvalue={}, typelist=[str], doc='Dictionary of environment variables that will be replaced in the running environment.'),
+        'env': SimpleItem(defvalue={}, typelist=[dict], doc='Dictionary of environment variables that will be replaced in the running environment.'),
```
alexanderrichards: seems that this has to be `str` since in the schema `_check_type` method it iterates over dicts and checks the key and value are the type specified in `typelist`...see below

https://github.com/ganga-devs/ganga/blob/b9dca4b3e069d2300e50b8757b90aea7f3dff085/ganga/GangaCore/GPIDev/Schema/Schema.py#L548-L551

mesmith75: Seems odd. So how do you specify the environment? Is it a string representation of a dict?

alexanderrichards: no looking at the tests it's a dict. but just the schema checking has a special case for dicts where it checks the content against the `typelist`

alexanderrichards: I agree it's confusing though

mesmith75 (approved):

mesmith75:
Let's merge this if it is ready. I am running a lot of Dirac executable jobs so I can give it a test before we release it.

ganga/GangaCore/Lib/Executable/Executable.py
egede: For the Dirac backend object, I notice that we simply do not have a `typelist`. https://github.com/ganga-devs/ganga/blob/b9dca4b3e069d2300e50b8757b90aea7f3dff085/ganga/GangaDirac/Lib/Backends/DiracBase.py#L112

egede (approved):

ganga/GangaCore/Lib/Executable/Executable.py
alexanderrichards: interesting... I'm guessing in that case it just does no type checking

PR #1861, opened 2021-03-03
Add archive checker to BKQuery

### LINKED ISSUE #1860: Check for archived paths
LHCb is archiving its old MC data sets. These will still appear in the BK so when people try and run jobs on them ganga will throw a replica error. Perhaps we can try and catch this and see if there LFNs in the archive SEs and raise a nice helpful error instead.

mesmith75:
Fixes #1860 

This operates in the BKQuery rather than the splitter. It gets the list of LFNs and replicas and checks if the SEs are all archives. If they are it raises an error. I added some options to not do the check.

It also makes a guess that the supplied query path is MC. If it is not it won't do the check as we really only worry about the archived MC data sets and it could take a long time for the large sets of collision data. That should catch most instances - people with more complex queries will not be caught but hopefully they are more aware of what they are doing.

The getReplicas should probably be broken up into chunks.

### DIFF
--- a/ganga/GangaDirac/Lib/Server/DiracCommands.py
+++ b/ganga/GangaDirac/Lib/Server/DiracCommands.py
@@ -404,7 +404,11 @@ def getServicePorts():
     ''' Get the service ports from the DiracAdmin based upon the Dirac config'''
     return DiracAdmin().getServicePorts()
 
-
+@diracCommand
+def isSEArchive(se):
+    ''' Ask if the specified SE is for archive '''
+    from DIRAC.DataManagementSystem.Utilities.DMSHelpers import DMSHelpers
+    return DMSHelpers().isSEArchive(se)
 
 @diracCommand
 def getSitesForSE(se):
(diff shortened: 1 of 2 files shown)


egede (commented):
I am a bit confused. Will this abort the submission if just some of the data is archived or will it only abort if *all* the data is archived?

mesmith75:
When you do a `BKQuery.getDataset()` it raises the exception if *all* the data is archived. If only some of the data is inaccessible it will get pulled up by the splitter at a later stage (the splitter only gets replicas available for jobs) but there you have the option to `ignoremissing`

egede:
> When you do a `BKQuery.getDataset()` it raises the exception if _all_ the data is archived. If only some of the data is inaccessible it will get pulled up by the splitter at a later stage (the splitter only gets replicas available for jobs) but there you have the option to `ignoremissing`

All good.

egede (dismissed):

alexanderrichards (dismissed):

mesmith75:
Just adjusted the options so you can not do any check, or do the check and only raise a warning if they are only archived.

egede (approved):

PR #1867, opened 2021-03-16
Change Dirac CPU time default

egede:


egede:
@alexanderrichards @mesmith75 Can you try to take a look at the failing tests in GangaGUI. When I test locally they pass. Not really sure what is going on ...

egede:
The change I made elsewhere is a completely trivial change, so that is certainly not a trigger.

alexanderrichards:
> The change I made elsewhere is a completely trivial change, so that is certainly not a trigger.

Yes this failure is completely unrelated. I'll take a look what's causing it.

alexanderrichards (dismissed):

egede:
OK, problem with failing GUI tests fixed. It was caused by the `SQLAlchemy` package moving from version `1.3.19` to `1.4.1`. The interesting thing is that this is not a direct dependency for us (and the `Flask` package version we use has not changed). In any case, I added it as an explicit dependency and pinned the version. Not clear to me if this is a problem in the way we use Flask or if it is a problem with Flask itself.

mesmith75 (approved):

alexanderrichards:
nice one, you beat me to it 🎉

alexanderrichards (approved):

PR #1869, opened 2021-03-20
Sample removal of logger.debug in loop

### LINKED ISSUE #1819: Trace down candidates for expensive debug statements to remove
In several places of the code, there are `debug` calls to the logger that are inside loops and/or cause expensive evaluations. As the statement is fully evaluated whether or not the log message is printed this is poor practise. The following needs to be done

- Identify `debug` statements that are either in loops or that have expensive evaluation (so just about anything beyond a simple string)
- Consider if it can be simply removed
- If staying in place, protect statement by checking for log level before evaluating.

gurudattapatil:
Is this a expensive debug statement? So i can move on with the issue #1819


egede:
No, I think this one is fine to keep. The default value of `AutoFlusherWaitTime` is 30, so this one will only get a call every 30 seconds.

egede (approved):
All good. Many thanks.

PR #1878, opened 2021-03-30
Update socket.io to latest version

Pushkar-s:
CLI seems to be stuck.  
Giving error:  ```The client is using an unsupported version of the Socket.IO or Engine.IO protocols.....```
Note: updating this ensures CLI works on safari. For other browsers, the web app still crashes.

### DIFF
--- a/ganga/GangaGUI/gui/static/js/slidecli.js
+++ b/ganga/GangaGUI/gui/static/js/slidecli.js
@@ -11,6 +11,6 @@ document.addEventListener('DOMContentLoaded', () => {
 // Show/Hide CLI
 let cliOpen = false;
 function toggleCli() {
-    document.querySelector('#cli-container').style['display'] = cliOpen ? 'none' : 'initial';
+    document.querySelector('#cli-container').style['visibility'] = cliOpen ? 'hidden' : 'initial';
     cliOpen = !cliOpen;
 }
--- a/ganga/GangaGUI/gui/templates/base.html
+++ b/ganga/GangaGUI/gui/templates/base.html
@@ -107,7 +107,7 @@
 
         {# Web CLI Container #}
         <div id="cli-container"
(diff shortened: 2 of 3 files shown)


egede:
I am hesitant in merging something that moves the CLI interface from being "broken" to be "partially broken".

mesmith75:
Which bit exactly is not working? How does one reproduce the problem?

egede:
If you start `ganga-gui` and then click `CLI` to the top right, it opens up an xterm inside your browser window. However, the terminal hangs and never gives a prompt.

Pushkar-s:
After I changed the display property CLI starts working for chrome and stops working for safari. 
If div(#cli-container) is not rendered along with other elements it doesn't work for chrome also I get an error in the browser console that 'cli.js' is not loaded in resources.
so removing the display property ensures that it is rendered along with other elements. 
I find it weird that this doesn't seem to reciprocate for safari. 'slidecli.js' goes missing from resources for safari after this change. 
Now,
Removing 'defer' from the script tag of 'slidecli.js' makes it work for safari without affecting chrome.
I am not entirely sure but this could be an issue due to the difference in javascript engines used by chrome and safari.

mesmith75:
Ah yes, that is not working. It worked when I opened it in a new tab, but that consumed all the CPU and crashed the browser.

Pushkar-s:
Looks like using visibility property instead of display completely fixes the issue. The problem was with rendering content I guess, Unlike visibility property display set to none considers the element not present in DOM.
The previous fix wasn't good as every time a page reloads the CLI pops up.

egede:
With the latest fix, I can get the CLI window to open and close in both firefox and chrome. However, I still can't type any commands into it - it just shows a line of text and the top and it is then black ... I wonder if we are using the correct tool here?

Pushkar-s:
Did you also include the first commit of updating the socketio version to the latest one in the cli.html file? since that is also required for the CLI to work.

egede:
> Did you also include the first commit of updating the socketio version to the latest one in the cli.html file? since that is also required for the CLI to work.

Hmm, I just made a fresh clone of Ganga and still see everything hanging when starting the CLI.

PR #1884, opened 2021-05-10
Clarify install instructions.

egede:


egede:
The documentation at [ReadTheDocs](https://ganga.readthedocs.io/en/latest/UserGuide/InstallAndBasicUsage.html) doesn't update automatically (I thought it would). So after this merge an update should be triggered.

alexanderrichards (approved):
Looks good to me

mesmith75 (approved):

PR #1883, opened 2021-05-05
Change splitter for subjobs

acampove:
Hi,

**Problem:** Given a job that has e.g. 100 subjobs each subjob might process 20 files. If one file is broken or unavailable in a specific site the whole 20 files won't be processed.

**Solution:** Allow for subjobs to be _resplit_ in such a way that the subjob will create 20 sub-subjobs, each processing one file. This should allow for the backend of the subjobs to be changed and therefore a different site to be used.

**Code:**

This fork allows to do:

```python
#It should work for any splitter
spl=ArgSplitter()
spl.args=[[1],[2],[3]]

#Pick up some job and subjob.
j=jobs(123)
sj=j.subjob[2]
sj.resubmit(splitter=spl)
```

The last line should allow the master job to create extra subjobs according to the `spl` splitter and submit them. The naming of the new subjobs should be as 

`123.2.0`
`123.2.1`
`123.2.2`

etc.

**Potential problem:** Although the jobs do run and finish, I do not seem to be able to run the monitoring on them by doing:

```python
j.jobs(997)
sj=j.subjobs[1]
slc=sj.subjobs
runMonitoring(slc)
```

image

Am I doing anything incorrectly or is this a problem in the code that needs to be fixed?

Cheers.

PR #1891, opened 2021-05-15
Add resplit for subjob

### LINKED ISSUE #1669: Enable resubmission of subjobs with a new splitter
If a job is created with a splitter, it will create multiple subjobs. If some of the subjobs fail, it is possible to resubmit them, but not with a modification of the splitter. It can be very inefficient to resubmit the entire job with a new splitter. It should be possible to take a number of subjob and essentially run a new splitter on it. This might then create a number of subjobs that replace the existing one.

Imagine that we have a job where it carries out some calculation for a given integer parameter. We make a splitter that parts them such that each subjob does three calculations.

```
j = Job(splitter=ArgSplitter([[1,2,3],[4,5,6],[7,8,9]])
j.submit()
```
Now it turns out that the subjob wirh `[7,8,9]` failed because it ran out of some resource (memory, CPU time, ...) but it can be fixed by splitting it up into three separate pieces. At the moment I can only do that by creating a copy of the subjob and re-split.

```
j2 = j.subjobs[2].copy()
j2.splitter = ArgSplitter([[7],[8],[9]])
j2.submit()
```
While this works, it is unfortunate as it destroys the bookkeeping part (I now have to remember that my calculation is in both `j` and `j2`. If instead I could do

```
j.subjobs[2].splitter = ArgSplitter([[7],[8],[9]])
j.subjobs[2].submit()
```
I would end up with a single jobs object `j`, that now has 5 subjobs (I removed one by resplitting and created 3 new ones). Bookkeepoing would be much better as a result.

mesmith75:
Fixes #1669 , inspired by #1883 .

Looks like:
```python
[14:17:25]                                                                                                                                                                         
Ganga In [7]: newsplitter = ArgSplitter(args = newargs)                                                                                                                            

[14:17:33]                                                                                                                                                                         
Ganga In [8]: jobs(14).subjobs(0).resplit(newsplitter)                                                                                                                             
INFO     removing: /tmp/masmith/tmpejoldzk2
INFO     removing: /tmp/masmith/tmp02bfpat6
INFO     job 14 status changed to "running"
INFO     Job 14 Running PostProcessor hook
INFO     job 14 status changed to "completed"
INFO     removing: /tmp/masmith/tmpfqb7xcix
INFO     Re-splitting Job: 14.0
INFO     submitting 2 subjobs
INFO     Preparing subjobs
INFO     SubJobConfigs: 2
INFO     rjobs: 2
INFO     submitting job 14.3 to Local backend
INFO     job 14 status changed to "submitting"
INFO     job 14 status changed to "submitted"
INFO     submitting job 14.4 to Local backend
INFO     job 14 status changed to "submitting"
INFO     job 14 status changed to "submitted"
Ganga Out [8]: 1

[14:17:48]                                                                                                                                                                         
Ganga In [9]: jobs(14).subjobs                                                                                                                                                     
Ganga Out [9]: 
Registry Slice: jobs(14).subjobs (5 objects)
--------------
    fqid |    status |      name | subjobs |    application |        backend |                             backend.actualCE |                       comment |  subjob status 
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
    14.0 |completed_ |           |         |     Executable |      Localhost |                           lxplus7103.cern.ch |            - has been resplit |            --- 
    14.1 | completed |           |         |     Executable |      Localhost |                           lxplus7103.cern.ch |                               |            --- 
    14.2 | completed |           |         |     Executable |      Localhost |                           lxplus7103.cern.ch |                               |            --- 
    14.3 | completed |           |         |     Executable |      Localhost |                           lxplus7103.cern.ch |             - resplit of 14.0 |            --- 
    14.4 | completed |           |         |     Executable |      Localhost |                           lxplus7103.cern.ch |             - resplit of 14.0 |            --- 
```

### DIFF
--- a/doc/UserGuide/Splitters.rst
+++ b/doc/UserGuide/Splitters.rst
@@ -105,3 +105,34 @@ The ``GangaDatasetSplitter`` is provided as an easy way of splitting over a numb
 
 
 If you check the output you will see the list of files that each subjob was given using ``j.subjobs()`` as above.
+
+Resplitting a job
+-----------------
+Sometimes a job that has been split will have some of the subjobs failing. This might for example be due to that they run out of CPU time and are required to be split into smaller units. To support this, it is possible to apply a new splitter to a subjob which is in the ``completed`` or ``failed`` state. In the example below can be seen how the first subjob is subsequently split into a further two subjobs. 
+
+.. code-block:: python
+    j = Job(splitter=ArgSplitter(args=[ [0],[0] ]))
+    j.submit()
+
+    # wait for jobs to complete
+    j.subjobs
+    
+    Registry Slice: jobs(8).subjobs (2 objects)
+    --------------
+        fqid |    status       |                       comment |
+    -------------------------------------------------------
+         8.0 | completed       |                               |
+         8.1 | completed       |                               |
+    
+    j.subjobs(0).resplit(ArgSplitter(args=[ [1], [1] ]))
+
+    # wait for jobs to complete
+    j.subjobs
+    
+    --------------
+        fqid |    status       |                       comment |
+    -------------------------------------------------------
+         8.0 |completed_failed |            - has been resplit |
+         8.1 | completed       |                               |
+         8.2 | completed       |              - resplit of 8.0 |
+         8.3 | completed       |              - resplit of 8.0 |
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -1531,7 +1531,6 @@ class Job(GangaObject):
             logger.info('submitting %s subjobs', len(rjobs))
         
             # Output Files
-            # validate the output files
             for this_job in rjobs:
                 (validOutputFiles, errorMsg) = this_job.validateOutputfilesOnSubmit()
                 if not validOutputFiles:
@@ -1546,14 +1545,9 @@ class Job(GangaObject):
             logger.debug("Job Configuration, Job %s:" % self.getFQID('.'))
 
             # prepare the master job with the correct runtime handler
-            # Calls rtHandler.master_prepare if it hasn't already been called by the master job or by self
-            # Only stored as transient if the master_prepare successfully
-            # completes
             jobmasterconfig = self._getJobMasterConfig()
 
             # prepare the subjobs with the runtime handler
-            # Calls the rtHandler.prepare if it hasn't already been called by the master job or by self
-            # Only stored as a transient if the prepare successfully completes
             jobsubconfig = self._getJobSubConfig(rjobs)
             logger.debug("# jobsubconfig: %s" % len(jobsubconfig))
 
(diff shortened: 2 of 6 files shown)


ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -1457,6 +1477,111 @@ def _doSplitting(self):
...
+            self.monitorPrepare_hook(jobsubconfig)
+
+            # submit the job
+            # master_submit has been written as the interface which ganga
+            # should call, not submit directly
+
+            r = self.backend.master_submit( rjobs, jobsubconfig, jobmasterconfig)
+
+            if not r:
+                raise JobManagerError('error during submit')
```
egede: Should we remove the newly created subjobs again if there is an error in the submission?

egede (changes requested):
Great to see this work. Some care will still have to be taken by a use to merge the correct output at the end, but I do not see that there is too much we can do about that.

- We should check that the initial subjob is in either a `completed` or a `failed` state. Otherwise a delayed error (and potentially an undefined state with new subjobs that are not submitted) would arise when freezing the original subjob.
- A test case should be created.

ganga/GangaCore/GPIDev/Lib/Job/Job.py
alexanderrichards: probably yes, as long as we have the original they were split from

egede:
Need to also check how this works with prepared applications.

ganga/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -166,10 +166,10 @@ def prepareCommand(app):
                                                 getName(opts_file))
 
     #Check if this was checked out with LbEnv or not
-    isLbEnv = False
-    with open(app.directory+'/Makefile', "r") as makefile:
-        if 'LbEnv' in makefile.read():
-            isLbEnv = True
+    isLbEnv = True
```
mesmith75: This fixes #1897 . I am not sure this ever worked and by now LbEnv has been mandatory for so long it shouldn't be an issue to remove the check. It breaks the resubmission with a prepared application as it looks in the build directory, which may not exist anymore if the application has been submitted already.

ganga/GangaLHCb/Lib/RTHandlers/GaudiExecRTHandlers.py
```diff
@@ -165,13 +165,7 @@ def prepareCommand(app):
...
                                                 getName(opts_file))
 
-    #Check if this was checked out with LbEnv or not
-    isLbEnv = False
-    with open(app.directory+'/Makefile', "r") as makefile:
-        if 'LbEnv' in makefile.read():
-            isLbEnv = True
-
-    sourceEnv = app.getWNEnvScript(isLbEnv)
+    sourceEnv = app.getWNEnvScript(True)
```
mesmith75, line 168: This fixes #1897 . LbEnv has been mandatory for so long now that this shouldn't be an issue.

egede (dismissed):

egede (approved):
Added some tests and documentation.

doc/UserGuide/Splitters.rst
```diff
@@ -105,3 +105,34 @@ The ``GangaDatasetSplitter`` is provided as an easy way of splitting over a numb
...
+    
+    j.subjobs(0).resplit(ArgSplitter(args=[ [1], [1] ]))
+
+    # wait for jobs to complete
+    j.subjobs
+    
+    --------------
+        fqid |    status       |                       comment |
+    -------------------------------------------------------
+         8.0 |completed_failed |            - has been resplit |
```
mesmith75, line 135: Is this right? It should be `completed_frozen`. Maybe there is a typo in the code.

egede: I just did a copy and paste from the screen ....

PR #1892, opened 2021-05-18
First steps towards supporting Python 3 LHCbDIRAC installations

chrisburr:
This is the first few things I've noticed when trying to use ganga with a Python 3 based installation of LHCbDIRAC. I'm probably not going to find time to fully check and finish this PR but hopefully this is useful nevertheless.

To test with Python 3 yourself you currently need to change to `/cvmfs/lhcb.cern.ch/lib/LbEnv-unstable` in `LHCbDIRACenv.py` and change `LHCbDiracVersion = v10.2.0a8` in your `~/gangarc`.

There is at lease one bug on the LHCbDIRAC side which makes bookkeeping queries fail and will be fixed by: https://gitlab.cern.ch/lhcb-dirac/LHCbDIRAC/-/merge_requests/1005

### DIFF
--- a/ganga/GangaDirac/Lib/Backends/DiracBase.py
+++ b/ganga/GangaDirac/Lib/Backends/DiracBase.py
@@ -183,7 +183,7 @@ class DiracBase(IBackend):
         self.extraInfo = None
         self.statusInfo = ''
         j.been_queued = False
-        dirac_cmd = """execfile(\'%s\')""" % dirac_script
+        dirac_cmd = """exec(open(\'%s\').read())""" % dirac_script
 
         try:
             result = execute(dirac_cmd, cred_req=self.credential_requirements)
@@ -224,7 +224,7 @@ class DiracBase(IBackend):
         self.extraInfo = None
         self.statusInfo = ''
         j.been_queued = False
-        dirac_cmd = """execfile(\'%s\')""" % myscript
+        dirac_cmd = """exec(open(\'%s\').read())""" % myscript
         submitFailures = {}
         try:
             result = execute(dirac_cmd, cred_req=self.credential_requirements, return_raw_dict = True, new_subprocess = True)
--- a/ganga/GangaDirac/Lib/Server/DiracCommands.py
+++ b/ganga/GangaDirac/Lib/Server/DiracCommands.py
@@ -261,8 +261,8 @@ def finished_job(id, outputDir=os.getcwd(), unpack=True, oversized=True, noJobDi
 def finaliseJobs(inputDict, statusmapping, downloadSandbox=True, oversized=True, noJobDir=True):
     ''' A function to get the necessaries to finalise a whole bunch of jobs. Returns a dict of job information and a dict of stati.'''
     returnDict = {}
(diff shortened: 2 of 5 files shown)


egede:
Hmm, for some readon I can't checkout this PR
```
[desktop] ~/.../ganga/src/ganga develop  1% gh pr checkout 1892
fatal: couldn't find remote ref refs/heads/py3-dirac
exit status 128
```
Any ideas?

chrisburr:
No idea...I've never seen `gh` fail but as it has I'd be tempted to use regular git.

Assuming you have SSH auth set up:
```bash
git remote add chrisburr git@github.com:chrisburr/ganga.git
git fetch --all
git checkout py3-dirac
```

alexanderrichards:
> Hmm, for some readon I can't checkout this PR
> 
> ```
> [desktop] ~/.../ganga/src/ganga develop  1% gh pr checkout 1892
> fatal: couldn't find remote ref refs/heads/py3-dirac
> exit status 128
> ```
> 
> Any ideas?

It's because it's not in our repo but Chris'

chrisburr:
> It's because it's not in our repo but Chris' and you don't have access to it.

`gh` should take care of adding the required remotes automatically

alexanderrichards:
> > It's because it's not in our repo but Chris' and you don't have access to it.
> 
> `gh` should take care of adding the required remotes automatically

The message it reports is that it's looking for the py3-dirac ref in the ganga repo which is correct as it's from your fork. I doubt it will automatically pull from there.

chrisburr:
> caveat : I don't use gh only regular git so I could be wrong on this but it's at least behaving seemingly as I would have expected

It can. I switch between PRs from forks a lot so it's the main reason I use `gh`. For example, 5144 is a PR from someone else's fork that I don't have write access to:
Screenshot 2021-05-19 at 09 45 08
(technically it's not adding remotes to do this, but that's an GitHub implementation detail)

That said, as it's not working, a vanilla git equivalent is in https://github.com/ganga-devs/ganga/pull/1892#issuecomment-843798841.

alexanderrichards:
good to know! 😉

egede:
My `gh` problems seems a red herring. I made a fresh clone of `ganga` and in that one `gh pr checkout 1892` worked without a problem. Who knows what was wrong with the other one.

egede:
@chrisburr Can you give me write access to the branch for this PR?

chrisburr:
You should already have access? If you're having trouble say and I'll invite you to my fork.
Screenshot 2021-05-20 at 07 54 47

chrisburr:
The latest version of the Python 3 client can now be requested for the production instance of LHCbDIRAC by using `lb-dirac prod.py3 ...` or, in the case of ganga, by setting `LHCbDiracVersion = prod.py3` in the gangarc.

I've had to make a small change to the logic used for setting up the DIRAC environment but otherwise this PR appears to be working for my tests.

egede:
Got merged as #1920

chrisburr:
Thanks @egede!

PR #1896, opened 2021-05-21
fix executable RT handler

mesmith75:


alexanderrichards (approved):
This was something I reported a while ago. Was there an issue relating to this that needs closing or did I forget to make one?

mesmith75:
You fixed the handler in GangaCore. LHCb has its own version.

alexanderrichards:
> You fixed the handler in GangaCore. LHCb has its own version.

you remember better what I've done than I do 🤣

PR #1899, opened 2021-05-28
Condor tail addition

gurudattapatil:
i have checked the working of this on simple condor jobs and seems to be working fine for it.

### DIFF
--- a/ganga/GangaCore/Lib/Condor/Condor.py
+++ b/ganga/GangaCore/Lib/Condor/Condor.py
@@ -797,4 +797,37 @@ class Condor(IBackend):
     updateMonitoringInformation = \
         staticmethod(updateMonitoringInformation)
 
+        job = self.getJobObject()
+
+        if not self.id:
+            logger.warning("Job %s not running" % job.id)
+            return False
+
+        idElementList = job.backend.id.split("#")
+        if 3 == len(idElementList):
+            if idElementList[1].find(".") != -1:
+                peekCommand = "condor_tail -f -name %s %s" % \
+                              (idElementList[0], idElementList[1])
+            else:
+                peekCommand = "condor_tail -f -name %s %s" % \
+                              (idElementList[0], idElementList[2])
+        else:
+            peekCommand = "condor_tail -f %s" % (idElementList[0])
+
+        status, output = subprocess.getstatusoutput(peekCommand)
+
+
+        if (status != 0):
+            logger.warning \
+                ("Return code '%s' peek job '%s' - Condor id '%s'" %
+                 (str(status), job.id, job.backend.id))
+            logger.warning("Tried command: '%s'" % peekCommand)
+            logger.warning("Command output: '%s'" % output)
+        else:
+            logger.info(output)
+
+        peekStatus = True
+
+        return peekStatus
+
 #_________________________________________________________________________
--- a/ganga/GangaCore/Lib/Condor/Condor.py
+++ b/ganga/GangaCore/Lib/Condor/Condor.py
@@ -797,6 +797,7 @@ class Condor(IBackend):
     updateMonitoringInformation = \
         staticmethod(updateMonitoringInformation)
 
+    def peek(self, filename=None, command=None):
         job = self.getJobObject()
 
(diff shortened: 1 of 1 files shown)


ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,38 @@ def timedetails(self):
     updateMonitoringInformation = \
         staticmethod(updateMonitoringInformation)
 
+    def peek(self, filename=None, command=None):
+        job = self.getJobObject()
+
+        if not self.id:
+            logger.warning("Job %s not running" % job.id)
+            return False
```
egede: We should use the job status to decide on this. But is the check actually required? Don't we only end up here if the job is in the `running` state?
```suggestion
        if job.status!='running':
            logger.warning("Job %s not running" % job.id)
            return False
```

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,38 @@ def timedetails(self):
...
+
+        idElementList = job.backend.id.split("#")
+        if 3 == len(idElementList):
+            if idElementList[1].find(".") != -1:
+                peekCommand = "condor_tail -f -name %s %s" % \
+                              (idElementList[0], idElementList[1])
```
egede: I suggest to use `f` strings instead as they are more readable. Same change in several other places in code.

```suggestion
                peekCommand = f"condor_tail -f -name {idElementList[0]} {idElementList[1]}"
```

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,38 @@ def timedetails(self):
     updateMonitoringInformation = \
         staticmethod(updateMonitoringInformation)
 
+    def peek(self, filename=None, command=None):
+        job = self.getJobObject()
+
```
egede, line 802: What happens if you call `peek` on a master job (where there is no actual job running, so nothing to `peek`). You need to check if your `job` object has subjobs, and if yes, return `False`.

egede (changes requested):

ganga/GangaCore/Lib/Condor/Condor.py
gurudattapatil: I will remove this as it may not be required.

ganga/GangaCore/Lib/Condor/Condor.py
gurudattapatil: Done

ganga/GangaCore/Lib/Condor/Condor.py
gurudattapatil: Done

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,38 @@ def timedetails(self):
...
+            logger.warning("Tried command: '%s'" % peekCommand)
+            logger.warning("Command output: '%s'" % output)
+        else:
+            logger.info(output)
+
+        peekStatus = True
+        if job.subjobs:
+            peekStatus= False
+            logger.error("Master Job does not have a peek status.")
+
```
egede: Should this not be up at the top at line 803 instead? Code also seems a bit inconsistent in if for a failure returns

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,38 @@ def timedetails(self):
...
+        status, output = subprocess.getstatusoutput(peekCommand)
+
+
+        if (status != 0):
+            logger.warning \
+                ("Return code '%s' peek job '%s' - Condor id '%s'" %
+                 (str(status), job.id, job.backend.id))
+            logger.warning("Tried command: '%s'" % peekCommand)
+            logger.warning("Command output: '%s'" % output)
```
egede, line 823: This should cause `peekStatus = False`

egede (changes requested):

Issue #1733, opened 2020-07-20
Limit ganga workers for Local backend


asnaylor:
I would like to be able to limit the new of jobs running on my local machine when using ganga. I was testing the `Local()` backend with `GenericSplitter()` and it spawn new processes until it couldn't anymore. I ran a job which had 30 files and it ran 30 instances of the executable script. Is there any way to control the number of concurrent processes without using `Task` objects?

egede:
This is currently not possible. Not a completely trivial change as it would require some sort of queue to be implemented for Local jobs

asnaylor:
Is this something the ganga devs are interested in adding? If so what would be the timeframe?

egede:
I have labelled this as a "good first issue" as it is quite decoupled from the rest of Ganga and could be picked up by an external contributor. I do not see it done by anybody from the core team in the short term. I think the best implementation would be by implementing the bulk submit function. That could then spawn an independent process that would limit the number of simultaneous subjobs completely independent of Ganga.

egede:
The number of jobs running simultaneously should be a configurable in the `[Local]` section of the configuration. Maybe the default value could be `auto` but a fixed number like `8` could be set as well. The meaning of `auto` would be that it is limited to the number of available cores.

asnaylor:
Thanks, this is great 👍🏻

PR #1900, opened 2021-05-30
Added batch_submit (#1733)

monalisha31:
Runs a specific number of subjobs at a time. The number of subjobs is decided by the user.  The changes were made in the master_submit function of Localhost.py
A simple test case:

```bash
j = Job()
j.splitter = GenericSplitter()
j.splitter.attribute = 'application.args'
j.splitter.values = [['hello', 1], ['world', 2], ['again', 3], ['again', 4], ['again', 5], ['again', 6], ['again', 7], ['again', 8], ['again', 9], ['again', 10],['again', 11]]
j.backend.batch_submit = 2
j.submit()
```


egede:
There is a misunderstanding of purpose here. The code in this PR submits the subjobs in groups but that is not what is required. What is required that only a specific number of jobs **run** at a given time. So if I have a job like

```python
j = Job()
j.application.exe='sleep'
j.splitter = GenericSplitter()
j.splitter.attribute = 'application.args'
j.splitter.values=[ [10] for i in range(20) ]
j.backend.batch_submit = 2
```
I should see that the whole job should take about 100 seconds to complete (20 subjobs sleeping 10 seconds each, but only 2 subjobs running in parallel).

monalisha31:
I misunderstood the parallel submit of subjobs parts before but now it's all clear. Thank you for the clarification. I will make the changes in the code.

egede:
Replaced by #1907

PR #1901, opened 2021-05-31
Condor tail addition clone

### LINKED ISSUE #734: Implement peek for running jobs on the Condor backend
Hi,

Not sure if this is already a feature of ganga, but it would be great if there was an easy way from to invoke 'condor_tail -f <jid>' from within the ganga prompt for a running condor job. Ideally doing "job.peek('stdout')" for a running condor job would execute this tailing the stdout!

Cheers
Will

gurudattapatil:


egede:
@mesmith75 Can I get you to try out this on the HTCondor system at CERN.

mesmith75:
Sure

mesmith75:
I am not sure this is working - it just seems to hang for me.
```python
[23:29:11]                                                                                                                                                                         
Ganga In [3]: jobs(694).subjobs(0).peek()                                                                                                                                          
```

mesmith75:
I don't think you want the -f option - it means you have a continuous follow of the stdout so the subprocess doesn't exit until the job does. Really you just want to peak at that moment in time. I also think it may be worth having a slightly longer buffer. 1024 bytes is not a lot. Eg for one of my fits I only get a couple of lines of readout:
```sh
condor_tail 5275280.0
al::init(partReco_2420_2016_Int[DststM,cosThetah]) using numeric integrator RooIntegrator1D to calculate Int(DststM)
[#1] INFO:NumericIntegration -- RooRealIntegral::init(bkgPDF_2016_bin4_Int[DststM,cosThetah]) using numeric integrator RooIntegrator1D to calculate Int(DststM)
[#1] INFO:NumericIntegration -- RooRealIntegral::init(ssmuPDF_2016_bin4_Int[DststM,cosThetah]) using numeric integrator RooIntegrator1D to calculate Int(DststM)
[#1] INFO:NumericIntegration -- RooRealIntegral::init(DstD_2420_2016_Int[DststM,cosThetah]) using numeric integrator RooIntegrator1D to calculate Int(DststM)
[#1] INFO:Minization -- RooNLLVar::evaluatePartition(2016_sig_bin4) first = 14947 last = 17438 Likelihood offset now set to 6645.98
[#1] INFO:NumericIntegration -- RooRealIntegral::init(ssmuPDFForFit_2016_bin4_Int[DststM,cosThetah]) using numeric integrator RooIntegrator1D to calculate Int(DststM)
[#1] INFO:Minization -- RooNLLVar::evaluatePartition(2016_ssmu_bin4) first = 314 last = 366 Likelihood offset now set to 36.5174
```

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,37 @@ def timedetails(self):
     updateMonitoringInformation = \
         staticmethod(updateMonitoringInformation)
 
+    def peek(self, filename=None, command=None):
```
mesmith75, line 800: So apparently `condor_tail` allows you to peak any filename in the `transfer_output_files` list. I wonder if it is worth giving the option to users to specify a filename here.

Also I would fill in the command to include some options so maybe a user can set the buffer length or peak the stderr

gurudattapatil:
Is the hardcoded maxbytes value ok or should it be available to change?

gurudattapatil:
> I am not sure this is working - it just seems to hang for me.
> 
> ```python
> [23:29:11]                                                                                                                                                                         
> Ganga In [3]: jobs(694).subjobs(0).peek()                                                                                                                                          
> ```

Did this resolve or still is a problem?

mesmith75:
Yes, now it doesn't hang.

The `maxbytes` should definitely be configurable - what you have now is a little long. You can add an option to the Config.

It should be set in Ganga's global configuration (i.e. what gets set in the `~/.gangarc`). An example of how to get to the configuration is here: https://github.com/ganga-devs/ganga/blob/9929fbde2e328d2b108f4cbb89e2721ad9f21e58/ganga/GangaCore/Lib/Condor/Condor.py#L536 

You will need to add the necessary option to when Ganga starts here: https://github.com/ganga-devs/ganga/blob/9929fbde2e328d2b108f4cbb89e2721ad9f21e58/ganga/GangaCore/__init__.py#L422

gurudattapatil:
I added the config for condor_tail -maxbytes and set default to 1Mb

ganga/GangaCore/__init__.py
```diff
@@ -424,6 +424,7 @@ def getLCGRootPath():
 
 condor_config.addOption('query_global_queues', True,
                  "Query global condor queues, i.e. use '-global' flag")
+condor_config.addOption('MaxBytes', "1000000", 'Max bytes option for peek option for Condor backend ')
```
mesmith75: Could you make the docstring a little more helpful
```suggestion
condor_config.addOption('MaxBytes', "1000000", 'Set maximum bytes to display with peek()')
```

mesmith75:
Ok this almost looks ready. Just the adjustment to the docstring. As I say I think it would also be helpful to make use of the filename if requested.

gurudattapatil:
I have done the doc string change, I will look into the filename use

gurudattapatil:
Done adding the filename to the condor_tail command

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -797,4 +797,40 @@ def timedetails(self):
...
+        job = self.getJobObject()
+        
+        queryCommand= " ".join\
+            (["-maxbytes" if getConfig("MaxBytes") else "1000000"])
+        if job.subjobs:
+            logger.error("Master Job does not have a peek status.")
+            peekStatus= False
+        name
```
mesmith75: This seems errant.

gurudattapatil: Oh sorry, that line shouldn't be there!

gurudattapatil:
Is there a reason for two tests failing?

egede:
See #1909 for why this is failing. HAs to wait for that one to be merged.

egede (approved):

PR #1902, opened 2021-05-31
finding free port for starting flask server

gurudattapatil:


egede (dismissed):
Looks good.

gurudattapatil:
Should I merge this?

egede:
> Should I merge this?

Yes, you are welcome.

alexanderrichards:
sorry for the delay (it was a bank holiday here in the UK on Monday). The only question I would have is that these ports were clearly hard coded (by default) to specific values. How will the clients know which new port number to use?

Clearly this is for the start_gui function so I'm not so worried about `gui_port` as the user will probably see something on his/her console when starting it but what about `internal_port`. I don't know what the client for `internal_port` is but are we sure that it will still function ok if the port is effectively random?

gurudattapatil:
I will edit the code to show the internal port as well if that is necessary. Should I do it?

egede:
No, the problem is a bit deeper. The GUI connects to the internal server, so if the GUI code doesn't know the port number, how will it connect? I can't quite remember the order things start in.

egede:
No, the problem is a bit deeper. The GUI needs to know the port number such that it can connect to the internal server.

alexanderrichards:
Github seems to be not loading properly and taking ages to update for me today for whatever reason but yes this is what I was afraid of @ulrik. If I understand what you're saying, then the gui **is** the client of the internal server so this port should remain hardcoded while the `gui_port` can be effectively randomised by the code in this PR?

egede:
Can we just have the internal server write the port number to a file in gangadir?

gurudattapatil:
I will try doing that.

mesmith75:
Is that not a security risk?

egede:
> Is that not a security risk?

Not more than many other acceptable security risks (and certainly less than having the port number fixed as is currently the case). The file should of course be protected by file permissions to the user only. In that way it is equivalent to the grid proxy that is stored in /tmp and protected by file permissions there.

gurudattapatil:
Will this work?

ulrik:
Hi - just in case you did no know… You must have made a mention of @ulrik at GitHub.

--
Med Venlig Hilsen

Ulrik Hardt Schønnemann
Partner
***@***.*** ***@***.***>
+ 45 6127 4152





Sivlandvænget 27B. 1. th, 5260 Odense
www.MedWare.dk <http://www.medware.dk/>

Følg os på LinkedIn <https://www.linkedin.com/company/medware-aps>
Vi gør opmærksom på, at denne e-mail kan indeholde fortrolig information. Hvis du ved en fejltagelse modtager e-mailen, beder vi dig venligst informere afsender om fejlen ved at bruge svarfunktionen. Samtidig beder vi dig slette e-mailen i dit system uden at videresende eller kopiere den. Selvom e-mailen og ethvert vedhæftet bilag efter vores overbevisning er fri for virus og andre fejl, som kan påvirke computeren eller it-systemet, hvori den modtages og læses, åbnes den på modtagerens eget ansvar. Vi påtager os ikke noget ansvar for tab og skade, som er opstået i forbindelse med at modtage og bruge e-mailen.

gurudattapatil:
Will this work?

egede (approved):

egede:
OK, I have checked this out. If I run `ganga --webgui`, then indeed everything works fine.

gurudattapatil:
I have merged the branch

PR #1906, opened 2021-06-03
adding condor testing framework

gurudattapatil:


gurudattapatil:
I am preparing condor testing framework. If there are any testing files for condor to check this, can you share them?

egede:
> I am preparing condor testing framework. If there are any testing files for condor to check this, can you share them?

No, we do not have any specific ones as we have not yet had this possibility. The first trivial test to implement will be something like
```python
j = Job(backend=Condor())
j.submit()
assert j.backend.id != ''
```
to simply check that a job submits without errors. I suggest to create a new directory called `Condor` inside the [test directory](https://github.com/ganga-devs/ganga/tree/develop/ganga/GangaCore/test) where we will then only run those tests when inside the virtual machine with that setup.

.github/workflows/ci_push_testing.yml
```diff
@@ -384,3 +384,29 @@ jobs:
...
+        run: |
+          python3 -m pip install --upgrade pip setuptools wheel
+          python3 -m pip install -e .[dev]
+      -name: Install Condor
+        run: |
+          yum install -y condor
+          yum condor_master
+      - name: Test with pytest
+        env:
+          GANGA_CONFIG_FILE: ~/.gangarc
```
egede: Can this line just be removed?

.github/workflows/ci_push_testing.yml
```diff
@@ -384,3 +384,29 @@ jobs:
...
+        run: python3 -m pytest --cov-report term-missing --cov ganga/GangaCore/Lib/Condor ganga/GangaCore/Lib/Condor/test
```
egede: This needs to be moved to the `ganga/GangaCore/test/Condor` directory.

.github/workflows/ci_push_testing.yml
```diff
@@ -384,3 +384,29 @@ jobs:
...
+      - name: Install git
+        run: yum install -y git
+      - name: Install dependencies
+        run: |
+          python3 -m pip install --upgrade pip setuptools wheel
+          python3 -m pip install -e .[dev]
+      -name: Install Condor
+        run: |
+          yum install -y condor
+          yum condor_master
```
egede: This doesn't look valid.

gurudattapatil: I will look into this, the sudo commands in this order work on ubuntu but i am unaware as of now how the work on CentOS

alexanderrichards (approved):

egede:
Seems to be working now. There is a new tester running in the CI test. It is based on a docker container with Condor inside.

PR #1907, opened 2021-06-06
Ability to limit simultaneous local jobs

### LINKED ISSUE #1592: Make backend wrappers work in both python2 and python3
The backend wrapper for the local backend https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template explcitly requires python2. The simple script (and the code injected into it) should be agnostic to python2/3 and the shebang changed to just say `/bin/env python`

### LINKED ISSUE #1733: Limit ganga workers for Local backend

monalisha31:
Runs a specific number of jobs at time. The code uses multiprocessing library to create a thread pool, spawning a fixed number of threads as provided by the user to run a specific number of jobs at a time in parallel. I have tested this code on a simple python code to compare the execution time. The link to the test code can be found [here](https://colab.research.google.com/drive/1sc-H7KLmxslz4kBqh_eNl06nN6QXJIkC?usp=sharingl) . I have used the following code to test this in ganga.

```python
j = Job()
j.application.exe='sleep'
j.splitter = GenericSplitter()
j.splitter.attribute = 'application.args'
j.splitter.values=[ [10] for i in range(20) ]
j.backend.batch_submit=3
j.submit()

```

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -2034,7 +2034,8 @@ class Job(GangaObject):
                         for jobs in self.getJobObject().subjobs:
                             # added this 10/8/2009 - now only kills subjobs
                             # which aren't finished.
-                            if jobs.status not in ['failed', 'killed', 'completed']:
+                            if jobs.status not in ['failed', 'killed', 'completed',
+                                                   'failed_frozen', 'completed_frozen']:
                                 jobs.updateStatus('killed', transition_update=transition_update)
                     self.updateStatus('killed', transition_update=transition_update)
                     #
--- a/ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
+++ b/ganga/GangaCore/Lib/Localhost/LocalHostExec.py.template
@@ -1,7 +1,5 @@
-#!/usr/bin/env python2
-from __future__ import print_function
+#!/usr/bin/env python3
 import os,os.path,shutil,tempfile
-import sys,time
 import glob
 import sys
 import mimetypes
@@ -100,6 +98,7 @@ try:
 except OSError as x:
     failurereport(statusfile, 'PROBLEM STARTING THE APPLICATION SCRIPT: \'%s\' \'%s\''%(execmd,str(x)))
                 
+print('WRAPPER: %d'% os.getpid(), file=statusfile)
 print('PID: %d'%child.pid, file=statusfile)
 statusfile.flush()
 
--- a/ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
+++ b/ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+import os
+import sys
+import subprocess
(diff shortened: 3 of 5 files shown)


egede:
This doesn't work. It simply just staggers the submission process, but as the job process is detached, the `Pool` implementation here doesn't block the next jobs from submitting. Take a look at the output of the test below.

First we define a job with 3 subjobs, and only allow one subjob to run at a time.
```python
j = Job()
j.application.exe='sleep'
j.splitter = GenericSplitter()
j.splitter.attribute = 'application.args'
j.splitter.values=[ [10] for i in range(3) ]
j.backend.batch_submit=1
```
From this, we would expect that subjob zero should start straight away, subjob 1 10 seconds later and subjob two 20 seconds later. However, as we see from below that is not the case.
```python
[11:12:15]
Ganga In [23]: [js.time for js in j.subjobs]
Ganga Out [23]: 
[
         Time (UTC)   Status
- - - - - - - - - - - - - - - - - - - - - 
2021/06/07 01:09:26 - backend_running
2021/06/07 01:09:26 - new
2021/06/07 01:09:26 - submitted
2021/06/07 01:09:26 - submitting
2021/06/07 01:09:28 - running
2021/06/07 01:09:36 - backend_final
2021/06/07 01:09:38 - final
,
 
         Time (UTC)   Status
- - - - - - - - - - - - - - - - - - - - - 
2021/06/07 01:09:26 - backend_running
2021/06/07 01:09:26 - new
2021/06/07 01:09:26 - submitted
2021/06/07 01:09:26 - submitting
2021/06/07 01:09:28 - running
2021/06/07 01:09:36 - backend_final
2021/06/07 01:09:38 - final
,
 
         Time (UTC)   Status
- - - - - - - - - - - - - - - - - - - - - 
2021/06/07 01:09:26 - backend_running
2021/06/07 01:09:26 - new
2021/06/07 01:09:26 - submitted
2021/06/07 01:09:26 - submitting
2021/06/07 01:09:28 - running
2021/06/07 01:09:36 - backend_final
2021/06/07 01:09:38 - final
]

```

monalisha31:
I searched about the cause of this problem. The error is due to the apply_async function. It never blocks the new jobs. It only helps in  parallel processing. The way to fix this problem will be to add a function that manages the batch job allotment to the thread pool which will allot the next batch of jobs to the pool as soon as the previous batch is finished.

monalisha31:
I will start working on it.

egede:
Let us talk at the meeting today. I do not think that simply modifying `apply_async` will work.

egede:
The strategy has changed slightly. Now the number of subjobs running in parallel on the Local backend is always limited. This is decided by the `batch` property of the `Local` backend object. By default it will run one subjob per CPU. The limit is not imposed if the master job is resubmitted.

alexanderrichards (approved):

mesmith75 (approved):

PR #1929, opened 2021-06-24
Make batch running of Local job deal with killing, resubmit and resplitting correctly

### LINKED ISSUE #1930: Resplitting and batching of Local jobs do not collaborate
While looking at why some tests have turned very slow, I uncovered that there is a problem with the two new features of resplitting and the batching of jobs on the Local backend. When a Local job with subjobs is created, it writes a wrapper script which then executes the subjob scripts. When a resplit is done this causes several problems.
- It overwrites the old wrapper script
- It fails on an attempt to create a directory which already exist
- There is a race condition for running subjobs.

egede:


egede:
It turns out that the slow testing was caused by actual bugs in the re-implemented Local backend. The side effect was that some tests had some very long timeouts. Tests could be improved to actually fail under this condition ...

egede:
Tests finish 20 minutes faster now!

alexanderrichards (approved):

PR #1603, opened 2020-03-04
Remove treat_as_inputfiles for inputdata

DumbMachine:
Removes code related to treat_as_inputfiles from all the `.py` files.

PR #1934, opened 2021-06-24
Add splitter to download input locally

MannyMoo:
The `SplitByFilesAndDownload` splitter adds `DiracFile` instances to a subjob's `inputfiles` after its creation, so that they're downloaded to the job's working directory. When the LFN->PFN xml catalog is generated on the working node, it finds the local copies and adds them so that they can be accessed by LFN. This is needed for jobs running in containers and using XRootD v3 based stacks (ie, slc5 builds), as they can't access files remotely.

As discussed with @egede, maybe it's better to add this as an option to the existing `SplitByFiles`, rather than adding a new splitter. Currently I get a warning on Ganga startup:
```
WARNING  Possible schema clash in class SplitByFilesAndDownload between SplitByFilesAndDownload and SplitByFiles
```

Note that this PR builds on top of the `container_build` branch that @mesmith75 is working on, as I needed that to be able to submit the containerised slc5 jobs. I can rebase this branch to include just my commits if necessary.

### DIFF
--- a/ganga/GangaLHCb/Lib/Applications/GaudiExec.py
+++ b/ganga/GangaLHCb/Lib/Applications/GaudiExec.py
@@ -443,6 +443,9 @@ class GaudiExec(IPrepareApp):
         """
         Return the script which wraps the running command in a correct environment
         """
+        j = self.getJobObject()
+        if j.virtualization:
+            return 'singularity exec -e --bind /cvmfs %s bash -c " source /cvmfs/lhcb.cern.ch/lib/LbEnv && source LbLogin.sh -c %s && ' % (j.virtualization.image, self.platform)
         if isLbEnv:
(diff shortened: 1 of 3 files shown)


mesmith75:
We need to decide what to do with this. I think it is a nice feature in principle - maybe if we add a `treat_as_inputfiles` attribute to the `LHCbDataset` and make the necessary changes to the splitter.

mesmith75:
Actually it was removed only very recently. I guess we want to revert this: https://github.com/ganga-devs/ganga/pull/1603/files

MannyMoo:
OK, that seems like a cleaner solution, so fine by me. I'm happy to test it out as well.

mesmith75:
Fixed by #1942

PR #1937, opened 2021-07-04
Allow for multi-core make-ing of the input sandbox

vlisovsk:
This introduces a new argument 'nMakeCores' (default=1) which is provided to the `make` command.

PR #1938, opened 2021-07-05
Thread the GaudiExec cmake

mesmith75:
Copy of #1937 with added configurable


mesmith75:
@vlisovsk I have copied it here with a slight amendment.

alexanderrichards (dismissed):

egede (changes requested):
To set a per application property via a general config option seems the wrong approach. You can just remove the config option and then use the `defaults_GaudiExec` config option to change the default.

alexanderrichards (approved):

egede (approved):

PR #1941, opened 2021-07-12
Ganga-GUI part 1

gurudattapatil:


alexanderrichards (approved):

egede (approved):
This all looks really good. I have a number of small usage issues and stylistic comments, but I would like to see this merged now and then deal with them as `issues` instead.

PR #1942, opened 2021-07-12
bring back treat as inputfiles

mesmith75:
Reverts #1603

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Tasks/CoreTransform.py
+++ b/ganga/GangaCore/GPIDev/Lib/Tasks/CoreTransform.py
@@ -111,6 +111,8 @@ class CoreTransform(ITransform):
                     unit.name = "Unit %d" % len(self.units)
                     unit.inputdata = GangaDataset(
                         files=filelist[fid:fid + self.files_per_unit])
+                    unit.inputdata.treat_as_inputfiles = self.inputdata[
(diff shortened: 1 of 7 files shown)


ganga/GangaCore/GPIDev/Lib/Tasks/CoreTransform.py
```diff
@@ -111,6 +111,8 @@ def createUnits(self):
                     unit.name = "Unit %d" % len(self.units)
                     unit.inputdata = GangaDataset(
                         files=filelist[fid:fid + self.files_per_unit])
+                    unit.inputdata.treat_as_inputfiles = self.inputdata[
+                        0].treat_as_inputfiles
```
alexanderrichards, line 115: strange line continuation here on the `[` was that intentional?

alexanderrichards (approved):

ganga/GangaCore/GPIDev/Lib/Tasks/CoreTransform.py
mesmith75: Yes - its a copy and paste of what was there before. I guess its a PEP8 thing

mesmith75:
@MannyMoo If you are willing to test this that would be great - maybe after we merge it to develop and update the virtualisation branch.

PR #1962, opened 2021-08-02
Code editor section added in GUI

monalisha31:
Currently, the code editor is working like this. A user can choose there specific language and save there code from the GUI. 

screencapture-0-0-0-0-5500-editor-2021-08-02-08_32_20

Execution of the code part will be added later. For now , A user can save there code in .txt format. A "save as" feature will be added later, it  will save the file in the format that is specified by the user.


monalisha31:
The current changes will enable the user to save the file in there desired format.  The user will enter the filename and file type (hello.py) in the save as section.
Screenshot from 2021-08-15 10-05-14

egede (commented):
The main code for the editor is there, but I would still say that it is not "functional" in the sense that it can be used together with the rest of Ganga. There is a list of (non-exhaustive) issues here
- It is not possible to open a file listed in a Ganga job in the editor.
- There seems in general to be no way to open/upload a file
- When a file is saved, it is saved on the web client side. In general if you want the Ganga process to use the file, it should be saved on the server side.
- There should be a way to execute what is in the editor inside the running Ganga process, so in the same way the `quick editor` works.
- After you have saved a file, it disappears and you are presented with the `default` javascript file again.

monalisha31:
I will work on these issues.

monalisha31:
Last week, I have added Ganga job list beside the editor where the user can edit a job. Next task will be to add a submit button, where a user can submit the edited job. After submission, a new job will be created. These are the changes made in the GUI. @egede Do i need to make any more changes in the attached UI?

screencapture-0-0-0-0-5500-editor-15-edit-2021-09-27-15_58_49
screencapture-0-0-0-0-5500-editor-2021-09-27-15_58_07

egede:
Hi @monalisha31 Sorry to take forever to look at this. I can now load a job into the editor as you describe. This is really good and would be a great way for a user to take a template job and then edit it to do the right thing for them. However, when I get the job into the editor, there is no way to save it as a job again. In fact I can't save it in any way or form. Do I miss something?

monalisha31:
Currently, the user can edit the job and save it as .txt file only. I am trying to add a save functionality that will save the edited job as a new Ganga job and the job data will be updated in the dashboard.

PR #1963, opened 2021-08-02
Ensure DiracCommands.getOutputDataInfo doesn't return dict_keys

chrisburr:
I suspect this fixes the issue reported in https://mattermost.web.cern.ch/lhcb/pl/a81pi1dqdjyf8jfyzp15ynapjc


egede:
Thanks for this fix. Will be merged by #1964

PR #1966, opened 2021-08-09
Added edit .gangarc section in the GUI

monalisha31:
A user can access this section through config section's edit gangarc button. The role of this section is to fetch, edit and save .gangarc file from the GUI. The aim here is to make the config section editable by allowing the user to make changes in the .gangarc file directly from the GUI. The webpage is working like this:

screencapture-0-0-0-0-5500-config-edit-2021-08-09-10_26_46


egede:
The initial check out of the code from Git is failing. Has the branch been deleted or is there a temporary error?

monalisha31:
The form text area is replaced with a code editor. Currently this section is working like

screencapture-0-0-0-0-5500-config-edit-2021-08-15-00_49_16

egede (approved):
This works really well. Will merge now.

PR #1970, opened 2021-08-11
adding the gui terminal to ganga and daemonization

gurudattapatil:
run.sh in gui folder works as discussed now working on creating a command to access the terminal easily


bin/gui-terminal
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+python3 ../ganga/GangaGUI/gui/webterm.py </dev/null &>/dev/null &
+sleep 15
+python3 ../ganga/GangaGUI/gui/autotype.py
```
egede: This can just be avoided, see suggested changes elsewhere.

bin/gui-terminal
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+python3 ../ganga/GangaGUI/gui/webterm.py </dev/null &>/dev/null &
```
egede: If you want to be able to kill the terminal that you start this script from and keep the GUI running, you will need to disown the process as well. Also some thought should be given to how the server process eventually will stop.

ganga/GangaGUI/gui/webterm.py
```diff
@@ -0,0 +1,138 @@
...
+        return json.dumps({"c": (cursor.x, cursor.y), "lines": lines})
+
+
+def open_terminal(command="bash", columns=80, lines=24):
+    p_pid, master_fd = pty.fork()
+    if p_pid == 0:  # Child.
+        argv = shlex.split(command)
+        env = dict(TERM="linux", LC_ALL="en_GB.UTF-8",
+                   COLUMNS=str(columns), LINES=str(lines))
+        os.execvpe(argv[0], argv, env)
```
egede:
```suggestion
def open_terminal(columns=80, lines=24):
    p_pid, master_fd = pty.fork()
    if p_pid == 0:  # Child.
        argv = shlex.split(command)
        env = dict(TERM="linux", LC_ALL="en_GB.UTF-8",
                   COLUMNS=str(columns), LINES=str(lines))
        env = {**os.environ, **env}
        os.execvpe("bash", ['bash', '-c', 'ganga --webgui'], env)
```

egede (changes requested):
- The `autotype.py` file I think can just be removed. 
- When I try to cut-and-paste from the terminal inside the browser, there is a tab character inserted in between all other characters. Any idea why this is the case?

egede (changes requested):
The `aiohttp` and the `pyte` packages are not added to `setup.py`.

bin/gui-terminal
gurudattapatil: autotype.py is used to type start ganga gui without user interaction from the web terminal, i was going to use selenium but it 
suffers with browser issues if drivers are not installed

gurudattapatil: I tried directly to input via javascript but this caused terminal to crash

bin/gui-terminal
gurudattapatil: I will look into this

gurudattapatil: actually </dev/null &>/dev/null & provides same functionality as disown process, this will also prevent unwanted logs to be printed.

egede: No, I do not think so. When you do this, the process group of the child process stays the same (so killing the parent will still kill the child). So I think you need to do the above AND disown the process.

gurudattapatil: Added the disown command

egede:
I looked into the problem with `pynput`. The dependency tree for installations give
```
pynput==1.7.3
  - evdev [required: >=1.3, installed: 1.4.0]
  - python-xlib [required: >=0.17, installed: 0.31]
    - six [required: >=1.10.0, installed: 1.16.0]
  - six [required: Any, installed: 1.16.0]
```
and it is the `evdev` package that is causing problems as it requires some kernel-header packages to be installed (and we can't assume that users have those). Seems yet another reason to avoid the `pynput` package.

gurudattapatil:
I will look into other libraries that will be useful to type in terminal to automatically start the gui

gurudattapatil:
removed autotype and other pynput

gurudattapatil:
time for typing can be adjusted according to need for now in autotype.py, i will work on automating this as soon as the terminal opens

gurudattapatil:
This branch is stable for merging, although the cli/terminal button in side pane may not work. 
@egede do we require any particular changes in this branch?

egede (approved):
This indeed seems to work now.

egede:
Has this PR been replaced by #2061? If yes, can I just close it?

gurudattapatil:
yes

PR #1976, opened 2021-08-20
Add append option to ArgSplitter

### LINKED ISSUE #1975: Append ArgSplitter args
`ArgSplitter` should append its run-time arguments rather than replace them. It would be better to be able to set up a job as 
```python
j = Job()
j.application.args = ['a','list','of','common','args']
j.splitter = ArgSplitter(args = [['args'],['for'],['each'],['subjob']])
```
Currently only the `ArgSplitter` args get passed to the subjobs

mesmith75:
Fixes #1975


egede (approved):
Good solution.

PR #1983, opened 2021-10-04
remove popen2

### LINKED ISSUE #876: popen2 module is deprecated
For each PBS job, I get a warning (could one get rid of it?):
######################################################################
/var/spool/torque/...: DeprecationWarning: The popen2 module is deprecated. Use the subprocess module.
import popen2
######################################################################

Original : #870

mesmith75:
Fixes #876

### DIFF
--- a/ganga/GangaCore/Lib/Batch/BatchScriptTemplate.py.template
+++ b/ganga/GangaCore/Lib/Batch/BatchScriptTemplate.py.template
@@ -3,7 +3,6 @@ from __future__ import print_function
 import shutil
 import os
 import time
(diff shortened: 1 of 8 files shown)


egede (approved):

mesmith75:
BTW there are a few references in `GangaAtlas` which this doesn't remove but I don't think we are claiming to still support that module

egede:
> BTW there are a few references in GangaAtlas which this doesn't remove but I don't think we are claiming to still support that module

No, it is not even changed to `python3`.

PR #1985, opened 2021-10-27
Update install scripts

mesmith75:
Update to a newer python version with LCG100. Also add in numpy and pandas for users.


ganga/GangaRelease/tools/ganga-cvmfs-install.sh
```diff
@@ -2,18 +2,22 @@
...
 
 . $1/bin/activate
 
 pip install --upgrade pip setuptools
 
 pip install git+https://github.com/ganga-devs/ganga.git@$1#egg=ganga[LHCb] 
 
+pip install numpy
+
+pip install pandas
```
alexanderrichards: do we depend on / need numpy or pandas?

alexanderrichards (dismissed):

ganga/GangaRelease/tools/ganga-cvmfs-install.sh
mesmith75: No - but some users have requested and it may be useful for them

alexanderrichards: ok

mesmith75: Actually I don't like this - which should add the LCG version to the `PYTHONPATH` instead.

egede (dismissed):
Given that his is the cvmfs installation that is affected by this, I guess it is OK to add pandas and numpy. However, it is a bit of a slippery slope.

mesmith75:
I think this is what we want. For some useful packages you need the compiled libraries so I have put in a little hack to add them to the `LD_LIBRARY_PATH` and then start ganga in a new process. The advantage is it won't change the environment for the user when they quit ganga.

egede (dismissed):

alexanderrichards (approved):

PR #1988, opened 2021-10-28
Remove ROOT application

mesmith75:
Removes ROOT RTHandlers from LHCb.


alexanderrichards (dismissed):

egede (commented):
Rather than just removing the ROOT application from DIRAC, should we remove it everywhere?

egede (dismissed):
I guess you will discover if there are some test cases that should be removed.

ganga/GangaCore/Lib/Executable/Executable.py
```diff
@@ -288,3 +288,28 @@ def prepare(self, app, appconfig, appmasterconfig, jobmasterconfig):
...
+    ascii_length = len(ascii_uppercase)
+    sample = []
+    for _ in range(6):
+        addToSample(sample, ascii_length)
+    assert(len(sample) == 6)
+
+    # seed is set to clock during import
+    return ''.join([str(a) for a in sample])
+
+
```
alexanderrichards, line 315: Out of context it seems strange that this was even in the ROOT module. Makes me wonder if it was only used in said module and maybe a candidate for removing altogether. Do you know where this is used?

alexanderrichards (approved):

ganga/GangaCore/Lib/Executable/Executable.py
```diff
@@ -288,3 +288,28 @@ def prepare(self, app, appconfig, appmasterconfig, jobmasterconfig):
 allHandlers.add('Executable', 'CREAM', LCGRTHandler)
 allHandlers.add('Executable', 'ARC', LCGRTHandler)
 allHandlers.add('Executable', 'Slurm', RTHandler)
+
+def randomString():
```
alexanderrichards, line 292: *I realised the previous comment didn't really show the whole code so more useful to see the function prototype*

alexanderrichards (approved):

ganga/GangaCore/Lib/Executable/Executable.py
alexanderrichards: I'm referring to the `randomString` function which unfortunately didn't show here in it's entirety.

mesmith75: Yes I agree it was weird. Would it be better put elsewhere?

alexanderrichards: for now, putting it in `Executable` seems as good as anything. My point was more that if it was only used in the now defunct `Root` module then maybe we don't need it at all?

mesmith75: Ah, It seems to be used in various places so I kept it.

PR #1992, opened 2021-11-02
Update testing python to 3.8.6

mesmith75:


alexanderrichards (dismissed):

mesmith75:
This is picking up a load of syntax warnings (`is` instead of `==`). I'll go through them and fix them in this PR. Do we want to make these warnings report as errors?

alexanderrichards (approved):

PR #1999, opened 2021-11-23
Bring back Condor requirements

### LINKED ISSUE #1998: Setting Condor requirements
Is it possible for the user to set Condor requirements?  It used to be possible via the [CondorRequirements class](https://github.com/ganga-devs/ganga/blob/23b1de3d4179508aa2f7b8b9867dc261d8e15036/ganga/GangaCore/Lib/Condor/CondorRequirements.py#L37).  The [Condor class](https://github.com/ganga-devs/ganga/blob/23b1de3d4179508aa2f7b8b9867dc261d8e15036/ganga/GangaCore/Lib/Condor/Condor.py#L49) includes a `requirements` item in its schema, but as far as I can see doesn't do anything with it.  Experimentally, any values set for `requirements` seem to be ignored :(

Thanks, Karl.

mesmith75:
Fixes #1998

### DIFF
--- a/ganga/GangaCore/Lib/Condor/Condor.py
+++ b/ganga/GangaCore/Lib/Condor/Condor.py
@@ -283,6 +283,8 @@ class Condor(IBackend):
         if self.globus_rsl:
             cdfDict['globus_rsl'] = self.globus_rsl
 
+        cdfDict['Requirements'] =  self.requirements.convert()
+
         return cdfDict
 
 
--- a/ganga/GangaCore/Lib/Condor/CondorRequirements.py
+++ b/ganga/GangaCore/Lib/Condor/CondorRequirements.py
@@ -115,7 +115,7 @@ the final requirement is the AND of all elements in the list
         if self.other:
             requirementList.extend(self.other)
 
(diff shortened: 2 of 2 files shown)


ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -283,6 +283,8 @@ def cdfPreamble(self, jobconfig):
         if self.globus_rsl:
             cdfDict['globus_rsl'] = self.globus_rsl
 
+        cdfDict['Requirements'] =  self.requirements.convert()
```
egede, line 286: The capital `R` in seems a bit odd when comparing to the rest of the `cdfDict`. Is it correct?

mesmith75: Yes I agree it looks odd. I tried this with setting `OpSys` to `Linux` or a random string. The former ran, whilst the second was stuck in idle because the scheduler couldn't match the requirements so I think it works.

Looking around I have seen examples with all of these (`executable`, `universe` etc) with lower case and other examples with upper case so I think it may be case-insensitive.

egede (approved):

PR #2001, opened 2022-01-25
DiracFile.put: check SE availability

chaen:
When using the default list of storage elements when doing `DiracFile.put`, check that the randomly chosen one is available for write

### DIFF
--- a/ganga/GangaDirac/Lib/Files/DiracFile.py
+++ b/ganga/GangaDirac/Lib/Files/DiracFile.py
@@ -754,7 +754,15 @@ class DiracFile(IGangaFile):
                 storage_elements = [self.defaultSE]
             else:
                 if configDirac['allDiracSE']:
-                    storage_elements = [random.choice(configDirac['allDiracSE'])]
+                    trySEs = list(configDirac['allDiracSE'])
+                    random.shuffle(trySEs)
+                    storage_elements = None
+                    for se in trySEs:
+                        if execute('checkSEStatus("%s", "%s")' % (se, 'Write')):
+                            storage_elements = [se]
+                            break
+                    if not storage_elements:
+                        raise GangaFileError("No SE allowed for Write")
                 else:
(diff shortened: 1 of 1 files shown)


chaen:
I see the CI failures are due to not being able to download `diracos`. However, the version of DIRAC and DIRACOS used are super old 

```
 Notice: 26 00:54:52 UTC dirac-install [NOTICE]  Retrieving http://diracproject.web.cern.ch/diracproject/tars/DIRAC-v7r1p19.tar.gz
Notice: 26 00:54:54 UTC dirac-install [NOTICE]  Retrieving http://diracproject.web.cern.ch/diracproject/tars/DIRAC-v7r1p19.md5
Notice: 26 00:54:55 UTC dirac-install [NOTICE]  Extract using system tar: /github/home/dirac_ui/DIRAC-v7r1p19.tar.gz
```

But I see it's hard coded in quite a few places

https://github.com/ganga-devs/ganga/blob/a7901c894837df563d724ecab17d9b884555543c/.github/workflows/ci_push_testing.yml#L277-L278

https://github.com/ganga-devs/ganga/blob/a7901c894837df563d724ecab17d9b884555543c/.github/workflows/ci_push_testing.yml#L147-L148

Any particular reasons for that ?

mesmith75:
I think there was a reason but I cannot remember - perhaps @egede  does? In any case i think the issue is that our certificate has expired.

egede:
I have a new certificate ready, so will take a look.

chaen:
I don't think the certificate expired has anything to do here, but you may want to bump the version of DIRAC tested ? And even use the python3 version :-)

egede:
The continued failures of the testing was due to
- Expired certificate in the GitHib secrets (now fixed)
- PR created by a non-core member (so no access to secrets). I fixed this by creating a temporary clone of the PR.

egede (approved):

mesmith75:
Thanks @chaen . The Dirac version was because we had problems with the version at Imperial, but we will look again.

PR #2007, opened 2022-02-10
Mac os compatibility

### LINKED ISSUE #2004: Splitting of local jobs not working on MacBook
On a MacBook (intel processor), if I submit a job created with:

```
j = Job()
j.splitter=ArgSplitter(args=[['Hello 1'], ['Hello 2']])
```

the traceback below is repeated continuously until I kill the process that's running `__jobscript__`.  This is using `ganga 8.5.0`, installed with `pip`.  The same job runs without problems on a Linux machine.

Any ideas?

Thanks, Karl.

```
  File "<string>", line 1, in <module>
  File "/opt/miniconda3/envs/test/lib/python3.10/multiprocessing/spawn.py", line 116, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/opt/miniconda3/envs/test/lib/python3.10/multiprocessing/spawn.py", line 125, in _main
    prepare(preparation_data)
  File "/opt/miniconda3/envs/test/lib/python3.10/multiprocessing/spawn.py", line 236, in prepare
    _fixup_main_from_path(data['init_main_from_path'])
  File "/opt/miniconda3/envs/test/lib/python3.10/multiprocessing/spawn.py", line 287, in _fixup_main_from_path
    main_content = runpy.run_path(main_path,
  File "/opt/miniconda3/envs/test/lib/python3.10/runpy.py", line 269, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/opt/miniconda3/envs/test/lib/python3.10/runpy.py", line 96, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/opt/miniconda3/envs/test/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/user/gangadir/workspace/user/LocalXML/6/input/__jobscript__", line 33, in <module>
    os.mkdir(outdir)
FileExistsError: [Errno 17] File exists: '/Users/user/gangadir/workspace/user/LocalXML/6/output'
```

egede:


ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
```diff
@@ -37,17 +37,18 @@ sys.stdout=open(join(outdir,'__syslog__'),'w')
...
+        pool.join()
 
 sys.stdout=sys.__stdout__
 sys.stderr=sys.__stderr__
 
 line='STOP: '+time.strftime('%a %b %d %H:%M:%S %Y',time.gmtime(time.time())) + os.linesep
 statusfile.writelines(line)
 statusfile.close()
-sys.exit()
+# sys.exit()
```
egede: Whatever. Should make no change.

setup.py
```diff
@@ -104,7 +104,7 @@ def run(self):
           "docker",
           "pymongo",
           "gdown",
-          "htcondor"
+#          "htcondor"
```
egede: We need to work out a way to make this optional? Can a dependency be made to fail silently?

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
```diff
@@ -37,17 +37,18 @@ sys.stdout=open(join(outdir,'__syslog__'),'w')
...
-finally:
-    pool.close()
-    pool.join()
+if __name__ == '__main__':
+    try:
+        pool = multiprocessing.Pool(batch_size)
+        pool.map(mp_worker, scriptnames_g)
+    finally:
+        pool.close()
+        pool.join()
```
egede: This looks odd and makes me think that there is a different problem. If this if statement is ever False the code will do nothing. That can't be the solution. Is there a problem with the shebang at the start instead?

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
```diff
@@ -8,7 +8,7 @@ from pathlib import Path
 from os.path import dirname, join, isdir
 
 def mp_worker(name):
-    print(f'START: {name} '+ time.strftime('%a %b %d %H:%M:%S %Y',time.gmtime(time.time())))
+    # print(f'START: {name} '+ time.strftime('%a %b %d %H:%M:%S %Y',time.gmtime(time.time())))
```
egede: Some different problem. The code itself is clearly valid python.

egede (commented):
Rebase caused a new branch ...

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
egede: @kh296 Can you have a look at which python interpreter the shebang actually starts for you. I suspect some mixed python environment to be the cause.

setup.py
alexanderrichards: good question about failing silently. I'm not sure about that however we can have optional dependencies. The problem however is that it's easy to optionally have an extra dependency but not sure how to optionally have less. I'll have to have a look

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
alexanderrichards: do you get a traceback?

setup.py
kh296: In the branch of the fork, I've modified [setup.py](https://github.com/kh296/ganga/blob/9df995563543e7dc9da3f5ff8e61268fbfda2d1f/setup.py#L96) to make inclusion of `htcondor` in the list `install_requires` conditional on platform.  I'm not sure if this can be propagated to pypi, but it at least allows installation from github.

I've also changed the way in which a failed import of `htcondor` is handled in [ganga/GangaCore/Lib/Condor/Condor.py](https://github.com/kh296/ganga/blob/9df995563543e7dc9da3f5ff8e61268fbfda2d1f/ganga/GangaCore/Lib/Condor/Condor.py#L21).  (The handling based on warnings didn't seem very robust.)

I've tested that, after these changes, the simple job splitting example:

```
j = Job()
j.application.args = []
j.splitter=ArgSplitter(args=[['Hello 1'], ['Hello 2']])
j.submit()
```

runs successfully for me on the Local backend for both MacOS and Linux, and runs on the Condor backend for Linux.

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
kh296: This doesn't generate a traceback, but the print-out is sent to the Ganga prompt, rather than to the log file.  Given the redefinition of `sys.stdout` in [/ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template](https://github.com/kh296/ganga/blob/9df995563543e7dc9da3f5ff8e61268fbfda2d1f/ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template#L36), I'm not sure why the output isn't being captured.

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
kh296: I'm running in a `conda` environment, created from an `environment.yml` file containing:

```
name: test
dependencies:
  - pip
  - python
  - pip:
    - -e git+https://github.com/kh296/ganga.git@MacOS_no_htcondor#egg=ganga
```

I've checked that both ganga and the job script are using the environment's installation of python (3.10.0).

setup.py
mesmith75: We already make the LHCb conditional so we can do the same for htcondor. Though I would argue the default should include condor as its in GangaCore.

kh296: The condition added in [setup.py](https://github.com/kh296/ganga/blob/9df995563543e7dc9da3f5ff8e61268fbfda2d1f/setup.py#L96) includes `htcondor` in the dependencies except for MacOS (for which `htcondor` isn't available on pypi):

```
if sys.platform != 'darwin':
      install_requires.append("htcondor")
```

There's probably a better way of doing this, and any solution that allows for installing ganga without requiring `htcondor` would work for me :)

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
egede: OK, so the shebang is not actually used when running this inside Ganga. As you see from the [run](https://github.com/ganga-devs/ganga/blob/5bd0d25f5d99597201db4da51e907838761c2589/ganga/GangaCore/Lib/Localhost/Localhost.py#L133) method in the local backend, it is started directly as a `python3` process. This should still evaluate to the same though. It might be that some tweak to how the process is started will fix things though. You can try to change that line to
```python
            import sys
            process = subprocess.Popen([sys.executable, scriptpath, 'subprocess'], stdin=subprocess.DEVNULL, start_new_session=True)
```
and see if that is fixing the problems and allow you to run with the template unchanged.

egede:
OK, I tried to push the changes discussed above to the `Popen` call. The change was done in both how the template was called and inside the template itself. I also backed out the weird `if __name__ == '__main__':` part.

setup.py
alexanderrichards: I agree with @mesmith75 that the default should be to include it since it's in GangaCore and I'm happy with the suggested fix to `setup.py` as long as it works installing from PyPI

ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template
kh296: I've tried running with [LocalHostExec_batch.py.template](https://github.com/ganga-devs/ganga/blob/MacOS_compatibility/ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template) unchanged, and with revised [Localhost.run()](https://github.com/ganga-devs/ganga/blob/5fe8339c7bc7b854845bc94f647c65dffcff0b1b/ganga/GangaCore/Lib/Localhost/Localhost.py#L132).  Unfortunately we return to the previous situation on MacOS: the job script seems to get stuck in an infinite loop, continuously trying, and failing, to create the multiprocessing pool.

kh296:
I'm not sure why the `if __name__ == `__main__:` bit should be needed, but on MacOS it definitely makes a difference.

egede:
I think what is going on is that when each of the new processes in the `ThreadPool` start, they somehow try to execute the script itself again, causing an infinite loop. The  `if __name__ == '__main__':` part protects against that. I have moved everything inside a function and put this `if` statement at the bottom, to make the purpose of it clearer and in line with usual constructs of this type.

@kh296 please give this another try.

kh296:
I confirm that subjobs run successfully on MacOS with the revised version of [LocalHostExec_batch.py.template](https://github.com/ganga-devs/ganga/blob/8b705ea8079b0c56a25837fa320e09045b68f629/ganga/GangaCore/Lib/Localhost/LocalHostExec_batch.py.template).

Thanks, Karl

setup.py
```diff
@@ -74,6 +74,31 @@ def run(self):
...
+    'Flask-SQLAlchemy>=2.4.3',
+    'gunicorn>=20.0.4',
+    'Flask-Login>=0.5.0',
+    'Flask-SocketIO==4.3.1',
+    'SQLAlchemy==1.3.19',
+    "docker",
+    "pymongo",
+    "gdown",
+]
+if sys.platform != 'darwin':
```
mesmith75, line 98: This doesn't fill me with joy but I guess its the only practical solution

mesmith75 (approved):

setup.py
egede: Yes, I looked around. To get a behaviour like "Try to install htcondor if available, but don't fail if not" seems only to be possible with the horrible hacks.

PR #2010, opened 2022-02-13
Fix non gaudirun GaudiExec

mesmith75:
A little bit of python 3 compatibility was missed when running GaudiExec with python.


mesmith75:
I think there may be one or two more of these lurking about

egede (approved):

PR #2011, opened 2022-02-13
fix put lfn

### LINKED ISSUE #2008: DiracFile put ignoring provided LFN
If you have a `DiracFile` with a local path, and try `df.put(lfn='/lhcb/user/path/to/some/lfn')` it ignores the LFN and makes its own.

mesmith75:
Fixes #2008


egede (commented):
LFNs have quite a few restrictions on them (length, valid characters). Do we want to protect the user. Also is there a (hidden) requirement that the basename of the LFN and the actual filename has to be the same or are they completely independent.

mesmith75:
That is a good point about extra protections.

For the filename it can be changed in the LFN. There is a warning in this function if the user does try to do that, but perhaps this fix bypasses that.

mesmith75:
I have added a check for filenames over 60 characters. Further up there is a warning about chaning the basename of the LFN, but it gives the option to continue which I think is correct.

egede (approved):

PR #2017, opened 2022-03-19
Don't raise an error if there is no sandbox for jobs killed by Dirac

### LINKED ISSUE #2015: Don't get sandbox from killed Dirac jobs
If the Dirac status is killed we shouldn't try and download the non-existent sandbox.

mesmith75:
Fixes #2015 

This intercepts the exception from the failed sandbox download if the backend status is `Killed`. I am checking with the Dirac developers if in fact one should *never* expect a sandbox from Dirac if the job is killed. If that is the case then I'll alter it to not even try downloading.


mesmith75:
Indeed, there should never be a sandbox for killed jobs - so I have prevented the download being attempted.

egede (approved):

alexanderrichards (approved):

PR #2020, opened 2022-03-29
Adjust DiracFile put filename length warning

mesmith75:
60 was a bit short - the GaudiExec sandboxes are often already longer than that.


egede (approved):
As we really don't know what the limit is, then this is as good as any.

PR #2027, opened 2022-05-25
Change report output type to LocalFile by default

### LINKED ISSUE #1984: Change report to use LocalFile by default

joj0s:
This PR changes the report function to use a LocalFile instead of a GoogleFile by default. It also includes a test case for it.

P.S. I have set up autopep8 to auto-run each time a Python file is saved. It fixed some whitespace issues in the `feedback_report.py` file. If you would like I can omit those from the PR.

Closes #1984

### DIFF
--- a/ganga/GangaCore/test/Unit/Utility/TestUtilitiesReport.py
+++ b/ganga/GangaCore/test/Unit/Utility/TestUtilitiesReport.py
@@ -24,6 +24,9 @@ class TestFeedbackReport(GangaUnitTest):
         self.assertTrue(os.path.exists(outputfile))
         self.assertGreater(os.path.getsize(outputfile), 0)
 
+        if (os.path.exists(outputfile)):
+            os.remove(outputfile)
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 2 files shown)


alexanderrichards:
Thanks for this.


> P.S. I have set up autopep8 to auto-run each time a Python file is saved. It fixed some whitespace issues in the `feedback_report.py` file. If you would like I can omit those from the PR.

This is fine. It means we gradually get all files done. It's probably too late this time as it's all a single commit but in future I would include the PEP8 corrections in their own commit so that it's clear which changes are related to the issue and which are cosmetic.

alexanderrichards (dismissed):

ganga/GangaCore/test/Unit/Utility/TestUtilitiesReport.py
```diff
@@ -0,0 +1,29 @@
...
+        j = Job()
+        j.submit()
+        self.assertTrue(run_until_completed(j))
+
+        report_result = report(j)
+        self.assertTrue(isinstance(report_result, LocalFile))
+
+        outputfile = os.path.join(report_result.localDir, report_result.namePattern)
+        self.assertTrue(os.path.exists(outputfile))
+        self.assertGreater(os.path.getsize(outputfile), 0)
```
egede, line 25: Will the file be deleted when the test is done. If not, we should clean up and remove it.

egede (approved):

PR #2033, opened 2022-06-17
Formatting and Unit Tests for GangaFile subclasses

### LINKED ISSUE #1627: Write Unit tests for the GangaFile classes
All the classes in https://github.com/ganga-devs/ganga/tree/develop/ganga/GangaCore/GPIDev/Lib/File that inherit from `IGangaFile` should have unit tests written and placed in a directory `GangaCore/test/Unit/Files`. The `mock` module should be used to prevent that the testing is waiting for any external services.

### LINKED ISSUE #2032: MassStorageFile remove method bugs
While testing the MassStorageFile class for #1627, I found a couple of bugs that cause the `remove` method to break.

- BUG 1:
[This line of code](https://github.com/ganga-devs/ganga/blob/d7b5f9db3b0e37a39a9e93a1df67599c65f54ab3/ganga/GangaCore/GPIDev/Lib/File/MassStorageFile.py#L516) should use the `remove` method instead of the `pop` method, since the local i variable is a filepath and not an index. Right now it causes a `TypeError`.

- BUG 2:
Even if the are multiple file locations to remove, [this loop](https://github.com/ganga-devs/ganga/blob/d7b5f9db3b0e37a39a9e93a1df67599c65f54ab3/ganga/GangaCore/GPIDev/Lib/File/MassStorageFile.py#L495) always exits after one iteration. This is due to `self.locations` being an instance of `GangaList` and not a regular list, and there is some fault in the implementation which I haven't yet discovered. I can confirm however that converting `self.locations` to a regular Python list using a list comprehension fixes the problem and loops as expected.

I can submit a PR to fix those, but I'm not sure how you want to deal with the second bug. Is it OK to use the quick fix I suggested or should I keep looking for the fault in `GangaList`?

joj0s:
This PR includes: 

- Autoformatting and import sorting via autopep8 for GangaFile subclasses
- Unit tests for all the classes
- Fixes for #2032 discovered via the unit tests

Closes #1627, closes #2032 

Note, despite autopep8 fixing a lot of whitespace and other common PEP8 issues, these classes still remain not completely PEP8 compliant. This is due to a lot of excessive line lengths and some unused imports and local variables which are not auto-fixable. If you'd like me to address those as well let me know

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/File/GoogleFile.py
+++ b/ganga/GangaCore/GPIDev/Lib/File/GoogleFile.py
@@ -1,22 +1,20 @@
-import re
+import base64
 import copy
 import glob
+import logging
 import os
 import pickle
+import re
 import stat
-import logging
-import base64
-
 from fnmatch import fnmatch
 
 import GangaCore.Utility.Config
-
-from GangaCore.GPIDev.Schema import Schema, Version, SimpleItem, ComponentItem
-from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
 from GangaCore.Core.exceptions import GangaFileError
-from GangaCore.Utility.logging import getLogger
-from GangaCore.GPIDev.Base.Proxy import isType, GPIProxyObjectFactory
+from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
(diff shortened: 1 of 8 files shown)


alexanderrichards (approved):

mesmith75 (approved):
Looks good to me

PR #2038, opened 2022-07-13
Add __class_getitem method required for python 3.9 and above

### LINKED ISSUE #2037: GangaList tests fail on newer Python versions: __class_getitem__ method not implemented
When attempting to fix tests and using Python version 3.9.0 locally, I came across the following failure:
```
======================================================================
FAIL: testAllListMethodsExported (ganga.GangaCore.test.Unit.GangaList.TestGangaList.TestGangaList)
Tests that all methods on list are exposed by GangaList
----------------------------------------------------------------------
  File "/home/jojos/Projects/cern/ganga/ganga/GangaCore/test/Unit/GangaList/TestGangaList.py", line 94, in testAllListMethodsExported
    self.assertFalse(missing_methods, 'Not all of lists methods are implemented: %s' % missing_methods)
AssertionError: ['__class_getitem__'] is not false : Not all of lists methods are implemented: ['__class_getitem__']
----------------------------------------------------------------------
```

This fails in the `develop` branch. The [`__class_getitem__`](https://peps.python.org/pep-0560/) method is a list method added on newer Python versions and has not been implemented by GangaList. Can confirm that the tests pass on Python version 3.7.9

egede:


egede:
The [data model](https://docs.python.org/3/reference/datamodel.html#class-getitem-versus-getitem) for python describes this method. Pretty obscure. AFAIU, simply returning the class object itself, means that the external type checking is not possible (but will not cause errors).

egede:
I tested in python 3.10 as well and all `GangaCore` tests work there. There is a huge number of warnings though coming from that some thread related methods are changing names.

joj0s (approved):

alexanderrichards (approved):

PR #2039, opened 2022-07-14
Autoformat codebase according to PEP8

joj0s:
This PR includes autoformatting for all non-externally contributed Ganga packages. Note that this doesn't mean that all files are now entirely PEP8 compliant. Only auto-fixable issues are included


egede:
Very encouraging that all tests are passing. It is a huge amount of code to review though. Most of the changes are completely trivial but I am a bit worried about the reordering of import statements. In most (all?) this is of course trivial but as we do things like creating globals and dynamically create new classes (the proxy classes in the GPI namespace), there is no guarantee that reordered imports will work.

We will also have to think about how we want to go ahead in the future. If we do nothing, the codebase will just drift again, so do we require compliance for any new PRs? If yes, how do we avoid creating PRs with mixed purposes like @joj0s warned against?

joj0s:
@egede that comment about circular imports is a very good one. When running autopep8 I disabled rule E402 that states all imports should be on top of file.

Note that before I did that the tests were failing. So the only import moving and sorting that should have happened is at the top level.

joj0s:
I have checked the changelog and I can confirm that the only import reordering that happens within this PR is top level.

egede (approved):

PR #2040, opened 2022-07-20
Asyncio compatible tests

### LINKED ISSUE #2036: Update tests to be compatible with asyncio
The current tests and testing utilities such as [run_until_state](https://github.com/ganga-devs/ganga/blob/d34b1c33558e43b1f93ae7621fa84a1618546186/ganga/GangaCore/testlib/monitoring.py#L13) are not compatible with the new asyncio based monitoring services. They should be updated to not fail using the new implementation.

joj0s:
The problems that caused the tests to fail with the new monitoring sevice were the following:

1. The test runner used the old service's disable/enable monitoring methods. I recreated this functionality for the new service. However, stopping the event loop would cause the thread to die. So I am using a simple `enabled` flag and stopping the `check_active_backends` method that runs continuously when that flag is `False`.
2. When running in the testing context, the running event loop wasn't automatically detected by other threads. To solve this, I added the `run_monitoring_task` method to the service which the rest of the services can use as a simple interface to send their monitoring tasks with. This can also be used to also choose different executors such as threads and processes.

### DIFF
--- a/ganga/GangaCore/Core/InternalServices/Coordinator.py
+++ b/ganga/GangaCore/Core/InternalServices/Coordinator.py
@@ -92,15 +92,9 @@ def _diskSpaceChecker():
 
 
 def disableMonitoringService():
-
-    # disable the mon loop
-    log.debug("Shutting down the main monitoring loop")
-    from GangaCore.Core.MonitoringComponent.Local_GangaMC_Service import _purge_actions_queue, stop_and_free_thread_pool
-    _purge_actions_queue()
-    stop_and_free_thread_pool()
     log.debug("Disabling the central Monitoring")
     from GangaCore.Core import monitoring_component
-    monitoring_component.disableMonitoring()
+    monitoring_component.disable()
 
 
 def disableInternalServices():
@@ -155,11 +149,7 @@ def disableInternalServices():
 
 def enableMonitoringService():
     from GangaCore.Core import monitoring_component
-    monitoring_component.alive = True
-    monitoring_component.enableMonitoring()
-    from GangaCore.Core.MonitoringComponent.Local_GangaMC_Service import _makeThreadPool, ThreadPool
-    if not ThreadPool or len(ThreadPool) == 0:
-        _makeThreadPool()
+    monitoring_component.enable()
     global servicesEnabled
     servicesEnabled = True
 
--- a/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
+++ b/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
@@ -75,6 +75,22 @@ class AsyncMonitoringService(GangaThread):
         for task in scheduled_tasks:
             task.cancel()
 
+    def disable(self):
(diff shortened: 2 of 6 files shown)


joj0s:
The tests are now all passing except for the integration ones. The issue is this: I am using asyncio's `all_tasks` method to collect all scheduled tasks and cancel them when necessary. The problem is that [from python 3.7 and above](https://github.com/python/cpython/issues/76431) this method uses a different signature so it will always fail on Python 3.6 (the integration tests run on Python 3.6.8). What would you suggest?

egede:
> The tests are now all passing except for the integration ones. The issue is this: I am using asyncio's `all_tasks` method to collect all scheduled tasks and cancel them when necessary. The problem is that [from python 3.7 and above](https://github.com/python/cpython/issues/76431) this method uses a different signature so it will always fail on Python 3.6 (the integration tests run on Python 3.6.8). What would you suggest?

python 3.6 in fact passed its EOL at the end of large year so we can in fact upgrade the minimum requirement. As python 3.7 was never used much, I think that we could actually upgrade the requirement to python 3.8.

egede:
https://endoflife.date/python

egede:
If you pull in changes from the develop branch, you should get the tests moved to python 3.8

joj0s:
@egede An update on this PR. The Python version update fixed the previous issues. Other tests are now failing. The reason is that currently only the local backend has had its `updateMonitoringInformation` method converted to async, and other backends that use [run_until_state](https://github.com/ganga-devs/ganga/blob/3f20173e6ed2104a3e6650b459176a9d37655fee/ganga/GangaCore/testlib/monitoring.py#L12) for their testing now expect an async method or they fail.

In order to fix that we decided to go ahead and work on implementing running synchronous monitoring tasks in thread pools as discussed previously.

ganga/GangaCore/Core/InternalServices/Coordinator.py
```diff
@@ -92,15 +92,9 @@ def _checker(): return True
...
 def disableMonitoringService():
-
-    # disable the mon loop
-    log.debug("Shutting down the main monitoring loop")
-    from GangaCore.Core.MonitoringComponent.Local_GangaMC_Service import _purge_actions_queue, stop_and_free_thread_pool
-    _purge_actions_queue()
-    stop_and_free_thread_pool()
     log.debug("Disabling the central Monitoring")
     from GangaCore.Core import monitoring_component
-    monitoring_component.disableMonitoring()
```
egede, line 103: Should this `disableMonitoring` method be removed?

joj0s: It is replaced by the `monitoring_component.disable()` method on the next line. Its new implementation is [here](https://github.com/ganga-devs/ganga/blob/6f23ac6c0cb9406d978d93a3993bfad5ebe331d1/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py#L94)

egede:
All good.

PR #2041, opened 2022-07-21
Change minimal python version to 3.8

egede:


mesmith75 (dismissed):

.github/workflows/ci_push_testing.yml
```diff
@@ -116,13 +110,11 @@ jobs:
...
     runs-on: ubuntu-latest
-    container: centos:7
+    container: centos/python-38-centos7
     steps:
     - uses: actions/checkout@v1
-    - name: Set up Python 3
-      run: yum install -y python3
     - name: Install git
-      run: yum install -y git
+      run: sudo yum install -y git
```
mesmith75: Is the `sudo` necessary? It has caused a failure

egede: Didn't work without, didn't work with it! I will interactively jump into container and try to work out what the problem is. Even if the python38 container is created by centos, it is a bit different to the standard centos7 container.

egede:
OK, should all be good now. Reverted to using the standard centos7 image as this means the same base image is used everywhere. GangaGUI is fragile as it depends on a very specific set of versions for the Flask related stuff.

alexanderrichards (approved):

mesmith75 (approved):

PR #2043, opened 2022-07-26
Update check-new-ganga.py

mesmith75:


mesmith75:
I'll change this not to hard-code the python version

alexanderrichards (dismissed):

alexanderrichards (approved):
I assume this will always choose the latest installed version?

mesmith75:
There should only be one python version in the install directory so this will pick whatever is there

PR #2047, opened 2022-08-05
Dummy remote backend for benchmarking

joj0s:
This PR adds a new testing backend to run benchmarks and possibly test network issue handling. It is a local backend clone that uses a Flask server to simulate status and outputfile information retrieval. The Flask server runs in a background thread and it shuts down when the app exits.

Only thing I am not sure about is handling the logging output. I thought it would make sense to print Flask events to the console so that you can see it as you are testing but I can switch to regular logging to file if you'd like.


ganga/GangaTest/Lib/TestRemoteBackend/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+config.addOption("SERVER_DEFAULT_DELAY", 0.1, "Artificial delay in seconds")
+config.addOption("ENABLE_FINALISATION", True,
+                 "Whether or not to simulate job finalisation via dummy output file download.")
+config.addOption("FINALISATION_DELAY", 10, "Job finalisation simulation delay in seconds.")
+
+
+from .server import DummyServer
+
+server = DummyServer()
+server.start()
```
egede: This is what is causing the database suite to fail if you look carefully at the output. The chain of events is something like
- The database testing happens inside a dedicated container
- When a test inherits from `GangaUnitTest` it is triggering the loading of the `GangaTest` package.
- This in turn causes the `DummyServer` to be started.
- For some reason this fails inside the container we use for DB testing. In itself an unrelated error that I do not think is related to Ganga.

I think the solution would be to delay the start of the `DummyServer` until it is actually required (i.e. that a job with that backend is created). Art the moment ANY test that inherits from `GangaUnitTest` will cause the server to start.

egede (changes requested):

ganga/GangaTest/Lib/TestRemoteBackend/__init__.py
joj0s: Right, I will move the server initialization and resubmit the changes.

joj0s:
@egede I made the server a Singleton by using a metaclass and moved the start process into the backend's init method. Tests are now passing,

joj0s:
Should I also fix the linter issues before merging? Keep in mind that linting output does not show up in this PR at the moment until we merge [PR #2048](https://github.com/ganga-devs/ganga/pull/2048)

ganga/GangaCore/Lib/Executable/Executable.py
```diff
@@ -290,6 +290,7 @@ def prepare(self, app, appconfig, appmasterconfig, jobmasterconfig):
 allHandlers.add('Executable', 'LCG', LCGRTHandler)
 allHandlers.add('Executable', 'gLite', gLiteRTHandler)
 allHandlers.add('Executable', 'TestSubmitter', RTHandler)
+allHandlers.add('Executable', 'DummyRemote', RTHandler)
```
egede: This line (and actually the one above as well) sits in the wrong place. The plugin structure of Ganga means that the GangaCore package should not need to know anything about the `GangaTest` package. Look how this is done in [`GangaDirac`](https://github.com/ganga-devs/ganga/blob/3ec76823744452a91edd82c72552f1ca2204be14/ganga/GangaDirac/Lib/RTHandlers/ExeDiracRTHandler.py#L216).

egede (commented):
Are the two template files actually identical to the ones from the `Local` backend. If yes, then we should be able to just take them from there instead - It is fine for `GangaTest` to depend on `GangaCore`, it is just the other direction that we do not like.

ganga/GangaCore/Lib/Executable/Executable.py
joj0s: Right, I will fix the above line as well then

joj0s:
> Are the two template files actually identical to the ones from the `Local` backend. If yes, then we should be able to just take them from there instead - It is fine for `GangaTest` to depend on `GangaCore`, it is just the other direction that we do not like.

Yes they are, in that case I will import them from GangaCore.

egede:
> Should I also fix the linter issues before merging? Keep in mind that linting output does not show up in this PR at the moment until we merge [PR #2048](https://github.com/ganga-devs/ganga/pull/2048)

No need to.

egede (approved):

egede:
@joj0s Just merge this whenever you like.

PR #2050, opened 2022-08-11
add check of uploaded Dirac proxy

### LINKED ISSUE #2049: Check for expiration of uploaded Dirac certificate
You can still submit jobs with a certificate that is about to expire - these then fail on the grid. Add a check for the expiry date.

mesmith75:
For submitting Dirac jobs this checks that the uploaded proxy is valid for at least 3 weeks as requested by the LHCbDirac people.

Fixes #2049


mesmith75:
Seems it has a problem with line breaks before and after the `and`

egede (approved):

PR #2052, opened 2022-08-12
Amend web location

### LINKED ISSUE #2051: Obsolete URL for documentation and support
Just a small thing, but the print out when starting Ganga includes a line referencing a URL that's no longer valid:
```
```
This is coming from [GangaCore.Runtime.bootstrap](https://github.com/ganga-devs/ganga/blob/50902fa181aec8c6eff5dbecec5b676589fbb81a/ganga/GangaCore/Runtime/bootstrap.py#L229).  The same URL shows up in a few other places (try repository search for "documentation and support"), I think all in the user documentation.

It could be useful for the URL to be upated.

mesmith75:
Changed the printout at startup to point to readthedocs and github to report issues.

Also changed the location in the comments of a few files.

I noticed some files have the license in them at the start - I would advocate removing that except for the `bootstrap.py` or `bin/ganga`

Fixes #2051


mesmith75:
To fix the linting I have had to make some fairly significant changes, particularly to `bootstrap.py`. In particular some of the logic was confusing - hopefully I have got it right with the changes I made but it may be worth another set of eyes. There also seemed to be some extraneous code that I have removed.

~Also, it still complains about some unused imports - I am not sure there is anything we can do about them as they test for the existence of IPython.~

Actually you can test for module existence. There is just one unused import on line 896 of `bootstrap.py` that I am not sure about as it wants to know the error if the import fails (it catches all exceptions).

egede (approved):

PR #2053, opened 2022-08-12
Ignore W503 in linting test

mesmith75:
I propose we explicitly ignore W503 as it is apparently outdated. We cannot have both W504 and W503 - they seem mutually exclusive but we are picking up both.


joj0s:
The rest of the flake8 config was moved to [setup.cfg](https://github.com/ganga-devs/ganga/blob/develop/setup.cfg) so I believe it would be best to move this rule there as well.

joj0s (approved):

egede (approved):

PR #2054, opened 2022-08-15
Update check-new-ganga.py

mesmith75:


egede (dismissed):
I assume that this is to fix the problem that happened with the release script overnight?

mesmith75:
yes - although that commit still wouldn't have worked. Now it should be done.

egede (approved):

PR #2056, opened 2022-08-16
Add autopep8 PR creation to CI

joj0s:


mesmith75:
If you are using autopep8 I think you need to set the max line length to match that in flake8 - I am not sure they are the same. Sorry if you already have and I missed it.

joj0s:
According to [the docs](https://github.com/hhatto/autopep8#configuration), autopep8 should read the info about rules and line length from the flake8 section in our setup.cfg file.

mesmith75:
ah very good

egede (approved):
Looks great.

mesmith75 (approved):

PR #2058, opened 2022-08-19
Fix CustomMerger documentation

### LINKED ISSUE #2021: CustomMerger documentation examples not updated to Python 3
The  `mymerger.py` template code in the [documentation](https://ganga.readthedocs.io/en/latest/UserGuide/PostProcessors.html) uses the old file() i/o function from Python2.7 namespace. Changing it to `open` would allow it to run with the newer versions as well as preserve backward compatibility if necessary.

mesmith75:
Fixes #2021

### DIFF
--- a/doc/UserGuide/PostProcessors.rst
+++ b/doc/UserGuide/PostProcessors.rst
@@ -93,8 +93,6 @@ In ``mymerger.py`` you must define a function called mergefiles(file_list,output
     :end-before: # -- POSTPROCESSORS MYMERGER STOP
     :dedent: 8
 
-.. code-block:: python
-
 
 This function would mimic the TextMerger, but with more control to the user. Note that the ``overwrite`` and
 ``ignorefailed`` flags will still work here as a normal merger object.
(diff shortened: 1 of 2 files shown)


egede (dismissed):

egede:
@joj0s I merged in the auto pep8 PR #2059 into here, but now the first check ("Linting") is failing. Do you understand why? It is not a real issue, and we can just merge it anyway, but it would be nice if everything was "green" if the correct workflow was followed.

joj0s:
@egede autopep8 can't fix every single pep8 violation automatically, even with aggressive changes. 

For example, it can't fix things like defining a variable and then not using it, so If there are issues like that then the linting check will still fail. I am not sure which specific issue there is still here though, maybe I should improve the output or restore the flake8 annotations we used?

egede:
Ahh, I see. I thought the failure was some technical problem and not just a linting failure. 

If the failure output can be restored in the automatically generated PR, it would be great. If we then do not fix it, we are at least aware of what we do not fix!

joj0s:
So I think the solution would be to run the linting check on the autopep8 PR but the thing is that GitHub doesn't let you run workflows on automated PRs to avoid endless recursive workflow runs. I will try [this workaround](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow) on my fork and let you know.

egede (approved):

PR #2061, opened 2022-08-31
cli

gurudattapatil:


gurudattapatil:
To use this use shell in bin/gui-terminal
Currently terminal runs on 127.0.0.1:11000, but working on configuring it to make it run on free port

egede (commented):
We requested a few more changes last Monday. Are the coming in?

gurudattapatil:
yes half of it is done, facing issue adding extra argument to the command, don't know exactly where the error is coming from.

gurudattapatil:
@egede Developed the whole thing, command would be like this 
`ganga-gui-terminal --port=13000 --flask=18000 --internal-port=24000`
The height of the terminal can be changed from [here](https://github.com/ganga-devs/ganga/blob/8e0c77443bea38d1549b1da4aab86c74d5c3166f/ganga/GangaGUI/gui/templates/cli1.html#L24)

gurudattapatil:
Will add the documentation this week

PR #2066, opened 2022-09-27
Implement variable monitoring poll rate for each backend

joj0s:
This PR adds the ability to use different poll rates to check on each backend. It uses the old monitoring service's logic, where there is a short base poll rate to check for active backends and a different poll rate for each active backend.

I also added a wrapper around the monitoring tasks to print deubugging information when errors happen during development.

Also please check out the commented out code on line 150. I thought about adding a tear down method to each backend that can be used when we want to tear down its monitoring setup (for example kill DIRAC subprocesses when there are no active DIRAC jobs to monitor left). If you agree with this idea I will add that to the IBackend interface.

### DIFF
--- a/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
+++ b/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
@@ -12,7 +12,7 @@ from GangaCore.Utility.logging import getLogger
 
 config = getConfig("PollThread")
 THREAD_POOL_SIZE = config['update_thread_pool_size']
-POLL_RATE = 2  # in seconds
+POLL_RATE = config['base_poll_rate']  # in seconds
 log = getLogger()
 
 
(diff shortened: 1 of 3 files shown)


ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
```diff
@@ -114,7 +114,10 @@ def _check_backend(self, backend):
 
     def _schedule_next_backend_check(self, backend):
         backend_name = getName(backend)
-        pRate = config.get(backend_name, config['default_backend_poll_rate'])
+        if backend in config:
+            pRate = config[backend_name]
+        else:
+            pRate = config['default_backend_poll_rate']
```
egede, line 120: All backends that we know about should have a corresponding config. So I do not think this check is required?

egede (approved):

ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
joj0s: @egede There are a couple of testing backends like the one I created and TestSavannah which I didn't find in the `PollThread` config. I would say it is good to have a fallback even it is redundant 99% of the time but if you would prefer it I can remove that check.

That being said, there was another issue with this code. Using the `default_backend_poll_rate` as a fallback (which is 30 seconds), caused [this integration test](https://github.com/ganga-devs/ganga/blob/229dc06a38e6093a85f0bcda54b9126d80e76a50/ganga/GangaCore/test/GPI/TestTimeStamps.py#L157) to fail and I am not sure why ([here are the workflow logs](https://github.com/ganga-devs/ganga/actions/runs/3136206338/jobs/5092866735)). The test now passes by using the `base_poll_rate` which is 2 seconds as a fallback instead. So do the GPI tests use a different backend to submit to? 

In any case I will try decreasing the base poll rate and see if that fixes it.

joj0s: Yes, decreasing the base poll rate to 15 seconds did fix that test. Should I leave it in as the default fallback poll rate?

egede (approved):

egede (approved):

PR #2069, opened 2022-09-28
New DIRAC concurrency model

joj0s:
This PR includes the following:
- A new DIRAC concurrency model using a subprocess with an internal thread pool
- The ability to use different environments and credential requirements within it
- Moving DIRAC status fetching to job finalisation


joj0s:
The only missing piece here is to make sure that the spawned DIRAC subprocesses are shut down when they are no longer needed. Right now they will remain open until Ganga exits.

My suggested solution would be to implement a universal hook that is called when a backend is no longer active. This would be added in PR #2066 as you can see in [the commented code here](https://github.com/ganga-devs/ganga/blob/75b0b503a8521b99986889be9dab992d833e66f3/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py#L143). Afterwards I could add the tear down logic there. Let me know if you think another way to do that is better.

alexanderrichards (approved):

egede (approved):

PR #2075, opened 2022-10-17
Add DIRAC executor process tear down

joj0s:
This PR adds the functionality to terminate a DIRAC process when there are no more jobs to monitor.

*Please note and review the following:*
I added a global signal handler that throws a `TerminationSignalException` whenever a SIGTERM is received. Since this is a change that can be used within the whole core codebase let me know if I should change anything. Specifically I am unsure about where in the code I should instantiate the signal handler (right now I am doing it in the bootstrap method) and whether the `TerminationSignalException` should be registered as a `GangaException` or a regular one.

### DIFF
--- a/ganga/GangaCore/Core/exceptions.py
+++ b/ganga/GangaCore/Core/exceptions.py
@@ -194,7 +194,8 @@ class GangaKeyError(GangaException, KeyError):
 
 class GangaTypeError(GangaException, TypeError):
     """
-    Class analogous to GangaKeyError. This class wraps TypeError so that users are prevented from seeing stack traces from known good exceptions thrown in Ganga code.
+    Class analogous to GangaKeyError. This class wraps TypeError so that users are prevented from seeing stack traces
+    from known good exceptions thrown in Ganga code.
     """
 
     def __init__(self, *args, **kwds):
--- a/ganga/GangaCore/Core/exceptions.py
+++ b/ganga/GangaCore/Core/exceptions.py
@@ -149,7 +149,7 @@ class RepositoryError(GangaException):
             logger.error("Shutting Down Repository_runtime")
             from GangaCore.Runtime import Repository_runtime
             Repository_runtime.shutdown()
-        except:
+        except BaseException:
             logger.error("Unable to disable Internal services, they may have already been disabled!")
 
 
--- a/ganga/GangaCore/Core/exceptions.py
+++ b/ganga/GangaCore/Core/exceptions.py
@@ -203,5 +203,5 @@ class GangaTypeError(GangaException, TypeError):
         self.kwds = kwds
 
 
-class TerminationSignalException(GangaException):
+class TerminationSignalException(Exception):
     pass
--- a/ganga/GangaCore/Core/exceptions.py
+++ b/ganga/GangaCore/Core/exceptions.py
@@ -201,3 +201,7 @@ class GangaTypeError(GangaException, TypeError):
         super(GangaException, self).__init__(args)
         TypeError.__init__(self, *args)
         self.kwds = kwds
+
+
+class TerminationSignalException(GangaException):
+    pass
--- a/ganga/GangaDirac/Lib/Backends/DiracBase.py
+++ b/ganga/GangaDirac/Lib/Backends/DiracBase.py
@@ -1298,7 +1298,6 @@ class DiracBase(IBackend):
         theseJobs = []
 
         if not monitoring_component.enabled:
-            print('NOPE')
             return
         for sj in allJobs:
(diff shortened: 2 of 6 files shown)


ganga/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -1293,6 +1298,7 @@ def finalise_jobs(allJobs, downloadSandbox=True):
         theseJobs = []
 
         if not monitoring_component.enabled:
+            print('NOPE')
```
alexanderrichards: do we want logging messages rather than a raw print?

ganga/GangaDirac/Lib/Backends/DiracBase.py
```diff
@@ -1311,6 +1317,10 @@ def finalise_jobs(allJobs, downloadSandbox=True):
         jobs = [stripProxy(j) for j in theseJobs]
         manager = AsyncDiracManager()
         for job in jobs:
+            print(f'Attempting to add from DIRAC task for {finished_job}')
+            if not monitoring_component.enabled:
+                print('NOPERZ')
```
alexanderrichards: again raw prints rather than logging messages.

joj0s: Yeah I left them in by accident when testing out, removed now, thanks for noticing!

ganga/GangaCore/Core/exceptions.py
```diff
@@ -201,3 +201,7 @@ def __init__(self, *args, **kwds):
         super(GangaException, self).__init__(args)
         TypeError.__init__(self, *args)
         self.kwds = kwds
+
+
+class TerminationSignalException(GangaException):
```
egede: A `GangaException` is an exception that can occur during the normal operation of `ganga` and will in general be passed back to the user uncaught. So I do *not* think this should be a `GangaException`

egede (commented):
How does this overlap with the `atexit` handler defined in ` Core/InternalServices/ShutdownManager.py`?

ganga/GangaCore/Core/exceptions.py
joj0s: I see, that makes sense. I will update it to subclass a base exeption in that case

joj0s:
> How does this overlap with the `atexit` handler defined in ` Core/InternalServices/ShutdownManager.py`?

@egede Please correct me if I am wrong but from what I understand, everything in the `ShutdownManager.py` module is meant to handle the graceful shutdown process of the main Ganga app. In our case, what we do is gracefully terminate created DIRAC subprocesses when we need to and the Shutdown Manager is not involved in any way in that since the main app continues to run as expected, so I don't see any overlap.

egede:
> @egede Please correct me if I am wrong but from what I understand, everything in the `ShutdownManager.py` module is meant to handle the graceful shutdown process of the main Ganga app. In our case, what we do is gracefully terminate created DIRAC subprocesses when we need to and the Shutdown Manager is not involved in any way in that since the main app continues to run as expected, so I don't see any overlap.

All good.

ganga/GangaDirac/Lib/Server/DiracProcessManager.py
```diff
@@ -91,18 +92,26 @@ async def execute(self, cmd, args_dict=None, return_raw_dict=False, env=None, cr
...
+        task_id = uuid.uuid4()
+        task_done = self.manager.AioEvent()
+        await self.task_queues[env_hash].coro_put((task_done, task_id, cmd, args_dict))
+
+        await task_done.coro_wait()
+        dirac_result = self.task_result_dicts[env_hash].get(task_id)
+        del self.task_result_dicts[env_hash][task_id]
+
+        returnable = self.parse_command_result(dirac_result, str(cmd), return_raw_dict)
+        print(f'Executed DIRAC command {cmd} with result {returnable}')
```
egede: Any reason for the direct print statement here rather than a logger?

egede (approved):

ganga/GangaDirac/Lib/Server/DiracProcessManager.py
joj0s: Looks like I didn't push the latest changes, hence the commented out code, will fix it, thanks for catching that!

joj0s:
Also, it looks like there a bunch of PEP8 violations which are not auto fixable. If you look at the linting logs, there are a lot of unused imports in `ganga/GangaCore/Core/__init__.py` . Do you want me to fix those as well before merging this?

Edit: On second thought, I think I should fix those after I submit the PR for the `runMonitoring` method because a few of those unused imports come from that commented code.

Issue #2077, opened 2022-10-20
udocker installation test is failing


egede:
The udocker installation test is failing. This is actually due to a bug in the code which does not deal correctly with a tarball not downloading.
```
(ganga) [laptop] ~/programming/ganga/ganga develop % pytest ganga/GangaCore/test/Unit/Utility/TestUtilitiesVirtualizationInstallUDocker.py
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.8.14, pytest-6.2.4, py-1.10.0, pluggy-0.13.1 -- /home/egede/programming/ganga/bin/python
cachedir: .pytest_cache
rootdir: /home/egede/programming/ganga/ganga, configfile: setup.cfg
plugins: pylint-0.18.0, mock-3.6.1, cov-2.12.1
collected 2 items                                                                                                                                                                                                                            

ganga/GangaCore/test/Unit/Utility/TestUtilitiesVirtualizationInstallUDocker.py::TestInstallUDocker::test_installUDocker_download_fail PASSED                                                                                           [ 50%]
ganga/GangaCore/test/Unit/Utility/TestUtilitiesVirtualizationInstallUDocker.py::TestInstallUDocker::test_installUDocker_success FAILED                                                                                                 [100%]

================================================================================================================== FAILURES ==================================================================================================================
_______________________________________________________________________________________________ TestInstallUDocker.test_installUDocker_success _______________________________________________________________________________________________

self = <TestUtilitiesVirtualizationInstallUDocker.TestInstallUDocker testMethod=test_installUDocker_success>, mock_subprocess_call = <MagicMock name='call' id='139888939321424'>

    @patch('subprocess.call')
    def test_installUDocker_success(self, mock_subprocess_call):
        mock_subprocess_call.side_effect = [0, 0]
        with tempfile.TemporaryDirectory() as dir:
>           installUdocker(dir)

ganga/GangaCore/test/Unit/Utility/TestUtilitiesVirtualizationInstallUDocker.py:15: 
[...]

PR #2078, opened 2022-10-20
Fix udocker

egede:
Solve problem in #2077.

### DIFF
--- a/.github/workflows/ci_push_testing.yml
+++ b/.github/workflows/ci_push_testing.yml
@@ -96,10 +96,14 @@ jobs:
           yum install -y rh-python38-python rh-python38-python-pip
       - name: Install dependencies
         run: |
-          /opt/rh/rh-python38/root/usr/bin/python -m pip install --upgrade pip setuptools wheel
-          /opt/rh/rh-python38/root/usr/bin/python -m pip install -e .[dev]
+          /opt/rh/rh-python38/root/usr/bin/python -m venv ~/venv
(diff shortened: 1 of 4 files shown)


joj0s:
It looks like udocker installation still fails in LHCb integration and Database controller tests. Maybe instead of using a hardcoded URL it would be better to install it via PyPI or git? Reference: https://indigo-dc.github.io/udocker/installation_manual.html

egede:
> It looks like udocker installation still fails in LHCb integration and Database controller tests. Maybe instead of using a hardcoded URL it would be better to install it via PyPI or git? Reference: https://indigo-dc.github.io/udocker/installation_manual.html

Yes, I agree. I think the LHCb integration tests are failing for a different reason though.

joj0s:
>  I think the LHCb integration tests are failing for a different reason though.

Indeed. I had a similar problem installing DIRAC when following the CI instructions, because they still use the old Python 2 installation. If you see the error message, it says that the Python 2 installation is no longer supported and that it cannot find `dirac_ui/bashrc`. In the newer installations, the source file has been moved to `/dirac_ui/diracos/diracosrc`. I believe installing DIRAC like how Alex does it in his [docker image](https://github.com/alexanderrichards/dirac_ui/blob/16a56df347cb3011deb6434425eb4379a37cc206/Dockerfile#L8) should fix this

alexanderrichards (approved):

PR #2100, opened 2023-02-04
Remove stale plugins fixes #2099

### LINKED ISSUE #2099: Remove plugins that have never been migrated to python3
The plugins have had plenty of time to upgrade to python3 now. So the ones that have not been updated should be regarded as stale and be removed from the `develop` branch. They can always be restored if required.

Plugins are located as directories [here](https://github.com/ganga-devs/ganga/tree/develop/ganga) and it should be identified which ones were never converted to python3. Candidates for removal are

- GangaAtlas
- GangaLSST
- GangaNA62
- GangaPanda
- GangaRobot
- GangaSAGA
- GangaService
- GangaSNOplus
- GangaSuperB

In each of them it should be checked if they indeed have never been updated to `python3` - properly best done by looking at the latest commits to the files.

abhijeetsharma200:
Deleted 6 Plugins which were incompatible with Python3.


egede:
A few more plugins were suggested to investigate for deletion. Did you conclude that they were in fact updated to python3 or did you just not look yet?

abhijeetsharma200:
@egede, I incorrectly made the assumption that the GangaLSST, GangaRobot and GangaSAGA have been migrated. The reasoning for the error is the following:

The latest significant code update to GangaLSST was on May 13, 2016, as seen in this [commit](https://github.com/ganga-devs/ganga/commit/f197fbef10a587bad21172c0c7bf3c791fc3b49f). The usage of except Exception as err in this commit led me to believe that the code has been migrated to Python3. This was a mistake, as "except Exception as err" is backwards compatible with Python 2.6 and 2.7.

The similar erroneous assumption that  GangaRobot and GangaSAGA were migrated to Python3 was solidified by the [commit](https://github.com/ganga-devs/ganga/commit/0c92fc3f360c5a5afa8372efc71eebcd4e02682f), which "fixes the uses of commas in 'except' and 'raise' statements" in GangaRobot and GangaSAGA among other files. And another [commit](https://github.com/ganga-devs/ganga/commit/b04375d409f6982cb41222bca80421333270cfe5), which changes the usage from "print" to "print()" in GangaRobot also lead me to believe that the plugins have been migrated.

I have realised my mistake and have updated the PR accordingly. Please let me know if this is fine.

ganga/GangaCore/__init__.py
```diff
@@ -891,7 +891,7 @@ def filefilter(fn):
     pwd_nam = pwd.getpwnam(user)
 
 groupid = grp.getgrgid(pwd_nam.pw_gid).gr_name
-groupnames = {'z5': 'lhcb', 'zp': 'atlas', 'zh': 'cms', 'vl': 'na62'}
```
egede: You might as well keep `atlas` and `na62` here.

ganga/GangaCore/__init__.py
```diff
@@ -903,12 +903,10 @@ def filefilter(fn):
 
 # From:
 # http://eos.cern.ch/index.php?option=com_content&view=article&id=87:using-eos-at-cern&catid=31:general&Itemid=41
-protoByExperiment = {'atlas': 'root://eosatlas.cern.ch',
-                     'cms': 'root://eocms.cern.ch',
+protoByExperiment = {'cms': 'root://eocms.cern.ch',
                      'lhcb': 'root://eoslhcb.cern.ch',
                      'alice': 'root://eosalice.cern.ch',
                      # These last 2 are guesses based on the standard
-                     'na62': 'root://eosna62.cern.ch',
```
egede: Also keep `atlas` and `na62` here.

egede (dismissed):
Looks good and essentially ready for merging. I left two very minor comments.

egede:
Testing reveals that we have not got the cleaning done quite correctly. I see
```
GangaCore/Utility/Runtime.py:175: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

config = {}

    def loadPlugins(config=None):
        import GangaTest.Framework.runner
        import GangaTest.Lib.GListApp
        import GangaTest.Lib.TFile
        import GangaTest.Lib.TestApplication
        import GangaTest.Lib.TestSubmitter  # TestSubmitter, TestSplitter here
>       import GangaTest.Lib.TestRobot
E       ModuleNotFoundError: No module named 'GangaTest.Lib.TestRobot'

GangaTest/__init__.py:13: ModuleNotFoundError
```
If this is the cause of all the errors or if there are other problems I do not know. You can run the tests locally by installing Ganga with the `[dev]` option, so `python -m pip install -U .[dev]` when you are in the directory where the `setup.py` file is. Then do something like
```
cd ganga/ganga
pytest GangaCore/test/Unit
```

abhijeetsharma200:
Hey @egede, you are right. I missed TestRobot and have fixed it. The tests are passing locally as well.

ganga/GangaCore/__init__.py
abhijeetsharma200: Made the changes

ganga/GangaCore/__init__.py
abhijeetsharma200: Made the changes as requested

egede (approved):
This is all good now. I can take over the final integration.

PR #2106, opened 2023-02-10
Change to lhcb-proxy-info

mesmith75:
Not sure there was a reason to not be using it. It is better as it is part of the standard LHCb environment so you don't need to execute it inside the Dirac one.


egede (approved):

mesmith75:
No, this doesn't work
```python
WARNING  exit status [127] of command lhcb-proxy-info
WARNING  command lhcb-proxy-info  not found
```
I guess this is being done inside the Dirac environment. Although I am not sure I understand how the `lhcb-proxy-init` then works OK.

PR #2118, opened 2023-02-23
Removed function as defined in Issue 2109

### LINKED ISSUE #2109: Broken method in Job class
The method [_outputfieldCopyable](https://github.com/ganga-devs/ganga/blob/45d523f0895c2992a4c9bc5aba1fae6b5a3867f8/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L138) in the Job class is manifestly broken as it doesn't return anything and has no side effects. So the method should just be removed.

A bit of care should be taken as it is used twice in the [schema](https://github.com/ganga-devs/ganga/blob/45d523f0895c2992a4c9bc5aba1fae6b5a3867f8/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L207) of the Job class. These two places will have `None` assigned where in principle a bool is expected. I think that False can just be assigned instead.

abhijeetsharma200:
Fixed Issue 2109

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -168,15 +168,6 @@ class JobInfo(GangaObject):
         self.submit_counter += 1
 
 
-def _outputfieldCopyable():
-    if 'ForbidLegacyOutput' in getConfig('Output'):
-        if getConfig('Output')['ForbidLegacyOutput']:
-            outputfieldCopyable = 0
-    else:
-        outputfieldCopyable = 1
-    return outputfieldCopyable
-
-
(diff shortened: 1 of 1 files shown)


abhijeetsharma200:
@egede It looks like the cause of the tests is the unavailable X509 certificate. How do I go about getting the certificate so I don't have to trouble you to run these tests

egede:
> @egede It looks like the cause of the tests is the unavailable X509 certificate. How do I go about getting the certificate so I don't have to trouble you to run these tests

I changed your status in github such that you no longer have to create a fork. So please create a pull request without the fork. Then the testing should run as expected.

abhijeetsharma200:
Alright. I will do that. Let me close this PR then and make a new one

PR #2119, opened 2023-02-23
Remove LegacyOutput option for jobs

### LINKED ISSUE #2109: Broken method in Job class

abhijeetsharma200:
Fixed issue 2109


egede (commented):
Can you check if the 'ForbidLegacyOutput' of the configuration is used elsewhere. If not, it should be removed. Just search for the name globally in the source.

mesmith75:
I don't think we want to do this. The function is not broken in the develop branch: https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L171

mesmith75 (changes requested):
The function here is returning a value as given by the config. I don't think we want to change this.

abhijeetsharma200:
> The function here is returning a value as given by the config. I don't think we want to change this.

@mesmith75 the function does have now have a return value but it didn't when the issue was created. I added the return value as this file was failing linting on this function with `Local Variable assigned but unused`.

So, I do not think this function does anything and can be removed without issues but as of right now it does give a boolean output which is desired.

mesmith75:
The function clearly does something. The file referenced in the issue is some odd commit but does not reflect what is in develop and has been for a long time.

mesmith75:
Ah ok, now I see, sorry for the noise

ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -2567,10 +2558,6 @@ def __setattr__(self, attr, value):
 
             if value != []:
 
-                if getConfig('Output')['ForbidLegacyOutput']:
```
mesmith75, line 2570: I would assume here we want to keep the error message by default. I think that `ForbidLegacyOutput` is by default `True` (at least going by the change you make by getting rid of the function above) - in which case we should keep reporting an error if a user tries to use the legacy output.

ganga/GangaCore/GPIDev/Lib/Job/Job.py
```diff
@@ -2593,10 +2580,6 @@ def __setattr__(self, attr, value):
 
             if value is not None:
 
-                if getConfig('Output')['ForbidLegacyOutput']:
```
mesmith75, line 2596: Same here

ganga/GangaCore/GPIDev/Lib/Tasks/ITransform.py
```diff
@@ -490,11 +646,6 @@ def __setattr__(self, attr, value):
 
             if value != []:
 
-                if getConfig('Output')['ForbidLegacyOutput']:
```
mesmith75, line 493: And here

abhijeetsharma200: Fair enough. I will fix it over the weekend.

egede:
Given that we are removing the whole ability to use `LegacyOutput` we should probably remove the entries in the `Job` schema as well. I believe that we can remove
- `inputsandbox`
- `outputsandbox`
- `outputdata`

The output below confirms that this is indeed the case
```python
[09:42:05]
Ganga In [14]: j.inputsandbox = ["sdfsd"]
ERROR    Use of job.inputsandbox is forbidden, please use job.inputfiles
ERROR    GangaException: Use of job.inputsandbox is forbidden, please use job.inputfiles

[09:42:10]
Ganga In [15]: j.outputsandbox = ["sdfsd"]
ERROR    Use of job.outputsandbox is forbidden, please use job.outputfiles

[09:42:19]
Ganga In [16]: j.outputdata = GangaDataset()
ERROR    Use of job.outputdata is forbidden, please use job.outputfiles
```

We should check through the codebase and clean up places where there are references to these schema attributes.

ganga/GangaDirac/Dirac.ini
```diff
@@ -25,6 +25,20 @@ creds_poll_rate = 24*3600
...
+        'mkdir_cmd': 'nsmkdir'
+    },
+    'backendPostprocess': {
+        'LSF': 'WN',
+        'LCG': 'client',
+        'Interactive': 'WN',
+        'Localhost': 'WN',
+        'CREAM': 'client', 'Dirac':'client'
+    }
+}
```
egede: See the GangaDirac testing. There is a passing error in this file now.

egede:
> Given that we are removing the whole ability to use `LegacyOutput` we should probably remove the entries in the `Job` schema as well. I believe that we can remove
> 
>     * `inputsandbox`
> 
>     * `outputsandbox`
> 
>     * `outputdata`
> 
> 
> The output below confirms that this is indeed the case
> 
> ```python
> [09:42:05]
> Ganga In [14]: j.inputsandbox = ["sdfsd"]
> ERROR    Use of job.inputsandbox is forbidden, please use job.inputfiles
> ERROR    GangaException: Use of job.inputsandbox is forbidden, please use job.inputfiles
> 
> [09:42:10]
> Ganga In [15]: j.outputsandbox = ["sdfsd"]
> ERROR    Use of job.outputsandbox is forbidden, please use job.outputfiles
> 
> [09:42:19]
> Ganga In [16]: j.outputdata = GangaDataset()
> ERROR    Use of job.outputdata is forbidden, please use job.outputfiles
> ```
> 
> We should check through the codebase and clean up places where there are references to these schema attributes.

I think we will do this as a separate PR. So the only thing required to update here is the failing test caused by the typo in `Dirac.ini`

egede (dismissed):

ganga/GangaCore/__init__.py
```diff
@@ -1,6 +1,6 @@
 # System Imports
 import os
-import re
+# import re
```
mesmith75: I think this should be deleted if the import is commented

egede (dismissed):

ganga/GangaCore/GPIDev/Lib/Tasks/ITransform.py
```diff
@@ -14,45 +14,197 @@
 import os
 from GangaCore.GPIDev.Lib.Tasks.ITask import addInfoString
 from GangaCore.GPIDev.Lib.Tasks.common import getJobByID
-from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
-from GangaCore.GPIDev.Lib.File.File import File
+# from GangaCore.GPIDev.Adapters.IGangaFile import IGangaFile
```
mesmith75: Similarly here

abhijeetsharma200:
I think this should be good to go. I will then create another PR for the changes suggested by @egede and @mesmith75

egede (approved):

egede:
@mesmith75 I think you need to approve this again to allow merging.

egede:
@mesmith75 I think you need to approve this again to allow merging.

mesmith75 (approved):

PR #2127, opened 2023-03-06
[Draft] Allow full_print to output in json

shalearkane:
#### What is `full_print_json`?
Instead of just outputting a description of the object like `full_print` or individual properties in a human-readable way, we output a JSON so that it can read by external tools.

#### Motivation
Since in future we will be enabling Ganga to be used in conjunction with other workflow management tools, we might need to share job info to and from Ganga. Instead of writing wrapper scripts that parse the output of Ganga, we can allow Ganga to export information in JSON which can then be consumed by any tool.

#### Advantages
- Straightforward to implement and does not touch core Ganga classes. So maintenance should be a non-issue.
- There will be no need to import Ganga classes in a script to run Ganga programmatically so that we can interact with the job, i.e. by `python script.py`. All of the details of the job is JSON exportable.
- There will be no need to write wrapper functions to parse the output of Ganga repl (since that is the only way/API to interact with jobs)

#### Disadvantages
- *Ganga already has internal APIs that we can use. So this function will make no sense once those are exposed.*
- Current implementation outputs every attribute for now. So it will be slow to evaluate for complex objects.
- Potentially useless function since the usefulness is only limited to interfacing with Ganga by other tools which for now is rarely used.
- Interactive mode is still present.
- Since this could become the standard for other tools to interface with Ganga, we might need to develop a JSON Schema so that random scripts do not break on new releases.

#### Current Approach to implement `full_print_json`(WIP)
Modify existing `full_print` command to output JSON. For this to be more machine-friendly instead of humans, we might export more information and we might drop indentation altogether. That way it can be fast to process and more useful for external tools.

#### Current usage
In my evaluation task, instead of getting the output in an ad-hoc manner, I am using `full_print_json` to get the information (JobID, JobStatus, OutputDirectory, etc) by calling `ganga script.py` where `script.py` is
```python
j = Job()
j.submit()
full_print_json(j)
```

I am currently trying out newer ways to interact with Ganga. This is one of the approaches that seems to be have some potential, so I am creating a draft pull request.

### DIFF
--- a/ganga/GangaCore/GPIDev/Base/Objects.py
+++ b/ganga/GangaCore/GPIDev/Base/Objects.py
@@ -29,7 +29,7 @@ from GangaCore.Utility.Profiling import cpu_profiler, mem_profiler, call_counter
 
 
 def _getName(obj):
-    """ Return the name of an object based on what we prioritise 
+    """ Return the name of an object based on what we prioritise
     Name is defined as the first in the list of:
     obj._name, obj.__name__, obj.__class__.__name__ or str(obj)
     Args:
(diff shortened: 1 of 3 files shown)


egede:
Please provide reasoning for opening or closing. Otherwise we will waste time reviewing.

shalearkane:
I closed the issue because I thought that I would refine the PR i.e. 
- remove the interactive mode code because that doesn't make sense 
- improve the formatting
- add a single-line mode so that a `json.loads(file.readline())` will be enough to parse

But later I decided to reopen because I thought that this idea should go through a discussion before refining because we may find other ways to achieve the same thing as I found out that there are APIs that provides JSON data.

egede:
We discussed at today's meeting. 
- Idea of output is JSON format is a good one. 
- We would prefer to see it  integrated into the existing `full_print` method. Maybe it could take a second argument like `full_print(j, format='json')`,`full_print(j, format='python')` and maybe in the future `full_print(j, format='yaml')`. The `python` one should be the default to keep backwards compatibility.
- There [is already some code](https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaGUI/api/routes.py) in Ganga dealing with serialisation into json. We should not reinvent the wheel. Maybe the code present in that file can be factored out to something that can go into the utils area of `GangaCore`.
- Note that the code referred to here allows for a server to be started that has a RESTful interface. That might be of interest as well.

egede:
@shalearkane Unless any statement on progress within the next few days, this will be closed.

PR #2132, opened 2023-03-15
Remove environment mangling

### LINKED ISSUE #2104: subprocess.Popen does not work with the centos 8 environment
The centos 8 environment has some odd functions defined that don't work with `subproccess.Popen`

mesmith75:
Fixes #2104

We should leave the environments as `os._Environ` objects. It stop the mangling of the environment that causes #2104


ganga/GangaCore/Utility/Shell.py
```diff
@@ -117,13 +117,11 @@ def __init__(self, setup=None, setup_args=[]):
         """
 
         if setup is not None:
-            self.env = dict(os.environ)
+            self.env = os.environ
             execute('source {0} {1}'.format(setup, " ".join(setup_args)), shell=True, env=self.env, update_env=True)
 
         else:
-            # bug #44334: Ganga/Utility/Shell.py does not save environ
```
alexanderrichards, line 123: I assume that the `bug #44334` is a reference to something very old that we don't care about.

alexanderrichards (dismissed):
This is interesting as on the face of it we seem to simply be copying the `os.environ` so I wouldn't have expected any difference between this and the new code where we use the `os.environ` directly. I imagine the reason this was do was to create a copy of the environment to avoid polluting it.

egede (commented):
Both `GangaDirac` and `GangaLHCb` tests are failing. The `GangaDirac` failures seems directly caused by the changes here. The `GangaLHCb` ones are hard to understand but might just be caused by the `Dirac` failures.

PR #2139, opened 2023-03-23
Avoid reading configuration file multiple times

vedanshbhartia:
The load_config_files function of GangaCore Config init would read system config variables in a loop, and set session variables in each loop cycle. This would cause the config file to be read multiple times. This commit moves the reading of user config and setting of session variables outside the loop.


mesmith75 (approved):

alexanderrichards (approved):

egede:
@vedanshbhartia Thanks for fixig this problem. I have created a clone that will be merged (to work around the CI from forks problem). This PR will be closed.

PR #2140, opened 2023-03-24
Fix to the problem of reading config multiple times (clone)

### LINKED ISSUE #2129: importing ganga from python prompt causes config to be executed multiple times
When doing `import ganga.ganga` from a standard python prompt the config file is read multiple times.
```python
In [1]: import ganga.ganga
INFO     reading config file /home/egede/.gangarc
INFO     reading config file /home/egede/.gangarc
INFO     reading config file /home/egede/.gangarc
INFO     reading config file /home/egede/.gangarc
INFO     reading config file /home/egede/.gangarc
```
This might be related to each of the of the plugins defined in the `.gangarc` file. Printing a traceback at the point where the [INFO statements above appear](https://github.com/ganga-devs/ganga/blob/db078fb3ee9f76f45ec5b23c249cd03512b717f8/ganga/GangaCore/Utility/Config/Config.py#L866) might reveal the problem.

egede:
The load_config_files function of GangaCore Config init would read system config variables in a loop, and set session variables in each loop cycle. This would cause the config file to be read multiple times. This commit moves the reading of user config and setting of session variables outside the loop.

### DIFF
--- a/ganga/GangaCore/Utility/Config/__init__.py
+++ b/ganga/GangaCore/Utility/Config/__init__.py
@@ -48,12 +48,16 @@ def load_config_files():
     """
     from GangaCore.Utility.Config import getConfig, setSessionValuesFromFiles
     from GangaCore.Runtime import GangaProgram
+
+    user_config = os.environ.get('GANGA_CONFIG_FILE') or os.path.expanduser('~/.gangarc')
+    config_files = GangaProgram.get_config_files(user_config)
+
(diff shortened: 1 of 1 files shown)


egede:
@alexanderrichards @mesmith75 Please review the PR clone here as well.

alexanderrichards (approved):

mesmith75 (approved):

PR #2145, opened 2023-03-29
Modernise Shell.py

### LINKED ISSUE #2105: Refactor use of shell command
Much of the `ganga` code is from before the subprocess module was fully written. This means that the codebase has a number of different ways that it deals with system calls. Some of this is causing trouble as seen in #2104 . First point of this might be to modernise
 - [Shell.py](https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Utility/Shell.py)
 - [execute.py](https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Utility/execute.py)

shalearkane:
Partially Fixes #2105 
- remove timing code
- simplify getting output

### DIFF
--- a/ganga/GangaCore/Utility/Shell.py
+++ b/ganga/GangaCore/Utility/Shell.py
@@ -37,8 +37,6 @@ import os
 import re
 import stat
 import tempfile
-import time
-import signal
 import subprocess
 
 from GangaCore.Utility.execute import execute
@@ -51,7 +49,8 @@ def expand_vars(env):
     """
     This function takes a raw dictionary which describes the environment and sanitizes it slightly
     This makes some attempt to take multi-line aliases and functions and make then into single line strings.
-    At best we don't like bash functions being assigned to variables here but we shouldn't crash when some users have bad a env
+    At best we don't like bash functions being assigned to variables here but we shouldn't crash when some users
+    have bad a env
     Args:
         env (dict): dictionary describing the environment which is to be sanitized
     """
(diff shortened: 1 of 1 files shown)


shalearkane:
@mesmith75 could you please review the changes and check if my approach is okay. I will then update the remaining files in the same manner.

shalearkane:
@mesmith75  Also could you please give a hint about what this piece of code does:
https://github.com/ganga-devs/ganga/blob/99c18db0f7946e61522dce003b7d256c3a57e0cc/ganga/GangaCore/Utility/execute.py#L287-L299

As far my understanding, it runs the python code in `eval_includes`. We also pass the current local variables that are in the function. So does that mean that while refactoring, the local variables that were already present in the function should not be changed because that might break code that processes local variables (such as `stdout`) within `eval_includes`?

egede:
> @mesmith75 Also could you please give a hint about what this piece of code does:
> 
> https://github.com/ganga-devs/ganga/blob/99c18db0f7946e61522dce003b7d256c3a57e0cc/ganga/GangaCore/Utility/execute.py#L287-L299
> 
> As far my understanding, it runs the python code in `eval_includes`. We also pass the current local variables that are in the function. So does that mean that while refactoring, the local variables that were already present in the function should not be changed because that might break code that processes local variables (such as `stdout`) within `eval_includes`?

I am not sure what it does, but it looks horribly complicated. I think a better approach is to trace down the places that the `execute` function is called and then for each of those understand what we actually try to do and if there is a simpler way to accomplish it.

egede:
@shalearkane You want to do further work here within the next few days, or should we merge in the already very helpful changes?

shalearkane:
I will create a new pull request for `execute.py` in a new PR. You can merge the current changes.

PR #2158, opened 2023-04-20
Partially fixes #2110 : Removes the LCG backend and related code

### LINKED ISSUE #2110: Clean up legacy in Job class schema
The Job class has some very old legacy fields in the [schema](https://github.com/ganga-devs/ganga/blob/45d523f0895c2992a4c9bc5aba1fae6b5a3867f8/ganga/GangaCore/GPIDev/Lib/Job/Job.py#L207). The fields `inputsandbox`, `outputsandbox` and `outputdata` can be removed. 

The rest of the code base should be checked and places where these are used should be removed as it all deals with legacy ways of specifying a job.

The `ForbidLegacyOutput` config option should be removed as well. This last bit is related to #2109.

vedanshbhartia:
Removes the LCG backend and related code


vedanshbhartia:
1. Is the VOMS proxy LCG specifc? Should it be removed?
2. How do I run unit tests locally? When I invoke the tests using `python -m pytest`, no test passes and python crashes. I got the same result using both python 3.10 and 3.9

mesmith75:
Could you make the title a bit more informative? Helps with the release notes

egede:
Take a look at how the testing is done in the [Continuous Integration](https://github.com/ganga-devs/ganga/blob/99c18db0f7946e61522dce003b7d256c3a57e0cc/.github/workflows/ci_push_testing.yml#L101). Note the `[dev]` when installing the dependencies.

vedanshbhartia:
> Take a look at how the testing is done in the [Continuous Integration](https://github.com/ganga-devs/ganga/blob/99c18db0f7946e61522dce003b7d256c3a57e0cc/.github/workflows/ci_push_testing.yml#L101). Note the `[dev]` when installing the dependencies.

Thanks a lot Ulrik, I was able to run the tests locally! I have updated the PR with the final changes, it should be good to review now.

egede (commented):
So this removes both the `LCG` and the `Cream` backend. The `LCG` backend is for sure broken and can be removed, but the `Cream` backend is at least in principle useable. I wonder if we should not remove it anyway as I doubt is has any practical use?? For discussion at a developers meeting.

egede:
We agreed that this can be merged. A clone of the PR will be made to facilitate the testing.

egede:
Clone was merged.

PR #2162, opened 2023-05-22
Modernise shell execute (clone)

egede:


alexanderrichards (dismissed):

egede:
Had to re-sign the gridpp-VO agreement. Still failing but I guess it might take a short time to migrate. Just try to re-run DIRAC tests  bit later.

egede:
@shalearkane  There is actually a real problem here (and not just a problem with the testing framework). The testing shows that the `info` method in [DiracProxy.py](https://github.com/ganga-devs/ganga/blob/6fdf1fd9be4dedbf59fcf3dc5b77dbafcadb92d3/ganga/GangaDirac/Lib/Credentials/DiracProxy.py#L91) doesn't behave as expected. As this deals exactly with the changed `Shell` code, I expect it to be the cause. Can you try to give some insight. I expect the problem is either with `env` or `cmd1` in the Shell object.

alexanderrichards (approved):

PR #2163, opened 2023-06-05
Let the latest version of Dirac install rather than pinning it.

egede:


egede:
So this PR is fixing the issue with the failing DIRAC tests. Should be merged asap and then added into the other pending PRs.

alexanderrichards (approved):

PR #2168, opened 2023-06-09
Added with open context manager when opening file - File left…

### LINKED ISSUE #2167: File left open when obtaining timestamps
When obtaining the timestamps for a Local job, the status file will be left open in most cases. The [current implementation](https://github.com/ganga-devs/ganga/blob/6fdf1fd9be4dedbf59fcf3dc5b77dbafcadb92d3/ganga/GangaCore/Lib/Localhost/Localhost.py#L191) should be changed to using a `with` clause.

matheus-LLC:
As per [egede] recommendation under issue #2167 I've put with open file context manager instead of just opening the file directly.


egede:
This is now merged.

PR #2169, opened 2023-06-10
Added with context manager when opening file (clone)

egede:
Clone to make sure that testing with secrets can run as well.

### DIFF
--- a/ganga/GangaCore/Lib/Localhost/Localhost.py
+++ b/ganga/GangaCore/Lib/Localhost/Localhost.py
@@ -188,25 +188,23 @@ class Localhost(IBackend):
         try:
             p = join(j.outputdir, '__jobstatus__')
             logger.debug("Opening output file at: %s", p)
-            f = open(p)
+            with open(p) as f:
+                for l in f:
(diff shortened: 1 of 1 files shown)


egede (commented):
Thanks for implementing this.

egede:
The autopep8 PR was not created for some reason. Maybe related to https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/. @joj0s can you take a look?

joj0s:
@egede Absolutely! By the looks for it, the issue with this PR is that it only includes a single linting violation (ambiguous variable name `l`) which is not automatically fixable so an autopep8 PR can not be created and it must be resolved manually. 

However, you are correct that the current `set-output` command we use has become deprecated in favor of [environment files](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files). So we should indeed migrate to using those. I have created issue #2170 and assigned myself, will submit a PR shortly.

alexanderrichards (approved):

PR #2176, opened 2023-07-10
Refactor out dependency on execute utility

### LINKED ISSUE #2105: Refactor use of shell command

shalearkane:
Fixes #2105 

Right now, it just removes the use of execute utility except for these cases:
1. `DiracUtilities.execute` uses the `python_setup` feature
2. `DiracUtilities.get_env` uses the `update_env` feature

It also removes the old timer code with a modern equivalent and the `eval_includes` functionality.


egede (approved):
Works all good in my testing

PR #2186, opened 2023-08-14
fix removeLFNs

### LINKED ISSUE #2185: Dirac removeLFNs is broken
It clearly does not work: https://github.com/ganga-devs/ganga/blob/3345d6f7733efcd044fa9cd273069bc1de15cd9a/ganga/GangaDirac/Lib/Backends/DiracUtils.py#L240 

Should be a quick fix.

mesmith75:
Fixes #2185


egede (commented):
Why is the credential requirement for this function (and the others that are exported to the GPI) None? Should it not be `DiracProxy`?

mesmith75:
I don't know. What we should do though is add the `@require_credential` decorator to some of these.

mesmith75:
I think the point is that normally the credential is associated with the backend whereas these are standalone functions.

egede (approved):

PR #2194, opened 2023-08-25
Stop Dirac process after creating credential

mesmith75:
We need to stop the currently running Dirac process if we create a new credential. In this way, the next time a Dirac command is run a new process gets started and connects with the new credential.


egede:
To make this work for proxies that are renewed external to Ganga, would it be better to make the dirac proxy decorator discover if the proxy has been renewed and then kill the server process?

alexanderrichards (approved):

PR #2199, opened 2023-09-11
Fixing forked PR tests

### LINKED ISSUE #2126: Enable PRs from forks to run the standard tests with secrets after permission is given
At the moment there is an administrative overhead associated with PRs arriving from forks. This is caused by that the CI testing is not given access to the secrets (so the DIRAC tests can't run) and do not have the ability to create new PRs (so the autopep8 mechanism doesn't run). This is clearly an important safeguard to avoid fishing of the secrets, but if the code is manually reviewed first it should be possible to allow.  

This is a known problem and there are a few different solutions.
- [Adding a `safe to test label`](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)
- [Use environment protection](https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci). (I do not fully understand this solution from just reading the page ...)

An important thing is that we want to avoid having an extra manual step in the CI testing if the PR is not from a fork. In other words the requirement of approval before the CI testing starts should only be for PRs coming from forks.

alexanderrichards:


egede:
Is this supposed to be finished? I notice that the tests did not run. Is that a sign that there is a problem or is there some "chicken and egg problem" where it needs to be merged before it can work correctly?

alexanderrichards:
> Is this supposed to be finished? I notice that the tests did not run. Is that a sign that there is a problem or is there some "chicken and egg problem" where it needs to be merged before it can work correctly?

I need to look into this a bit more but my initial thoughts, like yours were that it need merging before it can take effect

egede (dismissed):
OK, I will approve it. You will still need to make the merge on the command line in order to circumvent the check on tests to pass. And then be ready to revert it, if it just stops the testing completely.

alexanderrichards:
Ok, I have merged this so lets see what happens. It will only trigger on a PR though

alexanderrichards:
Seems to work as advertised.

PR #2201, opened 2023-09-12
Update README.rst

alexanderrichards:
This is a commit just to create a new PR and we have no intention of merging this.

### DIFF
--- a/.github/workflows/ci_push_testing.yml
+++ b/.github/workflows/ci_push_testing.yml
@@ -15,6 +15,14 @@ jobs:
     - name: Approve
       run: echo For security reasons, all pull requests need to be approved first before running any automated CI.
 
+  checkstatus:
+    name: status check
+    runs-on: ubuntu-latest
+    needs: [approve]
+    if: always()
+    steps:
+    - name: check
+      run: echo ${{needs.approve.result}}
   
   lint:
     name: Linting
--- a/README.rst
+++ b/README.rst
@@ -43,3 +43,4 @@ testing
 testing
 testing
 testing
+testing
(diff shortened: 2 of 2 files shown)


alexanderrichards:
closing this PR as it correctly asked me to approval before continuing

alexanderrichards:
re-opening to further discussion. Do we want the linting to run before we have to approve the rest or should we keep it as is where approval is needed for everything?

alexanderrichards:
in fact we could only have approval for those bits that need access to secrets?

egede:
The best would be to have it only for the parts needing secrets. In that way an outsider can discover most problems without our involvement.

I am a bit confused about how it actually will work. I thought it would not ask for approval if you were a developer within the team, but the comments above seems to indicate that it asked you to approve your own PR?

alexanderrichards:
All PRs run the same github action "CI Pust Testing". To allow the jobs within to access the secrets we use the on: `pull_request_target` which will by default run only the code from the base which is obviously safe. We then modify the checkout command to checkout the head of the PR and hence the potentially unsafe code. In order to police the running of this code we say that it has to run the tests in the environment `Integrate Pull Request` which we protect from running until approval is given by either myself, you or mark.

In practise this means that before the tests (that define to run in the environment) will run we get an email and have to click an approve button.

alexanderrichards:
> I thought it would not ask for approval if you were a developer within the team,

I have checked the box that says "allow administrators to bypass protection rules". I guess this means that it will still wait for approval but that we should be able to click a button somewhere to allow it to proceed without.  At that point though it's easy enough to just click approve since unlike with the code reviews we can approve ourselves

egede:
I believe this can just be closed. Reopen if required.

PR #2205, opened 2023-10-17
fix googlefile

mesmith75:
I think having a field as visible, copyable and protected doesn't work.


alexanderrichards (dismissed):

mesmith75:
The change to `gangadoc.py` fixes the core integration test. It is just an update to make it compatible with the latest `pydoc`

alexanderrichards (approved):

PR #2223, opened 2023-12-07
Update AFS credential

mesmith75:
This works for both lxplus7 and lxplus9. It is a bit awkward as the output of `tokens` is not the same between the two so the regex needs some adjustment.

It is also a bit awkward as the unit test needs a slight modification to explicitly `mock` the `exists` function for the AFS token.

As lxplus is now el9 by default I would like to get this merged and released fairly rapidly


egede (approved):

egede:
I imagine that this should be merged into the branch that the DEV version is created from as well.

PR #2226, opened 2023-12-07
Update setup.py

mesmith75:


egede (approved):
You could consider adding python 3.12 here as well.

alexanderrichards (approved):

PR #2228, opened 2023-12-09
stop mangling the environments

mesmith75:
This keeps the enviroment as the `os._Environ` type which can cope with the more complex environments of newer bash shells - in particular multiline functions.

### DIFF
--- a/ganga/GangaCore/Utility/Shell.py
+++ b/ganga/GangaCore/Utility/Shell.py
@@ -115,12 +115,9 @@ class Shell(object):
             setup_args (list): list of strings which are executed directly with a ' ' character spacing
         """
 
+        self.env = os.environ
         if setup is not None:
-            self.env = os.environ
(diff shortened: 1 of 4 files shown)


ganga/GangaCore/Utility/Shell.py
```diff
@@ -116,13 +116,11 @@ def __init__(self, setup=None, setup_args=[]):
...
         if setup is not None:
-            self.env = dict(os.environ)
+            self.env = os.environ
             execute('source {0} {1}'.format(setup, " ".join(setup_args)), shell=True, env=self.env, update_env=True)
 
         else:
-            # bug #44334: Ganga/Utility/Shell.py does not save environ
-            env = dict(os.environ)
-            self.env = expand_vars(env)
+            self.env = os.environ
```
egede: Seems like the `self.env = os.environ` can just be put up before the if clause and then just have the `execute` part inside the if clause.

egede (dismissed):

egede (approved):

PR #2243, opened 2023-12-19
Update the condor testing to use the latest version

### LINKED ISSUE #2237: Condor tests are failing
The Condor CI tests are failing with the error
```
https://research.cs.wisc.edu/htcondor/repo/9.1/el7/release/repodata/repomd.xml: [Errno 14] HTTPS Error 404 - Not Found
```
I imagine it is due to Condor no longer being available on centos7. Solution is presumably to upgrade the tests to run on `el9` instead. The CI code to change starts [here](https://github.com/ganga-devs/ganga/blob/01f17862a2e033e9d4b9a066dc45c45c55cd08c6/.github/workflows/ci_push_testing.yml#L172).

egede:


egede:
Note that the tests run with the CI code from the `develop` branch. I think this is a side effect of the changes @alexanderrichards made. In general that is perfectly fine, but of course causes a bit of trouble when you try to actually fix the testing code. In this case the `condor` tests are still failing even if I expect that this PR will fix it.

egede:
Closes #2237

mesmith75 (approved):

PR #2244, opened 2023-12-20
Remove unfinished GUI code from Ganga

### LINKED ISSUE #2241: Remove the non-functional GUI code
The GUI code in `ganga` was never developed into a functional thing. It is not realistic for somebody to pick up on this code and develop it further. Hence we are better off backing it out of Ganga

egede:
Closes #2241


mesmith75:
I think one file is still to be removed: https://github.com/ganga-devs/ganga/blob/develop/bin/ganga-gui

mesmith75 (approved):

alexanderrichards (approved):

PR #2251, opened 2024-01-08
Fix condor tests

mesmith75:
Last attempt before I give up. Does this really need the virtualenv?


egede (approved):
If it still doesn't work, I will give it another go. It complained in a different way without the virtualenv.

mesmith75:
Looks like it may have worked

PR #2282, opened 2024-02-11
Add logging functionality to `InteractiveScriptTemplate.py.template`

### LINKED ISSUE #2174: Interactive backend should save a log file
Would be helpful to have a `stdout` saved in the gangadir for Interactive jobs.

samadpls:
resolved #2174 
Added log file functionality for Interactive backend

### DIFF
--- a/ganga/GangaCore/Lib/Interactive/InteractiveScriptTemplate.py.template
+++ b/ganga/GangaCore/Lib/Interactive/InteractiveScriptTemplate.py.template
@@ -17,6 +17,17 @@ import time
 sys.path.insert( 0, '###GANGA_PYTHONPATH###' )
 
 statfileName = os.path.join( '###OUTPUTDIR###', '__jobstatus__' )
+logfile = os.path.join('###OUTPUTDIR###', 'interactive_log.txt')
+
+try:
+    statfile = open(statfileName, 'w')
+    logfile = open(logfile, 'w')
+except IOError as x:
+    print('ERROR: Unable to write status file or log file.')
+    print('ERROR: ', x)
+    raise
+finally:
+    statfile.close()
 try:
    statfile = open( statfileName, 'w' )
 except IOError as x:
--- a/ganga/GangaCore/Lib/Interactive/InteractiveScriptTemplate.py.template
+++ b/ganga/GangaCore/Lib/Interactive/InteractiveScriptTemplate.py.template
@@ -14,38 +14,38 @@ import time
 
 ###WNSANDBOX_SOURCE###
 
-sys.path.insert( 0, '###GANGA_PYTHONPATH###' )
+sys.path.insert(0, '###GANGA_PYTHONPATH###')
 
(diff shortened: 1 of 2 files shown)


ganga/GangaCore/Lib/Interactive/InteractiveScriptTemplate.py.template
```diff
@@ -17,6 +17,17 @@ import time
 sys.path.insert( 0, '###GANGA_PYTHONPATH###' )
 
 statfileName = os.path.join( '###OUTPUTDIR###', '__jobstatus__' )
+logfile = os.path.join('###OUTPUTDIR###', 'interactive_log.txt')
+
+try:
+    statfile = open(statfileName, 'w')
```
egede: Why is the `statfile` opened here. It seems to be opened again a few lines later.

ganga/GangaCore/Lib/Interactive/InteractiveScriptTemplate.py.template
```diff
@@ -17,6 +17,17 @@ import time
...
+    logfile = open(logfile, 'w')
+except IOError as x:
+    print('ERROR: Unable to write status file or log file.')
+    print('ERROR: ', x)
+    raise
+finally:
+    statfile.close()
```
egede: Should that not be `logfile`?

egede (commented):
This pull request doesn't seem to do anything. It open a logfile, but there is never anything written to it. The behaviour that we are looking for is that the output after the job has finished should be exactly the same as for the `Local` backend. So the `outputdir` should contain the files `stdout` and `stderr` with the standard output and error respectively. However, we want it at the same time to be printed to the interactive terminal. So behaviour should be a bit like the `tee` linux command.

As a secondary comment, it would be good to rewrite the opening and closing of files to use the more modern construction of the `with` command.

egede:
There should also be a test case written that shows that it works. Please ask for further advice when getting to that point.

samadpls:
Hey, @egede, thanks for your response. I apologize for my mistake, but I have updated the code, and it is now saving output in their respective files and matching the output of the local backend. However, I am still unsure about the correct directory for writing the test cases. Could you please clarify that for me? Thanks. 

Edit: One integration test is currently failing and I'm trying to determine the potential reason. It would be helpful if you could also point out the reason for the test case failure.

egede (commented):
This looks good. Please add a test case in https://github.com/ganga-devs/ganga/tree/develop/ganga/GangaCore/test/GPI in a file called TestInteractive.py. Look at a file like https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/test/GPI/TestArgSplitter.py to see how to do ths in an easy way. The test should confirm that both stdout and stderr are created with the correct content.

egede (dismissed):
Great work. Many thanks. This will be merged and made part of the next release.

egede (approved):
This looks good and ready to merge.

egede:
Please delete the branch your side to avoid pushing to the already merged branch.

PR #2290, opened 2024-02-22
Dirac test

egede:


egede:
The DIRAC testing framework has been updated from what was an ancient process. At the same time migrated to alma9. Will need to be merged with failing tests as we always test with the workflow from the develop branch.

mesmith75 (approved):

Issue #2300, opened 2024-02-27
Virtualization.rst is probably showing incorrect docker commands when calling the Docker class through job.virtualization


dg1223:
Passing arguments such as `Docker("docker://fedora:latest")` throws `TypeError`.

**Steps to reproduce**

```
j = Job(name='dockertest')
j.virtualization = Docker("docker://fedora:latest")
```
**Error**

```
ERROR    !!Unknown/Unexpected ERROR!!
ERROR    If you're able to reproduce this please report this to the Ganga developers!
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[30], line 1
----> 1 j.virtualization = Docker("docker://fedora:latest")

File ~/Documents/GitHub/gsoc-2024/ganga/gangaenv/lib/python3.8/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py:883, in GPIProxyClassFactory.<locals>._init(self, *args, **kwds)
    881 elif arg_len < len(getfullargspec(pluginclass.__init__)[0]):
    882     clean_args = (stripProxy(arg) for arg in args)
--> 883     instance = pluginclass(*clean_args)
    884 else:
    885     # In the future we will just pass the args to the classes directly and throw excepions, but for now we're trying to maintain old behavior
    886     logger.warning("Cannot use arguments: '%s' for constructing class type '%s'. Ignoring." %
    887                    (args, getName(pluginclass)))

TypeError: __init__() missing 1 required positional argument: 'mode'
```

Same issue if I run `j.virtualization = Docker("fedora:latest")`

The following lines in the documentation [Virtualization.rst](https://github.com/ganga-devs/ganga/blob/develop/doc/UserGuide/Virtualization.rst) contains these instructions:

**line 42**
`j.virtualization = Docker("docker://fedora:latest")`

**line 81**
`j.virtualization = Docker("image:tag")`
e.g.: `j.virtualization = Docker("fedora:latest")`

The command that works is:
`j.virtualization = Docker(image="fedora:latest")`

If this is a simple case of updating the documentation, I can do it soon. Thanks

egede:
Yes, just update documentation.

mesmith75:
Fixed by #2314

PR #2308, opened 2024-03-01
Fixed erroneous virtualization commands in Virtualization.rst (#2300)

dg1223:
Command that works is of the pattern:
`j.virtualization = Docker(image="image:tag")`


egede:
Would it be possible to fix the code such that `Docker('fedora:latest')` would work instead of `Docker(image='fedora:latest')`. From reading the code, it is not obvious to me what is wrong.

dg1223:
> Would it be possible to fix the code such that `Docker('fedora:latest')` would work instead of `Docker(image='fedora:latest')`. From reading the code, it is not obvious to me what is wrong.

Okay, I'll take this on as the next issue. Thanks

egede (approved):

Issue #2309, opened 2024-03-04
Investigate docker commands that do not currently work with virtualization and make them work


dg1223:
Related issue and PR: #2300, #2308

The following docker command patterns throw `TypeError` in Ganga.

`j.virtualization = Docker("docker://image:tag")`
and
`j.virtualization = Docker("image:tag")`

The documentation for virtualization [Virtualization.rst](https://github.com/ganga-devs/ganga/blob/develop/doc/UserGuide/Virtualization.rst) has been updated with the command that currently works:
`j.virtualization = Docker(image="image:tag")`

**We need to see if we can fix the code such that `Docker('image:tag')` would work instead of `Docker(image='image:tag')`.**

**Commands to reproduce issue**
```
j = Job(name='dockertest1')
j.virtualization = Docker("docker://fedora:latest")
```
```
j = Job(name='dockertest2')
j.virtualization = Docker("fedora:latest")
```
**Error**

```
ERROR    !!Unknown/Unexpected ERROR!!
ERROR    If you're able to reproduce this please report this to the Ganga developers!
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[30], line 1
----> 1 j.virtualization = Docker("docker://fedora:latest")

File ~/Documents/GitHub/gsoc-2024/ganga/gangaenv/lib/python3.8/site-packages/ganga/GangaCore/GPIDev/Base/Proxy.py:883, in GPIProxyClassFactory.<locals>._init(self, *args, **kwds)
    881 elif arg_len < len(getfullargspec(pluginclass.__init__)[0]):
    882     clean_args = (stripProxy(arg) for arg in args)
--> 883     instance = pluginclass(*clean_args)
    884 else:
    885     # In the future we will just pass the args to the classes directly and throw excepions, but for now we're trying to maintain old behavior
    886     logger.warning("Cannot use arguments: '%s' for constructing class type '%s'. Ignoring." %
    887                    (args, getName(pluginclass)))

TypeError: __init__() missing 1 required positional argument: 'mode'

```

egede:
The problem is with the `__init__` function in [`Docker.py`](https://github.com/ganga-devs/ganga/blob/54b2397f3eebd56da2280d8186a956adac725268/ganga/GangaCore/Lib/Virtualization/Docker.py#L39). Simply removing `mode` from the list of arguments and the the subsequent assignment line, I think will resolve the issue.

egede:
[...]

PR #2314, opened 2024-03-30
fix issues with some malfunctioning docker commands (#2309)

dg1223:
The following Docker commands now work correctly:

```
d = Docker()
d = Docker('fedora:latest')
d = Docker(image='fedora:latest')
d = Docker(image='fedora:latest', mode='P1')
```

In addition, added unit tests for each of these commands.

Please note that this fix does not remove the `__init__` function from [Docker,py](https://github.com/ganga-devs/ganga/blob/develop/ganga/GangaCore/Lib/Virtualization/Docker.py) as suggested [here](https://github.com/ganga-devs/ganga/issues/2309#issuecomment-1979881949). Instead, it sets default values for the `image` and `mode` parameters in the `__init__` function in Docker.py.


egede (approved):

egede:
@dg1223 Thanks for adding the test cases as well.

dg1223:
> @dg1223 Thanks for adding the test cases as well.

No problem at all!

PR #2321, opened 2024-04-12
WIP: Add apptainer to GaudiExec

### LINKED ISSUE #2320: Problems with running slc6 based GaudiExec
For some very old applications (things requiring slc6) you need to use apptainer as they are not functional with el9.

Should be a very short addition to the run line command.

mesmith75:
Fixes #2320

### DIFF
--- a/ganga/GangaLHCb/Lib/Applications/GaudiExec.py
+++ b/ganga/GangaLHCb/Lib/Applications/GaudiExec.py
@@ -195,7 +195,7 @@ class GaudiExec(IPrepareApp):
         'nMakeCores': SimpleItem(defvalue=1,
                                  doc='Number of cores to be provided via the "-j" option to the "make" command'\
                                      'when building the ganga-input-sandbox'),
-
+        'apptainerBuild' : SimpleItem(defvaule=False, doc="Run the build command in apptainer"),
         # Prepared job object
         'is_prepared': SimpleItem(defvalue=None, strict_sequence=0, visitable=1, copyable=1, hidden=0,
                                   typelist=[None, ShareDir], protected=0, comparable=1,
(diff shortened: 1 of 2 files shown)


ganga/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -564,7 +564,11 @@ def execCmd(self, cmd):
             if cmd != 'make':
                 rc, stdout, stderr = _exec_cmd(cmd_file.name, self.directory)
         else:
-            rc, stdout, stderr = _exec_cmd(cmd_file.name, self.directory)
+            if self.apptainerBuild or 'slc6' in self.platform:
+                cmd_to_run = 'apptainer exec --env "PATH=$PATH" --bind $PWD --bind /cvmfs:/cvmfs:ro /cvmfs/cernvm-prod.cern.ch/cvm4 ' + cmd_file.name
```
egede: I am worried about the use if this image. There is no guarantee that /cvmfs/cernvm-prod.cern.ch is mounted at a given location. Is there not an official LHCb image for centos7?

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
```diff
@@ -195,7 +195,7 @@ class GaudiExec(IPrepareApp):
         'nMakeCores': SimpleItem(defvalue=1,
                                  doc='Number of cores to be provided via the "-j" option to the "make" command'\
                                      'when building the ganga-input-sandbox'),
-
+        'apptainerBuild' : SimpleItem(defvaule=False, doc="Run the build command in apptainer"),
```
egede: This name might come back and bite us. Setting this to True will get a very specific centos7 image which the name doesn't indicate. Maybe have a second option that includes the the name of the actual image.

ganga/GangaLHCb/Lib/Applications/GaudiExec.py
mesmith75: There is but I don't think it is on CVMFS. I guess we could add it to our won `/cvmfs` location if we really needed it.

egede:
There might be a way to satisfy both the issues raised here.
- An option is introduced to give the location.
- The default location is `/cvmfs/cernvm-prod.cern.ch/cvm4` but other options could be `gitlab-registry.cern.ch/lhcb-docker/os-base/centos7-devel:latest`
- The error message if things go wrong might be a bit more explaining of what can be changed.

mesmith75:
Yes, I think you are  right

heistera:
Perhaps, in case cvmfs ist not accessible,  it is possible to download an image in a similar way like the following:

Apptainer("docker://gitlab-registry.cern.ch/lhcb-core/lbdocker/slc6-build:latest")

The same location has an image which looks like the source of the image used for cvmfs ...

egede (approved):

mesmith75:
I'll merge this now and add on the `prepareGaudiExec` adjustment in a separate PR

PR #2329, opened 2024-04-23
Fix apptainer call in GaudiExec

laf070810:
The `_exec_cmd` function will empty the environment variables, causing `apptainer command not found`, thus suggesting adding an environment setup like other `_exec_cmd` calls.


egede (approved):
I reproduce problem and confirm fix. Thanks.

PR #2333, opened 2024-04-24
Fix to allow Gaudi slc6 jobs to run inside apptainer

egede:
Modified template such that Local jobs runs in login environment and not environment of the Ganga session


egede:
Added better implementation from @mesmith75

PR #2338, opened 2024-05-13
Update to make sure that Local jobs run in fresh environment

egede:


mesmith75 (dismissed):
My only hesitation is some of the odd function stuff you get in el9. For example:
```
BASH_FUNC_ml%%=() {  module ml "$@"
}
BASH_FUNC_which%%=() {  ( alias;
 eval ${which_declare} ) | /usr/bin/which --tty-only --read-alias --read-functions --show-tilde --show-dot $@
}
BASH_FUNC_module%%=() {  local _mlredir=1;
 if [ -n "${MODULES_REDIRECT_OUTPUT+x}" ]; then
 if [ "$MODULES_REDIRECT_OUTPUT" = '0' ]; then
 _mlredir=0;
 else
 if [ "$MODULES_REDIRECT_OUTPUT" = '1' ]; then
 _mlredir=1;
 fi;
 fi;
 fi;
 case " $@ " in 
 *' --no-redirect '*)
 _mlredir=0
 ;;
 *' --redirect '*)
 _mlredir=1
 ;;
 esac;
 if [ $_mlredir -eq 0 ]; then
 _module_raw "$@";
 else
 _module_raw "$@" 2>&1;
 fi
}
BASH_FUNC_scl%%=() {  if [ "$1" = "load" -o "$1" = "unload" ]; then
 eval "module $@";
 else
 /usr/bin/scl "$@";
 fi
}
BASH_FUNC__module_raw%%=() {  eval "$(/usr/bin/tclsh '/usr/share/Modules/libexec/modulecmd.tcl' bash "$@")";
 _mlstatus=$?;
 return $_mlstatus
}
_=/usr/bin/printenv
```
is this taken care of appropriately?

egede:
Thanks @mesmith75 . Indeed functions were not dealt with correctly. And also while the login was run again, the environment was in fact not cleaned. Update should protect against both of these cases. A test is added as well.

mesmith75 (approved):

mesmith75:
I think this looks good. In general though I am not a big fan of these gymnastics we do with converting environments to dicts etc. Python has an `environ` object which should deal with this kind of thing seamlessly, although I think it would be a considerable effort to rework everything.

PR #2339, opened 2024-05-16
Update Condor.py

mesmith75:
The `runenv` is missing from the Condor script which breaks it when using virtualisation.


egede (commented):
Would it be better to add this line to the injected code in the [vitualisation](https://github.com/ganga-devs/ganga/tree/e2a6d67f793d6a121485badea32d967b070888ee/ganga/GangaCore/Lib/Virtualization) classes instead?

mesmith75:
I don't think so - in all the other templates it is already there. It is just because Condor doesn't use any of the standard templates for some reason.

egede (approved):
It was actually more a comment on all the other templates ... anyway that should not stop this bug fix.

PR #2344, opened 2024-05-20
Re-add runMonitoring

### LINKED ISSUE #2340: Option to monitor a slice
Re-add the option to monitor a slice of jobs with `runMonitoring`

mesmith75:
Fixes #2340

### DIFF
--- a/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
+++ b/ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
@@ -74,6 +74,7 @@ class AsyncMonitoringService(GangaThread):
         if job_slice and len(found_active_backends)==0:
             log.debug("No active backends found with a job slice. Turning off the monitoring loop")
             self.enabled = False
+            return
 
         # If a backend is newly found as active, trigger its monitoring
         previously_active_backends = self.active_backends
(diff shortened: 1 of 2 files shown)


ganga/GangaCore/Core/MonitoringComponent/MonitoringService.py
```diff
@@ -226,3 +226,28 @@ def stop(self):
...
+        Note:         
+          This method is meant to be used in Ganga scripts to request monitoring on demand. 
+        """
+
+        log.debug("runMonitoring")
+        if not self.alive:
+            log.error("Cannot run the monitoring loop. It has already been stopped")
+            return False
+
+        self.enabled = True
```
mesmith75, line 253: I am not wholly enthused by this. You have to set `enabled` to `True` to monitor the jobs but in reality it is not monitoring everything.

egede (commented):
What happens if I call `runMonitoring` when the monitoring is already running automatically. Will it do nothing, start a second parallel operation (and maybe cause trouble), do nothing, or make the whole thing crash?

mesmith75:
It does nothing

mesmith75:
So this currently monitors one job and then monitors all of them.

mesmith75:
This is ready for review.

alexanderrichards (approved):

PR #2366, opened 2024-06-19
Fix master kill

### LINKED ISSUE #2358: Job master kill error with failed subjobs
This needs fixing:
```python
[11:27:10]
Ganga In [12]: jobs(2317).kill()
ERROR    failed to kill job 2317: JobStatusError: forbidden status transition of job 2317 from "failed" to "killed"
ERROR    JobError: failed to kill job 2317: JobStatusError: forbidden status transition of job 2317 from "failed" to "killed"
```

mesmith75:
Fixes #2358

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -2462,7 +2462,7 @@ class Job(GangaObject):
                     raise JobError(msg)
             try:
                 if self.backend.master_kill():
-                    have_failures = False
+                    have_failures ='killed' 
                     ############
                     # added as part of typestamp prototype by Justin
                     if not self._getParent():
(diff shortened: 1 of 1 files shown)


egede (commented):
Is this related to the problem of killing a master job that has some failed subjobs? If yes, I do not see how?

mesmith75:
This is looping over the subjobs. When it gets to the last one, then it also updates the master job status. There is another `updateStatus` two lines later for the master job which tries to update the already updated master status and gives the error. It is better to do this than remove the second one, as the second one is also used for jobs without subjobs.

egede (dismissed):

mesmith75:
You are right though, I'll check again to be sure it is doing exactly what I think it should be

mesmith75:
You are right - this wasn't the real issue. The problem is that the master job ends up in `failed` once the last subjob is killed but then the following master job update tries to put it into `killed`, causing the issue.

I am not sure what is best - have the master job switch to `killed` or stay at `failed`.

alexanderrichards (dismissed):

mesmith75:
This is slightly less elegant, but more precise for what we want

egede (approved):

PR #2393, opened 2025-01-08
nd280 updates for 2024/2025

NickHastings:
This branch has two minor feature updates for ND280 processing and some code cleanups.

### DIFF
--- a/ganga/GangaND280/ND280Control/runND280RDP.py
+++ b/ganga/GangaND280/ND280Control/runND280RDP.py
@@ -17,14 +17,14 @@ from GangaCore.GPIDev.Lib.File import FileBuffer
 from GangaCore.GPIDev.Schema import Schema, SimpleItem, Version
 from GangaCore.Utility.Config import getConfig
 from GangaCore.Utility.files import expandfilename
+from GangaCore.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
+from . import ND280Configs
 
 shared_path = os.path.join(
     expandfilename(getConfig('Configuration')['gangadir']),
     'shared',
     getConfig('Configuration')['user'],
 )
-
-from . import ND280Configs
 
 
 class runND280RDP(IPrepareApp):
@@ -80,19 +80,6 @@ class runND280RDP(IPrepareApp):
     _category = 'applications'
     _name = 'runND280RDP'
     _exportmethods = ['prepare']
-    _GUIPrefs = [
-        {'attribute': 'args', 'widget': 'String_List'},
-        {'attribute': 'cmtsetup', 'widget': 'String'},
-        {'attribute': 'confopts', 'widget': 'String'},
-        {'attribute': 'env', 'widget': 'DictOfString'},
-    ]
-
-    _GUIAdvancedPrefs = [
-        {'attribute': 'args', 'widget': 'String_List'},
-        {'attribute': 'cmtsetup', 'widget': 'String'},
-        {'attribute': 'confopts', 'widget': 'String'},
-        {'attribute': 'env', 'widget': 'DictOfString'},
-    ]
 
     def __init__(self):
         super(runND280RDP, self).__init__()
@@ -218,8 +205,6 @@ class gLiteRTHandler(IRuntimeHandler):
             app.env,
         )
 
-
-from GangaCore.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
 
 allHandlers.add('runND280RDP', 'LSF', RTHandler)
 allHandlers.add('runND280RDP', 'Local', RTHandler)
--- a/ganga/GangaND280/ND280Control/runND280RDP.py
+++ b/ganga/GangaND280/ND280Control/runND280RDP.py
@@ -18,6 +18,7 @@ from GangaCore.GPIDev.Schema import Schema, SimpleItem, Version
 from GangaCore.Utility.Config import getConfig
 from GangaCore.Utility.files import expandfilename
 from GangaCore.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
+from GangaCore.Lib.Virtualization import Apptainer
 from . import ND280Configs
 
 shared_path = os.path.join(
@@ -75,6 +76,16 @@ class runND280RDP(IPrepareApp):
                 comparable=1,
                 doc='Location of shared resources. Presence of this attribute implies the application has been prepared.',
             ),
+            'container': SimpleItem(
+                defvalue=None, doc='Path to container', typelist=['type(None)', 'str']
+            ),
+            'mounts': SimpleItem(
+                defvalue=[],
+                doc='Container paths to mount',
+                typelist=['str'],
+                sequence=1,
+                strict_sequence=0,
+            ),
         },
     )
     _category = 'applications'
@@ -88,6 +99,12 @@ class runND280RDP(IPrepareApp):
         args = convertIntToStringArgs(self.args)
 
         job = self.getJobObject()
+        if self.container:
+            job.virtualization = Apptainer()
+            job.virtualization.image = self.container
+
+        if len(self.mounts) > 0:
+            job.virtualization.mounts = {mount: mount for mount in self.mounts}
 
         if self.cmtsetup == []:
             raise ApplicationConfigurationError('No cmt setup script given.')
--- a/ganga/GangaND280/Highland/Highland.py
+++ b/ganga/GangaND280/Highland/Highland.py
@@ -16,6 +16,7 @@ from GangaCore.GPIDev.Adapters.IRuntimeHandler import IRuntimeHandler
 from GangaCore.GPIDev.Schema import Schema, SimpleItem, Version
 from GangaCore.Utility.Config import getConfig
 from GangaCore.Utility.files import expandfilename
+from GangaCore.GPIDev.Adapters.ApplicationRuntimeHandlers import allHandlers
 
 shared_path = os.path.join(
     expandfilename(getConfig('Configuration')['gangadir']),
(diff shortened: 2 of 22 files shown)


alexanderrichards (dismissed):

NickHastings:
Had a quick look but I don't see what needs to be done to get the failing checks to pass. Any hints?

egede:
@NickHastings You are seeing a weakness of our CI. If the PR branch of your fork doesn't allow for the Ganga developers to edit (a box that you can tick or not), then the PEP8 compliance testing will fail. As all your changes are within the `nd280` part, I think it is really up to you if you want to fix the PEP8 compliance or not. I suggest that we just merge this PR without it.

ganga/GangaND280/ND280Control/runND280RDP.py
```diff
@@ -155,7 +151,15 @@ def configure(self, masterappconfig):
...
         # Create the bash script and put it in input dir.
+        self._scriptname = job.inputdir + 'runND280.sh'
         script = '#!/bin/bash\n'
+        if self.container:
+            bs = ' '.join([f'--bind {b}' for b in self.binds])
+            script += 'AM_IN_CONTAINER="$1"\n'
+            script += 'if [ "$AM_IN_CONTAINER" = "" ]; then\n'
+            script += f'    exec apptainer exec {bs} {self.container} {self._scriptname} "1"\n'
+            script += 'fi\n'
+
```
egede: Ganga has a general support for running jobs inside a container. It will be more flexible than what you implement here.  I copy the documentation of the `Apptainer` class here for reference.

```
class Apptainer(GPIProxyObject)
 |  Apptainer(*args, **kwds)
 |  
 |      The Apptainer class can be used for either Apptainer or Docker images.
 |      It requires that apptainer is installed on the worker node.
 |  
 |      For Apptainer images you provide the image name and tag from Apptainer
 |      hub like
 |  
 |        j=Job()
 |        j.application=Executable(exe=File('my/full/path/to/executable'))
 |        j.virtualization = Apptainer("shub://image:tag")
 |  
 |      Notice how the executable is given as a `File` object. This ensures that it
 |      is copied to the working directory and thus will be accessible inside the
 |      container.
 |  
 |      The container can also be provided as a Docker image from a repository. The
 |      default repository is Docker hub.
 |  
 |        j.virtualization = Apptainer("docker://gitlab-registry.cern.ch/lhcb-core/lbdocker/centos7-build:v3")
 |  
 |        j.virtualization = Apptainer("docker://fedora:latest")
 |  
 |      Another option is to provide a `GangaFile` Object which points to a
 |      apptainer file. In that case the apptainer image file will be copied to
 |      the worker node. The first example is with an image located on some shared
 |      disk. This will be effective for running on a local backend or a batch
 |      system with a shared disk system.
 |  
 |        imagefile = SharedFile('myimage.sif', locations=['/my/full/path/myimage.sif'])
 |        j.virtualization = Apptainer(image= imagefile)
 |  
 |      while a second example is with an image located in the Dirac Storage
 |      Element. This will be effective when using the Dirac backend.
 |  
 |        imagefile = DiracFile('myimage.sif', lfn=['/some/lfn/path'])
 |        j.virtualization = Apptainer(image= imagefile)
 |  
 |      If the image is a private image, the username and password of the deploy
 |      token can be given like the example below. Look inside Gitlab setting for
 |      how to set this up. The token will only need access to the images and
 |      nothing else.
 |  
 |        j.virtualization.tokenuser = 'gitlab+deploy-token-123'
 |        j.virtualization.tokenpassword = 'gftrh84dgel-245^ghHH'
 |  
 |      Directories can be mounted from the host to the container using key-value
 |      pairs to the mounts option. If the directory is not available on the host, a
 |      warning will be written to stderr of the job and no mount will be attempted.
 |  
 |        j.virtualization.mounts = {'/cvmfs':'/cvmfs'}
 |  
 |      By default the container is started in apptainer with the `--nohome`
 |      option. Extra options can be provided through the `options` attribute. See
 |      the Apptainer documentation for what is possible.
 |  
 |      If the apptainer binary is not available in the PATH on the remote node - or has a different name,
 |      it is possible to give the name of it like
 |  
 |        j.virtualization.binary='/cvmfs/oasis.opensciencegrid.org/mis/apptainer/current/bin/apptainer'
 |  
```

NickHastings: @egede thanks for the info about the Apptainer class. I knew there would be a cleaner way to do this but I needed to get something up and running  on a *very* short timescale, and my little hack that you see above fitted in with that. I'll back that commit out of the PR shortly.

egede: If what you want is to run inside a `centos7` container for a Grid job (or in fact anywhere with `/cvmfs` access), there is a very efficient way of doing this. For a given job, 
```
j.virtualization = Apptainer()
j.virtualization.image = '/cvmfs/cernvm-prod.cern.ch/cvm4'
```
is all that is required. It takes advantage of that this `cvmfs` location has an unpacked container image. Read [the documentation](https://cvmfs.readthedocs.io/en/stable/cpt-containers.html#distributing-container-images-on-cernvm-fs) for some more information about how this works.

NickHastings: I have a local .sif file that contains not only the OS but also the executable. I tried using the method documented above:
```
imagefile = SharedFile('myimage.sif', locations=['/my/full/path/myimage.sif'])
j.virtualization = Apptainer(image= imagefile)
```
but using that if I try to run a test task with two jobs they fail without a particularly useful error messages:

```[07:53:33]
Ganga In [25]: tasks(363).run()

[07:53:38]
Ganga In [26]: ERROR    Parallel Job Submission Failed: list index out of range
ERROR    Parallel Job Submission Failed: list index out of range
ERROR    JobStatusError: forbidden status transition of job 217487 from "new" to "failed"
```
Any hints on how to debug this?

egede: Try to make a minimally failing job. So do this outside the task framework. Then take a look at the wrapper script that is stored in the inputdir of the job. You can post it here for help as well. When you use the SharedFile object, you depend on that the file can be read in the same absolute file location on the worker node. Is that the case?

NickHastings: The sif file is available at that path on all nodes. The Task gets as far as creating the job input and output directories, but they are both empty.

I had a quick try to create jobs without the Task.

```
imagefile = SharedFile('nd280softwaremaster_centos7-13.28.sif',locations=['/gpfs/group/t2k/nd280/RealDataProcessing/containers/nd280softwaremaster_centos7-13.28.sif'])
j = Job(name='Tester', virtualization=Apptainer(image=imagefile), application=Executable(exe='uname', args=['-a']))
j.submit()
```
Inspecting the __jobscript__ that it created I see that it does:
```
os.system('cp /gpfs/group/t2k/nd280/RealDataProcessing/containers/nd280softwaremaster_centos7-13.28.sif .')
```
Why does it make a local copy of the sif file?

Checking the `output/__jobstatus__` I see:
```START: Tue Jan 21 00:45:36 2025
EXITCODE: -9999
FAILED: Tue Jan 21 09:45:37 2025
PROBLEM STARTING THE APPLICATION SCRIPT: '['apptainer', '-q', 'exec', '--bind', '/tmp/tmpoj_uarbk:/work_dir', '--no-home', '--bind', '/cvmfs:/cvmfs', 'nd280softwaremaster_centos7-13.28.sif', 'uname', '-a']' '[Errno 2] No such file or directory: 'apptainer''
FILES FOUND ARE: ['_python', 'nd280softwaremaster_centos7-13.28.sif', '__GangaInputData.txt__', 'stdout', 'stderr', '__syslog__', '.xdg']
```
So the job seems not to have inherited the $PATH from my shell (since the apptainer executable is in a rather non-standard location). Anyway that problem would seem to be completely unrelated to the problem with the Task since that doesn't even get as far as writing anything to the input directory.

It would be nice I there were some debugging flags/options I could turn on (running `ganga --debug` didn't reveal anything).

egede: For the Task related problem, maybe @mesmith75 can help. I know very little about it.

For the `SharedFile` object, I see that it is indeed copied to the working directory before it is used. Not ideal behaviour in this case. I think you can avoid that by simply writing 
```
imagefile = '/gpfs/group/t2k/nd280/RealDataProcessing/containers/nd280softwaremaster_centos7-13.28.sif'
```

However, even if this works, it is not optimal. Before `apptainer` can use a `sif` file, it has to unpack the filesystem stored in it (this is typically done in `/tmp`. It would most likely be better to have the container unpacked in advance. What is your setup. An isolated batch system with shared NFS or running on the grid, or something different again?

NickHastings: > For the Task related problem, maybe @mesmith75 can help. I know very little about it.
> 
> For the `SharedFile` object, I see that it is indeed copied to the working directory before it is used. Not ideal behaviour in this case. I think you can avoid that by simply writing
> 
> ```
> imagefile = '/gpfs/group/t2k/nd280/RealDataProcessing/containers/nd280softwaremaster_centos7-13.28.sif'
> ```
Thanks, can confirm that that prevents the `cp`.

I also found that making that change in the main code solved my problems :-)

> However, even if this works, it is not optimal. Before `apptainer` can use a `sif` file, it has to unpack the filesystem stored in it (this is typically done in `/tmp`. It would most likely be better to have the container unpacked in advance. What is your setup. An isolated batch system with shared NFS or running on the grid, or something different again?

Its the main compute cluster and batch system at KEK (not grid). File system is GPFS (IBMs clustered file system). I'm not so concerned about optimization at this point since I don't currently plan on using containers in production. This was a (hopefully) one off processing using an old software release that would only build on an old OS.

Anyway, it _seems_ so be working now. I'll just do a little more testing and then remove the WIP tag.

Thanks for your help.

NickHastings:
Hmm, all the discussion/comments after the first few days seems to have disappeared...

egede:
I resolved the discussion, but it should still be there (but slightly hidden).

egede:
@NickHastings Any updates?

NickHastings:
Sorry, forgot about this. Just now I rebased from develop, so I think it should be fine to merge.

egede (approved):

PR #2411, opened 2025-04-06
returning original object of outputfiles

### LINKED ISSUE #2190: Job outputfiles cannot be extended
It seems that one cannot add files to an existing list of outputfiles, only define a completely new list. This does not seem right to me:
```python
Ganga In [6]: type(jobs(997).outputfiles)
Ganga Out [6]: GangaCore.GPIDev.Base.Proxy.GangaList

[21:02:04]
Ganga In [7]: jobs(997).outputfiles = []

[21:02:10]
Ganga In [8]: jobs(997).outputfiles
Ganga Out [8]: []

[21:02:12]
Ganga In [9]: jobs(997).outputfiles.append(LocalFile('model.root'))

[21:02:17]
Ganga In [10]: jobs(997).outputfiles
Ganga Out [10]: []

```

sidharth-sinhasane:
fixes: #2190
Job output files can now be extended dynamically.

### DIFF
--- a/ganga/GangaCore/GPIDev/Lib/Job/Job.py
+++ b/ganga/GangaCore/GPIDev/Lib/Job/Job.py
@@ -554,34 +554,7 @@ class Job(GangaObject):
             return object.__getattribute__(self, name)
 
         if name == "outputfiles":
-            currentOutputFiles = object.__getattribute__(self, name)
-            currentUnCopyableOutputFiles = object.__getattribute__(
-                self, "non_copyable_outputfiles"
-            )
-
-            files = []
-            files2 = []
-
-            for f in currentOutputFiles:
-                if f.containsWildcards() and hasattr(f, "subfiles") and f.subfiles:
-                    files.extend(f.subfiles)
-                else:
-                    files.append(f)
-
-            for f in currentUnCopyableOutputFiles:
-                if f.containsWildcards() and hasattr(f, "subfiles") and f.subfiles:
-                    files2.extend(f.subfiles)
-                else:
-                    files2.append(f)
-
-            files3 = GangaList()
-            files3.extend(files)
-            files3.extend(files2)
-
-            # FIXME THIS SHOULD NOT HAVE TO BE HERE! (It does else we end up with really bad errors and this is wrong!)
-            files3._setParent(self)
-
-            return addProxy(files3)
+            return object.__getattribute__(self, name)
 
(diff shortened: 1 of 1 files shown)


egede (commented):
I am afraid that this will not work. In `outputfiles` we can have wildcards. So consider the following

```
j = Job()
j.application.exe='touch'
j.application.args=['abcd.txt']
j.outputfiles=['*.txt']
```

If I print the job object now, I see
```
Ganga In [11]: j.outputfiles
Ganga Out [11]: [LocalFile(namePattern='*.txt', localDir='')]
```
so the wildcard is still there. When I now submit the job and look at it after it has completed, I see something different
```
j.submit()
# wait
Ganga In [17]: j.outputfiles
Ganga Out [17]: [LocalFile(namePattern='abcd.txt', localDir='/home/egede/gangadir/workspace/egede/LocalXML/3/output')]
```
so you see that it now shows the actual file that was created and not the wildcard. However, if I copy the job (to create a new one)
```
j = j.copy()
Ganga In [18]: j.outputfiles
Ganga Out [18]: [LocalFile(namePattern='*.txt', localDir='')]
```
you see that the wildcard is back. This is the correct behaviour (i.e. the new job is a copy of the unsubmitted one).

I do not believe that your change here will have that behaviour. I am not sure if we have tes cases for this, but I will start them now.

egede:
As expected, the testing indeed shows that there now are multiple errors related to the use of wildcards for `outputfiles`.

sidharth-sinhasane:
While running [Gangacore/test/Unit](https://github.com/ganga-devs/ganga/tree/develop/ganga/GangaCore/test/Unit) and [Gangacore/test/GPI](https://github.com/ganga-devs/ganga/tree/develop/ganga/GangaCore/test/GPI) same test cases fail with or without these changes. I think we need to research more into it.

sidharth-sinhasane:
Tried returning proxy of original object.

outputfiles are now can be extended.
Screenshot from 2025-04-08 11-46-24


Also behaving as expected with wildcards.
Screenshot from 2025-04-08 11-37-55

egede (commented):
I do not think this latest change will work neither. In the last of your examples, the `outputfiles` should have "resolved" the wildcard and showed `abcd.txt` instead.

sidharth-sinhasane:
I tried same with inputfiles 
image
showed similar output as outputfiles(unresolved).

PR #2414, opened 2025-04-19
Cleaning legacy code

### LINKED ISSUE #2110: Clean up legacy in Job class schema

sidharth-sinhasane:
Description:

This PR is a step toward cleaning up legacy code as discussed in #2110.

- Removed references to inputSandbox and outputSandbox.

All existing tests are passing:

-  GangaCore/test/

-  GangaCore/test/GPI

More related clean-up commits will follow in this PR.

Please let me know if I’m heading in the right direction or if any adjustments are needed.

Partially fixes: #2110

### DIFF
--- a/ganga/GangaGaudi/Lib/RTHandlers/GaudiRunTimeHandler.py
+++ b/ganga/GangaGaudi/Lib/RTHandlers/GaudiRunTimeHandler.py
@@ -22,25 +22,6 @@ class GaudiRunTimeHandler(IRuntimeHandler):
     """This is the application runtime handler class for Gaudi applications 
     using the local, interactive and LSF backends."""
 
-    def master_prepare(self, app, appmasterconfig):
-        inputsandbox, outputsandbox = master_sandbox_prepare(app, appmasterconfig, ['inputsandbox'])
-        return StandardJobConfig(inputbox=unique(inputsandbox),
-                                 outputbox=unique(outputsandbox))
-
-    def prepare(self, app, appsubconfig, appmasterconfig, jobmasterconfig):
-
-        inputsandbox, outputsandbox = sandbox_prepare(
-            app, appsubconfig, appmasterconfig, jobmasterconfig)
-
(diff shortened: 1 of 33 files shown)


mesmith75:
I am not sure we want to be removing all the `prepare` and `master_prepare` functions

sidharth-sinhasane:
I’ve run both tests after removing the prepare and master_prepare functions, and the changes passed successfully. Based on this, I’m assuming it’s safe to proceed with the changes. Please let me know if this assumption is incorrect.

egede:
It looks to me as there is some confusion here. The schema entries `inputsandbox` and `outputsandbox` of the `Job` class are indeed legacy code that should be removed. But the concept of the `inputsandbox` and `outputsandbox` still exist in Ganga - it is just that the user do not directly interact with them. A lot of th code that is deleted in this PR relates to the those internal sandboxes and should stay.

sidharth-sinhasane:
> It looks to me as there is some confusion here. The schema entries `inputsandbox` and `outputsandbox` of the `Job` class are indeed legacy code that should be removed. But the concept of the `inputsandbox` and `outputsandbox` still exist in Ganga - it is just that the user do not directly interact with them. A lot of th code that is deleted in this PR relates to the those internal sandboxes and should stay.

I'll go ahead and revert those specific changes.

To address this properly, I’ll go through the files one by one. I’m marking this PR as a draft for now, and as I make incremental commits, please review them. We can proceed collaboratively from there.

PR #2420, opened 2025-06-11
added ganga import to executable

glk0:
Goal: fix an issue with the executable not working when using the [uv](https://github.com/astral-sh/uv) package manager.
Origin of issue:  `ganga` executable adds  `GangaCore` to `sys.path`  using:

https://github.com/ganga-devs/ganga/blob/ee0f0cbfc323ce011154a632587fc5e1d44b5346/bin/ganga#L43-L45
https://github.com/ganga-devs/ganga/blob/ee0f0cbfc323ce011154a632587fc5e1d44b5346/bin/ganga#L56-L59

This assumes the relative path the executable and library is as in the codebase. `uv` breaks that assumption, copying the executable elsewhere.  `__init__.py` in the `ganga` library has:

https://github.com/ganga-devs/ganga/blob/ee0f0cbfc323ce011154a632587fc5e1d44b5346/ganga/__init__.py#L4-L5

So  `import ganga` in the executable fix the issue.

### DIFF
--- a/bin/ganga
+++ b/bin/ganga
@@ -68,6 +68,7 @@ except KeyboardInterrupt:
 del standardSetup
 # %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+import ganga
 from GangaCore.Core.exceptions import GangaException
 import GangaCore.Runtime
 GangaCore.Runtime._prog = None


mesmith75:
I am not sure I understand this - when I install with `pip` the executable works fine. Is this just a quirk of a different package manager?

glk0:
In a 

> I am not sure I understand this - when I install with `pip` the executable works fine. Is this just a quirk of a different package manager?

In a certain way ... I have 2 ganga installs on my machine: one in global environment and the other in a venv. `file $(which ganga)` outputs:

```
/home/glk0/playganga/.venv/bin/ganga: Python script, ASCII text executable
```
for the venv and 
```
/home/glk0/.local/bin/ganga: symbolic link to /home/glk0/.local/share/uv/tools/ganga/bin/ganga
```
in the second case, the rest of the library is there with the executable while in the first case, only the executable file is there (going up one directory, you will not find the rest of  `ganga`'s source, which makes `from GangaCore import ...` fail even after `standardSetup`. I'm not sure if I can call it a quirk;  I've not tried with other package managers though.

glk0:
By the way should have marked this as draft. The fix is not working yet, some issue with configuration modifications.

PR #2430, opened 2025-07-29
Pass container as Apptainer constructor argument

NickHastings:
Previously was doing:
a = Apptainer()
a.image = '/path/to/foo.sif'

Now it seems that the base class requires
a = Apptainer('/path/to/foo.sif')

The checking the git history it's not clear to me when this change was introduced.


egede (approved):
At some point there was indeed a change made here to ensure that we could run unpackaged containers located in `/cvmfs`.

NickHastings:
I just made commit --amend (not change to code) that removes my unnecessary ramblings in the commit message.

Is it ok to squash the two commits together and then force push?

EDIT: I have trouble remembering policies of different projects.

PR #2432, opened 2025-09-04
removed use of some deprecated functions

glk0:
PR collecting various contributions tangential to main work: local testing fixes, removing deprecated features, etc

### DIFF
--- a/ganga/GangaCore/Runtime/bootstrap.py
+++ b/ganga/GangaCore/Runtime/bootstrap.py
@@ -1216,7 +1216,7 @@ under certain conditions; type license() for details.
         # The customisation of the interactive shell is done in this extension
         cfg.InteractiveShellApp.extensions = ['GangaCore.Runtime.ganga_extension']
 
-        IPython.start_ipython(argv=[], user_ns=local_ns, local_ns=local_ns, module=GangaCore.GPI, config=cfg)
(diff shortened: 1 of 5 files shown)


ganga/GangaCore/Runtime/bootstrap.py
```diff
@@ -1216,7 +1216,7 @@ def launch_IPython(local_ns):
         # The customisation of the interactive shell is done in this extension
         cfg.InteractiveShellApp.extensions = ['GangaCore.Runtime.ganga_extension']
 
-        IPython.start_ipython(argv=[], user_ns=local_ns, local_ns=local_ns, module=GangaCore.GPI, config=cfg)
```
mesmith75, line 1219: What is the purpose of this change?

mesmith75:
Is this all just linting? What is the script doing and when is it used?

ganga/GangaCore/Runtime/bootstrap.py
glk0: > What is the purpose of this change?

The old call convention is deprecated.

glk0:
The script is for local lint & fix, during local testing. Added with removing some deprecated APIs used to produce noisy outputs when running Ganga.

glk0:
I removed the linting part. Just the removal of various deprecated stuffs leading to some annoying messages.

PR #2422, opened 2025-07-07
adding deprecation utils

glk0:
Add ganga deprecation decorators, plus tests.

PR #2435, opened 2025-09-22
added deprecation management tooling

glk0:
This PR adds the tooling to manage deprecations. It inclue utility functions that allow to locate all the deprecated classes and functions in the codebase, parse the relevant informations and output it as JSON, etc. It relies on the `@deprecated` decorator from PEP 702 and `@deprecation_deadline` decorator from GangaCore.Utility.Deprecation introduced in https://github.com/ganga-devs/ganga/pull/2422 .


egede:
@alexanderrichards Do you have any idea why the `GangaDirac Unit` test is failing here. It looks to me like the proxy is OK but it then fails with some SSL stuff during setup?

scripts/deprecation_report.py
```diff
@@ -0,0 +1,10 @@
+import sys
+from ganga.GangaCore.Utility.Deprecation import generate_deprecation_json_report
+
+root = "./ganga"
```
egede: This should be dynamically extracted from the code that is executed. You can use the `__file__` attricute for this.

ganga/GangaCore/Utility/Deprecation.py
```diff
@@ -46,6 +60,9 @@ def decorator(obj):
         if not deprecated_msg:
             raise RuntimeError("This function or class is not marked as deprecated. Use @deprecation_dealine like:\n "
```
egede: Typo `dealine` -> `deadline`

egede (commented):
- Can you provide an example of how the `json` output looks like. 
- There should be some test cases that shows that it is working (i.e the `deprecation_report` should be tested.
- The use of `deprecation_report` should be integrated into the testing framework. So a test should be written that fails if there is code around which is past its expiry date.

ganga/GangaCore/Utility/Deprecation.py
```diff
@@ -104,3 +121,224 @@ def disable_deprecation_warnings():
...
+
+            if deprecation_deadline_dec is not None:
+                self.generic_visit(deprecation_deadline_dec)
+                try:
+                    kw = {k: v for k, v in (
+                        # Here we use `eval` to extract the arguments because the arguments are either
+                        # string literals or use the `date` function imported in this module. If the user
+                        # respect the format defined in the documentation of the deprecation utilies, this
+                        # should extract the correct information.
+                        map(lambda x: (x.arg, eval(ast.unparse(x.value))),
```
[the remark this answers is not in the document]
glk0: We trust the code that inside the codebase, so this is irrelevant. Wanted to be broad and allow use of relatively complex expression involving dates, which will be difficult to achieve with custom parsing.

ganga/GangaCore/Utility/Deprecation.py
```diff
@@ -104,3 +121,224 @@ def disable_deprecation_warnings():
...
+    """A NodeVisitor to locate all the deprecation notices and parse
+    them into DeprecationNotice objects
+    """
+
+    def __init__(self, file_path):
+        self.deprecation_instances = []
+        self.file_path = file_path
+
+    def visit_FunctionDef(self, node):
+        notice = self._check_decorators(node, "function")
```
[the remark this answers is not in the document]
glk0: Left over from a previous iteration, corrected.

glk0:
> Can you provide an example of how the json output looks like.

Here's a prettified output:

```json
[
  {
    "name": "list_plugins",
    "type": "function",
    "file_path": "GangaCore/Runtime/GPIFunctions.py",
    "lineno": 86,
    "reason": "use plugins ('category')",
    "deadline": "2025-12-31",
    "last_version": "8.4.4"
  },
  {
    "name": "applications",
    "type": "function",
    "file_path": "GangaCore/Runtime/GPIFunctions.py",
    "lineno": 92,
    "reason": "use plugins ('applications')",
    "deadline": null,
    "last_version": null
  },
  {
    "name": "backends",
    "type": "function",
    "file_path": "GangaCore/Runtime/GPIFunctions.py",
    "lineno": 97,
    "reason": "use plugins ('backend')",
    "deadline": null,
    "last_version": null
  }
]

```

Shows the functions to which we have applied the `@deprecated` decorator in `GPIFunctions.py`. Those were already marked for removal, so I figure I might try the new method there.

scripts/deprecation_report.py
glk0: Fixed

ganga/GangaCore/Utility/Deprecation.py
glk0: Fixed, thanks

scripts/deprecation_report.py
```diff
@@ -0,0 +1,11 @@
+import sys
```
egede: I think this can just be removed. It is sufficient that it runs as part of the unit test.

glk0: Fixed.

PR #2442, opened 2025-11-18
Python314 install and run

### LINKED ISSUE #2407: Ganga fails to install with python3.13
There is not yet a pythin3.13 compatible version of the `htcondor` package available. This in turn prevents Ganga to install via pip. From reading in the `htcondor` newsgroup, it is caused by some changes in how the `wheel` package works in 3.13 and it will take a bit of time to solve.

We can either make the `htcondor` installation optional (as any production system is anyway unlikely to use python 3.13 at the moment), or we can just wait for the problem to solve itself and in the meantime not have Ganga working with 3.13.

alexanderrichards:
closes #2407

This PR makes `htcondor` optional as well as fixing deprecation warnings and other little niggles when running Ganga under Python 3.14

-----

**Note: Will need to check that the updates to run under Python3.14 don't prevent ganga running under our min version.**

### DIFF
--- a/ganga/GangaCore/Lib/Condor/Condor.py
+++ b/ganga/GangaCore/Lib/Condor/Condor.py
@@ -42,8 +42,8 @@ import warnings
 with warnings.catch_warnings():
     warnings.simplefilter("ignore")
     try:
-        import htcondor
-        import classad
+        import htcondor2 as htcondor
+        import classad2 as classad
     except ModuleNotFoundError:
         logger.debug("The htcondor module not available for import. This will cause use of the Condor backend to fail.")
 
@@ -119,6 +119,8 @@ class Condor(IBackend):
 
         # Get the credential into the scheduler
         col = htcondor.Collector()
+
+        # Not sure this is needed in new API
         credd = htcondor.Credd()
         if credential_store.matches(AfsToken()):
             credd.add_user_cred(htcondor.CredTypes.Kerberos, None)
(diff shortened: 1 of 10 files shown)


egede:
I came across that htcondor has changed its python bindings. I suspect this is the reason for our problems. See [manual](https://htcondor.readthedocs.io/en/latest/apis/python-bindings/api/version2/migration-guide.html) on migration.

alexanderrichards:
Thanks @egede I made the changes for the htcondor migration and ganga seems happy to start. One thing that I find interesting is that I can pip install htcondor fine in Python3.14 today where as I'm sure that It couldn't find a version yesterday (This was the original problem). Maybe htcondor is now available and we simply need the migration patch?

alexanderrichards:
Yay fixed the Condor tests as well 🎉

ganga/GangaCore/Lib/Condor/Condor.py
```diff
@@ -119,9 +119,11 @@ def master_submit(self, rjobs, subjobconfigs, masterjobconfig, keep_going=False,
 
         # Get the credential into the scheduler
         col = htcondor.Collector()
-        credd = htcondor.Credd()
-        if credential_store.matches(AfsToken()):
-            credd.add_user_cred(htcondor.CredTypes.Kerberos, None)
+
+        # Not sure this is needed in new API
```
mesmith75, line 123: Are you sure this isn't needed. I can't see anything in the documentation to suggest the AFS tokens is automatically passed with the job submission

alexanderrichards: In short, no I'm not sure. I was getting problems with the method not existing in the new api and when I look it up I saw suggestions that it would automaticaly take the credentials from the environment if you have previously issues a kinit.

I will check again now that the Condor tests actually seem to be working.

mesmith75: Ah great - I'll have a go with the CERN Condor

PR #2446, opened 2026-01-20
Attempt to fix the interactive logger

Lalshivam:
The fix was straightforward and involved two key changes:

1. Moved Finalisation Logs from INFO to DEBUG
The noisy shutdown/finalisation messages that were spamming the interactive console were changed from logger.info()
 to logger.debug():

   -Before (spammy in interactive mode)
       log.info("Ganga is now attempting to shut down...")

   -After (only visible in DEBUG mode)
       log.debug("Ganga is now attempting to shut down...")

# Why This Works:
-INFO level remains active in interactive mode (not globally suppressed)
-Important user messages still appear at INFO level
-Internal finalisation chatter is now at DEBUG level, only visible when debugging
-Users who need to see finalisation details can use --debug flag

This approach is cleaner than the previous workaround of suppressing all INFO logs globally, which broke legitimate INFO messages that users needed to see.

### DIFF
--- a/ganga/GangaCore/Core/GangaRepository/container_controllers.py
+++ b/ganga/GangaCore/Core/GangaRepository/container_controllers.py
@@ -238,9 +238,11 @@ def native_handler(database_config, action, gangadir):
     if action not in ["start", "quit"]:
         raise NotImplementedError("Illegal Opertion on container")
     if action == "start":
-        logger.info("Native Database detection, skipping startup")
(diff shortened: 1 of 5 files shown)


mesmith75:
All of these things we want to keep as `info`. These are not the ones related to the job finalisation I am afraid.

egede:
Please avoid AI generated PRs.

